Issue #5, opened 2019-04-03
Support  saving to and load from s3 location


parano:
e.g.
```python
my_bentoml_model.save("s3://...")
```
```python
import bentoml
bentoml.load("s3://...")
```

PR #9, opened 2019-04-03
Save and Load model from S3 location

yubozhao:
Allow user to save and load model from a S3 location.

Load from cli:
```
bentoml serve --model-path=s3://bucket_name/key_to_model_archive
```

Load from file system:
```
import bentoml

ms = bentoml.load('s3://bucket/key/to/model')
```

https://github.com/bentoml/BentoML/issues/5

How was this patch tested?
--------------------------
pytest locally

### DIFF
--- a/bentoml/cli/__init__.py
+++ b/bentoml/cli/__init__.py
@@ -14,6 +14,10 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
 import click
 import os
 import tempfile
--- a/bentoml/cli/__init__.py
+++ b/bentoml/cli/__init__.py
@@ -16,12 +16,10 @@
 
 import click
 import os
+import tempfile
 from bentoml.loader import load
 from bentoml.server import BentoModelApiServer
-from bentoml.utils.s3 import check_is_s3_path, download_from_s3
-
-
-TEMPORARY_BENTOML_DIR_PATH = '/tmp/bentoml_tmp/download'
+from bentoml.utils.s3 import is_s3_path, download_from_s3
 
 
 @click.group()
@@ -46,9 +44,10 @@ def serve(model_path, port, storage_type='file'):
         if not os.path.isabs(model_path):
             model_path = os.path.abspath(model_path)
     elif storage_type == 's3':
-        if not check_is_s3_path(model_path):
+        if not is_s3_path(model_path):
             raise ValueError('Incorrect s3 path format')
-        downloaded_file_path = download_from_s3(model_path, TEMPORARY_BENTOML_DIR_PATH)
+        temp_dir = tempfile.mkdtemp()
+        downloaded_file_path = download_from_s3(model_path, temp_dir)
         model_path = downloaded_file_path
 
     model_service = load(model_path)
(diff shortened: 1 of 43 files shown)


bentoml/cli/__init__.py
```diff
@@ -18,6 +18,10 @@
 import os
 from bentoml.loader import load
 from bentoml.server import BentoModelApiServer
+from bentoml.utils.s3 import check_is_s3_path, download_from_s3
+
+
+TEMPORARY_BENTOML_DIR_PATH = '/tmp/bentoml_tmp/download'
```
parano: use tempfile.mkdtemp instead

bentoml/cli/__init__.py
```diff
@@ -30,15 +34,22 @@ def cli():
 @cli.command()
 @click.option('--model-path', type=click.Path(exists=True), required=True)
 @click.option('--port', type=click.INT)
-def serve(model_path, port):
+@click.option('--storage-type', type=click.STRING)
```
parano: why is this option necessary? you can tell it by looking at model-path right?

bentoml/loader.py
```diff
@@ -16,13 +16,17 @@
...
+import uuid
 
 from ruamel.yaml import YAML
 
 from bentoml.service import BentoService
 from bentoml.version import __version__ as BENTOML_VERSION
+from bentoml.utils.s3 import check_is_s3_path, download_from_s3
 
 
+TEMPORARY_BENTOML_DIR_PATH = '/tmp/bentoml_tmp/download'
```
parano: same as above

bentoml/loader.py
```diff
@@ -82,17 +86,27 @@ def load_bentoml_config(path):
     return bentoml_config
 
 
-def load(path, lazy_load=False):
+def load(path, lazy_load=False, storage_type='file'):
```
parano: similarly this parameter is probably not necessary

bentoml/utils/s3.py
```diff
@@ -0,0 +1,112 @@
...
+
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+import boto3
+from six.moves.urllib.parse import urlparse
+
+
+def check_is_s3_path(path):
```
parano: nit: rename to "is_s3_path",  "check" is more like asserting the given path is a s3 path

bentoml/model.py
```diff
@@ -266,28 +283,48 @@ def save(self, base_path, version=None, project_base_dir=None, copy_entire_proje
...
+                except Exception as e:
+                    raise e
+
+    def load(self, path=None, storage_type='file'):
         # TODO: add model.env.verify() to check dependencies and python version etc
 
         if self.__class__._bento_module_path is not None:
             # When calling load from pip installled bento model, use installed
             # python package for loading and the same path for '/artifacts'
+
```
parano: remove

bentoml/cli/__init__.py
yubozhao: It is easy to detect. Still want user explicitly declare it, so they are aware of it.  Also, when we support more remote storage type, this will come handy

bentoml/utils/s3.py
yubozhao: Yep, make sense

bentoml/cli/__init__.py
yubozhao: Cool, didn't know this lib exist. Will change to this

bentoml/cli/__init__.py
parano: agree but better to make it behave similar to other data science toolkits, e.g.
http://pandas.pydata.org/pandas-docs/stable/user_guide/io.html#reading-remote-files

Issue #94, opened 2019-04-19
TypeError: 'TfKerasModelArtifact' object is not iterable


sayakpaul:
Hi there. Here's my packaging code:
```python
%%writefile text_classification_service.py
import pandas as pd
from tensorflow import keras
from sklearn.preprocessing import LabelEncoder
from sklearn.feature_extraction.text import CountVectorizer
from string import digits
from bentoml import api, env, BentoService, artifacts
from bentoml.artifact import TfKerasModelArtifact
from bentoml.handlers import JsonHandler

@artifacts([TfKerasModelArtifact('model')])
@env(conda_dependencies=['tensorflow', 'pandas', 'scikit-learn'])
class TextClassificationService(BentoService):
    
    def vectorizer():
        vectorizer = CountVectorizer(stop_words=None, lowercase=True,
                             ngram_range=(1, 1), min_df=2, binary=True)
        
        train = pd.read_csv('https://raw.githubusercontent.com/Nilabhra/kolkata_nlp_workshop_2019/master/data/train.csv')
        vectorizer.fit_transform(train['text'])
        return vectorizer
    
    def remove_digits(s):
        remove_digits = str.maketrans('', '', digits)
        res = s.translate(remove_digits)
        return res
    
    @api(JsonHandler)
    def predict(self, parsed_json):
        text = parsed_json['text']
        text = remove_digits(text)
        vectorizer = vectorizer()
        text = vectorizer.transform(text)
        prediction =  self.artifacts.model.predict_classes(text)
        response = {'Sentiment': prediction}
        return response
```

And when I am building the archive using - 
```python
from text_classification_service import TextClassificationService

svc = TextClassificationService.pack(model=model)
saved_path = svc.save('/tmp/bento')
print(saved_path)
```

It gives me the following error (with trace):
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-28-39d1075ec116> in <module>
      1 from text_classification_service import TextClassificationService
      2 
----> 3 svc = TextClassificationService.pack(model=model)
      4 saved_path = svc.save('/tmp/bento')
      5 print(saved_path)

/miniconda3/lib/python3.7/site-packages/bentoml/service.py in pack(cls, *args, **kwargs)
    320         artifacts = ArtifactCollection()
    321 
--> 322         for artifact_spec in cls._artifacts_spec:
    323             if artifact_spec.name in kwargs:
[...]

PR #97, opened 2019-04-21
FIx tf keras inferenc issue for rest server

yubozhao:
Fixed issue https://github.com/bentoml/BentoML/issues/94.

*Issue:*  
Tensorflow Keras model not able to predict in Flask app. With Error: `Tensor Tensor("dense_2/Sigmoid:0", shape=(?, 1), dtype=float32) is not an element of this graph.`

*Possible Cause:*
The Tensorflow keras model loading and prediction might not be in the same thread, because flask's usage of multi thread.  

*Solution:*
Create a TF model wrapper that will load and predict with the same session and graph.

References:
* https://stackoverflow.com/questions/51127344/tensor-is-not-an-element-of-this-graph-deploying-keras-model
* https://github.com/tensorflow/tensorflow/issues/14356

How was this patch tested?
--------------------------
https://github.com/bentoml/BentoML/issues/94

### DIFF
--- a/bentoml/artifact/tf_keras_model_artifact.py
+++ b/bentoml/artifact/tf_keras_model_artifact.py
@@ -39,12 +39,13 @@ class TfKerasModelArtifact(ArtifactSpec):
         return _TfKerasModelArtifactInstance(self, model)
 
     def load(self, path):
-        try:
-            from tensorflow.python.keras.models import load_model
-        except ImportError:
-            raise ImportError("tensorflow package is required to use TfKerasModelArtifact")
+        #   try:
+        #       from tensorflow.python.keras.models import load_model
+        #   except ImportError:
+        #       raise ImportError("tensorflow package is required to use TfKerasModelArtifact")
 
-        model = load_model(self._model_file_path(path))
+        model = _TfKerasModelWrapper(self._model_file_path(path))
+        # model = load_model(self._model_file_path(path))
         return self.pack(model)
 
 
@@ -58,8 +59,8 @@ class _TfKerasModelArtifactInstance(ArtifactInstance):
         except ImportError:
             raise ImportError("tensorflow package is required to use TfKerasModelArtifact")
 
-        if not isinstance(model, training.Model):
-            raise ValueError('Expected `model` argument to be a `Model` instance')
+        #   if not isinstance(model, training.Model):
+        #       raise ValueError('Expected `model` argument to be a `Model` instance')
 
         self._model = model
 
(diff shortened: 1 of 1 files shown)


bentoml/artifact/tf_keras_model_artifact.py
```diff
@@ -68,3 +69,30 @@ def save(self, dst):
...
+            with self.session.as_default():
+                self.model = load_model(model)
+
+    def predict(self, x):
+        with self.graph.as_default():
+            with self.session.as_default():
+                result = self.model.predict(x)
+        return result
+
+    def predict_classes(self, x):
```
parano: this is not a `training.Model` method, only `Sequential` has it

parano: also make sure to pass down all parameters for this type of wrapper  `def predict_classes(self, x, batch_size=32, verbose=0):`

bentoml/artifact/tf_keras_model_artifact.py
```diff
@@ -68,3 +69,30 @@ def save(self, dst):
...
+    def __init__(self, model):
+        try:
+            import tensorflow as tf
+            from tensorflow.python.keras.models import load_model
+        except ImportError:
+            raise ImportError("tensorflow package is required to use TfKerasModelArtifact")
+        self.session = tf.Session()
+        self.graph = tf.get_default_graph()
+        with self.graph.as_default():
+            with self.session.as_default():
```
parano: is this necessary?

bentoml/artifact/tf_keras_model_artifact.py
parano: I think might be better to force gloabl default graph for entire API function instead of wrapping individual keras model methods here

yubozhao: I was afraid of using it as global default graph, but I will test that route.

bentoml/artifact/tf_keras_model_artifact.py
yubozhao: I think so, you need to load the model in the same session/graph

bentoml/artifact/tf_keras_model_artifact.py
yubozhao: Another pro of wrapping the individual keras model is there is no additional actions from user.  If we use global graph, user have to make prediction call within that graph context.  I think it would be just strange to users to use a var that 'comes out' from nowhere.

```python
...
def predict(self, df):
      with graph.as_default():    # user has no knowledge of where this come from.
              result = self.artifacts.model.predict(df)
```

bentoml/artifact/tf_keras_model_artifact.py
```diff
@@ -68,3 +69,30 @@ def save(self, dst):
...
+
+class _TfKerasModelWrapper(object):
+
+    def __init__(self, model):
+        try:
+            import tensorflow as tf
+            from tensorflow.python.keras.models import load_model
+        except ImportError:
+            raise ImportError("tensorflow package is required to use TfKerasModelArtifact")
+        self.session = tf.Session()
```
parano: shouldn't this be:
```
self.graph = tf.get_default_graph()
sess = tf.Session(graph=graph)
```?

bentoml/artifact/tf_keras_model_artifact.py
parano: @yubozhao oh I didn't mean "global" graph, I was referring to self.graph here. But instead of wrapping on Keras model, wrap it around the API function.

parano: Btw, I have some new changes on how user API function access artifacts, that might make this implementation easier.

yubozhao: Oh. I got you.  Also. I totally forgot to mention to you that this PR is not mean to be merged. I was able to get the tf keras model working with this, but I don’t think this is a good solution and there are going to be potential problems down the line. This PR’s goal is start to have a conversation around this

PR #113, opened 2019-04-24
Add gunicorn support for bentoml

yubozhao:
* Use gunicon instead of flask for DockerFile
  - take 2 optional arguments
    1. -w, --workers, default number of core / 2 + 1
    2. -p, --port, default 5000

```bash
cd archive_path
python __init__.py --port 4000 --workers 3
```

How was this patch tested?
--------------------------

### DIFF
--- a/bentoml/archive/archiver.py
+++ b/bentoml/archive/archiver.py
@@ -117,7 +117,6 @@ import os
 import sys
 
 from bentoml import archive
-from bentoml.server import BentoAPIServer
 from bentoml.cli import create_bentoml_cli
 
 __VERSION__ = "{pypi_package_version}"
--- a/bentoml/archive/archiver.py
+++ b/bentoml/archive/archiver.py
@@ -308,7 +308,8 @@ def _save(bento_service, dst, version=None):
     with open(os.path.join(path, 'Dockerfile'), 'w') as f:
         f.write(
             BENTO_SERVER_SINGLE_MODEL_DOCKERFILE_TEMPLATE.format(
-                conda_env_name=bento_service.env.get_conda_env_name(), service_name=bento_service.name))
+                conda_env_name=bento_service.env.get_conda_env_name(),
+                service_name=bento_service.name))
 
     # write bentoml.yml
     bentoml_yml_content = BENTOML_CONFIG_YAML_TEMPLATE.format(
--- a/bentoml/archive/archiver.py
+++ b/bentoml/archive/archiver.py
@@ -107,16 +107,23 @@ ENV PATH /opt/conda/envs/$conda_env/bin:$PATH
 
 RUN conda install pip && pip install -r /bento/requirements.txt
 
-# Run bento server with path to bento archive
-CMD ["bentoml serve /bento"]
+# Run Gunicorn server with path to module.
+CMD ["python /bento/{service_name}/__init__.py"]
 """
 
 # TODO: improve this with import hooks PEP302?
 INIT_PY_TEMPLATE = """\
+from __future__ import unicode_literals
 import os
 import sys
 
+import argparse
+import multiprocessing
+import gunicorn.app.base
+from gunicorn.six import iteritems
+
 from bentoml import archive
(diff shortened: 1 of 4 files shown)


bentoml/archive/archiver.py
```diff
@@ -127,10 +134,53 @@ def _parse_requirements(file_path):
...
 def load():
     return archive.load(__module_path)
 
+
+parser = argparse.ArgumentParser()
+parser.add_argument('-p', '--port')
+parser.add_argument('-w', '--workers')
+
+
+class GunicornApplication(gunicorn.app.base.BaseApplication):
```
parano: could you add some tests for this?

bentoml/archive/archiver.py
```diff
@@ -127,10 +134,53 @@ def _parse_requirements(file_path):
...
 cli=create_bentoml_cli(__module_path)
 
+
 def load():
     return archive.load(__module_path)
 
+
+parser = argparse.ArgumentParser()
+parser.add_argument('-p', '--port')
+parser.add_argument('-w', '--workers')
```
parano: add arg documentation here?

bentoml/archive/archiver.py
yubozhao: yeah, let me do that very quick

bentoml/archive/archiver.py
yubozhao: yep, good suggestion

bentoml/archive/archiver.py
```diff
@@ -127,10 +134,53 @@ def _parse_requirements(file_path):
...
+            self.cfg.set(key.lower(), value)
+
+    def load(self):
+        return self.application
+
+
 __all__ = ['__version__', '{service_name}', 'load']
+
+
+if __name__ == '__main__':
```
parano: On second thought, it might be better to put this inside a `bentoml` command. for example `bentoml serve-gunicorn ./bento_archive`. That way it's easier to re-use this code if we need to(as it won't be in a template), and a lot easier to test.

setup.py
```diff
@@ -24,7 +24,8 @@
     'Werkzeug',
     'pathlib2',
     'requests',
-    'packaging'
+    'packaging',
+    'gunicorn'
```
parano: make this an optional dependency but always install it in docker - thoughts?

bentoml/cli/__init__.py
```diff
@@ -29,6 +33,26 @@
 from bentoml.cli.serverless import generate_serverless_bundle
 
 
+class GunicornApplication(gunicorn.app.base.BaseApplication):  # pylint: disable=abstract-method
```
parano: move this to under bentoml/server folder?

bentoml/cli/__init__.py
```diff
@@ -91,6 +115,20 @@ def serve(port, archive_path=installed_archive_path):
         server = BentoAPIServer(model_service, port=port)
         server.start()
 
+    # Example Usage: bentoml serve-gunicorn ./SAVED_ARCHIVE_PATH --port=PORT --workers=WORKERS
+    @bentoml_cli.command()
+    @conditional_argument(installed_archive_path is None, 'archive-path', type=click.STRING)
+    @click.option('-p', '--port', type=click.INT, default=BentoAPIServer._DEFAULT_PORT)
+    @click.option('-w', '--workers', type=click.INT, default=(multiprocessing.cpu_count() // 2) + 1)
```
parano: move `multiprocessing.cpu_count() // 2) + 1` out as a separate method and add some comments about the choice here?

setup.py
```diff
@@ -42,8 +42,9 @@
 cv2 = [ 'opencv-python' ]
 pytorch  = [ 'torch', 'torchvision' ]
 tensorflow = [ 'tensorflow' ]
+gunicorn = [ 'gunicorn' ]
```
parano: also need to add this to the generated Dockerfile:
```
RUN conda install pip && pip install gunicorn
RUN pip install -r /bento/requirements.txt
```

Issue #92, opened 2019-04-19
Add support for H2O.ai models


sayakpaul:
The H2O.ai implementations of linear models, tree-based models and monotonically constrained models are really good. Moreover, H2O.ai supports AutoML as well. 

Since BentoML has support for Scikit-Learn, Keras, TensorFlow and so son, I thought it might be a good idea to add H2O.ai models in this list.

Nevertheless, you guys are doing a pretty awesome job and I am preparing demo notebook to make myself accustomed with the workflow of BentoML. Will share once completed.

parano:
Thanks @sayakpaul, it's a great idea to add H2O.ai model support, I'm adding it to our roadmap list. It seems straight forward to add a new artifact type from our side with h2o python module's save_model and load_model method.

Let us know how it goes with your demo notebook, would love to get your feedback.

sayakpaul:
Yes, sure. It took me 7 minutes (yes I measured the time) to wrap Scikit-Learn model into a REST API using BentoML and then test it in Postman. I have pushed my notebook [here](https://github.com/sayakpaul/BentoML-Explorations) (I will fix the references in some time). I am now working on wrapping a tf-keras model for a sentiment classification problem (dataset is scraped from Zomato :P). Will share it once done.

PR #133, opened 2019-05-02
Add h2o support

yubozhao:
Add h2o support
* H2oModelArtifact
* Example h2o notebook, classification prostate

https://github.com/bentoml/BentoML/issues/92

How was this patch tested?
--------------------------
n/a


bentoml/artifact/h2o_artifact.py
```diff
@@ -0,0 +1,72 @@
...
+        super(_H2oModelArtifactInstance, self).__init__(spec)
+        self._model = model
+
+    def save(self, dst):
+        try:
+            import h2o
+        except ImportError:
+            raise ImportError("h2o package is required to use H2oModelArtifact")
+
+        # H2o will generate model into a directory instead of to file
```
parano: this doesn't seem necessary to me, we can just use h2o's API to save and load from the path, it doesn't need to be a "file". Also I'm afraid there are cases

examples/h2o-classification/h2o-example.ipynb
```diff
@@ -0,0 +1,218 @@
...
+    "# H2O Classification"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# !curl -i https://raw.githubusercontent.com/multicode/h2o-notebook/master/prostate.csv -o prostate.csv"
```
parano: would be great to use bash's if statement here to make this notebook "run-all-able", similar to what we did with the sklearn example

examples/h2o-classification/h2o-example.ipynb
```diff
@@ -0,0 +1,218 @@
...
+    "import bentoml\n",
+    "\n",
+    "print(bentoml.__file__)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# H2O Classification"
```
parano: maybe add a few more markdown cells with some context/comments to this notebook?

examples/h2o-classification/h2o-example.ipynb
```diff
@@ -0,0 +1,218 @@
...
+   "source": [
+    "%%writefile h2o_model_service.py\n",
+    "import pandas as pd\n",
+    "import h2o\n",
+    "import bentoml\n",
+    "from bentoml.artifact import H2oModelArtifact\n",
+    "from bentoml.handlers import DataframeHandler\n",
+    "\n",
+    "@bentoml.artifacts([H2oModelArtifact('model')])\n",
+    "@bentoml.env(conda_dependencies=[\"h2o\", \"pandas\"])\n",
```
parano: I think this env config actually won't install h2o properly, you will need to add the conda-forge channel and also `conda install conda-py` just glancing through the document.  would be great to include a docker build in this notebook as well, cause that way we can actually test these things

examples/h2o-classification/h2o-example.ipynb
```diff
@@ -0,0 +1,218 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import bentoml\n",
+    "import pandas as pd\n",
+    "\n",
+    "df = pd.read_csv(\"https://raw.githubusercontent.com/multicode/h2o-notebook/master/prostate.csv\")\n",
+    "# Load exported bentoML model archive from path\n",
+    "bento_model = bentoml.load(saved_path)\n",
+    "bento_model.predict(df)"
```
parano: could you also add 'curl' and 'cli' example?

examples/h2o-classification/h2o-example.ipynb
parano: yup just tested this out, 'conda install h2o' only installs the java package. It seems to also install the python package in python2.7 - an easy workaround is to make it a pip dependency, because `pip install h2o` seems to work.

yubozhao: Yep. need to commit that change to reflect this.

bentoml/artifact/h2o_artifact.py
yubozhao: Since we only have information on the "directory" instead of the actual model file. We would need to do this hack.

parano: this is what I menat:
```
h2o_saved_path = h2o.save_model(model=self._model, path=dst)
os.rename(h2o_saved_path, self.spec._model_file_path(dst))
```

examples/h2o-classification/h2o-example.ipynb
```diff
@@ -0,0 +1,254 @@
...
+   "source": [
+    "%%writefile h2o_model_service.py\n",
+    "import pandas as pd\n",
+    "import h2o\n",
+    "import bentoml\n",
+    "from bentoml.artifact import H2oModelArtifact\n",
+    "from bentoml.handlers import DataframeHandler\n",
+    "\n",
+    "@bentoml.artifacts([H2oModelArtifact('model')])\n",
+    "@bentoml.env(requirements_txt='./requirements.txt')\n",
```
parano: consider
```
conda_pip_dependencies=['h2o']
# or
requirements_txt="""h2o"""
```

examples/h2o-classification/h2o-example.ipynb
yubozhao: done

bentoml/artifact/h2o_artifact.py
yubozhao: done

examples/h2o-classification/h2o-example.ipynb
yubozhao: done, no longer apply

examples/h2o-classification/h2o-example.ipynb
yubozhao: done

examples/h2o-classification/h2o-example.ipynb
yubozhao: done

examples/h2o-classification/h2o-example.ipynb
yubozhao: done

yubozhao:
@parano ready for another review

sayakpaul:
Great going guys. Good to see this implemented. Cannot wait to try this.

PR #135, opened 2019-05-07
Delete deployment command

yubozhao:
```bash
bentoml delete-deployment SERVICE_NAME --platform=aws-lambda
```

How was this patch tested?
--------------------------


bentoml/deployment/serverless.py
```diff
@@ -92,3 +98,28 @@ def deploy_with_serverless(platform, archive_path, extra_args):
     output_path = generate_serverless_bundle(bento_service, platform, archive_path, extra_args)
     subprocess.call(['serverless', 'deploy'], cwd=output_path)
     return output_path
+
+
+def delete_serverless_deployment(platform, service_name, additional_info):
+    yaml = YAML()
+    temp_dir = tempfile.mkdtemp()
```
parano: try `with TempDirectory as tempdir:` in `utils/tempdir.py`? the main difference is that it adds a prefix indicating the directory is related to bentoml package and it cleans up the directory afterward

bentoml/cli/__init__.py
```diff
@@ -144,6 +145,23 @@ def deploy(archive_path, platform, region, stage):
                 'Deploying with "--platform={platform}" is not supported in the current version of BentoML'
                 .format(platform=platform))
 
+    # Example useage: bentoml delete-deployment SERVICE_NAME --platform=aws-lambda
+    @_cli.command()
+    @click.argument('service-name', type=click.STRING)
+    @click.option('--platform', type=click.Choice([
+        'aws-lambda', 'aws-lambda-py2', 'gcp-function', 'aws-sagemaker', 'azure-ml', 'algorithmia'
+    ]), required=True)
+    @click.option('--region', type=click.STRING, required=True)
```
parano: will the `bentoml deploy` command fail if it has already been deployed to a different region in the same account? for simple deployment I thought it might be easier to just use the service name as "key" for delete and status

bentoml/cli/__init__.py
```diff
@@ -144,6 +145,23 @@ def deploy(archive_path, platform, region, stage):
                 'Deploying with "--platform={platform}" is not supported in the current version of BentoML'
                 .format(platform=platform))
 
+    # Example useage: bentoml delete-deployment SERVICE_NAME --platform=aws-lambda
+    @_cli.command()
+    @click.argument('service-name', type=click.STRING)
```
parano: I think it makes more sense to use the path to bento archive here, we can load the service name from an archive path. 'service-name' is kinda internal concept and doesn't mean anything to the user.

bentoml/deployment/aws_lambda_template.py
```diff
@@ -60,13 +61,13 @@ def generate_serverless_configuration_for_aws(apis, output_path, additional_opti
...
     else:
-        serverless_config['provider']['region'] = 'us-west-2'
+        serverless_config['provider']['region'] = DEFAULT_AWS_REGION
 
     if additional_options.get('stage', None):
         serverless_config['provider']['stage'] = additional_options['stage']
         logger.info(('Using user defined AWS stage: {0}', additional_options['stage']))
     else:
-        serverless_config['provider']['stage'] = 'dev'
+        serverless_config['provider']['stage'] = DEFAULT_DEPLOY_STAGE
```
parano: also add a log statement here `logger.info(('Using default AWS stage: {0}', DEFAULT_DEPLOY_STAGE)`

bentoml/deployment/aws_lambda_template.py
```diff
@@ -60,13 +61,13 @@ def generate_serverless_configuration_for_aws(apis, output_path, additional_opti
         serverless_config['provider']['region'] = additional_options['region']
         logger.info(('Using user defined AWS region: {0}', additional_options['region']))
     else:
-        serverless_config['provider']['region'] = 'us-west-2'
+        serverless_config['provider']['region'] = DEFAULT_AWS_REGION
```
parano: same here

bentoml/deployment/gcp_function_template.py
```diff
@@ -54,9 +55,14 @@ def generate_serverless_configuration_for_google(bento_service, apis, output_pat
     if additional_options.get('region', None):
         serverless_config['provider']['region'] = additional_options['region']
         logger.info(('Using user defined Google region: {0}', additional_options['region']))
+    else:
+        serverless_config['provider']['region'] = DEFAULT_GCP_REGION
```
parano: same here

bentoml/deployment/gcp_function_template.py
```diff
@@ -54,9 +55,14 @@ def generate_serverless_configuration_for_google(bento_service, apis, output_pat
...
+
     if additional_options.get('stage', None):
         serverless_config['provider']['stage'] = additional_options['stage']
         logger.info(('Using user defined Google stage: {0}', additional_options['stage']))
+    else:
+        serverless_config['provider']['stage'] = DEFAULT_DEPLOY_STAGE
```
parano: same here

bentoml/deployment/serverless.py
```diff
@@ -92,3 +98,28 @@ def deploy_with_serverless(platform, archive_path, extra_args):
     output_path = generate_serverless_bundle(bento_service, platform, archive_path, extra_args)
     subprocess.call(['serverless', 'deploy'], cwd=output_path)
     return output_path
+
+
+def delete_serverless_deployment(platform, service_name, additional_info):
```
parano: `additional_info.stage` is not used?

parano: might be cleaner to just list out parameters than hiding them in a dict `delete_serverless_deployment(platform, service_name, region, stage):`

bentoml/deployment/serverless.py
```diff
@@ -92,3 +98,28 @@ def deploy_with_serverless(platform, archive_path, extra_args):
...
+        config = {
+            "service": service_name,
+            "provider": {
+                "name": "aws",
+                "region": additional_info.get('region', DEFAULT_AWS_REGION)
+            }
+        }
+    yaml.dump(config, Path(saved_path))
+    subprocess.call(['serverless', 'remove'], cwd=temp_dir)
+    return
```
parano: would be great to have this function to return True when the deletion was successful and False when it failed to delete

bentoml/deployment/serverless.py
```diff
@@ -92,3 +98,28 @@ def deploy_with_serverless(platform, archive_path, extra_args):
...
+    elif platform == 'aws-lambda' or platform == 'aws-lambda-py2':
+        config = {
+            "service": service_name,
+            "provider": {
+                "name": "aws",
+                "region": additional_info.get('region', DEFAULT_AWS_REGION)
+            }
+        }
+    yaml.dump(config, Path(saved_path))
+    subprocess.call(['serverless', 'remove'], cwd=temp_dir)
```
parano: is there a way to figure out if this command run is successful or not?

bentoml/cli/__init__.py
yubozhao: make sense

bentoml/cli/__init__.py
yubozhao: check out the comment below.  I think your next comment raise a good point. Let's use archive_path for now.

bentoml/cli/__init__.py
yubozhao: done

bentoml/deployment/aws_lambda_template.py
yubozhao: done

bentoml/deployment/aws_lambda_template.py
yubozhao: done

bentoml/deployment/gcp_function_template.py
yubozhao: done

bentoml/deployment/gcp_function_template.py
yubozhao: done

bentoml/deployment/serverless.py
yubozhao: done

bentoml/deployment/serverless.py
yubozhao: done

bentoml/deployment/serverless.py
```diff
@@ -92,3 +99,31 @@ def deploy_with_serverless(platform, archive_path, extra_args):
...
+                "name": "aws",
+                "region": DEFAULT_AWS_REGION if region is None else region,
+                "stage": DEFAULT_AWS_DEPLOY_STAGE if stage is None else stage
+            }
+        }
+    yaml = YAML()
+    with TempDirectory() as tempdir:
+        saved_path = os.path.join(tempdir, 'serverless.yml')
+        yaml.dump(config, Path(saved_path))
+        subprocess.call(['serverless', 'remove'], cwd=tempdir)
```
parano: check the output see if the subprocess call is successful or not, and return true or false correspondingly

bentoml/deployment/serverless.py
```diff
@@ -92,3 +99,31 @@ def deploy_with_serverless(platform, archive_path, extra_args):
     output_path = generate_serverless_bundle(bento_service, platform, archive_path, extra_args)
     subprocess.call(['serverless', 'deploy'], cwd=output_path)
     return output_path
+
+
+def delete_serverless_deployment(platform, archive, region, stage):
+    bento_server = load(archive)
```
parano: nit: * bento_service

parano: Actually, it looks like you don't need to load the entire archve, just need to load the config:
```
service_name =  load_bentoml_config(archive_path)["metadata"]["service_name"]
```

bentoml/cli/__init__.py
```diff
@@ -144,6 +145,23 @@ def deploy(archive_path, platform, region, stage):
...
+    @_cli.command()
+    @click.argument('archive-path', type=click.STRING)
+    @click.option('--platform', type=click.Choice([
+        'aws-lambda', 'aws-lambda-py2', 'gcp-function', 'aws-sagemaker', 'azure-ml', 'algorithmia'
+    ]), required=True)
+    @click.option('--region', type=click.STRING, required=True)
+    @click.option('--stage', type=click.STRING)
+    def delete_deployment(archive_path, platform, region, stage):
+        if platform in SERVERLESS_PLATFORMS:
+            delete_serverless_deployment(platform, archive_path, region, stage)  # pylint:disable=too-many-function-args
```
parano: Recommend to check the deployment status given (platform, archive_path, region, stage) before calling delete, if there is no existing deployment found it should stop here and let the user know.

bentoml/deployment/serverless.py
parano: let's create a serverless class for accessing these commands? that way we can put the logic of checking the output to determine command status in there. And it makes it much easier to mock and add tests.

yubozhao: yep

yubozhao: done

bentoml/cli/__init__.py
yubozhao: done

bentoml/deployment/serverless.py
yubozhao: done

bentoml/deployment/serverless.py
yubozhao: done

bentoml/deployment/serverless.py
yubozhao: done

bentoml/deployment/serverless/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+        yaml = YAML()
+        with TempDirectory() as tempdir:
+            saved_path = os.path.join(tempdir, 'serverless.yml')
+            yaml.dump(config, Path(saved_path))
+
+            with subprocess.Popen(['serverless', 'remove'], cwd=tempdir, stdout=PIPE, stderr=PIPE) as proc:
+                response = parse_serverless_response(proc.stdout.read().decode('utf-8'))
+                logger.debug('Serverless response: %s', '\n'.join(response))
+                if self.platform == 'google-python':
+                    pass
```
parano: `pass` ?

bentoml/deployment/serverless/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+        elif platform == 'aws-lambda' or platform == 'aws-lambda-py':
+            self.region = DEFAULT_AWS_REGION if region is None else region
+            self.stage = DEFAULT_AWS_DEPLOY_STAGE if stage is None else stage
+
+    def generate_bundle(self):
+        output_path = generate_bentoml_deployment_snapshot_path(self.bento_service.name, self.platform)
+        Path(output_path).mkdir(parents=True, exist_ok=False)
+
+        # Calling serverless command to generate templated project
+        with subprocess.Popen(['serverless', 'create', '--template', self.provider, '--name', self.bento_service.name],
```
parano: just curious what's the difference between this and 
```
p = subprocess.Popen(...)
stdout, stderr = p.communicate
```
which seems more common?

bentoml/deployment/serverless/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+    str_list = serverless_response.strip().split('\n')
+    error = [s for s in str_list if 'Serverless Error' in s]
+    if error:
+        error_pos = str_list.index(error[0])
+        error_message = str_list[error_pos + 1]
+        raise BentoMLException(error_message)
+    return str_list
+
+class ServerlessDeployment(object):
+    """Managing deployment operations for serverless
```
parano: nice work pulling this part out 💯

bentoml/deployment/serverless/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+        self.archive_path = archive_path
+        self.bento_service = load(archive_path)
+        if platform == 'google-python':
+            self.region = DEFAULT_GCP_REGION if region is None else region
+            self.stage = DEFAULT_GCP_DEPLOY_STAGE if stage is None else stage
+        elif platform == 'aws-lambda' or platform == 'aws-lambda-py':
+            self.region = DEFAULT_AWS_REGION if region is None else region
+            self.stage = DEFAULT_AWS_DEPLOY_STAGE if stage is None else stage
+
+    def generate_bundle(self):
```
parano: maybe change to `_generate_bundle `?

bentoml/deployment/serverless/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+        self.platform = platform
+        self.provider = SERVERLESS_PROVIDER[platform]
+        self.archive_path = archive_path
+        self.bento_service = load(archive_path)
+        if platform == 'google-python':
+            self.region = DEFAULT_GCP_REGION if region is None else region
+            self.stage = DEFAULT_GCP_DEPLOY_STAGE if stage is None else stage
+        elif platform == 'aws-lambda' or platform == 'aws-lambda-py':
+            self.region = DEFAULT_AWS_REGION if region is None else region
+            self.stage = DEFAULT_AWS_DEPLOY_STAGE if stage is None else stage
```
parano: add an else catch all statement here and raise?

parano:
overall looks great other than those a few comments. there are still a bunch of linter tests failing tho.

In folllow up PR could you add some integration test? similar to those in serverless framework? https://github.com/serverless/serverless/blob/master/tests/integration/aws/general/nested-handlers/tests.js

Those tests will need AWS/Gcloud information and we will only be able to run locally. But it's still a good idea to have them and for now, we can run manually before each release.

bentoml/deployment/serverless/__init__.py
yubozhao: 💯

bentoml/deployment/serverless/__init__.py
yubozhao: I was toying with the idea to let user just generate again.  I will make the _ prefix

bentoml/deployment/serverless/__init__.py
yubozhao: yep, make sense

bentoml/deployment/serverless/__init__.py
yubozhao: For our current use cases, we don't have a lot of interactions with the process, this style is fine.  We don't have to terminate the process ourselves. 

For more advanced use cases, we probably want to utilize the style in your comment's code block, since we can have more control over input/output like add more input into the process (username/password/etc).

bentoml/deployment/serverless/__init__.py
yubozhao: I need setup google cloud function and test

bentoml/deployment/serverless/__init__.py
yubozhao: done

bentoml/deployment/serverless/__init__.py
yubozhao: done

PR #156, opened 2019-06-06
Specify api_server deps separately

parano:
Use  ```pip install bentoml[api_server]``` when building API server image, which will install extra deps including gunicorn

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,6 +33,7 @@ install_requires = [
     "click",
     "pandas",
     "dill",
+    "prometheus_client",
     "python-json-logger",
     "boto3",
     "pathlib2",
--- a/setup.py
+++ b/setup.py
@@ -73,6 +73,7 @@ tests_require += cv2
 
 extras_require = {
     "all": dev_all,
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -12,68 +12,75 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
```
parano: included style change from `black`

setup.py
```diff
@@ -83,35 +90,29 @@
...
+            "bentoml=bentoml.cli:cli"
+        ]
     },
     project_urls={
-        'Bug Reports': 'https://github.com/bentoml/BentoML/issues',
-        'Source Code': 'https://github.com/bentoml/BentoML'
-    }
+        "Bug Reports": "https://github.com/bentoml/BentoML/issues",
+        "Source Code": "https://github.com/bentoml/BentoML",
+        "Gitter Chat Room": "https://gitter.im/bentoml/BentoML",
```
parano: also added gitter chat room link to pypi page

PR #173, opened 2019-06-20
Logging config

parano:
* added global logging configuration
* make it configurable via config file (or env var)

n/a

How was this patch tested?
--------------------------
tested locally


bentoml/config/__init__.py
```diff
@@ -40,6 +40,15 @@ def expand_env_var(env_var):
             env_var = interpolated
 
 
+def parameterized_config(template):
+    """Generates a configuration from the provided template + variables defined in
+    current scope
+    :param template: a config content templated with {{variables}}
```
yubozhao: Let's use the Google docstring style.  That way we can be consistent with rest of the project and generate good documentation.

Google docstring format:
"""Summary

Long Summary

Args:
    param 1:  description

Returns:
	type: description
Raises:

bentoml/server/bento_api_server.py
```diff
@@ -100,11 +116,14 @@ def feedback_view_func():
     if len(data.keys()) <= 1:
         return Response(response="Missing feedback data", status=400)
 
-    log_feedback(feedback_logger, data)
+    bento_service["service_name"] = bento_service.name
+    bento_service["service_version"] = bento_service.version
```
yubozhao: Do we included the bento-service in the data send to feedback_logger?

yubozhao: you mean data["service_name"] and data["service_version"] right?

yubozhao:
LGTM, just that small fix for feedback function

Issue #175, opened 2019-06-20
BentoService#save should copy local python modules


parano:
**Describe the bug**
this was a regression introduced when we refactor the archive module, will take a look and add tests

parano:
https://github.com/bentoml/BentoML/pull/180

Issue #61, opened 2019-04-13
Add more info logs to BentoService#save


parano:
- The find module call is slow and not printing anything is pretty bad UX
- It's better to print out what the library is doing under the hood for the user, e.g. what python source code files are being copied, what artifacts are being saved etc.

parano:
https://github.com/bentoml/BentoML/pull/180

PR #180, opened 2019-06-21
BentoService#save should copy local python modules #175

parano:
* fix the bug described in #175 
* also #61

### DIFF
--- a/bentoml/archive/py_module_utils.py
+++ b/bentoml/archive/py_module_utils.py
@@ -21,6 +21,7 @@ import re
 import sys
 import inspect
 import importlib
+import logging
 from shutil import copyfile
 from modulefinder import ModuleFinder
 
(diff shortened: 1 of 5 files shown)


bentoml/archive/py_module_utils.py
```diff
@@ -121,17 +141,18 @@ def copy_used_py_modules(target_module, destination):
...
+    )
 
     # Look for dependencies that are not distributed python package, but users'
     # local python code, all other dependencies must be defined with @env
     # decorator when creating a new BentoService class
     user_packages_and_modules = {}
     for name, module in iteritems(finder.modules):
         if name == "bentoml" or name.startswith("bentoml."):
-            # Remove BentoML library from dependent modules list
-            break
```
parano: this was the bug :(

PR #181, opened 2019-06-22
Enhance the input_parameter in DataframeHandler for docs generation.

leonsim:

Update DataframHandler.

How was this patch tested?
--------------------------
Tested locally.

image

### DIFF
--- a/bentoml/handlers/dataframe_handler.py
+++ b/bentoml/handlers/dataframe_handler.py
@@ -51,20 +51,52 @@ class DataframeHandler(BentoHandler):
             records.
         typ (string): Type of object to recover for read json with pandas. Default is
             frame
-        input_columns ([string]): A list of column names that input data need to have.
+        input_dtypes ({string:string}): A dict of column name and data type.
 
     Raises:
-        ValueError: Incoming data is missing required columns in input_columns
+        ValueError: Incoming data is missing required columns in input_dtypes
         ValueError: Incoming data format is not handled. Only json and csv
     """
 
     def __init__(
-        self, orient="records", output_orient="records", typ="frame", input_columns=None
+        self, orient="records", output_orient="records", typ="frame", input_dtypes=None
     ):
         self.orient = orient
         self.output_orient = output_orient or orient
         self.typ = typ
-        self.input_columns = input_columns
+        self.input_dtypes = input_dtypes
+
+    def _get_type(self, item):
+        if item.startswith('int'):
+            return 'integer'
+        if item.startswith('float') or item.startswith('double'):
+            return 'number'
+        if item.startswith('str') or item.startswith('date'):
+            return 'string'
+        if item.startswith('bool'):
+            return 'boolean'
+        return 'object'
+
+    @property
+    def request_schema(self):
+        default = {"application/json": {"schema": {"type": "object"}}}
+        if self.input_dtypes is None:
+            return default
+
+        if isinstance(self.input_dtypes, dict):
+            return {
+                "application/json": {  # For now, only declare JSON on docs.
+                    "schema": {
+                        "type": "object",
+                        "properties": {
+                            k: {"type": "array", "items": {"type": self._get_type(v)}}
+                            for k, v in self.input_dtypes.items()
+                        },
+                    }
+                }
+            }
+
+        return default
 
     def handle_request(self, request, func):
         orient = request.headers.get("orient", self.orient)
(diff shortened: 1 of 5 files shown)


yubozhao:
Will this work with the `check_dataframe_column_contains` in dataframe handler?

leonsim:
> Will this work with the `check_dataframe_column_contains` in dataframe handler?

Yes, for now it's still Ok.
The `input_columns` usage changed by this PR though:
from `["col1", "col2"]` changed to `{"col1": "int", "col2": "int"}` .

The `check_dataframe_column_contains` still works because inside itself, it do a `set` transferring first.

yubozhao:
Sweet. Other than that. Looks good to me. Let’s merge it

bentoml/handlers/dataframe_handler.py
```diff
@@ -66,6 +66,36 @@ def __init__(
         self.typ = typ
         self.input_columns = input_columns
 
+    def _get_type(self, item):
```
parano: I know it's probably less common in ML use cases, but would be nice to also add `boolean` and `date` type here? I think both OpenAPI spec and pandas dataframe have those

bentoml/handlers/dataframe_handler.py
```diff
@@ -66,6 +66,36 @@ def __init__(
         self.typ = typ
         self.input_columns = input_columns
```
parano: maybe just rename 'input_columns' to 'input_dtype' now?

examples/quick-start/iris_classifier.py
```diff
@@ -9,7 +9,7 @@
 @env(conda_pip_dependencies=["scikit-learn"])
 class IrisClassifier(BentoService):
 
-    @api(DataframeHandler)
+    @api(DataframeHandler, input_columns={"col1":"int8","col2":"int8","col3":"int8","col4":"int8"})
```
parano: I think in this example it should be "float64"?

bentoml/handlers/dataframe_handler.py
```diff
@@ -66,6 +66,36 @@ def __init__(
...
+            return 'string'
+        return 'object'
+
+    @property
+    def request_schema(self):
+        default = {"application/json": {"schema": {"type": "object"}}}
+        if self.input_columns is None:
+            return default
+
+        if isinstance(self.input_columns, dict):
```
parano: Might be good to support list as input as well?

for example in the quick start, if you look at the input dtypes:
```pd.DataFrame(X).dtypes.to_dict()```

the columns names is basically the index:
```{0: dtype('float64'),
 1: dtype('float64'),
 2: dtype('float64'),
 3: dtype('float64')}
```

In those cases it might be easier to simply do:
```@api(DataframeHandler, input=['float', 'float', 'float', 'float'])```

and API client will send similar json format to the API endpoint:
```
curl -i \
--header "Content-Type: application/json" \
--request POST \
--data '[[5.1, 3.5, 1.4, 0.2]]' \
localhost:5000/predict
```

Comment (leonsim) [reply bentoml/handlers/dataframe_handler.py]: Hi, this is cool, but I haven't found a solution to implement this in a less complicated way.
Can I extend this in next PRs?

Also, if we support list type, the `check_dataframe_column_contains` needs to refactor too, as pointed out by @yubozhao

Comment (parano) [reply bentoml/handlers/dataframe_handler.py]: yea sounds great, let's do it and refactor check_dataframe_column_contains in follow up PR

File: examples/quick-start/bentoml-quick-start-guide.ipynb [hunk 5]
```diff
@@ -77,7 +77,7 @@
     "@env(conda_pip_dependencies=[\"scikit-learn\"])\n",
     "class IrisClassifier(BentoService):\n",
     "\n",
-    "    @api(DataframeHandler)\n",
+    "    @api(DataframeHandler, input_dtypes={\"col1\":\"float64\",\"col2\":\"float64\",\"col3\":\"float64\",\"col4\":\"float64\"})\n",
```
parano: maybe remove it from the quick_start guide for now, until we iterate on this a bit more?

leonsim:
To be safe, I'll merge this after today's release. 😄

bentoml/handlers/dataframe_handler.py
parano: btw, I was thinking of a simple conversion here when seeing an array:
```
if isinstance(input_dtypes, list):
  input_dtypes = dict((index, dtype) for index, dtype in enumerate(ic))
```
pd.Dataframe.read_json should already handling the reading part, both way works fine

leonsim: Good idea.

The question here is I'm not sure how it works on Open API that we support two formats for input parameters, ie, one is `[[1, 2]]`, another is `{"col1":[1],"col2":[2]}`.
Basically no matter user put a list or a dict into `input_dtypes`, we need to support the above 2 formats on API docs. Is my understanding right?

PR #190, opened 2019-07-02
API document improvements

parano:
* Fixed a few small things in the API docs
* added build script in dev doc

### DIFF
--- a/bentoml/artifact/text_file_artifact.py
+++ b/bentoml/artifact/text_file_artifact.py
@@ -25,11 +25,11 @@ class TextFileArtifact(ArtifactSpec):
     """Abstraction for saving/loading string to/from text files
 
     Args:
-        name (string): Name of the artifact
-        file_extension (string): The file format artifact will be saved as.
-            Default is .txt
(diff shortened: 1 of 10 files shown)


bentoml/artifact/text_file_artifact.py
```diff
@@ -25,11 +25,11 @@ class TextFileArtifact(ArtifactSpec):
     """Abstraction for saving/loading string to/from text files
 
     Args:
-        name (string): Name of the artifact
-        file_extension (string): The file format artifact will be saved as.
-            Default is .txt
-        encoding (string): The encoding will be used for saving/loading text.
-            Default is utf8
+        name (str): Name of the artifact
+        file_extension (:obj:`str`: optional): The file extention used for the saved
```
yubozhao: I think there is a typo.  Instead of `:` it should be `,` separate between type and optional

yubozhao: Screen Shot 2019-07-02 at 4 33 06 PM

PR #191, opened 2019-07-03
Fix error: 'ImageHandler' object has no attribute 'request_schema'

leonsim:

Fix an error about request_schema not found for ImageHandler

Nope

How was this patch tested?
--------------------------
Added unit tests.

### DIFF
--- a/bentoml/server/bento_api_server.py
+++ b/bentoml/server/bento_api_server.py
@@ -141,9 +141,7 @@ def get_docs(bento_service):
             post=OrderedDict(
                 tags=["app"],
                 description=api.doc,
-                requestBody=OrderedDict(
-                    required=True, content=api.handler.request_schema
-                ),
+                requestBody=OrderedDict(required=True, content=api.request_schema),
                 responses=default_response,
             )
         )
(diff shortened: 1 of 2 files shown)


bentoml/server/bento_api_server.py
```diff
@@ -141,9 +141,7 @@ def get_docs(bento_service):
             post=OrderedDict(
                 tags=["app"],
                 description=api.doc,
-                requestBody=OrderedDict(
-                    required=True, content=api.handler.request_schema
-                ),
+                requestBody=OrderedDict(required=True, content=api.request_schema),
```
yubozhao: Since we add default request_schema for all handlers.  We can make the part on BentoServiceAPI's property more straight forward now.  Other than that, good job!

https://github.com/bentoml/BentoML/blob/11725d46cd512bfa4006fd75afd98f819471aff7/bentoml/service.py#L95

```
    @property
    def request_schema(self):
        if hasattr(self.handler, 'request_schema'):
            return self.handler.request_schema
        return {"type": "object"}

```

leonsim: Sure. Maybe we can even delete that property?

PR #209, opened 2019-07-17
ImageHandler update

parano:
* allow post image directly with '--data-binary', for example:
```
curl -X POST "http://127.0.0.1:5000/predict" -H "Content-Type: image/png" --data-binary @Sample_Image.png
```
* 👆 also fixed posting image with the OpenAPI debugging web UI, user can upload image from the web UI for testing API endpoint now:
image

* fix ImageHandler request_schema for OpenAPI docs

* add request_schema for FastaiImageHandler


parano:
@leonsim could you take a look at the OpenAPI part and @yubozhao for the ImageHandler changes

PR #218, opened 2019-07-18
Include handler's dependencies to the archived env yml

yubozhao:
Include handler's dependencies as well, when we are archiving bento service

### DIFF
--- a/bentoml/handlers/base_handlers.py
+++ b/bentoml/handlers/base_handlers.py
@@ -61,6 +61,10 @@ class BentoHandler:
     def request_schema(self):
         return {"application/json": {"schema": {"type": "object"}}}
 
+    @property
+    def handler_dependencies(self):
+        return []
+
     def handle_clipper_strings(self, inputs, func):
         """Handle incoming input data for clipper cluster.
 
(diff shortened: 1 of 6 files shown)


bentoml/handlers/base_handlers.py
```diff
@@ -61,6 +61,10 @@ def handle_aws_lambda_event(self, event, func):
     def request_schema(self):
         return {"application/json": {"schema": {"type": "object"}}}
 
+    @property
+    def handler_dependencies(self):
```
parano: rename *pip_dependencies

bentoml/handlers/image_handler.py
```diff
@@ -90,6 +90,10 @@ def request_schema(self):
             },
         }
 
+    @property
+    def handler_dependencies(self):
+        return ['imageio', 'opencv-python']
```
parano: remove opencv-python for now, revisit lambda case in separate PR

bentoml/service.py
```diff
@@ -159,6 +159,12 @@ def _config_service_apis(self):
                     BentoServiceAPI(self, api_name, api_doc, handler, func)
                 )
 
+                # pylint:disable=no-member
+
+                # Add handler's dependencies to the env requirements
+                self._env.add_handler_dependencies(handler.handler_dependencies)
```
parano: move to _env

PR #226, opened 2019-07-24
Usage tracking

yubozhao:

### DIFF
--- a/bentoml/cli/__init__.py
+++ b/bentoml/cli/__init__.py
@@ -75,6 +75,7 @@ def create_bento_service_cli(archive_path=None):
     @conditional_argument(archive_path is None, "archive-path", type=click.STRING)
     @click.pass_context
     def run(ctx, api_name, archive_path=archive_path):
+        track_cli('run')
         model_service = load(archive_path)
 
         try:
@@ -92,8 +93,6 @@ def create_bento_service_cli(archive_path=None):
                 )
             )
 
-        track_cli(api_name, model_service, ctx.args)
-
         api.handle_cli(ctx.args)
 
     # Example Usage: bentoml info /SAVED_ARCHIVE_PATH
@@ -106,9 +105,8 @@ def create_bento_service_cli(archive_path=None):
         """
         List all APIs defined in the BentoService loaded from archive
         """
+        track_cli('info')
         model_service = load(archive_path)
-
-        track_cli('info', model_service)
 
         service_apis = model_service.get_service_apis()
         output = json.dumps(
@@ -128,9 +126,8 @@ def create_bento_service_cli(archive_path=None):
     )
     @conditional_argument(archive_path is None, "archive-path", type=click.STRING)
     def docs(archive_path=archive_path):
+        track_cli('docs')
         model_service = load(archive_path)
-
-        track_cli('docs', model_service)
 
         print(json.dumps(get_docs(model_service), indent=2))
 
@@ -146,11 +143,9 @@ def create_bento_service_cli(archive_path=None):
         default=BentoAPIServer._DEFAULT_PORT,
         help="The port to listen on for the REST api server, default is 5000.",
     )
-    @click.pass_context
-    def serve(ctx, port, archive_path=archive_path):
+    def serve(port, archive_path=archive_path):
+        track_cli('serve')
         model_service = load(archive_path)
-
-        track_cli('serve', model_service, ctx.args)
 
         server = BentoAPIServer(model_service, port=port)
         server.start()
@@ -171,11 +166,9 @@ def create_bento_service_cli(archive_path=None):
         help="Number of workers will start for the gunicorn server",
     )
     @click.option("--timeout", type=click.INT, default=60)
-    @click.pass_context
-    def serve_gunicorn(ctx, port, workers, timeout, archive_path=archive_path):
+    def serve_gunicorn(port, workers, timeout, archive_path=archive_path):
+        track_cli('serve_gunicorn')
         model_service = load(archive_path)
-
-        track_cli('serve_gunicorn', model_service, ctx.args)
 
         server = BentoAPIServer(model_service, port=port)
         gunicorn_app = GunicornApplication(server.app, port, workers, timeout)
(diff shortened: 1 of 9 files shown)


bentoml/service.py
```diff
@@ -412,7 +412,9 @@ def version(self):
 
     def save(self, *args, **kwargs):
         from bentoml import archive
+        from bentoml.utils.usage_stats import track_archive
```
parano: just import at top of the file.

bentoml/utils/usage_stats.py
```diff
@@ -0,0 +1,113 @@
...
+import requests
+
+from bentoml import config, load
+import bentoml
+
+AMPLITUDE_URL = "https://api.amplitude.com/httpapi"
+API_KEY = '7f65f2446427226eb86f6adfacbbf47a'
+
+
+def get_artifact_handler_info(bento_service):
```
parano: maybe just ```get_bento_service_info```? might be good to also add `env` staff here

bentoml/cli/__init__.py
```diff
@@ -91,6 +92,8 @@ def run(ctx, api_name, archive_path=archive_path):
                 )
             )
 
+        track_cli(api_name, model_service, ctx.args)
```
parano: change `api_name` here to just run, we can log api_name separately...  api_name variable is user-defined and we can't easily use that

bentoml/cli/__init__.py
```diff
@@ -104,6 +107,9 @@ def info(archive_path=archive_path):
         List all APIs defined in the BentoService loaded from archive
         """
         model_service = load(archive_path)
+
+        track_cli('info', model_service)
```
parano: no need to do this in every command, just add it to the command group definition, which will be invoked for every click command

bentoml/cli/deployment.py
```diff
@@ -72,6 +73,8 @@ def add_deployment_commands(cli):
     def deploy(
         archive_path, platform, region, stage, api_name, instance_type, instance_count
     ):
+        track_deployment(platform, archive_path)
```
parano: this seems unnecessary to me since we already track CLI commands. Let's get rid of it for now and we can add something like this when we have python APIs for deployment staff (e.g. the clipper style python API or later on REST APIs for creating/managing deployment)

bentoml/utils/usage_stats.py
```diff
@@ -0,0 +1,113 @@
...
+        artifact_types.append(artifact_instance.__class__.__name__)
+
+    for api in bento_service.get_service_apis():
+        handler_types.append(api.handler.__class__.__name__)
+
+    return {"handler_types": handler_types, "artifact_types": artifact_types}
+
+
+def track(event_type, info):
+    info["py_version"] = "{major}.{minor}.{micro}".format(
```
parano: move this and `bento_version`, `platform` to top of this file as constants?

bentoml/utils/usage_stats.py
```diff
@@ -0,0 +1,113 @@
...
+    event_data = {
+        "api_key": API_KEY,
+        "event": json.dumps(event_info)
+    }
+
+    try:
+        requests.post(AMPLITUDE_URL, data=event_data)
+        return
+    except Exception:  # pylint:disable=broad-except
+        # silently fail
```
parano: should still log it with `logger.error`

bentoml/utils/usage_stats.py
```diff
@@ -0,0 +1,113 @@
...
+        "user_id": str(uuid.uuid4()),
+        "event_properties": info
+    }]
+    event_data = {
+        "api_key": API_KEY,
+        "event": json.dumps(event_info)
+    }
+
+    try:
+        requests.post(AMPLITUDE_URL, data=event_data)
```
parano: let's add a 1-second timeout to this? just in case it hangs for too long when used in slow/limited internet connection.

bentoml/utils/usage_stats.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def track_deployment(platform_name, archive_path):
+    if config['core'].getboolean("usage_tracking"):
+        try:
+            service = load(archive_path)
+            info = get_artifact_handler_info(service)
+            info['deployment_platform'] = platform_name
+            return track("deploy", info)
+        except Exception:  # pylint:disable=broad-except
```
parano: remove this try/catch ?

bentoml/cli/__init__.py
yubozhao: we should able to have access to api_name at this point. But I think you are right now change it to `run` to standardize events.

parano: i mean can’t easily use that for analytics

bentoml/utils/usage_stats.py
yubozhao: done

bentoml/cli/deployment.py
yubozhao: done

bentoml/cli/__init__.py
yubozhao: done

bentoml/utils/usage_stats.py
yubozhao: done

bentoml/utils/usage_stats.py
yubozhao: done

bentoml/utils/usage_stats.py
yubozhao: done

bentoml/utils/usage_stats.py
parano: actually - let's use `logger.info` here, since the message is not important for user

parano:
Another thing we should do is to opt-out or mock the API call in dev/test environment, we don't want to get spam'd every time a test run is triggered.

bentoml/utils/usage_stats.py
yubozhao: done

bentoml/service.py
yubozhao: done

bentoml/utils/usage_stats.py
yubozhao: done

PR #236, opened 2019-07-30
Accept list for env's requirements_txt parameter

yubozhao:
Accept list of string for the requirements_txt parameter.

### DIFF
--- a/bentoml/service_env.py
+++ b/bentoml/service_env.py
@@ -21,7 +21,6 @@ from sys import version_info
 from ruamel.yaml import YAML
 
 from bentoml.utils import Path
-from bentoml.exceptions import BentoMLException
 from bentoml.version import __version__ as BENTOML_VERSION
 
 PYTHON_VERSION = "{major}.{minor}.{micro}".format(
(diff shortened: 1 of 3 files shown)


bentoml/service_env.py
```diff
@@ -152,20 +152,22 @@ def set_setup_sh(self, setup_sh_path_or_content):
...
         if requirements_txt_file.is_file():
             with requirements_txt_file.open("rb") as f:
-                self._requirements_txt = f.read()
+                content = f.read()
+                module_list = content.decode('utf-8').split('\n')
+                self._pip_dependencies += module_list
         else:
-            self._requirements_txt = requirements_txt_path_or_content.encode(
-                "utf-8"
+            raise BentoMLException(
```
parano: Use ValueError

parano: Actually - just remove the if statement in front, and let it raise FileNotFoundError

parano:
Looks good besides the last comment - could you in separate PR also update all existing examples to use this instead of conda_pip_dependencies? Thanks

Issue #123, opened 2019-04-25
Add --with-conda option to CLI command serve and run


parano:
Allow user to load and run a BentoArchive in its own conda environment.
We can use the BentoService name and version string to ensure unique conda env for every BentoArchive.

PR #239, opened 2019-07-31
Add --with-conda optional parameter for cli commands.

leonsim:

Add an optional parameter for `bentoml serve`

#123 

How was this patch tested?
--------------------------
Tested locally.


bentoml/cli/__init__.py
```diff
@@ -74,8 +86,47 @@ def bentoml_cli(ctx, verbose, quiet):
     )
     @click.argument("api-name", type=click.STRING)
     @conditional_argument(archive_path is None, "archive-path", type=click.STRING)
+    @click.option(
+        '--with-conda',
+        is_flag=True,
+        default=False,
+        help="Run API server on a Conda environment.",
```
parano: nit ".. in a BentoML managed Conda environment"

bentoml/cli/__init__.py
```diff
@@ -74,8 +86,47 @@ def bentoml_cli(ctx, verbose, quiet):
...
+            env_path = os.path.join('/tmp', env_name + '.yaml')
+            yaml.dump(config['env']['conda_env'], Path(env_path))
+
+            pip_req = os.path.join(archive_path, 'requirements.txt')
+
+            subprocess.call(
+                'command -v conda >/dev/null 2>&1 || {{ echo >&2 "--with-conda '
+                'parameter requires conda but it\'s not installed."; exit 1; }} && '
+                'conda env update -n {env_name} -f {env_file} && '
+                'conda init bash && '
```
parano: instead of init bash and activate, let's just do:
```
/Users/.../anaconda3/envs/{env_name...}/bin/bentoml {api_name} .... 
```

should work the same right?

bentoml/cli/__init__.py
```diff
@@ -152,7 +235,44 @@ def serve(port, archive_path=archive_path):
...
+                'eval "$(conda shell.bash hook)" && '
+                'conda activate {env_name} && '
+                '{{ [ -f {pip_req} ] && pip install -r {pip_req} || echo "no pip '
+                'dependencies."; }} &&'
+                'bentoml serve_gunicorn {archive_path} -p {port} -w {workers} '
```
parano: Is there a way to get all the CLI args from `click`? this makes it harder for us to add CLI args because we always need to remember to add it here - it's fine if there's no easy way to do it

examples/quick-start/bentoml-quick-start-guide.ipynb
```diff
@@ -15,11 +15,62 @@
...
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 1,
    "metadata": {
     "scrolled": true
    },
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stdout",
```
parano: let's clear the output cells of this notebook - we plan to just keep output cells for the notebooks in gallery repo

examples/quick-start/bentoml-quick-start-guide.ipynb
```diff
@@ -152,9 +236,104 @@
...
+      "/Users/leon.sun/anaconda3/envs/IrisClassifier_2019_08_08_e1f4376a/lib/python3.7/site-packages/sklearn/base.py:306: UserWarning: Trying to unpickle estimator SVC from version 0.20.3 when using version 0.21.3. This might lead to breaking code or invalid results. Use at your own risk.\n",
+      "  UserWarning)\n",
+      " * Debugger is active!\n",
+      " * Debugger PIN: 130-753-966\n"
+     ]
+    }
+   ],
    "source": [
-    "!bentoml serve {saved_path}"
+    "!bentoml serve {saved_path} --with-conda"
```
parano: let's remove it here just to make sure quickstart guide runs in google colab - I will add more documentation and examples for --with-conda later on

bentoml/cli/__init__.py
leonsim: then it's difficult to handle those pip packages?

bentoml/cli/__init__.py
leonsim: Yes, I believe there must be a way to get it, but while reading this doc http://click.palletsprojects.com/en/7.x/api/ I haven't found anything useful for this...

examples/quick-start/bentoml-quick-start-guide.ipynb
leonsim: Yes sure

examples/quick-start/bentoml-quick-start-guide.ipynb
leonsim: Yes, mis-submitted this file...

bentoml/cli/__init__.py
leonsim: After a little digging into the source code, I find `click` doesn't preserve the raw parameters once parsed them. Think  we can either clone and patch `click` or submit an issue to `click`, both of which seem not easy ways.

bentoml/cli/__init__.py
parano: you can do the same thing with pip right?  `/Users/.../anaconda3/envs/{env_name...}/bin/pip install ...`

parano: actually, let's keep this.. so if the user has set a different default env dir for conda, everything still works.

parano: let's just use `conda activate` here -  do we need `conda init bash` and `eval "$(conda shell.bash hook)"` tho? it seems unnecessary if we already require the user to install conda and have `conda` command available right?

bentoml/cli/__init__.py
```diff
@@ -74,8 +86,47 @@ def bentoml_cli(ctx, verbose, quiet):
...
-    def run(ctx, api_name, archive_path=archive_path):
+    def run(ctx, api_name, archive_path=archive_path, with_conda=False):
+        if with_conda:
+            config = load_bentoml_config(archive_path)
+            metadata = config['metadata']
+            env_name = metadata['service_name'] + '_' + metadata['service_version']
+
+            yaml = YAML()
+            yaml.default_flow_style = False
+            env_path = os.path.join('/tmp', env_name + '.yaml')
```
parano: use bentoml/util/tempdir instead? or just python `tempfile` module

parano: "/tmp" may break cross-platform support

bentoml/cli/__init__.py
leonsim: The reason is `conda` has some issues when running in a subshell.
I found the solution here https://github.com/conda/conda/issues/7980#issuecomment-492784093

parano: 👍

parano:
One small thing I noticed was this subprocess call does not preserve the env variables - if user use env vars to set BentoML config, those will be lost in subprocess call. For example:

```
BENTOML__LOGGING__LOGGING_LEVEL=debug bentoml info ./examples/sklearn-sentiment-clf/model
```

leonsim:
> bentoml

This PR didn't change the `info` command, but other commands would drop env variables in subprocesses. Do you want to propagate those env variables to subprocess?

parano:
@leonsim  yea that would be great, I was just using `info` as an example, the `BENTOML__` prefix can be used to set any bentoml config: https://github.com/bentoml/BentoML/blob/master/bentoml/config/configparser.py#L47

PR #240, opened 2019-07-31
use pip_dependencies instead of conda_pip for examples

yubozhao:


parano:
LGTM!, @yubozhao  mind do the same for projects under gallery repo?

PR #246, opened 2019-08-05
Create Sagemaker deployment operator

yubozhao:


bentoml/deployment/sagemaker/__init__.py
```diff
@@ -339,17 +366,186 @@ def delete(self):
...
+
+def get_sagemaker_client(region):
+    return boto3.client('sagemaker', region)
+
+
 class SageMakerDeploymentOperator(DeploymentOperatorBase):
     def apply(self, deployment_pb):
-        # deploy code.....
+        deployment_spec = deployment_pb.spec
+        sagemaker_config = deployment_spec.deployment_operator_config.sagemaker_operator_config
```
parano: should it handle 'none' value here?

yubozhao: done

examples/quick-start/try_clipper.py
```diff
@@ -0,0 +1,85 @@
...
+import json
+
+import sys
+import os
+
+sys.path.insert(0, os.path.join(os.getcwd(), "..", ".."))
+import bentoml
+
+
+def test_predict_function(query, func, input_type):
```
parano: was this file meant for a different PR?

yubozhao: done

bentoml/deployment/sagemaker/__init__.py
```diff
@@ -339,17 +358,187 @@ def delete(self):
...
+
+        if sagemaker_config.region is not None:
+            region = sagemaker_config.region
+        else:
+            region = config.get('aws', 'default_region')
+
+        sagemaker_client = boto3.client('sagemaker', region)
+
+        repository = get_local()
+        bento = repository.get(deployment_spec.bento_name, deployment_spec.bento_version)
```
yubozhao: @parano Is this is the way I get archive from repo? or I have store available to me?

bentoml/deployment/sagemaker/__init__.py
```diff
@@ -70,6 +75,16 @@ def generate_aws_compatible_string(item):
     return re.sub(pattern, "-", item)
 
 
+def create_sagemaker_model_name(bento_name, bento_version):
+    return generate_aws_compatible_string("bentoml-" + bento_name + "-" + bento_version)
+
+
+def create_sagemaker_endpoint_config_name(bento_name, bento_version):
+    return generate_aws_compatible_string(
+        bento_name + "-" + bento_version + "-configuration"
```
leonsim: nit pick: use format to build string

bentoml/deployment/sagemaker/__init__.py
```diff
@@ -343,27 +358,216 @@ def delete(self):
...
+            'Creating',
+            'Updating',
+            'RollingBack',
+            'SystemUpdating',
+        ]:
+            service_state = DeploymentState.PENDING
+        elif endpoint_status == 'Deleting':
+            service_state = DeploymentState.INACTIVATED
+        elif endpoint_status == 'Failed':
+            service_state = DeploymentState.ERROR
```
leonsim: Should we surface errors to users for failed or undocumented status? Also feel using a dict might be better than many elif.

bentoml/deployment/sagemaker/__init__.py
```diff
@@ -343,27 +358,216 @@ def delete(self):
...
     def apply(self, deployment_pb, repo):
-        # deploy code.....
-        spec = deployment_pb.spec
-        bento_path = repo.get(spec.bento_name, spec.bento_version)
+        deployment_spec = deployment_pb.spec
+        sagemaker_config = deployment_spec.sagemaker_operator_config
+        if sagemaker_config is None:
+            raise BentoMLDeploymentException('Sagemaker configuration is missing.')
+
+        region = sagemaker_config.region or config.get('aws', 'default_region')
```
leonsim: think the second case has been handled above. do we still need it here?

bentoml/deployment/sagemaker/__init__.py
```diff
@@ -343,27 +358,216 @@ def delete(self):
...
         return ApplyDeploymentResponse(status=Status.OK(), deployment=res_deployment_pb)
 
-    def delete(self, deployment_pb, repo):
-        # delete deployment
+    def delete(self, deployment_pb, repo=None):
+        deployment_spec = deployment_pb.spec
+        sagemaker_config = deployment_spec.sagemaker_operator_config
+        if sagemaker_config is None:
+            raise BentoMLDeploymentException('Sagemaker configuration is missing.')
+        sagemaker_client = boto3.client('sagemaker', sagemaker_config.region)
```
leonsim: think we can prepare this client beforehand so we can reuse it for both delete and apply.

bentoml/deployment/sagemaker/__init__.py
yubozhao: Good suggestion

bentoml/deployment/sagemaker/__init__.py
yubozhao: na, we don't need it here.  It should be handled at CLI entry point

bentoml/deployment/sagemaker/__init__.py
yubozhao: Yeah, I feel the client is redundant as well.  I put these clients here, because each actions by operator are really separate tasks, and in a multi tenants setting, it could get messy.  So I end up putting the client a lot closer to the actions instead of some place before

bentoml/deployment/sagemaker/__init__.py
yubozhao: good suggestion on dict instead of if/elif. 
I think we could just let the undocumented status fail, it will be caught at the outer level and respond correctly.  Since we dump everything into info_json, we can properly process and surface that in the next iteration.

bentoml/deployment/sagemaker/__init__.py
```diff
@@ -343,27 +358,213 @@ def delete(self):
...
+        logger.info("AWS delete endpoint response: %s", delete_endpoint_response)
+        if delete_endpoint_response["ResponseMetadata"]["HTTPStatusCode"] == 200:
+            # We will also try to delete both model and endpoint configuration for user.
+            # Since they are not critical, even they failed, we will still count delete
+            # deployment a success action
+            model_name = create_sagemaker_model_name(
+                deployment_spec.bento_name, deployment_spec.bento_version
+            )
+            delete_model_response = sagemaker_client.delete_model(ModelName=model_name)
+            logger.info("AWS delete model response: %s", delete_model_response)
```
leonsim: I feel it's a bit noisy for users to see all the intermediate response. In the case of failure, it makes sense to show the error which is handled below, otherwise, maybe a simple sentence like _Modle is successfully deleted_ is sufficient. (Or maybe we log the response in Debug level?)

bentoml/deployment/sagemaker/__init__.py
leonsim: Got it. Think you are right for this.

bentoml/deployment/sagemaker/__init__.py
yubozhao: I think you are right on too much info.  This logs are left over from previous to work with CLI.  We should clean these up, might be another pr.

bentoml/deployment/sagemaker/__init__.py
```diff
@@ -70,6 +75,16 @@ def generate_aws_compatible_string(item):
     return re.sub(pattern, "-", item)
 
 
+def create_sagemaker_model_name(bento_name, bento_version):
+    return generate_aws_compatible_string("bentoml-" + bento_name + "-" + bento_version)
```
leonsim: can use format to build string if you agree it's better

bentoml/deployment/sagemaker/__init__.py
```diff
@@ -343,27 +358,213 @@ def delete(self):
...
+        )
+        logger.info("AWS create endpoint response: %s", create_endpoint_response)
+
+        res_deployment_pb = Deployment(state=DeploymentState())
         res_deployment_pb.CopyFrom(deployment_pb)
-        # res_deployment_pb.state = ...
+
+        # We are going to wait 10 mins for AWS to create sagemaker
+        start_time = time.time()
+        while (time.time() - start_time) < 600:
```
leonsim: nitpick: define 600 in a constant/variable.

bentoml/cli/deployment.py
```diff
@@ -354,19 +353,24 @@ def apply(
             operator=get_deployment_operator_type(platform),
         )
         if platform == 'aws_sagemaker':
+            if not api_name:
+                raise click.BadParameter(
```
parano: 👍

PR #253, opened 2019-08-06
Using versioneer for managing BentoML version

parano:
* https://github.com/warner/python-versioneer


parano:
ignoring pep8speaks's comment since versioneer.py is not our code -

PR #262, opened 2019-08-14
Yatai deployment service implementation WIP

parano:


bentoml/deployment/store.py
```diff
@@ -16,45 +16,92 @@
...
 
     id = Column(String, primary_key=True)
     name = Column(String, unique=True)
-    namespace = Column(String)
+    namespace = Column(String, nullable=False)
 
     spec = Column(JSON)
-    labels = Column(JSON)
-    annotation = Column(JSON)
+    labels = Column(JSON, nullable=False, default={})
```
yubozhao: I think labels and annotation fields are optional, correct?  If they are optional, then nullable won't need to set to be False

bentoml/deployment/store.py
```diff
@@ -16,45 +16,92 @@
...
+        labels=MessageToDict(deployment_pb.labels),
+        annotation=MessageToDict(deployment_pb.labels),
+    )
+
+
+def deployment_orm_obj_to_pb(deployment_obj):
+    return deployment_pb2.Deployment(
+        name=deployment_obj.name,
+        namespace=deployment_obj.namespace,
+        spec=ParseDict(deployment_obj.spec, deployment_pb2.DeploymentSpec),
```
yubozhao: I test this in ipython, and got 

```
AttributeError: 'google.protobuf.pyext._message.FieldProperty' object has no attribute 'SetInParent'
```
I got this error, because inside DeploymentSpec, we have additional messages like `SageMakerOperatorConfig` and others. We probably want to make sure parse those config with `ParseDict` first

bentoml/deployment/store.py
parano: Will need to test out how it works when doing filtering by label on null jsonb filed in postgres

parano: they are indeed optional, note the default value '{}'

bentoml/deployment/store.py
parano: oh thank for catching this, that was a typo, should be ```ParseDict(deployment_obj.spec, deployment_pb2.DeploymentSpec())```

Issue #264, opened 2019-08-14
Multiple images in ImageHandler


lambdaofgod:
**Is your feature request related to a problem? Please describe.**
ImageHandler doesn't support handling multiple images:

`return Response(response="Only support single file input", status=400)`

Issue #263, opened 2019-08-14
Cannot use list in ImageHandler's input_name


lambdaofgod:
**Describe the bug**
I need to create a model that compares two images - in ImageHandler documentation there is mention of input_name parameter which is supposedly string[] type, but when list argument is passed, it fails.

**To Reproduce**

```

@env(conda_pip_dependencies=['scikit-image'])
@ver(major=1, minor=0)
class ImageSimilarityScorer(BentoService):

    @api(ImageHandler, input_name=['original', 'compared'], accept_multiple_files=True)
    def predict(self, original, compared):
        return original[0, 0] == compared[0, 0]

```

BTW I also tried passing `input_name` as @api's keyword argument. Also it fails when I use tuple instead of list.

**Screenshots/Logs**
When running `bentoml serve` of my model I get
> Traceback (most recent call last):
>   File "/home/kuba/Projects/venvs/ml_server/lib/python3.6/site-packages/flask/app.py", line 2446, in wsgi_app
>     response = self.full_dispatch_request()
>   File "/home/kuba/Projects/venvs/ml_server/lib/python3.6/site-packages/flask/app.py", line 1951, in full_dispatch_request
>     rv = self.handle_user_exception(e)
>   File "/home/kuba/Projects/venvs/ml_server/lib/python3.6/site-packages/flask/app.py", line 1820, in handle_user_exception
>     reraise(exc_type, exc_value, tb)
>   File "/home/kuba/Projects/venvs/ml_server/lib/python3.6/site-packages/flask/_compat.py", line 39, in reraise
>     raise value
>   File "/home/kuba/Projects/venvs/ml_server/lib/python3.6/site-packages/flask/app.py", line 1949, in full_dispatch_request
>     rv = self.dispatch_request()
>   File "/home/kuba/Projects/venvs/ml_server/lib/python3.6/site-packages/flask/app.py", line 1935, in dispatch_request
>     return self.view_functions[rule.endpoint](**req.view_args)
>   File "/home/kuba/Projects/venvs/ml_server/lib/python3.6/site-packages/bentoml/server/bento_api_server.py", line 152, in docs_view_func
>     docs = get_docs(bento_service)
>   File "/home/kuba/Projects/venvs/ml_server/lib/python3.6/site-packages/bentoml/server/bento_api_server.py", line 142, in get_docs
>     requestBody=OrderedDict(required=True, content=api.request_schema),
[...]

PR #265, opened 2019-08-14
ImageHandler - handling multiple images

lambdaofgod:
Handling multiple images in ImageHandler

#264

How was this patch tested?
--------------------------
Since `handle_request` isn't covered in tests, I've tested simple service using curl:

```
@ver(major=1, minor=0)
class ImageSimilarityScorer(BentoService):

    @api(ImageHandler, input_name='image', accept_multiple_files=True)
    def predict(self, images):
        original = images[0]
        compared = images[1]
        return int(original[0, 0, 0] == compared[0,0, 0])
```

It returns expected value for two different random images I've got from the web

### DIFF
--- a/bentoml/handlers/image_handler.py
+++ b/bentoml/handlers/image_handler.py
@@ -138,12 +138,18 @@ class ImageHandler(BentoHandler):
                 file_names = [secure_filename(file.filename) for file in input_files]
                 for file_name in file_names:
                     check_file_format(file_name, self.accept_file_extensions)
-                input_streams = [BytesIO(input_file.read()) for input_file in input_files]
+                input_streams = [
+                    BytesIO(input_file.read()) for
+                    input_file in input_files
+                ]
             else:
                 raise ValueError(
                     "BentoML#ImageHandler unexpected HTTP request: %s" % request
                 )
-            input_data = tuple(imread(input_stream, pilmode=self.pilmode) for input_stream in input_streams)
+            input_data = tuple(
+                imread(input_stream, pilmode=self.pilmode)
+                for input_stream in input_streams
+            )
 
         result = func(input_data)
         result = get_output_str(result, request.headers.get("output", "json"))
--- a/bentoml/handlers/image_handler.py
+++ b/bentoml/handlers/image_handler.py
@@ -133,7 +133,17 @@ class ImageHandler(BentoHandler):
 
             input_data = imread(input_stream, pilmode=self.pilmode)
         else:
-            return Response(response="Only support single file input", status=400)
+            input_files = request.files.getlist(self.input_name)
+            if input_files:
+                file_names = [secure_filename(file.filename) for file in input_files]
+                for file_name in file_names:
+                    check_file_format(file_name, self.accept_file_extensions)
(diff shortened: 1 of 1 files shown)


parano:
thanks for submitting the fix @lambdaofgod - I tested it locally and it works as expected.   

A question - looking at your proposal here https://github.com/bentoml/BentoML/issues/263 I silight prefer that approach:
```
@env(conda_pip_dependencies=['scikit-image'])
@ver(major=1, minor=0)
class ImageSimilarityScorer(BentoService):

    @api(ImageHandler, input_name=['original', 'compared'], accept_multiple_files=True)
    def predict(self, original, compared):
        return original[0, 0] == compared[0, 0]
```
The main reason is I'm not sure how the order of images provided by the client will affect the result, if they are provided in an array. 

would something like the following work for your use case?
```
        else:
            input_files = [request.files.get(name) for name in self.input_name]
            if input_files:
                file_names = [secure_filename(file.filename) for file in input_files]
                for file_name in file_names:
                    check_file_format(file_name, self.accept_file_extensions)
                input_streams = [
                    BytesIO(input_file.read()) for
                    input_file in input_files
                ]
            else:
                raise ValueError(
                    "BentoML#ImageHandler unexpected HTTP request: %s" % request
                )
            input_data = [
                imread(input_stream, pilmode=self.pilmode)
                for input_stream in input_streams
            ]
            result = func(*input_data)
```

parano:
we will also need to update the `request_schema` method for OpenAPI/Swagger spec and API client to work properly, let me know if you are not familiar with OpenAPI, I can submit a follow-up PR to fix that as well

lambdaofgod:
Thanks for the swift reply.

What you suggest was indeed my first take on that problem. 

As to Swagger I haven't done anything similar, but if you point me in the right direction I could give it a try.

parano:
@lambdaofgod great! BentoML supports OpenAPI spec(it's not yet documented, sorry!) when you start an API server created with BentoML, it generates an endpoint that gives you the full API spec following the OpenAPI (or formerly Swagger) definition. Thus users can easily generate API clients in different languages that talks to BentoML model server.  You can read more about it [here](https://swagger.io/specification/)

When you start an API server locally, you can go to the index page in your browser, and it gives you a nice debugging interface that is also generated based on this OpenAPI endpoint. We use that to test out API server and it's pretty convenient:

image

In the ImageHandler file you updated, `request_schema` basically returns the OpenAPI spec of this API in JSON string. I think you will need to update the `multipart/form-data` section, currently it assumes `self.input_name` is just one string.

lambdaofgod:
@parano I've updated that. It seems like it might use some extra parameter validation, and also it was necessary to use tuples instead of lists to store `input_name`.

parano:
@lambdaofgod thanks for updating the PR, the OpenAPI related changes look great! what's the reason that tuple is necessary to store input_name? It seems to me an array could work too?

lambdaofgod:
@parano yeah, if I don't do that, I get the error like in #263

parano:
Got it, thanks again @lambdaofgod, merging this PR now! I will do some more QA and add related documentation before the next BentoML release, will let you know when it got released to PyPI.

PR #274, opened 2019-08-29
Update for operator apply

yubozhao:
Operator apply function will also handle update along with create.

Tested locally

### DIFF
--- a/bentoml/yatai/__init__.py
+++ b/bentoml/yatai/__init__.py
@@ -83,7 +83,8 @@ class YataiService(YataiServicer):
                 if deployment_orm.spec.operator != request.deployment.spec.operator:
                     return ApplyDeploymentResponse(
                         status=Status.ABORTED(
-                            'New deployment spec has different platform from existing one.'
+                            'New deployment spec has different platform from existing one. '
+                            'Please delete existing deployment and apply again'
                         )
                     )
                 previous_deployment = deployment_orm
@@ -92,10 +93,13 @@ class YataiService(YataiServicer):
                     request.deployment.name, request.deployment.namespace
                 ) as deployment:
                     deployment.spec = MessageToDict(request.deployment.spec)
+                    if request.deployment.labels:
+                        deployment.labels = dict(request.deployment.labels)
+                    if request.deployment.annotations:
+                        deployment.annotations = dict(request.deployment.annotations)
             else:
                 previous_deployment = None
-                # create or update deployment spec record
-                self.deployment_store.insert_or_update(request.deployment)
+                self.deployment_store.insert(request.deployment)
 
             # find deployment operator based on deployment spec
             operator = get_deployment_operator(request.deployment)
--- a/bentoml/yatai/__init__.py
+++ b/bentoml/yatai/__init__.py
@@ -91,10 +91,6 @@ class YataiService(YataiServicer):
                 request.deployment.state = previous_deployment.state
 
                 request.deployment.state.state = DeploymentState.PENDING
-                if not request.deployment.labels:
-                    request.deployment.labels = previous_deployment.labels
-                if not request.deployment.annotations:
-                    request.deployment.annotations = previous_deployment.annotations
 
             self.deployment_store.insert_or_update(request.deployment)
             # find deployment operator based on deployment spec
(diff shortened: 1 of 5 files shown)


bentoml/yatai/__init__.py
```diff
@@ -75,16 +75,35 @@ def ApplyDeployment(self, request, context=None):
...
-            # TODO: validate deployment config
+            deployment_orm = self.deployment_store.get(
+                request.deployment.name, request.deployment.namespace
+            )
+            if deployment_orm:
+                # check deployment platform
+                if deployment_orm.spec.operator != request.deployment.spec.operator:
+                    return ApplyDeploymentResponse(
+                        status=Status.ABORTED(
+                            'New deployment spec has different platform from existing one.'
```
parano: ++ 'try deleting existing deployment and apply again'

bentoml/yatai/__init__.py
```diff
@@ -75,16 +75,35 @@ def ApplyDeployment(self, request, context=None):
...
-            # create or update deployment spec record
-            self.deployment_store.insert_or_update(request.deployment)
+                with self.deployment_store.update_deployment(
+                    request.deployment.name, request.deployment.namespace
+                ) as deployment:
+                    deployment.spec = MessageToDict(request.deployment.spec)
+            else:
+                previous_deployment = None
+                # create or update deployment spec record
+                self.deployment_store.insert_or_update(request.deployment)
```
parano: this should be out of the if/else?

bentoml/yatai/__init__.py
```diff
@@ -75,16 +75,35 @@ def ApplyDeployment(self, request, context=None):
...
+                    return ApplyDeploymentResponse(
+                        status=Status.ABORTED(
+                            'New deployment spec has different platform from existing one.'
+                        )
+                    )
+                previous_deployment = deployment_orm
 
-            # create or update deployment spec record
-            self.deployment_store.insert_or_update(request.deployment)
+                with self.deployment_store.update_deployment(
```
parano: prefer to use insert_or_update - annotation & labels can be updated in this call as well

yubozhao: I will do more checks here and add information if it is available.  I end up with using update because I don’t want override state

bentoml/yatai/__init__.py
yubozhao: This should be just insert.

bentoml/yatai/__init__.py
```diff
@@ -75,16 +75,39 @@ def ApplyDeployment(self, request, context=None):
...
             )
 
-            # TODO: validate deployment config
-
-            # create or update deployment spec record
-            self.deployment_store.insert_or_update(request.deployment)
+            deployment_orm = self.deployment_store.get(
+                request.deployment.name, request.deployment.namespace
+            )
+            if deployment_orm:
```
parano: this naming is confusing - this is not an orm object but a protobuf object, simply use:

```
previous_deployment = self.deployment_store.get(...)
```

bentoml/yatai/__init__.py
parano: just do `request.deployment.state = ...` and `insert_or_update` at the end

bentoml/yatai/__init__.py
```diff
@@ -75,16 +76,34 @@ def ApplyDeployment(self, request, context=None):
...
+                            'New deployment spec has different platform from existing one. '
+                            'Please delete existing deployment and apply again'
+                        )
+                    )
+                request.deployment.state = previous_deployment.state
 
-            # create or update deployment spec record
-            self.deployment_store.insert_or_update(request.deployment)
+                request.deployment.state.state = DeploymentState.PENDING
+                if not request.deployment.labels:
```
parano: l94-l97 are probably not necessary - how can a user remove all labels if they want to?

yubozhao: Right now, there isn't way to remove labels yet.  Maybe that's something web could provide as a feature?

I am thinking persevering the old labels might be more important than removing extra labels for now, that's why i add it.

The scenario I keep come back to, is when they update the deployment with new bento tag (maybe the most common tasks compare to others).  I don't want the labels get wiped out, or I don't want they have to re-enter labels everytime, when they apply new bento tag.

PR #275, opened 2019-08-29
Add additional tracker events

yubozhao:


bentoml/archive/loader.py
```diff
@@ -21,7 +21,7 @@
 import tempfile
 import logging
 
-from bentoml.utils.usage_stats import track_loading
+from bentoml.utils.usage_stats import track_loading, track_load_start
```
parano: suggest rename: track_loading => track_load_finish, no need to change the actual tag in amptitude

bentoml/server/bento_api_server.py
```diff
@@ -289,4 +290,6 @@ def start(self):
         """
         Start an REST server at the specific port on the instance or parameter.
         """
+        track_server('flask')
```
parano: this event will be duplicated when running gunicorn server right?

bentoml/utils/usage_stats.py
```diff
@@ -86,6 +87,10 @@ def track_save(bento_service):
         return track("save", info)
 
 
+def track_load_start():
+    return track('load-start', {})
```
parano: let's move the logic of checking `if config['core'].getboolean("usage_tracking"):`  to track function above :)

parano: otherwise it's easy to forget

bentoml/utils/usage_stats.py
```diff
@@ -100,6 +105,19 @@ def track_cli(command, deploy_platform=None):
         return track('cli-' + command, info)
 
 
+def track_server(server_type):
+    start_time = time.time()
+    import atexit
```
parano: import at the top of the file

bentoml/server/bento_api_server.py
parano: anyway to workaround?

yubozhao: It is duplicated in each workers for gunicorn. I thought about putting it at server initialization. However I really want uptime for it, so it ends up on each worker. 

It’s is bad send the message multiple times. While I think it is kinda neat data point for how many workers

yubozhao: Open for suggestions though

bentoml/utils/usage_stats.py
yubozhao: Yep. Good catch

bentoml/archive/loader.py
```diff
@@ -21,7 +21,7 @@
 import tempfile
 import logging
 
-from bentoml.utils.usage_stats import track_loading
+from bentoml.utils.usage_stats import track_loading_finish, track_load_start
```
parano: nit: s/track_loading_finish/track_load_finish/ just to be consistent

bentoml/server/gunicorn_server.py
```diff
@@ -82,3 +83,7 @@ def load(self):
         bento_service = load(self.bento_archive_path)
         api_server = BentoAPIServer(bento_service, port=self.port)
         return api_server.app
+
+    def run(self):
+        track_server('gunicorn')
```
parano: just add an attribute "num_of_workers" here in the amplitude message, you can access self. num_of_workers

bentoml/server/bento_sagemaker_server.py
```diff
@@ -61,4 +62,5 @@ def __init__(self, bento_service, api_name, app_name=None):
         setup_routes(self.app, self.bento_service, api_name)
 
     def start(self):
+        track_server('sagemaker')
```
parano: notice the typo in comment above, "..an AWS sagemaker compatibility *REST* server."

PR #280, opened 2019-09-06
Separate create and apply cli command.

yubozhao:
Separate create and apply cli command.  Now create takes options while apply takes in yaml file


bentoml/cli/deployment.py
```diff
@@ -250,20 +246,60 @@ def apply(
...
                     namespace=namespace,
+                    message='Creating deployment...',
+                )
+                result.deployment.state.CopyFrom(result_state.state)
+
+            _echo('Finished create deployment {}'.format(name), CLI_COLOR_SUCCESS)
+            display_deployment_info(result.deployment, output)
+
+    @deploy.command(help='Apply model service deployment from yaml file')
+    @click.option("--file", callback=parse_yaml_file_or_string_callback)
```
parano:
```
@click.option("--file", "yaml_content", callback=...)
def apply(yaml_content, output, wait):
```

bentoml/cli/deployment.py
```diff
@@ -250,20 +246,60 @@ def apply(
...
+    @click.option('--output', type=click.Choice(['json', 'yaml']), default='json')
+    @click.option(
+        '--wait/--no-wait',
+        default=True,
+        help='Wait for apply action to complete or encounter an error.'
+        'If set to no-wait, CLI will return immediately. The default value is wait',
+    )
+    def apply(file, output, wait):
+        yaml_content = file
```
parano: s/yaml_content/deployment_yaml ?

bentoml/cli/deployment.py
parano: use click's `type=click.File`, which also handle "string" input properly, most unix shell program refer "-" to reading from stdin.  also you can remove the "read from string" case in the callback function

bentoml/cli/utils.py
```diff
@@ -0,0 +1,93 @@
...
+        spec.bento_name = spec_yaml['bento_name']
+        spec.bento_version = spec_yaml['bento_version']
+        return Deployment(
+            name=deployment_yaml['name'],
+            namespace=deployment_yaml['namespace'],
+            annotations=deployment_yaml.get('annotations'),
+            labels=deployment_yaml.get('labels'),
+            spec=spec,
+        )
+    except KeyError as e:
```
parano: let's use a separate function to validate input yaml instead of this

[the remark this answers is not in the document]
parano: similarly, labels, annotations are optional too

bentoml/cli/deployment.py
```diff
@@ -249,21 +256,75 @@ def apply(
...
+    )
+    @click.option('-o', '--output', type=click.Choice(['json', 'yaml']), default='json')
+    @click.option(
+        '--wait/--no-wait',
+        default=True,
+        help='Wait for apply action to complete or encounter an error.'
+        'If set to no-wait, CLI will return immediately. The default value is wait',
+    )
+    def apply(deployment_yaml, output, wait):
+        print('deployment yaml', deployment_yaml)
```
parano: rm

PR #287, opened 2019-09-13
BentoRepository implementation part 2

parano:


protos/repository.proto
```diff
@@ -36,27 +38,55 @@ message BentoMetadata {
...
   }
 
-  BentoUri uri = 1;
-  string created_at = 2;
-  string name = 3;
-  string version = 4;
-  BentoServiceEnv env = 5;
-  repeated BentoArtifact artifacts = 6;
-  repeated BentoServiceApi apis = 7;
+  string name = 1;
```
yubozhao: What's the different between name/version inside `BentoServiceMetadata` and `Bento`

protos/repository.proto
```diff
@@ -36,27 +38,55 @@ message BentoMetadata {
...
+    UNINITIALIZED = 0; // default
+    UPLOADING = 1;
+    DONE = 2;
+    ERROR = 3;
+    TIMEOUT = 4;
+  }
+  Status status = 1;
+  google.protobuf.Timestamp updated_at = 2;
+  int32 percentage = 3; // for uploading completion percentage
+  string message = 4; // for error message
```
yubozhao: should we just rename this to error_message instead, if only for error message?

bentoml/archive/archiver.py
```diff
@@ -61,20 +62,27 @@ def _get_artifacts_list(bento_service):
...
 
+    from bentoml.service import BentoService
+
+    if not isinstance(bento_service, BentoService):
+        raise BentoMLException(
+            "save_to_dir only work with instance of custom BentoService class"
+        )
+
+    if version is not None:
+        bento_service.set_version(version)
```
yubozhao: What happen if the version is None?  I don't think I find where we generate random version yet

parano: in BentoService#version

bentoml/repository/metadata_store.py
```diff
@@ -0,0 +1,187 @@
...
+    )
+    return BentoPB(
+        name=bento_obj.name,
+        version=bento_obj.version,
+        uri=bento_uri,
+        bento_service_metadata=bento_service_metadata_pb,
+    )
+
+
+class BentoMetadataStore(object):
```
yubozhao: Maybe because I was just reading protobuf stuff on Bento and BentoMetadata in repository.proto, the naming `BentoMetadataStore` is bit confusing.  Maybe to `BentoStore`?  I am not exactly sure.

bentoml/yatai/yatai_service_impl.py
```diff
@@ -0,0 +1,294 @@
...
+    def ApplyDeployment(self, request, context=None):
+        try:
+            # apply default namespace if not set
+            request.deployment.namespace = (
+                request.deployment.namespace or self.default_namespace
+            )
+
+            previous_deployment = self.deployment_store.get(
+                request.deployment.name, request.deployment.namespace
+            )
```
yubozhao: TODO: We will need to add validation back for apply deployment

bentoml/repository/metadata_store.py
parano: probably gonna merge it with the BentoRepository concept in follow up PR

Issue #283, opened 2019-09-09
Preserve request image files in logs directory


parano:
**Is your feature request related to a problem? Please describe.**
Currently, we don't preserve request image files, https://github.com/bentoml/BentoML/blob/master/bentoml/server/bento_api_server.py#L63

**Describe the solution you'd like**
Generate a UUID to every image received by api server, log UUID to prediction logs and save image file to logs/images directory and named by the generated UUID

parano:
done in #288

PR #288, opened 2019-09-14
Preserve request image files in logs directory

leonsim:

#283 

How was this patch tested?
--------------------------

### DIFF
--- a/bentoml/server/bento_api_server.py
+++ b/bentoml/server/bento_api_server.py
@@ -200,9 +200,9 @@ def bento_service_api_wrapper(api, service_name, service_version):
         all_paths = []
 
         if request.content_type.startswith(img_prefix):
-            filename = '{request_id}-{timestamp}.{ext}'.format(
-                request_id=request_id,
+            filename = '{timestamp}-{request_id}.{ext}'.format(
                 timestamp=int(time.time()),
+                request_id=request_id,
                 ext=request.content_type[len(img_prefix) :],
             )
             path = os.path.join(log_folder, filename)
@@ -214,9 +214,9 @@ def bento_service_api_wrapper(api, service_name, service_version):
             file = request.files[name]
             if file and file.filename:
                 orig_filename = secure_filename(file.filename)
-                filename = '{request_id}-{timestamp}-{orig_filename}'.format(
-                    request_id=request_id,
+                filename = '{timestamp}-{request_id}-{orig_filename}'.format(
                     timestamp=int(time.time()),
+                    request_id=request_id,
                     orig_filename=orig_filename,
                 )
                 path = os.path.join(log_folder, filename)
@@ -231,14 +231,14 @@ def bento_service_api_wrapper(api, service_name, service_version):
             # Assume there is not a strong use case for idempotency check here.
             # Will revise later if we find a case.
 
-            image_paths = None
+            image_paths = []
             if not config('logging').getboolean('disable_logging_image'):
                 image_paths = log_image(request, request_id)
 
             response = api.handle_request(request)
 
             request_log = {
-                "uuid": request_id,
+                "request_id": request_id,
                 "service_name": service_name,
                 "service_version": service_version,
                 "api": api.name,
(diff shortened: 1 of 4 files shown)


bentoml/server/bento_api_server.py
```diff
@@ -195,21 +194,54 @@ def bento_service_api_wrapper(api, service_name, service_version):
...
-                        "response": response.response,
-                    }
-                )
+
+            request_log = {
+                "uuid": request_id,
+                "service_name": service_name,
+                "service_version": service_version,
+                "api": api.name,
+                "request": _request_to_json(request),
```
parano: in the case of image request, can we add the path to saved files here?

parano:
looks like we also want to turn off this feature in tests environment, I did something similar in this PR here https://github.com/bentoml/BentoML/pull/287

bentoml/server/bento_api_server.py
```diff
@@ -195,21 +194,54 @@ def bento_service_api_wrapper(api, service_name, service_version):
     summary_name = str(service_name) + "_" + str(api.name)
     request_metric_time = Summary(summary_name, summary_name + " request latency")
 
+    def log_image(request, request_id):
+        img_prefix = 'image/'
+        log_folder = config['logging'].get('base_log_folder')
```
parano: another change after https://github.com/bentoml/BentoML/pull/287 is to use config('logging').get('base_log_folder')

leonsim: Ok, I'll rebase and update.

bentoml/server/bento_api_server.py
leonsim: good idea. will do.

bentoml/server/bento_api_server.py
```diff
@@ -194,21 +193,66 @@ def bento_service_api_wrapper(api, service_name, service_version):
...
+            request_log = {
+                "uuid": request_id,
+                "service_name": service_name,
+                "service_version": service_version,
+                "api": api.name,
+                "request": _request_to_json(request),
+                "response_code": response.status_code,
+            }
+
+            if image_paths is not None:
```
parano: image_paths is [] for normal json request right?

bentoml/server/bento_api_server.py
```diff
@@ -194,21 +193,66 @@ def bento_service_api_wrapper(api, service_name, service_version):
...
+            path = os.path.join(log_folder, filename)
+            all_paths.append(path)
+            with open(path, 'w') as f:
+                f.write(request.get_data())
+
+        for name in request.files:
+            file = request.files[name]
+            if file and file.filename:
+                orig_filename = secure_filename(file.filename)
+                filename = '{request_id}-{timestamp}-{orig_filename}'.format(
```
parano: consider '{timestamp}-{request_id}-{orig_filename}'? so when doing manual debugging, image files will be sorted nicely by time?

bentoml/server/bento_api_server.py
```diff
@@ -194,21 +193,66 @@ def bento_service_api_wrapper(api, service_name, service_version):
...
-                        "service_name": service_name,
-                        "service_version": service_version,
-                        "api": api.name,
-                        "request": _request_to_json(request),
-                        "response": response.response,
-                    }
-                )
+
+            request_log = {
+                "uuid": request_id,
```
parano: could you help rename this field to just request_id?  `"request_id": request_id`

bentoml/server/bento_api_server.py
```diff
@@ -59,12 +62,8 @@ def _request_to_json(req):
     """
     if req.content_type == "application/json":
         return req.get_json()
-    elif "image" in req.content_type:
-        return {"data": "dont handle"}
-    elif "video" in req.content_type:
-        return {"data": "dont handle"}
 
-    return {"data": req.get_data().decode("utf-8")}
+    return {"data": "ignore content type other than JSON."}
```
parano: just `return {}`? 

for a image handler api server, we don't want to pollute the logs with this message `{"data": "ignore content type other than JSON."}` right?

leonsim: Yes, since we have output the file paths, this won't be needed.

PR #306, opened 2019-09-19
use in-development bentoml package for deployment

yubozhao:
If `bentoml_deploy_version` is set in bentoml.cfg. When archiving service, it will use that version instead of the installed bentoml version.

When deploying to the cloud, if Bentoml is in editor mode(BentoML module is installed with `pip install -e`), we will generate tar.gz file of the BentoML module and placed it in the archive directory for deploying.  Deployed service will have the targz file's BentoML installed

### DIFF
--- a/bentoml/archive/templates.py
+++ b/bentoml/archive/templates.py
@@ -91,8 +91,8 @@ WORKDIR /bento
 RUN conda env update -n base -f /bento/environment.yml
 RUN pip install -r /bento/requirements.txt
 
-# Install wheel files
-RUN if [ -f /bento/install_bundled_dependencies.sh ] then /bin/bash -c /bento/install_bundled_dependencies.sh; fi
+# Install additional dependencies bundle
+RUN if [ -f /bento/install_bundled_dependencies.sh ]; then /bin/bash -c /bento/install_bundled_dependencies.sh; fi
 
 # run user defined setup script
 RUN if [ -f /bento/setup.sh ]; then /bin/bash -c /bento/setup.sh; fi
@@ -125,8 +125,8 @@ WORKDIR /opt/program
 RUN conda env update -n base -f /opt/program/environment.yml
 RUN pip install -r /opt/program/requirements.txt
 
-# Install wheel files
-RUN if [ -f /bento/install_bundled_dependencies.sh ] then /bin/bash -c /bento/install_bundled_dependencies.sh; fi
+# Install additional dependencies bundle
+RUN if [ -f /bento/install_bundled_dependencies.sh ]; then /bin/bash -c /bento/install_bundled_dependencies.sh; fi
 
 # run user defined setup script
 RUN if [ -f /opt/program/setup.sh ]; then /bin/bash -c /opt/program/setup.sh; fi
--- a/bentoml/archive/templates.py
+++ b/bentoml/archive/templates.py
@@ -91,6 +91,9 @@ WORKDIR /bento
 RUN conda env update -n base -f /bento/environment.yml
 RUN pip install -r /bento/requirements.txt
 
+# Install wheel files
+RUN if [ -f /bento/install_wheels.sh ] then /bin/bash -c /bento/install_wheels.sh; fi
+
 # run user defined setup script
 RUN if [ -f /bento/setup.sh ]; then /bin/bash -c /bento/setup.sh; fi
 
@@ -121,6 +124,9 @@ WORKDIR /opt/program
 # update conda base env
 RUN conda env update -n base -f /opt/program/environment.yml
 RUN pip install -r /opt/program/requirements.txt
+
+# Install wheel files
+RUN if [ -f /bento/install_wheels.sh ] then /bin/bash -c /bento/install_wheels.sh; fi
 
 # run user defined setup script
 RUN if [ -f /opt/program/setup.sh ]; then /bin/bash -c /opt/program/setup.sh; fi
--- a/bentoml/archive/templates.py
+++ b/bentoml/archive/templates.py
@@ -92,7 +92,7 @@ RUN conda env update -n base -f /bento/environment.yml
 RUN pip install -r /bento/requirements.txt
 
 # Install wheel files
-RUN if [ -f /bento/install_wheels.sh ] then /bin/bash -c /bento/install_wheels.sh; fi
+RUN if [ -f /bento/install_wheels.sh ] then /bin/bash -c /bento/install_bundled_dependencies.sh; fi
 
 # run user defined setup script
 RUN if [ -f /bento/setup.sh ]; then /bin/bash -c /bento/setup.sh; fi
--- a/bentoml/archive/templates.py
+++ b/bentoml/archive/templates.py
@@ -125,7 +125,7 @@ WORKDIR /opt/program
 RUN conda env update -n base -f /opt/program/environment.yml
 RUN pip install -r /opt/program/requirements.txt
 
-# Install additional dependencies bundle
+# Install additional pip dependencies inside bundled_pip_dependencies dir
 RUN if [ -f /bento/bentoml_init.sh ]; then /bin/bash -c /bento/bentoml_init.sh; fi
 
 # run user defined setup script
@@ -162,7 +162,7 @@ __all__ = ['__version__', '{service_name}', 'load']
 BENTO_INIT_SH_TEMPLATE = """\
 #!/bin/bash
 
-for filename in ./bundled_dependencies/*.tar.gz; do
+for filename in ./bundled_pip_dependencies/*.tar.gz; do
     [ -e "$filename" ] || continue
     pip install "$filename" --ignore-installed
 done
--- a/bentoml/archive/templates.py
+++ b/bentoml/archive/templates.py
@@ -92,7 +92,7 @@ RUN conda env update -n base -f /bento/environment.yml
 RUN pip install -r /bento/requirements.txt
 
 # Install additional dependencies bundle
-RUN if [ -f /bento/install_bundled_dependencies.sh ]; then /bin/bash -c /bento/install_bundled_dependencies.sh; fi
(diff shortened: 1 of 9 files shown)


bentoml/utils/__init__.py
```diff
@@ -75,3 +79,21 @@ def ProtoMessageToDict(protobuf_msg, **kwargs):
...
+
+
+def _is_pypi_release():
+    is_installed_package = hasattr(version_mod, 'version_json')
+    is_tagged = not BENTOML_VERSION.startswith('0+untagged')
+    is_clean = not version_mod.get_versions()['dirty']
+    return is_installed_package and is_tagged and is_clean
+
+
+def _is_bentoml_in_editor_mode():
```
parano: _is_bentoml_in_develop_mode


add a comment: `this is for BentoML developer to create BentoService containing custom develop branches of BentoML library, it gets triggered when BentoML module is installed via "pip install --editable ."`

bentoml/utils/__init__.py
```diff
@@ -75,3 +79,28 @@ def ProtoMessageToDict(protobuf_msg, **kwargs):
...
+def _is_bentoml_in_develop_mode():
+    try:
+        module_location, = importlib.util.find_spec('bentoml').submodule_search_locations
+    except AttributeError:
+        # python 2.7 doesn't have importlib.util, will fall back to imp instead
+        import imp
+        _, module_location, _ = imp.find_module('bentoml')
+
+    setup_py_path = os.path.abspath(os.path.join(module_location, '..', 'setup.py'))
+    if not _is_pypi_release() and os.path.isfile(setup_py_path):
```
parano: ```return not _is_pypi_release() and os.path.isfile(setup_py_path)```

bentoml/deployment/utils.py
```diff
@@ -44,3 +52,77 @@ def process_docker_api_line(payload):
...
+    try:
+        module_location, = importlib.util.find_spec('bentoml').submodule_search_locations
+    except AttributeError:
+        # python 2.7 doesn't have importlib.util, will fall back to imp instead
+        import imp
+        _, module_location, _ = imp.find_module('bentoml')
+    bentoml_setup_py = os.path.abspath(os.path.join(module_location, '..', 'setup.py'))
+
+    if not os.path.isfile(bentoml_setup_py):
+        raise KeyError('"setup.py" for Bentoml module not found')
```
parano: keyError not suitable for this case,  use `assert os.path.isfile(bentoml_setup_py), '"setup.py" for Bentoml module not found'`

bentoml/deployment/utils.py
```diff
@@ -44,3 +52,77 @@ def process_docker_api_line(payload):
...
+
+    # Create random directory inside bentoml module for storing the bundled
+    # targz file. Since dist-dir can only be inside of the module directory
+    date_string = datetime.now().strftime("%Y_%m_%d")
+    bundle_dir_name = '__bento_dev_{}'.format(date_string)
```
parano: since the directory will get overwritten if re-ran in the same day, let's just use "__bento_dev_tmp", and remove the directory if it exists to avoid potential issues with files that are not cleaned up properly

bentoml/deployment/utils.py
```diff
@@ -44,3 +52,77 @@ def process_docker_api_line(payload):
...
+
+    sandbox.run_setup(
+        bentoml_setup_py, ['sdist', '--format', 'gztar', '--dist-dir', bundle_dir_name]
+    )
+
+    # copy the generated targz to archive directory and remove it from
+    # bentoml module directory
+    source_dir = os.path.abspath(os.path.join(module_location, '..', bundle_dir_name))
+    dest_dir = os.path.join(archive_path, 'bundled_dependencies')
+    if os.path.isdir(source_dir):
```
parano: is this `if` necessary? in what case it is not a directory and what should we do to handle that case?

bentoml/deployment/utils.py
```diff
@@ -44,3 +52,77 @@ def process_docker_api_line(payload):
...
+        shutil.copytree(source_dir, dest_dir)
+        shutil.rmtree(source_dir)
+
+    # Include script for install targz file in archive directory
+    install_script_path = os.path.join(archive_path, 'install_bundled_dependencies.sh')
```
parano: let's move this part to `save_to_dir` method and make the script always saved in a Bento directory along with the dockerfile -- so, later on, we can add an API allow users to add their custom python libraries this way as well

and let's rename it to `bentoml_init.sh` so we need to add other bentoML internal staff that's required when initializing a container

bentoml/deployment/utils.py
```diff
@@ -44,3 +52,77 @@ def process_docker_api_line(payload):
...
+    with open(install_script_path, 'w') as f:
+        f.write(INSTALL_TARGZ_TEMPLATE)
+    permission = "755"
+    octal_permission = int(permission, 8)
+    os.chmod(install_script_path, octal_permission)
```
parano: is this necessary?

bentoml/deployment/utils.py
```diff
@@ -44,3 +52,77 @@ def process_docker_api_line(payload):
...
+
+    # remove the 'dirty' and 'hash' tags for bentoml in the environment.yml
```
parano: discussed offline, remove this part

bentoml/deployment/serverless/serverless_utils.py
```diff
@@ -129,10 +129,27 @@ def generate(self):
...
+            bundle_dependencies_path = os.path.join(
+                self.archive_path, 'bundle_dependencies'
+            )
+            shutil.copytree(bundle_dependencies_path, dest_bundle_path)
+            requirement_txt_path = os.path.join(tempdir, 'requirements.txt')
+
+            with open(requirement_txt_path, 'r+') as file:
+                content = file.read()
+                requirement_list = content.split('\n')
+                bundle_files = os.listdir(dest_bundle_path)
```
parano: for loop

bentoml/deployment/utils.py
```diff
@@ -44,3 +50,36 @@ def process_docker_api_line(payload):
...
+    )
+
+    # copy the generated targz to archive directory and remove it from
+    # bentoml module directory
+    source_dir = os.path.abspath(os.path.join(module_location, '..', bundle_dir_name))
+    dest_dir = os.path.join(archive_path, 'bundled_dependencies')
+    shutil.copytree(source_dir, dest_dir)
+    shutil.rmtree(source_dir)
+
+    return
```
parano: remove

bentoml/deployment/utils.py
```diff
@@ -44,3 +50,36 @@ def process_docker_api_line(payload):
...
+    except AttributeError:
+        # python 2.7 doesn't have importlib.util, will fall back to imp instead
+        import imp
+
+        _, module_location, _ = imp.find_module('bentoml')
+    bentoml_setup_py = os.path.abspath(os.path.join(module_location, '..', 'setup.py'))
+
+    assert os.path.isfile(bentoml_setup_py), '"setup.py" for Bentoml module not found'
+
+    # Create random directory inside bentoml module for storing the bundled
```
parano: update comment

bentoml/archive/templates.py
```diff
@@ -122,6 +125,9 @@ def _parse_requirements(file_path):
 RUN conda env update -n base -f /opt/program/environment.yml
 RUN pip install -r /opt/program/requirements.txt
 
+# Install additional dependencies bundle
```
parano: update comment, same above

bentoml/deployment/utils.py
```diff
@@ -44,3 +50,36 @@ def process_docker_api_line(payload):
...
+    # targz file. Since dist-dir can only be inside of the module directory
+    bundle_dir_name = '__bento_dev_tmp'
+    Path(bundle_dir_name).mkdir(exist_ok=True, parents=True)
```
parano: remove directory bundle_dir_name first, then just call os.mkdir

bentoml/archive/templates.py
```diff
@@ -151,3 +157,13 @@ def load():
 
 __all__ = ['__version__', '{service_name}', 'load']
 """
+
+
+BENTO_INIT_SH_TEMPLATE = """\
+#!/bin/bash
+
+for filename in ./bundled_dependencies/*.tar.gz; do
```
parano: suggest rename bundled_pip_dependencies

bentoml/deployment/serverless/serverless_utils.py
```diff
@@ -129,10 +129,27 @@ def generate(self):
...
         )
-        shutil.copy(os.path.join(self.archive_path, "requirements.txt"), tempdir)
+        requirement_txt_path = os.path.join(self.archive_path, 'requirements.txt')
+        shutil.copy(requirement_txt_path, tempdir)
         model_serivce_archive_path = os.path.join(tempdir, self.bento_name)
         model_path = os.path.join(self.archive_path, self.bento_name)
         shutil.copytree(model_path, model_serivce_archive_path)
+
+        if _is_bentoml_in_develop_mode():
+            dest_bundle_path = os.path.join(tempdir, 'bundle_dependencies')
```
parano: let's make the naming consistent, bundled_pip_dependencies

bentoml/deployment/serverless/serverless_utils.py
```diff
@@ -129,10 +129,27 @@ def generate(self):
             ],
             tempdir,
         )
-        shutil.copy(os.path.join(self.archive_path, "requirements.txt"), tempdir)
+        requirement_txt_path = os.path.join(self.archive_path, 'requirements.txt')
+        shutil.copy(requirement_txt_path, tempdir)
         model_serivce_archive_path = os.path.join(tempdir, self.bento_name)
```
parano: not related to this PR, suggest update naming - s/model_serivce_archive_path/bento_archive_path

PR #318, opened 2019-09-30
Make BentoML deploy version configurable

parano:
* Adding bentoml_deploy_version config to allow customizing bentoml module that being loaded when distributing a saved BentoService as PyPI, docker image or when deployed to AWS lambda/Sagemaker

* Use last PyPI release version as the target BentoML version when saving BentoService with a dev branch


parano:
ignoring the pep8speaks error above, since the configuring logging call must be above other imports

PR #321, opened 2019-10-01
Enhance serverless deployment

yubozhao:

How was this patch tested?
--------------------------

### DIFF
--- a/bentoml/deployment/serverless/aws_lambda.py
+++ b/bentoml/deployment/serverless/aws_lambda.py
@@ -26,7 +26,7 @@ from bentoml.utils import Path
 from bentoml.utils.whichcraft import which
 from bentoml.deployment.operator import DeploymentOperatorBase
 from bentoml.yatai.status import Status
-from bentoml.exceptions import BentoMLException
+from bentoml.exceptions import BentoMLException, BentoMLMissingDepdencyException
 from bentoml.proto.deployment_pb2 import (
     Deployment,
     ApplyDeploymentResponse,
@@ -133,7 +133,7 @@ def generate_handler_py(bento_name, apis, output_path):
 class AwsLambdaDeploymentOperator(DeploymentOperatorBase):
     def apply(self, deployment_pb, repo, prev_deployment=None):
         if which('docker') is None:
-            raise ValueError(
+            raise BentoMLMissingDepdencyException(
                 'Docker is not installed. Please visit '
                 'www.docker.com for instructions'
             )
--- a/bentoml/deployment/serverless/aws_lambda.py
+++ b/bentoml/deployment/serverless/aws_lambda.py
@@ -39,7 +39,7 @@ from bentoml.deployment.serverless.serverless_utils import (
     TemporaryServerlessContent,
     TemporaryServerlessConfig,
     parse_serverless_info_response_to_json_string,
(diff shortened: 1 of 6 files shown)


bentoml/deployment/serverless/aws_lambda.py
```diff
@@ -131,6 +132,11 @@ def generate_handler_py(bento_name, apis, output_path):
 
 class AwsLambdaDeploymentOperator(DeploymentOperatorBase):
     def apply(self, deployment_pb, repo, prev_deployment=None):
+        if which('docker') is None:
+            raise BentoMLMissingDepdencyException(
```
parano: this only test if docker client is installed, it may fail to connect to docker daemon, or docker daemon has not been started yet, let's add a common method for checking if docker is available?

yubozhao: yeah, lets do that

parano: one potential way to do that is run 'docker info' and parse the output

bentoml/deployment/serverless/serverless_utils.py
```diff
@@ -24,47 +24,42 @@
...
-    if parsed_version >= version.parse(MINIMUM_SERVERLESS_VERSION):
-        return
-    else:
-        raise ValueError(
-            "Incompatiable serverless version, please install version 1.40.0 or greater"
-        )
+
+def install_serverless_package():
+    if not os.path.isfile(SERVERLESS_BIN_COMMAND):
+        if which('npm') is None:
```
parano: can we check the minimum 'npm' and 'node' version required?

bentoml/deployment/serverless/serverless_utils.py
```diff
@@ -24,47 +24,42 @@
...
-
-    if parsed_version >= version.parse(MINIMUM_SERVERLESS_VERSION):
-        return
-    else:
-        raise ValueError(
-            "Incompatiable serverless version, please install version 1.40.0 or greater"
-        )
+
+def install_serverless_package():
+    if not os.path.isfile(SERVERLESS_BIN_COMMAND):
```
parano: We should still run install right? let's say in the next version of BentoML we want to use a newer version of serverless, this will prevent users from updating to new serverless

parano: I think if it's already installed the same version, NPM will skip.. I might be wrong tho

bentoml/deployment/serverless/serverless_utils.py
```diff
@@ -24,47 +24,42 @@
 from subprocess import PIPE
 
 from ruamel.yaml import YAML
-from packaging import version
 
+from bentoml.configuration import BENTOML_HOME
```
parano: use `from bentoml.configuration import _get_bentoml_home`, in case BENTOML_HOME got reset after loading BentoML module

bentoml/deployment/serverless/aws_lambda.py
yubozhao: yeah, I was reading their documentation, that's the recommend way to test.  Also, I tested locally.  Using subprocess.check_output with `docker info`  if Docker is not running, it will throw exception

PR #326, opened 2019-10-05
Remove unnecessary check for S3 client initialisation and implement feature: get pre-signed S3 URL.

leonsim:

How was this patch tested?
--------------------------

### DIFF
--- a/bentoml/repository/__init__.py
+++ b/bentoml/repository/__init__.py
@@ -118,24 +118,13 @@ class _LocalBentoRepository(BentoRepositoryBase):
 
 class _S3BentoRepository(BentoRepositoryBase):
     def __init__(self, base_url):
-        # Ensure bucket exist and server has permission to manage files under base_path
-        self.base_url = base_url
         self.uri_type = BentoUri.S3
 
         parse_result = urlparse(base_url)
-        bucket = parse_result.netloc
-        base_path = parse_result.path
+        self.bucket = parse_result.netloc
+        self.base_path = parse_result.path
 
-        s3_client = boto3.client("s3")
-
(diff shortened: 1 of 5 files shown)


bentoml/repository/__init__.py
```diff
@@ -118,28 +118,34 @@ def dangerously_delete(self, bento_name, bento_version):
...
+        self.base_path = parse_result.path
 
-        s3_client = boto3.client("s3")
+        self.s3_client = boto3.client("s3")
+
+    def add(self, bento_name, bento_version):
+        # Generate pre-signed s3 path for upload
+        expiration = 3600
+        object_name = os.path.join(
+            self.base_path, '{}-{}'.format(bento_name, bento_version)
```
parano: let's keep the name consistent with our local repo? ```object_name = ",".join([self.base_path, bento_name, bento_version])```  prefer this way over `os.path.join` because it may break things if we were to run the Yatai server on a windows server.

bentoml/repository/__init__.py
```diff
@@ -118,28 +118,34 @@ def dangerously_delete(self, bento_name, bento_version):
...
-        base_path = parse_result.path
+        self.bucket = parse_result.netloc
+        self.base_path = parse_result.path
 
-        s3_client = boto3.client("s3")
+        self.s3_client = boto3.client("s3")
+
+    def add(self, bento_name, bento_version):
+        # Generate pre-signed s3 path for upload
+        expiration = 3600
```
parano: make this a config? you can probably add a new "yatai" section
```
[yatai]
bento_uri_default_expiration = 3000
```

PR #329, opened 2019-10-09
Yatai feature: Implement retrieving and removing functionalities for s3 repo.

leonsim:

How was this patch tested?
--------------------------

### DIFF
--- a/bentoml/repository/__init__.py
+++ b/bentoml/repository/__init__.py
@@ -125,11 +125,17 @@ class _S3BentoRepository(BentoRepositoryBase):
 
         self.s3_client = boto3.client("s3")
 
+    @property
+    def _expiration(self):
+        return config('yatai').getint('bento_uri_default_expiration')
+
+    def _get_object_name(self, bento_name, bento_version):
+        return "/".join([self.base_path, bento_name, bento_version])
+
     def add(self, bento_name, bento_version):
         # Generate pre-signed s3 path for upload
-        expiration = config('yatai').getint('bento_uri_default_expiration')
 
-        object_name = "/".join([self.base_path, bento_name, bento_version])
+        object_name = self._get_object_name(bento_name, bento_version)
 
         try:
             response = self.s3_client.generate_presigned_post(
@@ -137,7 +143,7 @@ class _S3BentoRepository(BentoRepositoryBase):
                 object_name,
                 Fields=None,
                 Conditions=None,
-                ExpiresIn=expiration,
+                ExpiresIn=self._expiration,
             )
         except Exception as e:
             raise BentoMLRepositoryException(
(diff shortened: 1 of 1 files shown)


bentoml/repository/__init__.py
```diff
@@ -147,11 +153,31 @@ def add(self, bento_name, bento_version):
...
+        object_name = self._get_object_name(bento_name, bento_version)
+
+        try:
+            response = self.s3_client.generate_presigned_url(
+                'get_object',
+                Params={'Bucket': self.bucket, 'Key': object_name},
+                ExpiresIn=self._expiration,
+            )
+        except Exception as e:
+            raise BentoMLRepositoryException(
```
parano: I think generate_presigned_url returns None if it fails?

Could you make it check if the response is None and raise the same exception in that case?

bentoml/repository/__init__.py
```diff
@@ -147,11 +153,31 @@ def add(self, bento_name, bento_version):
...
 
     def dangerously_delete(self, bento_name, bento_version):
         # Remove s3 path containing related Bento files
-        raise NotImplementedError
+
+        object_name = self._get_object_name(bento_name, bento_version)
+
+        response = self.s3_client.delete_object(Bucket=self.bucket, Key=object_name)
+
+        DELETE_MARKER = 'DeleteMarker'
```
parano: add a comment about this field in boto3's response and what it means?

bentoml/repository/__init__.py
```diff
@@ -147,11 +153,31 @@ def add(self, bento_name, bento_version):
...
+            )
+        return response
 
     def dangerously_delete(self, bento_name, bento_version):
         # Remove s3 path containing related Bento files
-        raise NotImplementedError
+
+        object_name = self._get_object_name(bento_name, bento_version)
+
+        response = self.s3_client.delete_object(Bucket=self.bucket, Key=object_name)
```
parano: does this raise a ClientError when object no exist? can we improve the error message we show to users in that case?

bentoml/repository/__init__.py
leonsim: I'm following https://boto3.amazonaws.com/v1/documentation/api/latest/guide/s3-presigned-urls.html
So it will raise an exception in that case and the response is always a valid URL.

bentoml/repository/__init__.py
leonsim: different people think differently about this, I want to hear from you on this.
when trying to delete an object, if the object is not there, should we return success or failure?
what if a user tried to delete the same object twice?
(should we make endpoints idempotently safe? (even on what is idempotency, there are different opinions...))

bentoml/repository/__init__.py
leonsim: ok

bentoml/repository/__init__.py
parano: yes the doc says``` :return: Presigned URL as string. If error, returns None.``` it's probably better to give the user an error with message in that case, rather than return None

bentoml/repository/__init__.py
parano: I think it should return "deleted" when the object is not found, similar behavior to the s3 delete_object API.

My concern here was - we should try to catch ClientError around delete_object call, in case the request timeout or anything like that happened. And in that case, we want to know it failed to delete the object on the cloud platform.

bentoml/repository/__init__.py
leonsim: Oh, I find the sentence you mentioned in the doc. It's for the demo function `create_presigned_url `,  rather than the `generate_presigned_url` function. Apart from that, yeah I think it's better to provide an error than returning None.

bentoml/repository/__init__.py
leonsim: Got it. I'll wrap a try-catch then.

parano:
Thanks @leonsim, merging now!

bentoml/repository/__init__.py
parano: Got it, that make sense

PR #334, opened 2019-10-11
KerasModelArtifact Feature: Save As Weights & JSON

ghunkins:
This PR introduces the option for users of `KerasModelArtifact` to store only what is absolutely necessary for inference. This is particularly helpful for rapid, space-conscious deployments of Keras models that may have training configuration or history bloat.

Note: For `Keras` models that have already reduced to minimal information for inference, this option is still viable but will not reduce model storage size. Other options (i.e. translation to [`Tensorlow Lite`](https://www.tensorflow.org/lite/convert)) are necessary in this case.

No.

How was this patch tested?
--------------------------
`tox` passes on all supported environments. No tests for this specific feature have been written, but a successful deployment using this feature has been made.

Additionally, all files were auto-formatted using the provided formatting script.

### DIFF
--- a/bentoml/artifact/keras_model_artifact.py
+++ b/bentoml/artifact/keras_model_artifact.py
@@ -39,7 +39,13 @@ class KerasModelArtifact(BentoServiceArtifact):
         store_as_json_and_weights (bool): flag allowing storage of the Keras model as JSON and weights
     """
 
-    def __init__(self, name, custom_objects=None, model_extension=".h5", store_as_json_and_weights=False):
+    def __init__(
+        self,
+        name,
+        custom_objects=None,
+        model_extension=".h5",
+        store_as_json_and_weights=False,
+    ):
         super(KerasModelArtifact, self).__init__(name)
         self._model_extension = model_extension
         self._store_as_json_and_weights = store_as_json_and_weights
@@ -129,7 +135,7 @@ class KerasModelArtifact(BentoServiceArtifact):
                 if self._store_as_json_and_weights:
                     with open(self._model_json_path(path), 'r') as json_file:
                         model_json = json_file.read()
-                    model = keras.models.model_from_json(loaded_model_json)
+                    model = keras.models.model_from_json(model_json)
                     model.load_weights(self._model_weights_path(path))
                 # otherwise, load keras model via standard load_model
                 else:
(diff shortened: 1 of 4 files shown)


bentoml/artifact/keras_model_artifact.py
```diff
@@ -32,11 +32,23 @@
 class KerasModelArtifact(BentoServiceArtifact):
     """
     Abstraction for saving/loading Keras model
+
+    Args:
+        name (string): name of the artifact
+        custom_objects (dict): dictionary of Keras custom objects for model
+        store_as_json_and_weights (bool): flag allowing storage of the Keras model as JSON and weights
```
ghunkins: Added documentation as well!

tests/conftest.py
```diff
@@ -71,6 +72,7 @@ def predictFastaiImage(self, input_data):
         def predictFastaiImages(self, original, compared):
             return all(original.data[0, 0] == compared.data[0, 0])
 
+
```
ghunkins: Changes due to running the auto-formatting script. I can checkout these files if we don't want edits on them in this PR.

tests/test_service.py
```diff
@@ -38,6 +38,7 @@ def test_fastai_image_handler_pip_dependencies():
         # fast ai is required 3.6 or higher.
         assert True
     else:
+
```
ghunkins: Same auto-format comment as above.

tests/conftest.py
parano: No worries, just curious what version of `black` were you running"?

bentoml/artifact/keras_model_artifact.py
```diff
@@ -23,7 +23,7 @@
 
 try:
     import tensorflow as tf
-    from tensorflow.python import keras
+    import keras
```
parano: This may break for some users who only have installed Tensorflow but not the Keras package, although the original approach has some limitations too, we will do a follow-up PR to address this before next release

parano:
Thanks @ghunkins for the contribution! Merging now!

bentoml/artifact/keras_model_artifact.py
ghunkins: Should've posted a comment / question on this as well: when actually attempting to pass in my Keras model, `isinstance(data, keras.engine.training.Model)` from [Line 95](https://github.com/ollano-ops/BentoML/blob/38e551a0d6687d2272b2ec454fa9a3af3ecaacc4/bentoml/artifact/keras_model_artifact.py#L95) would return `False`. Switching to this fixed that. It looks as though `tensorflow.python.keras.engine.training.Model` and `keras.engine.training.Model` are not perfect aliases of this. 

I'll create an issue and see if I can put out a fix.

PR #365, opened 2019-11-04
Return S3 downloadable URL on upload and implement loading archive from S3.

leonsim:

image

How was this patch tested?
--------------------------

### DIFF
--- a/bentoml/archive/loader.py
+++ b/bentoml/archive/loader.py
@@ -23,8 +23,9 @@ import tarfile
 import logging
 import tempfile
 import requests
+from contextlib import contextmanager
 
-
+import boto3
 from bentoml.utils import dump_to_yaml_str
 from bentoml.utils.usage_stats import track_load_finish, track_load_start
 from bentoml.exceptions import BentoMLException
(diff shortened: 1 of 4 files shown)


bentoml/yatai/yatai_service_impl.py
```diff
@@ -300,7 +300,10 @@ def GetBento(self, request, context=None):
             bento_metadata_pb = self.bento_metadata_store.get(
                 request.bento_name, request.bento_version
             )
+
             if bento_metadata_pb:
+                uri = self.repo.get(bento_metadata_pb.name, bento_metadata_pb.version)
```
parano: add a comment here explaining that s3 link might be expired? which I assume is why we needed this line here

bentoml/archive/loader.py
```diff
@@ -30,6 +34,22 @@
...
+                    tar.extractall(path=tmpdir)
+                    return func(os.path.join(tmpdir, filename), **kwargs)
+
+        return func(archive_path, **kwargs)
+
+    return wrapper
+
+
+@download_remote_archive
 def load_bentoml_config(archive_path):
```
parano: perfer doing this in a more explicit way with at top of each functions:
```
if is_remote_archive(archive_path)
   archive_path = download_remote_archive(archive_path)
...
```

bentoml/yatai/python_api.py
```diff
@@ -118,16 +119,26 @@ def upload_bento_service(bento_service, base_path=None, version=None):
...
+                raise BentoMLException(
+                    "Error getting Bento with status code {} ".format(response.status)
+                )
+
             logger.info(
                 "Successfully saved Bento '%s:%s' to S3: %s",
                 bento_service.name,
                 bento_service.version,
-                response.uri.uri,
+                response.bento.uri.uri,
```
parano: actually I'd prefer to just print the original s3 address here, since users can always use that as a reference to verify it with aws cli or aws web console, and it's also easier to integrate with other tools.

One thing we need to change soon is from ```load(archive_path)``` to ```load(name, version)```, I think once we did that, we won't need the `saved_path` in your example to be the downloadable URL, so it remains the original s3 url.

One thing I do wish we can do now, is directory loading from an address like "s3://my_repo/my_svc/version_id/my_svc.tar.gz" if there user's machine already has aws configured. that way the loading can still work, it's just not via a presigned URL like we do with our yatai client.

leonsim: yep, but the original s3 address (`response.uri.uri`) is just a bucket domain name (like this `https://test-bento.s3-ap-northeast-1.amazonaws.com`) without the key.

+1 for `load(name, version)`

an address like "s3://my_repo/my_svc/version_id/my_svc.tar.gz" does look much simpler, but this file `python_api.py`  may be on a host without S3 keys (because the keys are configured at yatai server) and I assume the bucket is more frequently configured as a private bucket than a public bucket, then this will result in an Access Denied URL.

PR #388, opened 2019-11-19
ImageHandler accept file type check fix and unit tests

parano:
* fix accept format check - previously it will fail for filename with captialized extension e.g. '.PNG'
* when prediction request is malformated, API server should return 400 error response instead of raise exception with 500 internal error
* add unit tests for ImageHandler

### DIFF
--- a/bentoml/handlers/image_handler.py
+++ b/bentoml/handlers/image_handler.py
@@ -44,7 +44,7 @@ def verify_image_format_or_raise(file_name, accept_format_list):
     """
     if accept_format_list:
         _, extension = os.path.splitext(file_name)
-        if extension not in accept_format_list:
+        if extension.lower() not in accept_format_list:
             raise ValueError(
                 "Input file not in supported format list: {}".format(accept_format_list)
             )
--- a/bentoml/handlers/image_handler.py
+++ b/bentoml/handlers/image_handler.py
@@ -128,29 +128,37 @@ class ImageHandler(BentoHandler):
             response object
         """
         if request.method != "POST":
-            return Response(response="Only accept POST request", status=400)
+            return Response(
+                response="BentoML#ImageHandler only accept POST request", status=400
+            )
 
-        input_files = [request.files.get(filename) for filename in self.input_names]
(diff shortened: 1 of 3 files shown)


bentoml/handlers/image_handler.py
```diff
@@ -128,29 +128,37 @@ def handle_request(self, request, func):
             response object
         """
         if request.method != "POST":
-            return Response(response="Only accept POST request", status=400)
+            return Response(
+                response="BentoML#ImageHandler only accept POST request", status=400
```
leonsim: how about 405?

bentoml/handlers/image_handler.py
```diff
@@ -128,29 +128,37 @@ def handle_request(self, request, func):
...
+            for form_input_name in self.input_names
+            if form_input_name in request.files
+        ]
+        if input_files:
+            file_names = [secure_filename(file.filename) for file in input_files]
+            try:
+                for file_name in file_names:
+                    verify_image_format_or_raise(file_name, self.accept_image_formats)
+            except ValueError as e:
+                return Response(response=str(e), status=400)
```
leonsim: 👍

bentoml/handlers/image_handler.py
```diff
@@ -128,29 +128,37 @@ def handle_request(self, request, func):
             response object
         """
         if request.method != "POST":
-            return Response(response="Only accept POST request", status=400)
+            return Response(
+                response="BentoML#ImageHandler only accept POST request", status=400
+            )
 
-        input_files = [request.files.get(filename) for filename in self.input_names]
-        if len(input_files) == 1 and input_files[0] is None:
```
leonsim: There is one case we tried to cover previously:
if a user submits only one file, even without an incorrect filename, we can still process it.

parano: got you, good catch!

yubozhao: Yes.  I believe that is the case, when you submit only one file but with incorrect filename.  It looks like @parano is handling reading data from files with incorrect file names.

PR #401, opened 2019-11-26
AWS Lambda deployment improvements

yubozhao:


bentoml/deployment/aws_lambda/__init__.py
```diff
@@ -394,28 +402,24 @@ def describe(self, deployment_pb, yatai_service):
                     state = DeploymentState(state=DeploymentState.PENDING)
                     state.timestamp.GetCurrentTime()
                     return DescribeDeploymentResponse(status=Status.OK(), state=state)
-            except Exception as error:
+            except Exception as error:  # pylint: disable=broad-except
```
parano: this is not OK... handle the specific error by type

bentoml/deployment/aws_lambda/utils.py
```diff
@@ -41,44 +44,44 @@ def ensure_sam_available_or_raise():
...
             if 'wheel' in files:
-                logger.debug('removing file: ' + os.path.join(root, 'wheel'))
+                logger.debug('removing file: %s', os.path.join(root, 'wheel'))
                 os.remove(os.path.join(root, 'wheel'))
             if 'pip' in files:
-                logger.debug('removing file: ' + os.path.join(root, 'pip'))
+                logger.debug('removing file: %s', os.path.join(root, 'pip'))
                 os.remove(os.path.join(root, 'pip'))
-    except Exception as e:
+    except Exception as e:  # pylint: disable=broad-except
```
parano: why do we need to do this?

bentoml/deployment/aws_lambda/utils.py
```diff
@@ -31,7 +31,10 @@
 
 def ensure_sam_available_or_raise():
     try:
+        # pylint: disable=unused-import
         import samcli  # noqa
+
+        # pylint: enable=unused-import
```
parano: does this work? ```import samcli  # pylint: disable=unused-import```

parano: actually might be nice to add a minimal required version check here, something like:
```
assert list(map(int, samcli.__version__.split('.'))) > [0, 31, 1], "Upgrade sam cli with pip install -U ............."
```

parano:
Merging now, feel free to address comments in follow up!

PR #405, opened 2019-11-26
Add e2e test script for lambda deployment

yubozhao:
`python ./script/e2e_lambda_deployment.py`


scripts/e2e_lambda_deployment.py
```diff
@@ -0,0 +1,141 @@
...
+                headers={'Content-Type': 'application/json'},
+            )
+            if request_result.status_code != 200:
+                deployment_failed = True
+            if request_result.content.decode('utf-8') != '[0]':
+                deployment_failed = True
+        except Exception as e:
+            logger.error(str(e))
+            deployment_failed = True
+        test_endpoint_command = [
```
parano: not used?

scripts/e2e_lambda_deployment.py
```diff
@@ -0,0 +1,141 @@
...
+@artifacts([PickleArtifact('clf')])
+@env(pip_dependencies=['scikit-learn'])
+class IrisClassifier(BentoService):
+    @api(DataframeHandler)
+    def predict(self, df):
+        return self.artifacts.clf.predict(df)
+
+
+if __name__ == '__main__':
+    print('E2E DEPLOYMENT TEST FOR LAMBDA:::: Training iris classifier')
```
parano: use logger.info?, change logger to logger = logging.getLogger("bentoml.test")
 ?

scripts/e2e_lambda_deployment.py
```diff
@@ -0,0 +1,141 @@
...
+    loaded_service = load(saved_path)
+    sample_data = X[0:1]
+
+    deployment_failed = False
+    print(
+        'E2E DEPLOYMENT TEST FOR LAMBDA:::: Creating AWS Lambda test deployment '
+        'for iris classifier with BentoML CLI'
+    )
+    bento_name = '{}:{}'.format(loaded_service.name, loaded_service.version)
+    deployment_name = 'tests-lambda-e2e'
```
parano: let's add a random string to this name - note that we may move this to CI later

tests/deployment/aws_lambda/test_aws_lambda_deployment_operator.py
```diff
@@ -104,6 +104,32 @@ def return_predict_func(mock_load_service, mock_download_artifacts):
...
+    open(os.path.join(mock_bento_bundle_path, 'requirements.txt'), 'w').close()
+
+    init_sam_project(
+        mock_sam_project_path,
+        mock_bento_bundle_path,
+        mock_deployment_name,
+        mock_bento_name,
+        mock_api_names,
+    )
+    assert os.path.isfile(
```
parano: could you also add assertions for other init_sam_projec behaviors?

scripts/e2e_lambda_deployment.py
yubozhao: yeah, we just use `requests` package instead. removing it now

PR #406, opened 2019-11-27
add TensorflowKerasSavedModelArtifact

hrmthw:

* add TensorflowKerasSavedModelArtifact

Why proposed these changes?
-------------------------------
* After TensorFlow 2.0, keras is deeply embedded. It is recommended in the official starter tutorial.
* tf.keras saved models have different loading behavior from normal tf models.

no

How was this patch tested?
--------------------------
no now

### DIFF
--- a/bentoml/artifact/tf_savedmodel_artifact.py
+++ b/bentoml/artifact/tf_savedmodel_artifact.py
@@ -51,7 +51,7 @@ def _load_tf_keras_saved_model(path):
 
         TF2 = tf.__version__.startswith('2')
     except ImportError:
-        raise ImportError("Tensorflow package is required "
+        raise ImportError("Tensorflow 2.0 or above is required "
                           "to use TfKerasSavedModelArtifact")
 
     if TF2:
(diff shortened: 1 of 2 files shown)


bentoml/artifact/tf_savedmodel_artifact.py
```diff
@@ -115,6 +131,13 @@ def load(self, path):
         return self.pack(loaded_model)
 
 
+class TensorflowKerasSavedModelArtifact(TensorflowSavedModelArtifact):
```
parano: could you add docstring here similar to TensorflowSavedModelArtifact, with some example code?

hrmthw: will do

hrmthw: Will do!

bentoml/artifact/tf_savedmodel_artifact.py
```diff
@@ -115,6 +131,47 @@ def load(self, path):
...
+    >>>
+    >>>     @bentoml.api(JsonHandler)
+    >>>     def predict(self, json):
+    >>>         input_data = json['input']
+    >>>         prediction = self.artifacts.model.add(input_data)
+    >>>         return prediction.numpy()
+    >>>
+    >>> svc = TfKerasModelService()
+    >>>
+    >>> # Option 1: pack directly with Tensorflow trackable object
```
parano: this is no longer Tensorflow trackable object here right? should be `tf.keras.models.Model` or something like it?

bentoml/artifact/tf_savedmodel_artifact.py
parano: And could you also add this to the docs/artifacts list?  so it shows up in our docs site?

hrmthw:
I just found we actually do not need this PR.

According to this doc about TensorFlow Serving, https://tf.wiki/en/deployment/serving.html#keras-sequential-mode-models-deployment, tf.keras.Model shares the saving/loading API with normal tf.Module.

For sequential models and custom models, only `__call__` are wrapped and bound to auto-generated signature by default. In custom models, we can use `@tf.function` decorator to define the signatures of methods we want to call in the Servicer.

The tf.keras.models.load is just a special case.

PR #410, opened 2019-12-02
Optimize default metrics

leonsim:

image

How was this patch tested?
--------------------------

### DIFF
--- a/bentoml/server/bento_api_server.py
+++ b/bentoml/server/bento_api_server.py
@@ -23,6 +23,7 @@ import time
 import logging
 from functools import partial
 from collections import OrderedDict
+from timeit import default_timer
 
 from flask import Flask, jsonify, Response, request
 from werkzeug.utils import secure_filename
@@ -190,8 +191,12 @@ def bento_service_api_wrapper(api, service_name, service_version):
     """
     Create api function for flask route
     """
-    summary_name = str(service_name) + "_" + str(api.name)
-    request_metric_time = Summary(summary_name, summary_name + " request latency")
+    request_metric_time = Summary(
+        api.name,
+        api.name + " request latency",
+        namespace=service_name,
+        labelnames=('endpoint', 'code'),
+    )
 
     def log_image(req, request_id):
         img_prefix = 'image/'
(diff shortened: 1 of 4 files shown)


bentoml/server/bento_api_server.py
```diff
@@ -190,8 +191,12 @@ def bento_service_api_wrapper(api, service_name, service_version):
     """
     Create api function for flask route
     """
-    summary_name = str(service_name) + "_" + str(api.name)
-    request_metric_time = Summary(summary_name, summary_name + " request latency")
+    request_metric_time = Summary(
+        api.name,
+        api.name + " request latency",
+        namespace=service_name,
```
yubozhao: maybe namespace is not the most appropriate choice.  It collides with our deployment's namespace, that could lead to confusion. Maybe change it to `bento_service_name`

parano: it's prometheus' namespace...

parano: I'd recommend use `"BENTOML"` as default namespace, and make it configurable(add to the config list). And it would be better to include service name as part of the metric name, so the metric name will become `{service_name}_{api.name}`. I think most commonly people are just using `predict`, `classify` as API name, and it can get confusing if they have multiple BentoService deployed.

bentoml/server/bento_api_server.py
```diff
@@ -256,8 +261,19 @@ def wrapper():
...
 
             response.headers["request_id"] = request_id
+
             return response
 
+        start_time = default_timer()
+
+        response = handle()
+
+        request_metric_time.labels(
```
parano: Is this something we can do at flask level for all endpoints?  with something like before_request, after_request

guides/metrics.md
```diff
@@ -0,0 +1,6 @@
+
+## Defalut metrics to track
+
+- For each API endpoint:
+    - sum of requests
+    - count of requests with label of http response code
```
parano: could you also add other metrics you plan to add here in this list? it's ok if they are not yet implemented

bentoml/server/bento_api_server.py
parano: Note that we don't need a `Summary` object for each API endpoint, they are already part of the label  - correct me if I'm wrong about this

bentoml/server/bento_api_server.py
leonsim: Ok. Just thought it would be helpful when users want to query based on service_name when the Prometheus is shared among different Bento Services. Anyway, we can revise this later, for now, I'll change back to `BENTOML`.

bentoml/server/bento_api_server.py
leonsim: Yes, if we change the metric name to include the api name, then there is no need to add this `endpoint` label here.

leonsim: For flask level, I think that's possible. Do you have some concerns for doing it here, or are there some benefits to instrument this at flask level?

guides/metrics.md
leonsim: Sure, will do.

PR #425, opened 2019-12-11
Refactor the way of returning response of Bad Request(400) and Method…

leonsim:
… Not Allowed(405).

How was this patch tested?
--------------------------

### DIFF
--- a/bentoml/handlers/base_handlers.py
+++ b/bentoml/handlers/base_handlers.py
@@ -21,6 +21,8 @@ import json
 import pandas as pd
 import numpy as np
 
+from werkzeug.exceptions import BadRequest
+
 
 class BentoHandler:
     """Handler in BentoML is the layer between a user API request and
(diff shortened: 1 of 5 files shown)


bentoml/handlers/base_handlers.py
```diff
@@ -81,4 +83,4 @@ def get_output_str(result, output_format, output_orient="records"):
                 # when result is not JSON serializable
                 return json.dumps(str(result))
     else:
-        raise ValueError("Output format {} is not supported".format(output_format))
+        raise BadRequest("Output format {} is not supported".format(output_format))
```
parano: Can we define our own BadInput exception so it can be properly handled when loading a BentoServiceAPI for use with CLI or AWS Lambda?

parano:
looks like it's failling some related handler tests, I will try fix those in a separate PR as follow up

PR #435, opened 2019-12-16
Refactor instrumentation with Flask WSGI middleware.

leonsim:

How was this patch tested?
--------------------------

### DIFF
--- a/bentoml/server/middlewares.py
+++ b/bentoml/server/middlewares.py
@@ -1,0 +1,66 @@
+from timeit import default_timer
+
+from flask import Request
+
+from bentoml import config
+
+
+class InstrumentMiddleware:
+    def __init__(self, app, bento_service):
+        self.app = app
+        self.bento_service = bento_service
+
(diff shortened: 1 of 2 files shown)


bentoml/server/middlewares.py
```diff
@@ -0,0 +1,66 @@
...
+            namespace=namespace,
+            labelnames=['api_name', 'service_version'],
+        )
+
+    def __call__(self, environ, start_response):
+        req = Request(environ)
+        api_name = req.path.lstrip('/')
+        start_time = default_timer()
+
+        def start_response_wrapper(status, headers):
```
parano: start_response callable got called at the time of creating a response right?  why not just wrap those code the same way `metrics_request_in_progress ` is done below, wrapping around `self.app` call?

leonsim: Right.
The reason here is to get the response status code.

parano:
Ran tests locally and all passed. Merging and ignoring the travis build failure for now.

PR #454, opened 2019-12-30
adaptive micro batching

hrmthw:

How was this patch tested?
--------------------------

### DIFF
--- a/bentoml/bundler/loader.py
+++ b/bentoml/bundler/loader.py
@@ -144,7 +144,7 @@ def load_bento_service_class(bundle_path):
     return model_service_class
 
 
-def load(bundle_path):
+def load(bundle_path, skip_artifact=False):
     """Load bento service from local file path or s3 path
 
     Args:
@@ -159,12 +159,14 @@ def load(bundle_path):
         with _resolve_remote_bundle_path(bundle_path) as local_bundle_path:
             return load(local_bundle_path)
 
-    track_load_start()
+    if not skip_artifact:
+        track_load_start()
 
     svc_cls = load_bento_service_class(bundle_path)
-    svc = svc_cls()
+    svc = svc_cls(skip_artifact)
 
-    track_load_finish(svc)
+    if not skip_artifact:
+        track_load_finish(svc)
     return svc
 
 
--- a/bentoml/bundler/loader.py
+++ b/bentoml/bundler/loader.py
@@ -144,7 +144,7 @@ def load_bento_service_class(bundle_path):
     return model_service_class
 
 
-def load(bundle_path, skip_artifact=False):
+def load(bundle_path):
     """Load bento service from local file path or s3 path
 
     Args:
@@ -159,14 +159,12 @@ def load(bundle_path, skip_artifact=False):
         with _resolve_remote_bundle_path(bundle_path) as local_bundle_path:
             return load(local_bundle_path)
 
-    if not skip_artifact:
-        track_load_start()
+    track_load_start()
 
     svc_cls = load_bento_service_class(bundle_path)
-    svc = svc_cls(skip_artifact)
+    svc = svc_cls()
 
-    if not skip_artifact:
-        track_load_finish(svc)
+    track_load_finish(svc)
     return svc
 
 
--- a/bentoml/cli/__init__.py
+++ b/bentoml/cli/__init__.py
@@ -37,7 +37,6 @@ from bentoml.cli.aws_sagemaker import get_aws_sagemaker_sub_command
 from bentoml.cli.bento import add_bento_sub_command
 from bentoml.server import BentoAPIServer, get_docs
 from bentoml.server.marshal_server import MarshalServer
-from bentoml.server.gunicorn_server import GunicornBentoServer
 from bentoml.cli.click_utils import BentoMLCommandGroup, conditional_argument, _echo
 from bentoml.cli.deployment import get_deployment_sub_command
 from bentoml.cli.config import get_configuration_sub_command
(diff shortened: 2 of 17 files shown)


bentoml/bundler/loader.py
```diff
@@ -158,12 +158,14 @@ def load(bundle_path):
         with _resolve_remote_bundle_path(bundle_path) as local_bundle_path:
             return load(local_bundle_path)
 
-    track_load_start()
+    if not skip_artifact:
+        track_load_start()
```
parano: this and the `track_load_finish` below  are for usage tracking, not related to loading artifact

bentoml/utils/__init__.py
```diff
@@ -32,6 +34,26 @@
...
+def detect_free_port(host='localhost'):
+    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    sock.bind((host, 0))
+    port = sock.getsockname()[1]
+    sock.close()
+    return port
+
+
+@contextmanager
+def reserve_free_port(host='localhost'):
```
parano: 👍

bentoml/bundler/loader.py
hrmthw: I know. Loading service without artifact is different from the usage before. It take shorter time than g-unicorn workers obviously. I'm not sure if we could mix them up.

setup.py
```diff
@@ -39,6 +39,7 @@
     "tabulate",
     "humanfriendly",
     "alembic",
+    "aiohttp",
```
parano: does it matter for us to include aiohttp's optional dependency `cchardet`?

bentoml/server/marshal_server.py
```diff
@@ -0,0 +1,63 @@
...
+
+CONTENT_TYPE_LATEST = str("text/plain; version=0.0.4; charset=utf-8")
+
+marshal_logger = logging.getLogger("bentoml.marshal")
+logger = logging.getLogger(__name__)
+
+
+class MarshalServer:
+    """
+    BentoAPIServer creates a REST API server based on APIs defined with a BentoService
```
parano: update the docstring here?

bentoml/server/marshal_server.py
```diff
@@ -0,0 +1,63 @@
...
+import logging
+import multiprocessing
+
+from bentoml import config
+from bentoml.marshal import MarshalService
+
+
+CONTENT_TYPE_LATEST = str("text/plain; version=0.0.4; charset=utf-8")
+
+marshal_logger = logging.getLogger("bentoml.marshal")
```
parano: what's marshal_logger for?

bentoml/cli/__init__.py
```diff
@@ -259,31 +280,53 @@ def serve(port, bento=None, with_conda=False):
...
-                    timeout=timeout,
-                ),
-            )
-            return
-
-        from bentoml.server.gunicorn_server import GunicornBentoServer
+            def _run_api_server(_api_server_port):
+                run_with_conda_env(
+                    pip_installed_bundle_path,
+                    'bentoml serve_gunicorn {bento} -p {port} -w {workers} '
```
parano: consider just add `--enable-marshal` here in the cli arg? that way the marshal server will also be running in a conda environment and probably will simplify the code here

setup.py
hrmthw: no, not used

bentoml/handlers/dataframe_handler.py
```diff
@@ -67,10 +156,14 @@ class DataframeHandler(BentoHandler):
     """
 
     def __init__(
-        self, orient="records", output_orient="records", typ="frame", input_dtypes=None
+        self, orient="records", output_orient="records", typ="frame", input_dtypes=None,
+        micro_batch=False, mb_max_latency=0.8,
```
parano: Do we really need the `micro_batch` parameter here?

It seems to me, whether for `micro_batch` to be True or False, it is the same for the user-defined API callback function(in case of Dataframe handler where input is already a 'batch').  I think ideally, we want users to just define the callback function to process a batch of inputs so that BentoML can load the BentoService to run with micro-batch enabled or disabled.

bentoml/handlers/dataframe_handler.py
```diff
@@ -44,11 +48,96 @@ def _check_dataframe_column_contains(required_column_names, df):
             )
 
 
+def _dataframe_csv_from_input(raws, content_types):
+    n_row_sum = -1
+    for i, (data, content_type) in enumerate(zip(raws, content_types)):
+        if content_type.lower() == "application/json":
```
parano: also treat `data` as "application/json" when content_type is None?

bentoml/handlers/dataframe_handler.py
parano: Instead, we could add a constant `batch_mode_supported` to BentoHandler and default to false, that way on the marshal server-side, we can determine if a BentoServiceAPI endpoint should be loaded for batch process or not. This can be useful when we are rolling out this feature as some of the handlers does not yet support it.

setup.py
parano: It's an optional dependency used in aiohttp https://github.com/aio-libs/aiohttp/blob/63a0d1040533947ef4973b3baed87e5fb7870e5a/aiohttp/client_reqrep.py#L69 but I wasn't sure about the performance implication

hrmthw:
@parano ping

setup.py
hrmthw: Oh, I see. Maybe we can deal with it later with ujson.

parano:
Thanks @hrmthw, everything looks great, merging now!

PR #457, opened 2020-01-03
Deployment improvement - Add gunicorn worker per instance option for sagemaker

yubozhao:


bentoml/utils/validator/__init__.py
```diff
@@ -51,6 +51,10 @@
                     'region': {'type': 'string'},
```
parano: I saw `api_name` have `required: True`, do we also need that for instance_type and instance_count?

bentoml/deployment/sagemaker/__init__.py
```diff
@@ -377,21 +377,21 @@ def _create_sagemaker_model(
...
         "ExecutionRoleArn": execution_role_arn,
     }
-    default_worker_count = config().getint(
-        'apiserver', 'default_gunicorn_workers_count'
-    )
-    if default_worker_count > 0:
+
+    # Will set envvar, if user defined gunicorn worker is greater than 0.  EnvVar needs
+    # to be string instead of the int.
+    default_num_of_gunicorn_worker_count = spec.num_of_gunicorn_workers_per_instance
```
parano: IIUC, num_of_gunicorn_workers_per_instance is optional right?

shouldn't this be:
```
if spec.num_of_gunicorn_workers_per_instance:
   sagemaker_model_info['PrimaryContainer']['Environment']['GUNICORN_WORKER_COUNT'] = ....
```

parano: I suppose this code assumes we have something like a default value of "-1" for num_of_gunicorn_workers_per_instance,  but I don't see where we set that default value for this proto field

bentoml/deployment/sagemaker/templates.py
```diff
@@ -86,7 +86,9 @@
 from bentoml.server.utils import get_gunicorn_num_of_workers
 
 bento_server_timeout = os.environ.get('BENTO_SERVER_TIMEOUT', 60)
-bento_server_workers = int(os.environ.get('BENTO_SERVER_WORKERS', get_gunicorn_num_of_workers()))
+bento_server_workers = int(
```
parano: let's move this to a separate python file? so we got to run linter and unit test ?

parano: actually same for the DEFAULT_WSGI_PY and DEFAULT_NGINX_CONFIG above

bentoml/yatai/deployment_utils.py
```diff
@@ -57,19 +57,20 @@ def deployment_dict_to_pb(deployment_dict):
...
                 sagemaker_config.get('instance_count')
             )
         if sagemaker_config.get('instance_type'):
-            sagemaker_operator_config_pb.instance_type = sagemaker_config.get(
-                'instance_type'
-            )
+            sagemaker_config_pb.instance_type = sagemaker_config.get('instance_type')
+        if sagemaker_config.get('num_of_gunicorn_workers_per_instance'):
+            sagemaker_config_pb.num_of_gunicorn_workers_per_instance = \
+                sagemaker_config.get('num_of_gunicorn_workers_per_instance')
```
parano: let's add a unit test for `deployment_dict_to_pb`? it seems easy to forget about it when adding new field in proto message

bentoml/cli/deployment.py
```diff
@@ -219,9 +219,15 @@ def deployment():
         type=click.INT,
     )
     @click.option(
-        '--api-name',
-        help='User defined API function will be used for inference. Option applicable'
+        '--gunicorn-workers-per-instance',
```
parano: let's make this the same as it is in the proto message `num_of_gunicorn_workers_per_instance`? or make them both `gunicorn workers per instance`?

protos/deployment.proto
```diff
@@ -34,6 +34,7 @@ message DeploymentSpec {
     string instance_type = 2;
     int32 instance_count = 3;
     string api_name = 4;
+    int32 num_of_gunicorn_workers_per_instance = 5;
```
parano: add an inline comment to explain the behavior when the optional field is not presented?

parano: would be nice to also add an inline comment for api_name etc

bentoml/utils/validator/__init__.py
yubozhao: both `instance type` and `instance count` has default value in the bentoml.cfg, that's probably why we dont have them as required.

bentoml/deployment/sagemaker/__init__.py
yubozhao: yeah, this is assuming the user doesn't enter any non-positive value.  Since we already check the value during validation, i probably should update it to what you type out

bentoml/cli/deployment.py
yubozhao: I removed `num_of_`.  It sounds weird to me in CLI context.  I will just add it back.

bentoml/yatai/deployment_utils.py
yubozhao: We do have one basic test for `deployment_dict_to_pb`.  Yes, I agree we need to add more of it

PR #462, opened 2020-01-07
BentoHandler API output format refactor

parano:
* Removed the behavior of configuring API handler output type by setting HTTP request header
* Fixed issues with Numpy outputs from user API function
* Additional parameter validation for `orient` and `output_orient` in DataframeHandler
* Removed the leading "predictions" key in TensorHandler outputs
* Fixed all linter errors

### DIFF
--- a/bentoml/handlers/tensorflow_tensor_handler.py
+++ b/bentoml/handlers/tensorflow_tensor_handler.py
@@ -22,14 +22,14 @@ from flask import Response
 from bentoml.handlers.utils import (
     NestedConverter,
     tf_b64_2_bytes,
-    tf_tendor_2_serializable,
+    tf_tensor_2_serializable,
 )
-from bentoml.handlers.base_handlers import BentoHandler, get_output_str
+from bentoml.handlers.base_handlers import BentoHandler, api_func_result_to_json
 from bentoml.exceptions import BentoMLException, BadInput
 
 
 decode_b64_if_needed = NestedConverter(tf_b64_2_bytes)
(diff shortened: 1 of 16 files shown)


bentoml/handlers/tensorflow_tensor_handler.py
```diff
@@ -90,10 +90,9 @@ def _handle_raw_str(self, raw_str, output_format, func):
             raise NotImplementedError("column format 'inputs' is not implemented")
 
         if output_format == "json":
-            result_object = {"predictions": result}
```
parano: @hrmthw removed the leading `{"predictions": ...}` here just for consistency, lmk if you think this will cause other problems

hrmthw: It's for consistency of tensor-flow serving only. Removing it will not cause other problems.

bentoml/handlers/tensorflow_tensor_handler.py
```diff
@@ -22,14 +22,14 @@
...
-    tf_tendor_2_serializable,
+    tf_tensor_2_serializable,
 )
-from bentoml.handlers.base_handlers import BentoHandler, get_output_str
+from bentoml.handlers.base_handlers import BentoHandler, api_func_result_to_json
 from bentoml.exceptions import BentoMLException, BadInput
 
 
 decode_b64_if_needed = NestedConverter(tf_b64_2_bytes)
-decode_tf_if_needed = NestedConverter(tf_tendor_2_serializable)
```
hrmthw: tendor……

PR #479, opened 2020-01-17
CLI/Yatai server improvement - Enhance list operation with order-by-column, order and other options

yubozhao:


bentoml/yatai/client/deployment_api.py
```diff
@@ -446,27 +455,29 @@ def list_lambda_deployments(
...
         if list_result.status.status_code != status_pb2.Status.OK:
             return list_result
-        lambda_deployments = [
-            deployment
-            for deployment in list_result.deployments
-            if deployment.spec.operator == DeploymentSpec.AWS_LAMBDA
-        ]
-        del list_result.deployments[:]
-        list_result.deployments.extend(lambda_deployments)
+        # lambda_deployments = [
```
parano: remove this?

bentoml/deployment/store.py
```diff
@@ -146,17 +156,40 @@ def delete(self, name, namespace):
...
-    def list(self, namespace, filter_str=None, labels=None, offset=None, limit=None):
+    def list(
+        self,
+        namespace,
+        operator=None,
+        labels=None,
+        offset=None,
+        limit=None,
+        order_by=None,
+        ascending_order=None,
```
parano: This works but I believe you want the default value to be False? `ascending_order = False`

bentoml/deployment/store.py
```diff
@@ -146,17 +156,40 @@ def delete(self, name, namespace):
...
 
-    def list(self, namespace, filter_str=None, labels=None, offset=None, limit=None):
+    def list(
+        self,
+        namespace,
+        operator=None,
+        labels=None,
+        offset=None,
+        limit=None,
+        order_by=None,
```
parano: we could also just do `order_by='created_at'` here right?

bentoml/deployment/store.py
```diff
@@ -146,17 +156,40 @@ def delete(self, name, namespace):
...
                 )
 
-    def list(self, namespace, filter_str=None, labels=None, offset=None, limit=None):
+    def list(
+        self,
+        namespace,
+        operator=None,
+        labels=None,
+        offset=None,
+        limit=None,
```
parano: might be good to add a default limit as well, like `200`, and default offset `0`

protos/deployment.proto
```diff
@@ -171,11 +171,19 @@ message DescribeDeploymentResponse {
 }
 
 message ListDeploymentsRequest {
+  enum SORTABLE_COLUMN {
+    unset = 0;
+    created_at = 1;
```
parano: I think you forgot to handle ‘unset’ again 
IIRC the first item in an enum is the default value in protobuf? how about just 
```
created_at = 0;
name = 1;
```

parano: same for the other list rpc below

bentoml/repository/metadata_store.py
```diff
@@ -28,6 +28,8 @@
     Boolean,
     DateTime,
     UniqueConstraint,
+    text,
```
parano: text not used?

bentoml/deployment/store.py
parano: that way you can remove the `if limit` check below as well

bentoml/deployment/store.py
```diff
@@ -146,17 +156,40 @@ def delete(self, name, namespace):
...
+                query = query.filter(
+                    Deployment.spec['operator'].contains(operator_name)
+                )
+            if order_by:
+                order_by = ListDeploymentsRequest.SORTABLE_COLUMN.Name(order_by)
+            else:
+                order_by = 'created_at'
+            order_by_field = getattr(Deployment, order_by)
+            order_by_action = (
+                desc(order_by_field) if not ascending_order else order_by_field
```
parano: nit ```order_by_action = order_by_field if ascending_order else desc(order_by_field)```

bentoml/cli/deployment.py
```diff
@@ -250,28 +258,40 @@ def get(name, namespace, output):
         default=ALL_NAMESPACE_TAG,
     )
     @click.option(
-        '--limit', type=click.INT, help='Limit how many resources will be retrieved'
+        '-p',
+        '--platform',
+        type=click.Choice(['aws-sagemaker', 'aws-lambda']),
```
parano: should we use 'sagemaker' and 'lambda' to match the commands? we will need to do a mapping from cli to the operator type name in proto tho

bentoml/cli/deployment.py
```diff
@@ -250,28 +258,40 @@ def get(name, namespace, output):
...
+        help='platform',
     )
     @click.option(
-        '--filters',
-        type=click.STRING,
-        help='List resources containing the filter string in name',
+        '--limit', type=click.INT, help='Limit how many resources will be retrieved'
```
parano: 'The maximum amount of deployments to be listed at once`

bentoml/cli/bento.py
```diff
@@ -64,25 +63,16 @@ def _print_bentos_info(bentos, output_type):
             _print_bento_info(bento, output_type)
 
 
-def get_bento_sub_command():
+def add_bento_sub_command(cli):
```
parano: love how easy it is to change this design!

bentoml/cli/bento.py
```diff
@@ -125,22 +115,29 @@ def get(bento, limit, filters, output):
...
     @click.option(
-        '--filters',
-        type=click.STRING,
-        help='List resources containing the filter string in name',
+        '--offset', type=click.INT, help='How many BentoServices will be skipped'
     )
+    @click.option(
+        '--order-by', type=click.Choice(['created_at', 'name']), default='created_at',
+    )
+    @click.option('--ascending-order', is_flag=True)
```
parano: use "--asc/--desc" instead?  which are the same keywords found in SQL, look more familiar for people

bentoml/cli/deployment.py
```diff
@@ -250,28 +258,40 @@ def get(name, namespace, output):
...
     @click.option(
         '-l',
         '--labels',
         type=click.STRING,
         help='List deployments matching the giving labels',
     )
+    @click.option(
+        '--order-by', type=click.Choice(['created_at', 'name']), default='created_at',
+    )
+    @click.option('--ascending-order', is_flag=True)
```
parano: would be nice to have "--asc/--desc"

parano:
finished a pass, looking great!

bentoml/deployment/store.py
```diff
@@ -146,17 +155,38 @@ def delete(self, name, namespace):
...
-                query.offset(offset)
-            if filter_str:
-                query.filter(Deployment.name.contains(filter_str))
+                query = query.offset(offset)
+            if operator:
+                operator_name = DeploymentSpec.DeploymentOperator.Name(operator)
+                query = query.filter(
+                    Deployment.spec['operator'].contains(operator_name)
+                )
+            if order_by:
```
parano: no need for this now?

bentoml/repository/metadata_store.py
```diff
@@ -173,18 +175,30 @@ def dangerously_delete(self, bento_name, bento_version):
...
+            query = sess.query(Bento).limit(limit)
             if offset:
-                query.offset(offset)
-            if filter_str:
-                query.filter(Bento.name.contains(filter_str))
+                query = query.offset(offset)
             if bento_name:
                 # filter_by apply filtering criterion to a copy of the query
                 query = query.filter_by(name=bento_name)
+            if order_by:
```
parano: similarly no need for this `if else` here now

bentoml/repository/metadata_store.py
```diff
@@ -173,19 +175,35 @@ def dangerously_delete(self, bento_name, bento_version):
...
+            order_by_field = getattr(Bento, order_by)
+            order_by_action = (
+                order_by_field if ascending_order else desc(order_by_field)
+            )
+            query = query.order_by(order_by_action)
             if bento_name:
                 # filter_by apply filtering criterion to a copy of the query
                 query = query.filter_by(name=bento_name)
 
+            # We are not defaulting limit to 200 in the signature,
```
parano: 👍

PR #486, opened 2020-01-21
Fixed setup_sh feature in service_env

kevlar1818:
It appears that the setup_sh feature of the `env` decorator was broken.
We (@seantur and myself) implemented the functionality as described in the documentation
and added the tests seen below.

Additionally, we moved the setup.sh execution higher in the generated Dockerfile.
The reason for this was to allow the setup.sh script to add our private PyPI repository
to the image, which is required for the `pip install -r /bento/requirements.txt` line to
succeed for private packages. We would certainly welcome feedback on this solution
and discuss alternatives.

No.

How was this patch tested?
--------------------------
We used a Python 3.7.5 virtualenv and ran the following commands:
```
./scripts/lint.sh
./scripts/format.sh
pytest
```

### DIFF
--- a/bentoml/bundler/templates.py
+++ b/bentoml/bundler/templates.py
@@ -78,15 +78,15 @@ RUN conda install pip numpy scipy \\
 COPY . /bento
 WORKDIR /bento
 
+# run user defined setup script
+RUN if [ -f /bento/setup.sh ]; then /bin/bash -c /bento/setup.sh; fi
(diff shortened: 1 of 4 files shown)


parano:
Hey @kevlar1818, excited to see the contribution, everything looks great! The change to move up the line in Dockerfile makes sense, thanks for explaining. Merging now.

Issue #499, opened 2020-01-24
serve_gunicorn is now called with the `bento` param and not `bundle_path`


pdoyle5000:
**Describe the bug**
bundle_path is still passed to serve_gunicorn as a kwarg and now breaks as it is an unexpected parameter.  the bundle_path param needs to be switched the new new `bento` param as changed in the commit: https://github.com/bentoml/BentoML/commit/abfd77cc327c9d205ccbe34c6cb2fe3762cd4502

to avoid an unexpected parameter exception.  I am not confident in where along the dependency chain to make that alteration as to not affect the other severing/cli modes.

@kevlar1818 @seantur

parano:
@pdoyle5000 thanks for reporting the bug!

Passing bundle_path to server/gunicorn_serve should still work, we made it backward compatible by design. This error was due to the naming of the parameter was wrong, a bug introduced in https://github.com/bentoml/BentoML/commit/abfd77cc327c9d205ccbe34c6cb2fe3762cd4502
We are working on a fix now.

parano:
@pdoyle5000 Submitted a PR fixing the issue here https://github.com/bentoml/BentoML/pull/500

Since this bug breaks the docker API server run, we will follow up with a minor release 0.6.1 to include this bug fix. Thanks again for reporting the issue.

parano:
Release 0.6.1 is out now including the fix in #500 @pdoyle5000

PR #500, opened 2020-01-24
[ISSUE-499] bentoml serve_gunicorn wrong parameter 

parano:
This PR fixes the issue starting gunicorn server, as reported here: 
https://github.com/bentoml/BentoML/issues/499


parano:
Adding a non-related fix to the PR to make it pass the tests. It's due to a change from the latest release of `pip` that breaks one of our unit tests.

PR #510, opened 2020-01-30
replace [Service] with .get(Service, None) in bentoml/deployment/sage…

HenryDashwood:
…maker/__init__.py

Currently BentoML throws a key error if a user has certain roles attached to their account. For instance  Cognito_PollyPoolAuth_Role does not have a "Service" field. It has a "Federated" field. This change avoids the key error issue. 

No

How was this patch tested?
--------------------------
I have been able to deploy successfully with this change implemented.


HenryDashwood:
So today I learnt that if you fork a repository, make a pull request, and then make more changes, those subsequent changes also get added to the PR... :/

PR #517, opened 2020-02-10
Add handler configs to BentoServiceMetadata proto and bentoml.yml file

parano:


bentoml/handlers/dataframe_handler.py
```diff
@@ -67,7 +67,7 @@ class DataframeHandler(BentoHandler):
     """
 
     def __init__(
-        self, orient="records", output_orient="records", typ="frame", input_dtypes=None
+        self, orient="records", output_orient="records", typ="frame", input_dtypes=[]
```
hrmthw: https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html#anti-pattern

parano: 🤯🤯🤯

Issue #512, opened 2020-02-04
Problem importing handlers from external files using bentoml serve


omrihar:
**Describe the bug**
I have a custom handler I created to tackle the problem with CORS support in lambda deployments. It resides in a second file which I import in my main service definition file. This file also gets copied into the service environment. The problem, however, is that when I try to serve the model using `bentoml serve` I get an import error saying this module does not exist.

I've been able to solve this temporarily by adding the current directory of the service module to the python path using `sys.path.append(os.path.dirname(__file__))`, which is of course not a long-term solution.

**To Reproduce**
1. Import anything from a self-written module in the script defining the service class.
2. Build the model
3. Serve it using `bentoml serve ...`

**Expected behavior**
The model should work.

**Environment:**
 - OS: Manjaro Linux
 - Python/BentoML Version Python 3.7.6, BentoML-0.6.1

**Additional context**
As discussed on the BentoML Slack channel.

parano:
Thanks for reporting the issue @omrihar!

It's actually a regression introduced when we refactored the service loading path, due to the lack of related tests.  Here's the PR  including the fix and adding tests: https://github.com/bentoml/BentoML/pull/518

parano:
Closing this issue now as #518 has been merged. The upcoming release 0.6.2 will include the fix in #518

PR #518, opened 2020-02-10
[ISSUE-512] Fix appending saved path to sys.path when loading BentoService

parano:
* Fixing issue #512 

This is a regression introduced when we refactored bentoml.yml for pip installed BentoService to load.


bentoml/bundler/loader.py
```diff
@@ -107,6 +107,7 @@ def load_bento_service_class(bundle_path):
 
     # Prepend bundle_path to sys.path for loading extra python dependencies
     sys.path.insert(0, bundle_path)
```
parano: keeping this line for pip installed BentoService to load python dependencies properly

PR #520, opened 2020-02-10
Improve `bentoml list` cli output

parano:


bentoml/cli/utils.py
```diff
@@ -127,8 +127,11 @@ def _format_deployment_age_for_print(deployment_pb):
         # we will not show the age for those deployments
         return None
     else:
-        deployment_duration = datetime.utcnow() - deployment_pb.created_at.ToDatetime()
-        return humanfriendly.format_timespan(deployment_duration)
+        return humanfriendly_age_from_datetime(deployment_pb.created_at.ToDatetime())
+
+
+def humanfriendly_age_from_datetime(dt, detailed=False, max_unit=2):
```
parano: setting max_unit default to 2, this also effects `bentoml deployment list` cc @yubozhao

yubozhao: got it, thanks for point it out

Issue #506, opened 2020-01-29
Add click-completion support to bentoml


omrihar:
**Is your feature request related to a problem? Please describe.**
BentoML has a very large CLI implemented in click. To find out what commands are available, I need to use `--help` repeatedly for many subcommands.

**Describe the solution you'd like**
Add hooks for shell completion. One such option is click-completion, which is this project: https://github.com/click-contrib/click-completion. Adding it requires simply initializing it in the bentoml project.

parano:
Contributed by @lintingzhen here: https://github.com/bentoml/BentoML/pull/534 🎉

Will be shipped in the coming release! I will add some related documentation shortly.

PR #534, opened 2020-02-15
feat: support bash completion

lintingzhen:

support bash completion for bentoml cli.

https://github.com/bentoml/BentoML/issues/506

How was this patch tested?
--------------------------

### DIFF
--- a/bentoml/cli/__init__.py
+++ b/bentoml/cli/__init__.py
@@ -20,6 +20,7 @@ import re
 import os
 import json
 import click
+import click_completion
 import tempfile
 import subprocess
 from pathlib import Path
@@ -48,6 +49,7 @@ from bentoml.proto import status_pb2
 from bentoml.utils import status_pb_to_error_code_and_message
 from bentoml.exceptions import BentoMLException
 
+click_completion.init()
 
 def escape_shell_params(param):
     k, v = param.split('=')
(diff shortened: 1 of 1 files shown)


bentoml/cli/__init__.py
```diff
@@ -20,6 +20,7 @@
 import os
 import json
 import click
+import click_completion
```
parano: Thanks for the contribution @lintingzhen!

The CI tests are actually failing because of missing this library in a test environment.

Could you make it an optional dependency?
```
try:
  import click_completion
except ImportError:
   click_completion.init()
```
 and in the install command:
```
```
try:
  import click_completion
except ImportError:
  echo.error("'click_completion' is required for BentoML auto-completion, install it with `pip install click_completion`")
```
That way we don't need to add it to setup.py as a hard requirement and user can

**Comment by parano** [comment]
Could you also run the auto-formatting and linter script on your PR? `./scripts/format.sh` and `./script/lint.sh`, and make sure to address the issues pointed out by @pep8speaks  above?

Thanks

File: bentoml/cli/__init__.py (hunk 1 above)
Comment (lintingzhen) [reply bentoml/cli/__init__.py]: How to process `click_completion.DocumentedChoice` in argument?
```
    @bentoml_cli.command(
        help="Install shell command completion",
        short_help="Install shell command completion",
    )
    @click.option('--append/--overwrite',
                  help="Append the completion code to the file",
                  default=None)
    @click.argument('shell',
                    required=False,
                    type=click_completion.DocumentedChoice(click_completion.core.shells)
                    )
    @click.argument('path', required=False)
    def install_completion(append, shell, path):
```

File: bentoml/cli/__init__.py [hunk 2]
```diff
@@ -334,6 +343,28 @@ def serve_gunicorn(
...
+    @click.option('--append/--overwrite',
+                  help="Append the completion code to the file",
+                  default=None)
+    @click.argument('shell',
+                    required=False,
+                    type=shell_types
+                    )
+    @click.argument('path', required=False)
+    def install_completion(append, shell, path):
+        if 'click_completion' in sys.modules:
```
parano: simply ‘if click_completion:’ ?

lintingzhen: 用in sys.modules是不是更能表示是一个package，而不是一个普通的变量？

parano: sys.module can by modified by other python code before Loading this file where this variable is in local scope. It doesn’t make a big difference in this case but slightly more conventional

lintingzhen: You're right.

parano:
Thanks for the contribution and addressing comments @lintingzhen, merging now!

Issue #531, opened 2020-02-13
Auto pip dependencies extraction


parano:
Currently, BentoML users need to manually specify a list of pip dependency names when defining a BentoService class. It can be a tedious process especially for users that has a large list of dependencies. Tools like `pipreqs` are very handy but hard to directly integrate with BentoML. We want to propose a new feature that allow BentoML to automatically figure out the pip dependencies of a BentoService class and pin their versions.

The proposed API:

* Add `auto_pip_dependencies: Boolean` parameter to BentoServiceEnv class
* In `BentoServiceEnv#save`, infer all the pip dependencies and add them to the list before writing to `requirements.txt` file
* Previous attempt can be found here: https://github.com/bentoml/BentoML/compare/master...parano:bs-env-improvements?expand=1#

Issue #450, opened 2019-12-21
pip dependency version specified in BentoService '@env' should match package in current environment


parano:
When defining the pip dependencies of a BentoService, e.g.:

```
@env(pip_dependencies=['scikit-learn==0.22'])
class MyPredictionService(BentoService):
     ...
```

BentoML should verify that in the current python session, the scikit-learn package is presented and matches the specified version. A good place to add this verification would be:

https://github.com/bentoml/BentoML/blob/master/bentoml/service_env.py#L147

The ideal behavior is to show a warning to the user, that the specified dependency version is different from the version being used in the current python session.

lintingzhen:
The function `add_pip_dependencies` at https://github.com/bentoml/BentoML/blob/master/bentoml/service_env.py#L147 is not called anywhere.
So?

parano:
@lintingzhen yes we are not using it currently,  `BentoServiceEnv` is initialized in the `@env` decorator and it seems we don't want the user to modify it dynamically anyway, it might be a good idea to just remove those methods such as adding/removing pip dependencies or conda channel.

lintingzhen:
The function `add_pip_dependencies` at https://github.com/bentoml/BentoML/blob/master/bentoml/service_env.py#L147 is not called anywhere.
So?

PR #536, opened 2020-02-19
Support auto pip dependencies and check user dependency version

lintingzhen:

Support auto pip dependencies and check user dependency version.

https://github.com/bentoml/BentoML/issues/450
https://github.com/bentoml/BentoML/issues/531

How was this patch tested?
--------------------------

### DIFF
--- a/bentoml/service_env.py
+++ b/bentoml/service_env.py
@@ -17,6 +17,7 @@ from __future__ import division
 from __future__ import print_function
 
 import os
+import logging
 from sys import version_info
 import stat
 from pathlib import Path
@@ -24,6 +25,11 @@ from pathlib import Path
 from ruamel.yaml import YAML
 
 from bentoml.configuration import get_bentoml_deploy_version
+from bentoml.utils.pip_pkg import EPP_PKG_NOT_EXIST, EPP_PKG_VERSION_MISMATCH, \
+    parse_requirement_string, verify_pkg
+
+
+logger = logging.getLogger(__name__)
 
 PYTHON_VERSION = "{major}.{minor}.{micro}".format(
     major=version_info.major, minor=version_info.minor, micro=version_info.micro
@@ -100,6 +106,8 @@ class BentoServiceEnv(object):
         bentoml_deploy_version = get_bentoml_deploy_version()
         self._pip_dependencies = ["bentoml=={}".format(bentoml_deploy_version)]
         if pip_dependencies:
+            for dependency in pip_dependencies:
+                self.check_dependency(dependency)
             self._pip_dependencies += pip_dependencies
 
         self.set_setup_sh(setup_sh)
@@ -108,6 +116,17 @@ class BentoServiceEnv(object):
             self.add_conda_channels(conda_channels)
         if conda_dependencies:
             self.add_conda_dependencies(conda_dependencies)
+
+    @staticmethod
+    def check_dependency(dependency):
+        name, version = parse_requirement_string(dependency)
+        code = verify_pkg(name, version)
+        if code == EPP_PKG_NOT_EXIST:
+            logger.warning('%s package does not exist in the current python '
+                           'session', name)
+        elif code == EPP_PKG_VERSION_MISMATCH:
+            logger.warning('%s package version is different from the version '
+                           'being used in the current python session', name)
 
     def get_conda_env_name(self):
         return self._conda_env.get_name()
--- a/bentoml/service_env.py
+++ b/bentoml/service_env.py
@@ -17,6 +17,7 @@ from __future__ import division
 from __future__ import print_function
 
 import os
+import sys
 import logging
 from sys import version_info
 import stat
@@ -26,7 +27,7 @@ from ruamel.yaml import YAML
 
 from bentoml.configuration import get_bentoml_deploy_version
 from bentoml.utils.pip_pkg import EPP_PKG_NOT_EXIST, EPP_PKG_VERSION_MISMATCH, \
-    parse_requirement_string, verify_pkg
+    parse_requirement_string, verify_pkg, seek_pip_dependencies
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


bentoml/service_env.py
```diff
@@ -164,7 +187,27 @@ def save(self, path):
...
         with open(requirements_txt_file, "wb") as f:
-            pip_content = "\n".join(self._pip_dependencies).encode("utf-8")
+            dependencies_map = {}
+            for dep in self._pip_dependencies:
+                name, version = parse_requirement_string(dep)
+                dependencies_map[name] = version
+            if self._auto_pip_dependencies:
+                if "__main__" in sys.modules:
+                    main_module = sys.modules["__main__"]
+                    if hasattr(main_module, "__file__"):
```
parano: most of the time this is being called from a jupyter notebook, where the main module does not have a __file__ attribute

bentoml/service_env.py
```diff
@@ -164,7 +187,27 @@ def save(self, path):
         requirements_txt_file = os.path.join(path, "requirements.txt")
 
         with open(requirements_txt_file, "wb") as f:
-            pip_content = "\n".join(self._pip_dependencies).encode("utf-8")
+            dependencies_map = {}
+            for dep in self._pip_dependencies:
+                name, version = parse_requirement_string(dep)
+                dependencies_map[name] = version
+            if self._auto_pip_dependencies:
+                if "__main__" in sys.modules:
```
parano: Instead of starting from the main module, it should start from the module where this `bento_servce` is defined. See the previous attempt where it passes in the `bento_service` instance when doing `bentoservice._env.save(..)`: https://github.com/bentoml/BentoML/compare/master...parano:bs-env-improvements?expand=1#diff-a6531af7171884b899814e04a018e143

parano: For example, here is the code that starts searching for local python modules(non-pypi packages, such as imported individual python files or directories with __init__.py), and 
https://github.com/bentoml/BentoML/blob/master/bentoml/bundler/bundler.py#L106

It's a bit hacky to get the module file this way but it works for most cases:
https://github.com/bentoml/BentoML/blob/master/bentoml/bundler/py_module_utils.py#L98

bentoml/service_env.py
```diff
@@ -164,7 +187,27 @@ def save(self, path):
...
+                    if hasattr(main_module, "__file__"):
+                        project_path = os.path.split(main_module.__file__)[0]
+                        reqs, unknown_modules = seek_pip_dependencies(project_path)
+                        for package_name, package_version in reqs:
+                            dependencies_map[package_name] = package_version
+                        for module_name in unknown_modules:
+                            logger.warning("unknown package dependency for module: %s",
+                                           module_name)
+
+            pip_content = "\n".join(
```
parano: did you run `./script/format.sh` to auto-format this?

bentoml/utils/pip_pkg.py
```diff
@@ -0,0 +1,169 @@
...
+
+__pip_pkg_map = {}
+__pip_module_map = {}
+
+EPP_NO_ERROR = 0
+EPP_PKG_NOT_EXIST = 1
+EPP_PKG_VERSION_MISMATCH = 2
+
+
+def collect_pip_pkg_info():
```
parano: 👍

bentoml/service_env.py
parano: this is also why in our notebook examples, we recommend users to use `%writefile` magic keyword to write the BentoService class definition to a separate file.

bentoml/service_env.py
```diff
@@ -157,14 +189,37 @@ def set_requirements_txt(self, requirements_txt_path):
...
+                    )
+                    dependencies_map.update(reqs)
+                    for module_name in unknown_modules:
+                        logger.warning(
+                            "unknown package dependency for module: %s", module_name
+                        )
+
+            pip_content = "\n".join(
+                map(
+                    lambda nv: "{}=={}".format(nv[0], nv[1]) if nv[1] else nv[0],
```
parano: nit: consider moving this to a separate function? or rename `nv` to make it more readable? 

```lambda pkg_name, pgk_version: "{}=={}".format(pkg_name, pgk_version) if pgk_version else pgk_name```

File: bentoml/utils/pip_pkg.py [hunk 6]
```diff
@@ -0,0 +1,202 @@
...
+                        (dist._key, dist._version)
+                    )
+                else:
+                    self.setuptools_module_set.add(mn)
+
+        self.nonlocal_package_path = set()
+
+        self.searched_modules = {}
+        for m in pkgutil.iter_modules():
+            if m[1] not in self.searched_modules:
```
parano: any reason we do not use `if m.name not in self.searched_modules` here?

bentoml/utils/pip_pkg.py
```diff
@@ -0,0 +1,202 @@
...
+                path = m[0].path
```
parano: similarily this can be `m.module_finder.path` right?

bentoml/utils/pip_pkg.py
```diff
@@ -0,0 +1,202 @@
...
+
+        self.dependencies = {}
+        self.unknown_module_set = set()
+        self.parsed_module_set = set()
+
+    def do(self):
+        self.seek_in_file(self.target_py_file_path)
+
+    def seek_in_file(self, file_path):
+        if not os.path.isfile(file_path):
```
parano: why do we need this check here?  and should it show a warning when it happens?

parano:
Thanks again @lintingzhen , looks great overall, feel free to address the comments in follow up PR, merging now!

bentoml/service_env.py
lintingzhen: dict#items()返回的是tuple list，没法这么用，我改成列表生成式。

bentoml/utils/pip_pkg.py
lintingzhen: No need to check, remove it.

PR #569, opened 2020-03-19
microbatch: set semaphore based on workers

hrmthw:

How was this patch tested?
--------------------------


bentoml/cli/__init__.py
```diff
@@ -346,6 +346,7 @@ def serve_gunicorn(
                     port=port,
                     num_of_workers=microbatch_workers,
                     prometheus_lock=prometheus_lock,
+                    model_server_workers=workers or 3,
```
parano: 3 is actually not the default value, I think it should use this function instead to get the default `workers` value:
https://github.com/bentoml/BentoML/blob/master/bentoml/server/utils.py#L29
https://github.com/bentoml/BentoML/blob/master/bentoml/server/gunicorn_config.py#L7

hrmthw: @parano updated

bentoml/marshal/marshal.py
```diff
@@ -122,10 +122,18 @@ class MarshalService:
...
+    def __init__(
+        self,
+        bento_bundle_path,
+        target_host="localhost",
+        target_port=None,
+        target_count=2,
+    ):
         self.target_host = target_host
         self.target_port = target_port
+        self.target_count = target_count
```
parano: add a quick inline comments about `self.target_count` and `self._target_sema`?

bentoml/marshal/dispatcher.py
```diff
@@ -55,17 +54,19 @@ def __init__(self, interval, max_size, task_concurrency=2):
...
     @property
-    @lru_cache(maxsize=1)
     def outbound_sema(self):
         '''
-        semaphore should be created after forked(in the current event loop)
+        semaphore should be created after process forked
         '''
-        return asyncio.Semaphore(self.task_concurrency)
+        if self._sema is None:
+            self._sema = self.shared_sema and self.shared_sema() or asyncio.Semaphore(1)
```
parano: consider this for readability?
```
self._sema = self.shared_sema() if self.shared_sema else asyncio.Semaphore(1)
```

bentoml/cli/__init__.py
```diff
@@ -346,6 +346,7 @@ def serve_gunicorn(
                     port=port,
                     num_of_workers=microbatch_workers,
                     prometheus_lock=prometheus_lock,
+                    model_server_workers=workers,
```
parano: is it possible to start GunicornBentoServer before  GunicornMarshalServer? that way you should be able to get the actual worker count via `gunicorn_app.cfg.workers`

hrmthw: `gunicorn_app.run` will block itself.

bentoml/marshal/dispatcher.py
hrmthw: The later is always better than the former. I always forget it.

bentoml/cli/__init__.py
parano: Ohh got it

parano: I think a slightly better way to do this is:
```
model_server_workder_count = workers or get_gunicorn_num_of_workers()
```
and then pass it to both GunicornBentoServer and GunicornMarshalServer

That also fixes a current issue, for example when user set the CLI arg `--workers=5`, it still prints something like "get_gunicorn_num_of_workers: 3, calculated by cpu count", which is wrong and very confusing

hrmthw: Oh yeah, great!

bentoml/cli/__init__.py
```diff
@@ -333,19 +335,23 @@ def serve_gunicorn(
...
             with reserve_free_port() as api_server_port:
                 marshal_server = GunicornMarshalServer(
-                    target_host="localhost",
-                    target_port=api_server_port,
                     bundle_path=bento_service_bundle_path,
                     port=port,
-                    num_of_workers=microbatch_workers,
+                    workers=microbatch_workers,
                     prometheus_lock=prometheus_lock,
+                    outbound_host="localhost",
```
parano: 👍

bentoml/server/gunicorn_config.py
```diff
@@ -1,11 +1,8 @@
 import atexit
 from multiprocessing.util import _exit_function
 
-from bentoml.server.utils import get_gunicorn_num_of_workers
 from bentoml.utils.usage_stats import track_server_stop
 
-workers = get_gunicorn_num_of_workers()
```
parano: 👏

parano:
LGTM, thanks for the update @hrmthw, merging now!

PR #571, opened 2020-03-19
json handler: add adaptive batching support

hrmthw:
TODOs:

* add warning about the breaking change of json handler
* add `Bento-Multi-Input-Flag`
* update examples

### DIFF
--- a/bentoml/handlers/json_handler.py
+++ b/bentoml/handlers/json_handler.py
@@ -29,6 +29,10 @@ from bentoml.handlers.base_handlers import BentoHandler, api_func_result_to_json
 from bentoml.handlers.utils import concat_list
 
 
+class BadResult:
+    pass
+
+
 class JsonHandler(BentoHandler):
     """JsonHandler parses REST API request or CLI command into parsed_json(a
     dict in python) and pass down to user defined API function
@@ -65,6 +69,7 @@ class JsonHandler(BentoHandler):
             except (json.JSONDecodeError, UnicodeDecodeError):
                 responses[i] = SimpleResponse(400, None, "not a valid json input")
             except Exception:  # pylint: disable=broad-except
+                responses[i] = SimpleResponse(500, None, "internal server error")
                 import traceback
 
                 traceback.print_exc()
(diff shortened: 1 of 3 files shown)


bentoml/handlers/json_handler.py
```diff
@@ -45,8 +50,42 @@ def handle_request(self, request, func):
...
+        responses = [bad_resp] * len(requests)
+
+        for i, request in enumerate(requests):
+            try:
+                raw_str = request.data
+                parsed_json = json.loads(raw_str)
+                instances_list[i] = parsed_json
+            except (json.JSONDecodeError, UnicodeDecodeError):
+                responses[i] = SimpleResponse(400, None, "not a valid json input")
+            except Exception:  # pylint: disable=broad-except
```
parano: any specific reason that we need this? I think the API sever side does a broad except and will handle unexpected errors

bentoml/handlers/json_handler.py
```diff
@@ -45,8 +50,42 @@ def handle_request(self, request, func):
...
+            raise ValueError(
+                "The return value with JsonHandler must be list of jsonable objects, "
+                "and have same length as the inputs."
+            )
+
+        results = [merged_result[s] for s in slices]
+
+        for i, result in enumerate(results):
+            result_str = api_func_result_to_json(result)
+            responses[i] = SimpleResponse(200, dict(), result_str)
```
parano: isn't this going to rewrite the  `SimpleResponse(400, None, "not a valid json input")` above?

parano: maybe in line 58 initialize it as:
```
        responses = [None] * len(requests)
```
and here you can do something like:
```
if responses[I] is None:
   responses[i] = SimpleResponse(200, dict(), result_str)
```

hrmthw: Oh, I see. I'll make sure these handlers are as robust as DataframeHandler did in later commits.

bentoml/handlers/json_handler.py
hrmthw: To catch the error so that other normal requests could be properly returned.

hrmthw:
@parano updated

bentoml/handlers/json_handler.py
parano: 👍 got it

PR #596, opened 2020-04-08
Add fastai2_model_artifact

HenryDashwood:

Adds a fastai2_model_artifact

No

How was this patch tested?
--------------------------


parano:
@HenryDashwood thanks for updating the PR, I assume this is ready for review now? I will try it out on my end later today and update with you if I had any feedback.

HenryDashwood:
Cheers! There's a slight differences when loading saving the model as we have to use Sylvain's context manager:
```python3
with remove_patches_path():
    saved_path = service.save()
```
Would it be better if this were handled by Bentoml internally?

parano:
@HenryDashwood I'm planning to make the following change in BentoML to avoid that issue with fastcore's patch:

Instead of doing `yaml.dump(y, Path("/tmp/test.yml"))`, we will use:
```
with Path('/tmp/test.yaml', 'wb', encoding='utf8') as fp:
    yaml.dump(y, fp)
```

I will create a PR for that soon and test it together with your PR. I think after that change, the user won't need to use the `remove_patches_path` any more.

HenryDashwood:
Ah that's perfect then!

Issue #604, opened 2020-04-17
Add support for Fasttext Models


GCHQResearcher83493:
[Fasttext](https://fasttext.cc/) produces small, efficient models for text classification and representation learning.  It'd make a useful addition to the set of frameworks supported.

I've written up a Fasttext model artifact and forked it [here](https://github.com/GCHQResearcher83493/BentoML/blob/fasttext_model_artifact/bentoml/artifact/fasttext_model_artifact.py), and will open a pull request if no one objects?

PR #605, opened 2020-04-17
Support for Fasttext

GCHQResearcher83493:
Added an Artifact class for Fasttext models.

#604 

How was this patch tested?
--------------------------
Passed bentoml test suite & was able to build/deploy fasttext models to an endpoint.

### DIFF
--- a/bentoml/artifact/fasttext_model_artifact.py
+++ b/bentoml/artifact/fasttext_model_artifact.py
@@ -1,0 +1,51 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import os
+
+from bentoml.artifact import BentoServiceArtifact, BentoServiceArtifactWrapper
+from bentoml.exceptions import MissingDependencyException
+
+try:
+    import fasttext
+except MissingDependencyException as e:
+    raise e
+
+
+class FasttextModelArtifact(BentoServiceArtifact):
+    """
+    Abstraction for saving/loading fasttext models
+    
+    Args:
+        name (str): Name for the artifact
+
+    Raises:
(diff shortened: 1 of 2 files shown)


bentoml/artifact/fasttext_model_artifact.py
```diff
@@ -0,0 +1,51 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import os
+
+from bentoml.artifact import BentoServiceArtifact, BentoServiceArtifactWrapper
+from bentoml.exceptions import MissingDependencyException
+
+try:
```
parano, line 10: Looks like this is causing the unit tests to fail. Because this file got loaded whenever `bentoml/artifact/__init__.py` was imported. We are planning to change how these artifact classes handle optional dependencies, but as for now, could you change it to importing fasttext within the load/save function?  similar to how the xgboost artifact is implemented: https://github.com/bentoml/BentoML/blob/master/bentoml/artifact/xgboost_model_artifact.py

parano:
Thanks for the PR @GCHQResearcher83493!

One comment about dynamically importing fasttest, otherwise this looks pretty good!

bentoml/artifact/fasttext_model_artifact.py
GCHQResearcher83493: Thanks for the suggestion, I've updated the artifact to import later on.

parano:
Thanks for updating the PR @GCHQResearcher83493! The changes look great! Merging now!

parano:
hi @GCHQResearcher83493, thanks again for the contribution, would you be interested to also help contribute an example notebook using the Fasttext artifact you've created here?  feel free to reach me in the BentoML slack channel if you're interested!

PR #610, opened 2020-04-25
[Doc] Add Heroku deployment guide

yubozhao:


docs/source/deployment/heroku.rst
```diff
@@ -0,0 +1,147 @@
+Deploying to Heroku
+===================
+
+Heroku is a popular platform as a service based on managed container system. It provides
```
parano, line 4: add something here about BentoML? So we can share this page directly on twitter?

PR #615, opened 2020-04-29
[TESTING] add postgres dependency to part of test_requires

yubozhao:
package `psycopg2-binary` was missing for test yatai_server Postgres


parano:
add it here instead: https://github.com/bentoml/BentoML/blob/master/e2e_tests/requirements.txt

Issue #631, opened 2020-05-06
botocore.exceptions.ClientError


jjneojiajun:
Hey BentoML Maintainers,

I am currently facing an issue deploying my model to AWS Lambda, although i followed the docs completely. I have received the botocore.exceptions.ClientError indicating there's (NoSuchBucket) when calling the GetBuckAcl Operations and InvalidLocationConstraint when calling the CreateBucket Operations. Any tips in regards to this? Is there like a parameter that i can use to create this bucket?

![image](https://user-images.githubusercontent.com/22229998/81140501-1f3d4280-8f9c-11ea-9b0e-ce2306701b50.png)

Thanks!

yubozhao:
hi @jjneojiajun what is the region you are trying to deploy to?  

There is an ongoing issue on boto3 that create an S3 bucket will fail for region `us-east-1` that I am not aware,  https://github.com/boto/boto3/issues/125.  I tested deploy lambda with the `us-east-1` region and received the same error as you did.  

I am working on a fix for creating S3 bucket for the `us-east-1` region.

If you are not deploying to that region, could you provide more info on which region you are deploying to?

jjneojiajun:
> hi @jjneojiajun what is the region you are trying to deploy to?
> 
> There is an ongoing issue on boto3 that create an S3 bucket will fail for region `us-east-1` that I am not aware, [boto/boto3#125](https://github.com/boto/boto3/issues/125). I tested deploy lambda with the `us-east-1` region and received the same error as you did.
> 
> I am working on a fix for creating S3 bucket for the `us-east-1` region.
> 
> If you are not deploying to that region, could you provide more info on which region you are deploying to?

Hi @yubozhao, yes I was trying to deploy to us-east-1 region. So I guess this is part of the bug mix that's required. I will close this thread then. Thank you! :)

spencersmith:
I'm still getting this issue after the first deployment to lambda. 

Running version 0.8.1 or bentoml. 

The first deployment worked for the most part, however subsequent deployments do not.

parano:
cc @yubozhao could you help take a look?

yubozhao:
Hi, @spencersmith, I tested deploying and updating with the quick start example BentoService with `us-east-1` region. Unfortunately, I wasn't able to reproduce the boto3 ClientError about InvalidLocationConstraint error for creating Bucket.

Can you share your boto3 version and what region you are deploying into?

parano:
[...]

PR #633, opened 2020-05-06
Fix create s3 bucket issue for us-east-1 region

yubozhao:
For creating s3 bucket for `us-east-1` region, you can not specify the region in the request https://github.com/boto/boto3/issues/125.

### DIFF
--- a/bentoml/utils/s3.py
+++ b/bentoml/utils/s3.py
@@ -76,15 +76,25 @@ def create_s3_bucket_if_not_exists(bucket_name, region):
             # `LocationConstraint` is set to `us-east-1` region.
             # https://github.com/boto/boto3/issues/125.
             # This issue still show up in  boto3 1.13.4(May 6th 2020)
-            # solution is not specify a region, if the region is `us-east-1`
-
-            if region != 'us-east-1':
+            try:
                 s3_client.create_bucket(
                     Bucket=bucket_name,
                     CreateBucketConfiguration={'LocationConstraint': region},
                 )
-            else:
-                s3_client.create_bucket(Bucket=bucket_name)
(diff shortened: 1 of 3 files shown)


bentoml/utils/s3.py
```diff
@@ -71,10 +71,20 @@ def create_s3_bucket_if_not_exists(bucket_name, region):
...
-                CreateBucketConfiguration={'LocationConstraint': region},
-            )
+
+            # NOTE: boto3 will raise ClientError(InvalidLocationConstraint) if
+            # `LocationConstraint` is set to `us-east-1` region.
+            # https://github.com/boto/boto3/issues/125.
+            # This issue still show up in  boto3 1.13.4(May 6th 2020)
+            # solution is not specify a region, if the region is `us-east-1`
+
+            if region != 'us-east-1':
```
parano, line 81:
```python
create_bucket_configuration = {'LocationConstraint': region} if region is not "east-1" else {}
s3_client.create_bucket(
                    Bucket=bucket_name,
                 CreateBucketConfiguration=create_bucket_configuration,
                )
```

e2e_tests/aws_lambda/test_lambda_deployment.py
```diff
@@ -25,7 +25,7 @@ def test_aws_lambda_deployment(iris_clf_service):
         '-b',
         iris_clf_service,
         '--region',
-        'us-west-2',
+        'us-east-1',
```
parano, line 28: no need to... it will just make our test run slower. If you want, add a unit test to cover the behavior

parano:
This is a potential fix for https://github.com/bentoml/BentoML/issues/631

PR #669, opened 2020-05-17
[CI] Python 3.8 Support

bojiang:


bojiang:
info: `https://github.com/python/cpython/pull/19917`

PR #679, opened 2020-05-17
Generate a presigned-url instead of presigned-post for S3

withsmilo:
(Thanks for sending a pull request! Please make sure to read the contribution guidelines, then fill out the blanks below.)

What changes were proposed in this pull request?
------------------------------------------------
1. I set the signature version of the S3 client `s3v4` as default. If user needs to change it, the user can change it via a new environment variable named `BENTOML_S3_SIGNATURE_VERSION`.
2. I changed from `generate_presigned_post` to `generate_presigned_url` because the former might have a bug for MinIO or other S3-compatible services.

Does this close any currently open issues?
------------------------------------------
maybe #676 

How was this patch tested?
--------------------------
Our company has an internal S3-compatible service. I succeeded in testing there.

PR #681, opened 2020-05-18
Fix the internal uri for S3

withsmilo:

This patch will resolve [the S3 issue](https://github.com/bentoml/BentoML/pull/679#issuecomment-629907043).

How was this patch tested?
--------------------------

### DIFF
--- a/bentoml/repository/__init__.py
+++ b/bentoml/repository/__init__.py
@@ -171,7 +171,11 @@ class _S3BentoRepository(BentoRepositoryBase):
                 "Not able to get pre-signed URL on S3. Error: {}".format(e)
             )
 
-        return BentoUri(type=self.uri_type, uri=response)
+        return BentoUri(
+            type=self.uri_type,
+            uri='s3://{}/{}'.format(self.bucket, object_name),
+            additional_fields=response,
+        )
 
     def get(self, bento_name, bento_version):
         # Return s3 path containing uploaded Bento files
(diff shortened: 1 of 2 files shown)


parano:
Hi @withsmilo, are you able to run `bentoml serve IrisClassifier:latest` with this branch and your company's s3 service? I found that with AWS s3, after uploading the tar.gz file via the `generate_presigned_url`, I can download the file from s3 but trying to "untar" the file will fail. That might have been the reason we used `generate_presigned_post` initially, I will look into it a bit more.

withsmilo:
@parano 
I succeeded to download the file from s3, untar the file, and start a API server with this patch. The reason you failed to "untar" probably seems to be using `files = files` instead of `data = fileobj` when calling requests.

parano:
@withsmilo I see, just tested locally and that fixed the issue for uploading to AWS s3 as well, thanks for the quick fix!

parano:
There's another issue related to this - when loading the model, BentoML is currently not using the pre-signed get-object URL for the download. Instead, it uses boto3 directly and requires the client-side to have s3 credentials. I will follow up with a fix for that.

withsmilo:
@parano 
Actually, I also modified [bentoml/bundler/loader.py](https://github.com/bentoml/BentoML/blob/master/bentoml/bundler/loader.py#L48) to apply an `endpoint_url` to the `boto3.client`. However, I couldn't know if this was correct, so I didn't include that file to this patch. Would you please check?

parano:
Yes! I think after my upcoming PR, endpoint_url will no longer need to specified in the `bentoml/bundler/loader.py`, as the client will use the pre-signed URL generated [here](https://github.com/bentoml/BentoML/blob/10adaa75dbbfc1d4c82b511f2379edf16464d7ae/bentoml/repository/__init__.py#L186) instead. I will ping you to take a look once the PR is out!

PR #683, opened 2020-05-18
Use pre-signed S3 url for downloading saved bundle from s3 service

parano:
* BentoML's model registry is designed to support s3, and only the server-side need to be configured with s3 credentials
* Currently the client-side use boto3 for s3 file download, which requires the client to also have the s3 credentials in order to load or run a BentoML bundle saved in s3
* This PR fixes this issue by using a pre-signed s3 URL to download 
* Updated BentoUri Protobuf definition. Replaced `additional_fields` with just `presigned_url`

How was this patch tested?
--------------------------
* All e2e tests are passing on my local machine now
* Manually tested with AWS s3 and MinIO deployment,  both`save` and `load` works as expected

cc @withsmilo @yubozhao

### DIFF
--- a/bentoml/bundler/loader.py
+++ b/bentoml/bundler/loader.py
@@ -25,7 +25,6 @@ import tempfile
 from contextlib import contextmanager
 from urllib.parse import urlparse
 
-import boto3
 from bentoml.utils.s3 import is_s3_url
 from bentoml.utils.usage_stats import track_load_finish, track_load_start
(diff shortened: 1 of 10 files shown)


withsmilo:
@parano 
I tested your PR, which seems to work well with our s3-compatible service..

parano:
Thanks for helping with testing the PR @withsmilo!  I will run some more tests on my end tomorrow and create a new release!

Issue #697, opened 2020-05-19
The CLI 'bentoml delete' deletes the saved bundle, but shows an error message


withsmilo:
**Describe the bug**
I'm testing BentoML connected the S3-like service.
```bash
$ bentoml list
BENTO_SERVICE                         AGE                           APIS                       ARTIFACTS
IrisClassifier:20200519220101_721D08  56.55 seconds                 predict<DataframeHandler>  model<SklearnModelArtifact>, test_list<PickleArtifact>
IrisClassifier:20200519214443_46F9DD  17 minutes and 14.65 seconds  predict<DataframeHandler>  model<SklearnModelArtifact>, test_list<PickleArtifact>
```
When I tries to call `bentoml delete IrisClassifier:20200519214443_46F9DD`, the command showed me an error message.
```bash
$ bentoml delete -y IrisClassifier:20200519214443_46F9DD
Failed to delete Bento IrisClassifier:20200519214443_46F9DD INTERNAL:
BentoService IrisClassifier:20200519214443_46F9DD deleted
```
However, the `IrisClassifier:20200519214443_46F9DD` disappeared on the Yatai UI.
<img width="1241" alt="bentoml_yatai_200519" src="https://user-images.githubusercontent.com/3915543/82329949-15e0bb00-9a1d-11ea-9ec2-f17ad7ed1c64.png">
It also disappeared from our S3-like service.
<img width="708" alt="bentoml_s3_200519" src="https://user-images.githubusercontent.com/3915543/82329945-137e6100-9a1d-11ea-896a-56699e9b6252.png">
The deletion itself seems to have been successful. But why do I get an error message like that?

**Expected behavior**
The CLI `bentoml delete` should not show an error message.

**Environment:**
 - OS: MacOS Catalina 10.15.4
 - Python/BentoML Version : Python 3.6.10 / BentoML-0.7.7

**Additional context**
N/A

PR #700, opened 2020-05-19
[FIX] properly handle s3 delete_object response

yubozhao:

Fix https://github.com/bentoml/BentoML/issues/697

This PR adds additional handles to the s3 delete object response that is not documented.  The undocumented response was causing `bentoml delete` display internal error, even when the metadata and file obj are deleted.

https://github.com/bentoml/BentoML/issues/697

How was this patch tested?
--------------------------

### DIFF
--- a/bentoml/repository/__init__.py
+++ b/bentoml/repository/__init__.py
@@ -24,6 +24,8 @@ from pathlib import Path
 from abc import abstractmethod, ABCMeta
 from urllib.parse import urlparse
 
+from botocore.exceptions import ClientError
+
 from bentoml import config
 from bentoml.exceptions import YataiRepositoryException
 from bentoml.utils.s3 import is_s3_url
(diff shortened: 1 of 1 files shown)


bentoml/repository/__init__.py
```diff
@@ -202,15 +204,58 @@ def dangerously_delete(self, bento_name, bento_version):
...
+            #        'date': 'Tue, 19 May 2020 19:46:57 GMT',
+            #        'server': 'AmazonS3'
+            #     },
+            #     'RetryAttempts': 0
+            #   }
+            # }
+            # An open issue on github: https://github.com/boto/boto3/issues/759
+            if DELETE_MARKER in response:
+                return response[DELETE_MARKER]
+            elif 'ResponseMetadata' in response:
```
parano, line 235: there's no return value in this `elif` branch?

bentoml/repository/__init__.py
```diff
@@ -202,15 +204,58 @@ def dangerously_delete(self, bento_name, bento_version):
...
+                    self.s3_client.head_object(Bucket=self.bucket, Key=object_name)
+                    raise YataiRepositoryException(
+                        f'Fail to delete {bento_name}:{bento_version} in bucket: '
+                        f'{self.bucket}'
+                    )
+                except ClientError as e:
+                    error_response = e.response.get('Error', {})
+                    error_code = error_response.get('Code', 'Unknown')
+                    if error_code != '404':
+                        raise e
```
parano, line 249: should it return True here?

bentoml/repository/__init__.py
```diff
@@ -202,15 +204,58 @@ def dangerously_delete(self, bento_name, bento_version):
...
+            #       'x-amz-request-id': '332EE9F7AB555D2B',
+            #        'date': 'Tue, 19 May 2020 19:46:57 GMT',
+            #        'server': 'AmazonS3'
+            #     },
+            #     'RetryAttempts': 0
+            #   }
+            # }
+            # An open issue on github: https://github.com/boto/boto3/issues/759
+            if DELETE_MARKER in response:
+                return response[DELETE_MARKER]
```
parano, line 234: Could we make this and LocalRepository's behavior consistent? The return value ```response ['DELETE_MARKER']``` is not used, let's just use `return` here



```


```

bentoml/repository/__init__.py
parano: nvm, I think it should be consistent with LocalRepository implementation, see the other comment

bentoml/repository/__init__.py
parano: Could we add this?
```
if DELETE_MARKER in response and response[DELETE_MARKER] is False:
   logger.warning(f"BentoML has deleted service '{bento_name}:{bento_version}' from YataiService records,
but it failed to delete the saved bundle files stored in s3://{self.bucket}/{object_name}, the
files may have already been deleted by the user.")
```

bentoml/repository/__init__.py
```diff
@@ -202,15 +204,58 @@ def dangerously_delete(self, bento_name, bento_version):
...
+            elif 'ResponseMetadata' in response:
+                # Note: Use head_object to 'check' is the object deleted or not.
+                # head_object only try to retrieve the metadata without returning
+                # the object itself.
+                try:
+                    self.s3_client.head_object(Bucket=self.bucket, Key=object_name)
+                    raise YataiRepositoryException(
```
parano, line 241: Use the same warning message in the previous comment here? instead of raising an exception?

bentoml/repository/__init__.py
```diff
@@ -202,15 +204,58 @@ def dangerously_delete(self, bento_name, bento_version):
...
+                        f'Fail to delete {bento_name}:{bento_version} in bucket: '
+                        f'{self.bucket}'
+                    )
+                except ClientError as e:
```
parano, line 245: add an inline comment here `# expected ClientError with Code 404, as target object should be deleted and 'head_object' should raise`

bentoml/repository/__init__.py
```diff
@@ -202,15 +204,58 @@ def dangerously_delete(self, bento_name, bento_version):
...
+                    error_response = e.response.get('Error', {})
+                    error_code = error_response.get('Code', 'Unknown')
```
parano, line 247: nit: error_response.get('Code', None)

bentoml/repository/__init__.py
parano: Add an inline comment?
```
# 'Error code 404 means target file object does not exist, as expected after delete_object call'
```

parano:
```
if error_code == '404':
    # 'Error code 404 means target file object does not exist, as expected after delete_object call'
    return
else:
    # unexpected boto3 ClientError
    raise e
```

Issue #754, opened 2020-06-03
Automatic Update of BentoML Server Endpoint with new Model


ssakhavi:
**Is your feature request related to a problem? Please describe.**
I have a model that has been deployed via docker. I have a new model and would like to add a new endpoint with a new version number to access the new endpoint. I don't want to rebuild the docker image and I want it to be done automatically.

**Describe the solution you'd like**
Assuming that we've deployed the model using a container and the repository is mounted on storage (e.g. S3), is it possible to update the server without the need to do it manually? Something similar to what TensorFlow serving is doing. 

**Describe alternatives you've considered**
Tensorflow serving can automatically detect the presence of a new SavedModel as long it's in the watched folder. 

**Additional context**

parano:
Thanks for the suggestion @ssakhavi  we are looking to support this as part of https://github.com/bentoml/BentoML/issues/754

Note that there are in fact lots of drawbacks to the Tensorflow-serving workflow. It almost works exactly like the old days where web developers will upload their PHP files directly to a server over FTP and expect Apache to load those scripts and website got updated. It is extremely error-prone especially for something like BentoML where dependencies and service logic can also get updated.  It is very convenient for testing out a model in the development phase and that's the reason we plan to support it, we'd never suggest users use that for serious production workloads. Building a comprehensive CI/CD pipeline that rebuilds the docker image, tests the new image, and then uses something like canary deployment to update the production deployment is probably the right way to go.

PR #787, opened 2020-06-11
Hot reload with OpenAPI support

### LINKED ISSUE #758: BentoML Artifacts Refactor: Support unload/load/pack dynamically

There have been a number of user requests related to loading and unloading models artifacts dynamically lately and this is to add this support.

Use cases:
* Speed up BentoService development workflow - currently user always needs to figure out how to save the model and load it back, then test it with the BentoService class. Ideally, the user should be able to do that with BentoML's model artifact class and build their BentoService class iteratively.
* Online learning - a new model is being training on the fly
* Composing model pipeline dynamically and loading models lazily 
* Automatically load new versions of artifact trained from a local directory, similar to the TensorFlow serving design https://github.com/bentoml/BentoML/issues/754 

WIP

n/a

**Additional context**
n/a

akainth015:
## Description
Fixes #758 and adds hotswap support by switching to asynchronous operations for the server.

## Motivation and Context
Hotswapping trained models with BentoML is very useful for the reasons detailed in the issue above. This means they can be switched out even after the server has been started.

## How Has This Been Tested?
Still in progress

## Types of changes

- [ ] Documentation
- [ ] Test, CI, or build
- [ ] None

## Components (if applicable)
- [x] Model Server (mico-batching, logging, metrics, tracing, benchmark, OpenAPI)
- [x] YataiService (model management, deployment automation)
- [ ] Documentation

## Checklist:

### DIFF
--- a/bentoml/saved_bundle/config.py
+++ b/bentoml/saved_bundle/config.py
@@ -20,6 +20,7 @@ from pathlib import Path
 from ruamel.yaml import YAML
 
 from bentoml import __version__ as BENTOML_VERSION
+from bentoml import config
 from bentoml.configuration import get_bentoml_deploy_version
 from bentoml.utils import dump_to_yaml_str
 from bentoml.yatai.proto.repository_pb2 import BentoServiceMetadata
@@ -34,6 +35,8 @@ metadata:
 """
 
 logger = logging.getLogger(__name__)
+DEFAULT_MAX_LATENCY = config("marshal_server").getint("default_max_latency")
+DEFAULT_MAX_BATCH_SIZE = config("marshal_server").getint("default_max_batch_size")
 
 
 def _get_apis_list(bento_service):
@@ -180,8 +183,15 @@ class SavedBundleConfig(object):
                         else:
                             api_metadata.input_config[k] = v
                 else:
-                    api_metadata.mb_max_latency = api_config["mb_max_latency"]
-                    api_metadata.mb_max_batch_size = api_config["mb_max_batch_size"]
+                    if 'mb_max_latency' in api_config:
+                        api_metadata.mb_max_latency = api_config["mb_max_latency"]
+                    else:
+                        api_metadata.mb_max_latency = DEFAULT_MAX_LATENCY
+
+                    if 'mb_max_batch_size' in api_config:
+                        api_metadata.mb_max_batch_size = api_config["mb_max_batch_size"]
+                    else:
+                        api_metadata.mb_max_batch_size = DEFAULT_MAX_BATCH_SIZE
 
                 if "input_config" in api_config:
                     for k, v in api_config["input_config"].items():


parano:
Hi @akainth015 - #758 is a work-in-progress proposal to introduce setup/loading/unloading hooks to the artifacts API so that individual artifacts can implement those methods and allow more customization. I will fill in more details on the design proposal and use cases on that issue soon. 

I'm not sure building an async version of BentoAPIServer solves that issue, but it is an interesting idea. BentoML currently has an async layer built with aiohttp that does request handling and micro-batching(https://docs.bentoml.org/en/latest/guides/micro_batching.html).  I think this can be useful when the user does not need micro-batching. Although the user may lose the ability to run multiple model workers.

bentoml/server/bento_api_server.py
```diff
@@ -41,17 +40,20 @@
...
-<div id="swagger-ui-container"></div>
-<script src="/static/swagger-ui-bundle.js"></script>
-<script>
-    SwaggerUIBundle({{
-        url: '{url}',
-        dom_id: '#swagger-ui-container'
-    }})
-</script>
+    <div id="swagger-ui-container"></div>
+    <script src="https://unpkg.com/swagger-ui-dist@3/swagger-ui-bundle.js"></script>
```
parano, line 50: One benefit of not using CDN is that the user can use this web UI even without the internet. A few early users requested this a while back so we changed it including "swagger-ui-bundle.js" in /static and use it instead

akainth015: Yeah, I was having syntax errors with the local version of the script, not intended to be a permanent change (that's why I didn't change the style sheet)

akainth015:
> Hi @akainth015 - #758 is a work-in-progress proposal to introduce setup/loading/unloading hooks to the artifacts API so that individual artifacts can implement those methods and allow more customization. I will fill in more details on the design proposal and use cases on that issue soon.
> 
> I'm not sure building an async version of BentoAPIServer solves that issue, but it is an interesting idea. BentoML currently has an async layer built with aiohttp that does request handling and micro-batching(https://docs.bentoml.org/en/latest/guides/micro_batching.html). I think this can be useful when the user does not need micro-batching. Although the user may lose the ability to run multiple model workers.

Thank you for letting me know about all this! I am starting to see that just using an asynchronous server does not fix the inherent issue, but I have a few other ideas for how to accomplish this. If you have already had some thoughts, I would love to hear them!
1. Make the service thread safe, so that the API server can be run in a background thread
2. Probably no code changes required, but a packed service can be restarted behind a load balancer if there are multiple instances of the server and switched out with zero downtime using a [Kubernetes rolling update](https://kubernetes.io/docs/tutorials/kubernetes-basics/update/update-intro/)

parano:
hi, @akainth015  I think with tools like Kubernetes rolling update and other blue/green deployment tools available, it is already possible for DevOps to swap the BentoML API server container behind an endpoint and it is the recommended approach, see https://github.com/bentoml/BentoML/issues/754#issuecomment-640177122  

IMHO the hot loading model implementation within the python process can be a simple blocking call, as it is meant for easy development/testing rather than being used in mission-critical production environments. In which case, a comprehensive CI/CD workflow is always preferred.

From what I heard, the other issues mentioned in https://github.com/bentoml/BentoML/issues/758 especially reloading artifacts while developing BentoService is actually a lot more important for the project, because the user may need to re-train the model and re-run the packaging code every time they make some code change in their BentoService code. And it slows down the development workflow a lot.  I will put some notes from previous discussions on that issue and share it with you soon.

Issue #775, opened 2020-06-10
add a retrieve cli cmd 


iancoffey:
**Is your feature request related to a problem? Please describe.**

I want to use bentoml in my CI pipeline, and to do so, I want to be able to have a `bentoml retrieve` type cmd. This command would retrieve the model and artifacts and put them in the specified directory. This creates a really interesting capability for the rest of the pipeline to make use of the service. EG:

`bentoml retrieve ModelServe:20200610145522_D08399 --dir /tmp/modelserve`

Would populate this target directory with the same contents as a local model repository would have, but wont do anything else with the model.

**Describe the solution you'd like**

Implement `bentoml retrieve` command

**Describe alternatives you've considered**

My alternative is to write a helper/wrapper tool, but it should probably just live in the cli imo.

**Additional context**
[slack convo](https://bentoml.slack.com/archives/CKC2JP3TM/p1591733350030700)

PR #810, opened 2020-06-17
Add retrieve cli cmd

iancoffey:
## Description

This PR adds a `retrieve` method to the cli, allowing the bento cli to retrieve a services artifacts from YataiService. This will small change enable new interesting use cases for CI.

closes https://github.com/bentoml/BentoML/issues/775

## Motivation and Context

As noted in https://github.com/bentoml/BentoML/issues/775, this small addition will create some super interesting possibilities for CI pipelines.

## How Has This Been Tested?
Locally and on my developer k8s cluster

## Types of changes

- [ ] Documentation
- [ ] Test, CI, or build
- [ ] None

## Components (if applicable)
- [x] BentoService (model packaging, dependency management, handler definition)
- [ ] Documentation

## Checklist:

### DIFF
--- a/bentoml/cli/bento.py
+++ b/bentoml/cli/bento.py
@@ -248,14 +248,14 @@ def add_bento_sub_command(cli):
             _echo(
                 f'BentoService {name}:{version} not found - '
                 f'{error_code}:{error_message}',
-                CLI_COLOR_ERROR
+                CLI_COLOR_ERROR,
             )
             return
 
         if get_bento_result.bento.uri.s3_presigned_url:
             bento_service_bundle_path = get_bento_result.bento.uri.s3_presigned_url
         else:
-            bento_service_bundle_path =  get_bento_result.bento.uri.uri
+            bento_service_bundle_path = get_bento_result.bento.uri.uri
 
         safe_retrieve(bento_service_bundle_path, target_dir)
 
--- a/bentoml/cli/bento.py
+++ b/bentoml/cli/bento.py
@@ -257,6 +257,6 @@ def add_bento_sub_command(cli):
         else:
             bento_service_bundle_path =  get_bento_result.bento.uri.uri
 
-        bento_path = safe_retrieve(bento_service_bundle_path, target_dir)
+        safe_retrieve(bento_service_bundle_path, target_dir)
 
-        click.echo('Service %s artifact directory => %s' % (name, bento_path))
+        click.echo('Service %s artifact directory => %s' % (name, target_dir))
--- a/bentoml/cli/bento.py
+++ b/bentoml/cli/bento.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import click
+import os
 from google.protobuf.json_format import MessageToJson
 from tabulate import tabulate
 
@@ -25,6 +26,7 @@ from bentoml.yatai.proto import status_pb2
 from bentoml.utils import pb_to_yaml, status_pb_to_error_code_and_message
 from bentoml.utils.usage_stats import track_cli
 from bentoml.yatai.client import YataiClient
+from bentoml.saved_bundle import safe_retrieve
 
 
 def _print_bento_info(bento, output_type):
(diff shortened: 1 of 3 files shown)


bentoml/cli/__init__.py
```diff
@@ -420,6 +421,44 @@ def install_completion(append, shell, path):
...
             )
 
+    @bentoml_cli.command(
+        help='Retrieves BentoService artifacts into a target directory',
+        short_help="Retrieves BentoService artifacts into a target directory",
+    )
+    @conditional_argument(pip_installed_bundle_path is None, "bento", type=click.STRING)
+    @click.option(
+        '--target_dir',
+        help="Directory to put artifacts into. Defaults to pwd.",
```
parano, line 431: nit:  ''Default to pwd" => "Defaults to the current working directory"

bentoml/cli/__init__.py
```diff
@@ -420,6 +421,44 @@ def install_completion(append, shell, path):
                 "install it with `pip install click_completion`"
             )
 
+    @bentoml_cli.command(
```
parano, line 424: it might make more sense to move this to the `cli/bento.py`, which has all the CLI commands for managing saved bundle?

The line ```bento_service_bundle_path = resolve_bundle_path(bento, pip_installed_bundle_path)``` should be safe to remove as well because it is only required for CLI commands if user `pip install` a saved bundle, e.g. with the quick start example:

```bash
# run the quickstart guide code within the bentoml repository directory
python ./guides/quick-start/main.py

# pip installed the saved bundle
pip install $(bentoml get IrisClassifier:latest -q | jq -r ".uri.uri")

# now the CLI for this saved bundle will be available, it contains CLI commands defined in "create_bento_service_cli"
IrisClassifier run predict  --input="[[1,2,3,4]]"
```

bentoml/saved_bundle/loader.py
```diff
@@ -158,6 +159,25 @@ def load_bento_service_class(bundle_path):
...
+
+    Returns:
+        string: location of safe local path
+    """
+    if _is_remote_path(bundle_path):
+        with _resolve_remote_bundle_path(bundle_path) as local_bundle_path:
+            shutil.copytree(local_bundle_path, target_dir)
+            return target_dir
+
+    return bundle_path
```
parano, line 178: Shouldn't it also copy this local `bundle_path`  to the `target_dir`?

iancoffey: Thats a dood point...I suppose it should, for consistency sake. 👍

bentoml/cli/__init__.py
```diff
@@ -420,6 +421,44 @@ def install_completion(append, shell, path):
...
+            error_code, error_message = status_pb_to_error_code_and_message(
+                get_bento_result.status
+            )
+            raise BentoMLException(
+                f'BentoService {name}:{version} not found - '
+                f'{error_code}:{error_message}'
+            )
+            return
+
+        bento_service_bundle_path = resolve_bundle_path(
```
parano, line 454: After moving to `bento.py`, this can actually get rid of `resolve_bundle_path ` and use the `get_bento_result ` directly:

```
            if get_bento_result.bento.uri.s3_presigned_url:
                # Use s3 presigned URL for downloading the repository if it is presented
                return get_bento_result.bento.uri.s3_presigned_url
            else:
                return get_bento_result.bento.uri.uri
```

resolve_bundle_path also resolved when `bento` arg is a local path(e.g. `bentoml serve $DIR` vs `bentoml serve IrisClassifier:latest`), which is not applicable here

bentoml/cli/__init__.py
```diff
@@ -420,6 +421,44 @@ def install_completion(append, shell, path):
...
+            bento, pip_installed_bundle_path
+        )
+
+        bento_path = safe_retrieve(bento_service_bundle_path, target_dir)
+
+        click.echo('Service %s artifact directory => %s' % (name, bento_path))
```
parano, line 460: It looks like you were trying to achieve two things here:

1. if this bento is stored locally, return the path where it is stored
2. if this bento is stored remotely, download it and copy to target_dir

Would it make more sense to just add a separate command for printing the saved path?

Currently, we've been using `jq` to do this in our documentation:
```bash
$ bentoml get IrisClassifier:latest -q | jq -r ".uri.uri"

/Users/chaoyu/bentoml/repository/IrisClassifier/20200616144911_7CE204
```
Would be nice to add a command or option for this, for example:
```bash
$ bentoml get IrisClassifier:latest --print-location

/Users/chaoyu/bentoml/repository/IrisClassifier/20200616144911_7CE204
```
The CI pipeline can use the output to determine if this is a local path or if it needs to be downloaded with `retrieve` command:
```
$ bentoml retrieve IrisClassifier:latest --target-dir /foo/bar
```

iancoffey: I like this idea of adding `--print-location` - but since this PR doesn't depend on this feature (uri info is already exposed via get as you mentioned), perhaps this addition to `get` could be added in a new issue/pr? What do you think?

parano: Yes, that sounds great! I mentioned it just in case it is something useful for your CI pipeline use case. I will create a separate issue for that.

iancoffey: awesome, ty

bentoml/cli/__init__.py
iancoffey: Moved to bento.py, thanks

bentoml/saved_bundle/loader.py
iancoffey: Now it will also copy this local bundle_path to the target_dir

bentoml/cli/bento.py
```diff
@@ -218,3 +220,43 @@ def delete(bentos, yes):
...
+            return
+
+        if get_bento_result.bento.uri.s3_presigned_url:
+            bento_service_bundle_path = get_bento_result.bento.uri.s3_presigned_url
+        else:
+            bento_service_bundle_path =  get_bento_result.bento.uri.uri
+
+        bento_path = safe_retrieve(bento_service_bundle_path, target_dir)
+
+        click.echo('Service %s artifact directory => %s' % (name, bento_path))
```
parano, line 262: should this be  ``` click.echo('Service %s artifact directory => %s' % (name, target_dir))``` instead?  if safe_retrieve guarantee to retrieve bundle to target_dir

parano:
Thanks for updating the PR @iancoffey, everything looks great besides one minor comment! You also need to run `./dev/format.sh` to format the code changes, which should make the travis-ci pass.

bentoml/cli/bento.py
iancoffey: Hmmmm. It could be - they should hypothetically be the same. Ill update this.

parano: I was a bit confused cause I saw the `return bundle_path` line in the `safe_retrieve`, maybe we just need to change that to `return target_dir ` instead?

iancoffey: I removed the return var and just output target_dir instead - seems simple enough. LMK if there is a better way :)

Issue #747, opened 2020-06-01
Validate artifact name string


parano:
In BentoML, every artifact within a Service was given a unique name(unique within the Service itself).  The artifact name is used as the property name for accessing this artifact after loading from disk, e.g. `self.artifact.model`. The artifact name is also used as the sub-folder name in which related artifact files will be stored when saving a Service bundle, e.g. ``

Currently, when defining a Service, the artifact name can be defined with any string, even empty string or characters that may cause issues in other systems. BentoML should do a simple validation when the user initializes an artifact class, and make sure the artifact name is a valid string.

This can be added here https://github.com/bentoml/BentoML/blob/v0.7.8/bentoml/artifact/artifact.py#L36

parano:
Thanks @AlexDut for the contribution, this will be released in 0.8.2

PR #817, opened 2020-06-18
Artifact string name validation

AlexDut:
## Description

Add simple validation for artifact's 'name' attribute. As this name is used as a property to access the artifact, the string should respect Python's variable naming conventions. Those are the following:

- A variable name must start with a letter or the underscore character.
- A variable name cannot start with a number.
- A variable name can only contain alpha-numeric characters and underscores.
- Variable names are case-sensitive.

The check is done in `BentoServiceArtifact` __init__ method.
Some tests were added for the validation.

## Motivation and Context

Issue [#747](https://github.com/bentoml/BentoML/issues/747)

## How Has This Been Tested?

Following the `DEVELOPMENT.md` file:
```bash
$ ./travis/unit_tests.sh
```

## Types of changes

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation
- [x] Test, CI, or build
- [ ] None

## Components (if applicable)
- [x] Model Artifact (model serialization, multi-framework support)
- [ ] Documentation

## Checklist:

- [ ] My change requires a change to the documentation.

### DIFF
--- a/tests/artifact/test_bento_service_artifact.py
+++ b/tests/artifact/test_bento_service_artifact.py
@@ -1,0 +1,51 @@
+import pytest
+from bentoml.artifact import BentoServiceArtifact
+
+
+def test_valid_artifact_name():
+    name = "_test"
+    artifact = BentoServiceArtifact(name)
+    assert artifact.name == "_test"
+
+    name = "test"
+    artifact = BentoServiceArtifact(name)
+    assert artifact.name == "test"
+
+    name = "TEST00"
+    artifact = BentoServiceArtifact(name)
+    assert artifact.name == "TEST00"
+
+    name = "__"
+    artifact = BentoServiceArtifact(name)
+    assert artifact.name == "__"
+
+    name = "thisIs012Along__erTest"
+    artifact = BentoServiceArtifact(name)
+    assert artifact.name == "thisIs012Along__erTest"
+
+
+def test_unvalid_artifact_name():
+    name = ""
+    with pytest.raises(ValueError) as e:
+        artifact = BentoServiceArtifact(name)
+    assert "'name' must start with a letter" in str(e.value)
+
+    name = "_"
+    with pytest.raises(ValueError) as e:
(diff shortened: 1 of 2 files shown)


tests/artifact/test_bento_service_artifact.py
```diff
@@ -0,0 +1,51 @@
...
+    assert artifact.name == "thisIs012Along__erTest"
+
+
+def test_unvalid_artifact_name():
+    name = ""
+    with pytest.raises(ValueError) as e:
+        BentoServiceArtifact(name)
+    assert "'name' must start with a letter" in str(e.value)
+
+    name = "_"
```
parano, line 33: Looks like this case does fit the requirement description: `'name' must start with a letter (a-z|A-Z) or underscore character and can only contain alpha-numeric characters and underscores`

Could you take a look and see if the `benotoml.utils.isidentifier` method work for this check?

parano: Actually, since we dropped support for PY2, we can probably just use the `name.isidentifier()` here. The error message could be improved as well, e.g.:

```
Artifact name must be a valid python identifier, a string is considered a valid identifier if it only contains alphanumeric letters (a-z) and (0-9), or underscores (_). A valid identifier cannot start with a number, or contain any spaces.
```

AlexDut: I changed the regex to `name.isidentifier()` and the error message with the one you suggested. 
Btw, `benotoml.utils.isidentifier` function was failing for the check and the check with accents (same for `name.isidentifier()`).
Those, 2 tests were not fitting the requirements of a python identifier. I was a bit to restrictive when I read the definition and actually you can define variables names with accents or using the Underscore(_) character even if it is not recommended.

parano: Thanks for updating the PR @AlexDut, surprised that characters with accents work for `isidentifier`, but I think technology as long as it is a valid python identifier we can use it as a property name for accessing the artifact.

PR #828, opened 2020-06-22
Support multiple images in the same request

### LINKED ISSUE #806: Multiple Image Inputs support

The new ImageInput with micro-batching support does not allow multiple image inputs, comparing to the old ImageHandler

Add `bentoml.adapters.MultiImageInput` input adapter type:

```
class MultiImageInput(BaseInputAdapter):
    """
    Warning: this is a not yet implemented, follow issue #806 for updates
    Transform incoming images data from http request, cli or lambda event into numpy
    array.
    Args:
        input_names (string[]]): A tuple of acceptable input name for HTTP request.
            Default value is (image,)
        accept_image_formats (string[]):  A list of acceptable image formats.
            Default value is loaded from bentoml config
            'apiserver/default_image_input_accept_file_extensions', which is
            set to ['.jpg', '.png', '.jpeg', '.tiff', '.webp', '.bmp'] by default.
            List of all supported format can be found here:
            https://imageio.readthedocs.io/en/stable/formats.html
        pilmode (string): The pilmode to be used for reading image file into numpy
            array. Default value is 'RGB'.  Find more information at:
            https://imageio.readthedocs.io/en/stable/format_png-pil.html
    Raises:
        ImportError: imageio package is required to use ImageInput
    Example usage:
        ```
        class MyService(BentoService):
            @bentoml.api(input=MultiImageInput(input_names=('imageX', 'imageY')))
            def predict(self, image_groups):
                for image_group in image_groups:
                    image_array_x = image_group['imageX']
                    image_array_y = image_group['imageY']
        ```
    """

```
see placeholder file here: https://github.com/bentoml/BentoML/pull/804

`bentoml.adapters.MultiFileInput`

**Additional context**
N/A

akainth015:
## Description

Supports multi-image requests through `multipart/form-data` for more complex use cases

## Motivation and Context

Fixes #806 

## How Has This Been Tested?

Documentation has been tested, but the input adapter itself needs testing both locally and as part of AWS Lambda

## Types of changes

- [ ] Documentation
- [ ] Test, CI, or build
- [ ] None

## Components (if applicable)
- [x] BentoService (model packaging, dependency management, handler definition)
- [ ] Documentation

## Checklist:

- [x] I have updated the documentation accordingly.


bentoml/adapters/multi_image_input.py
```diff
@@ -39,59 +50,96 @@ class MultiImageInput(BaseInputAdapter):
...
-
         :param event: AWS lambda event data of the python `dict` type
         :param func: user API function
         """
-        raise NotImplementedError
+        content_type = event.headers['content-type']
+        if "multipart/form-data" in content_type:
+            files = {}
+
+            request = decoder.MultipartDecoder(event.body, content_type)
```
akainth015, line 131: This relies on the content type including the multipart boundary, which should be a safe assumption, but I thought I should note it anyways.

setup.py
```diff
@@ -42,6 +42,7 @@
     "python-dateutil>=2.1,<2.8.1",
     "python-json-logger",
     "requests",
+    "requests_toolbelt",
```
akainth015, line 45: As far as I am able to tell, this is the easiest way to parse a multipart request with just its contents (since we don't have a `Request` from Lambda)

bentoml/adapters/multi_image_input.py
```diff
@@ -39,59 +50,96 @@ class MultiImageInput(BaseInputAdapter):
...
+
+    def read_file(self, name: str, file):
+        safe_name = secure_filename(name)
+        verify_image_format_or_raise(safe_name, self.accepted_image_formats)
+        return imread(file, pilmode=self.pilmode)
+
+    def handle_batch_request(
+        self, requests: Iterable[SimpleRequest], func
+    ) -> Iterable[SimpleResponse]:
+        raise NotImplementedError(
```
bojiang, line 95: We should implement this handler first. The reason we made former ImageInput(LegacyImageInput) deprecated is we should implement this.

bojiang: https://github.com/bentoml/BentoML/blob/master/bentoml/adapters/image_input.py#L163

Take this as an example. When running with micro-batch enabled, batching server would queue some requests and pass them to `handle_batch_request` as a batch(list of SimpleRequest). The `handle_batch_request` is expected to handle these in efficient way, taking the advantage of batching.
https://github.com/tensorflow/serving/tree/master/tensorflow_serving/batching

bentoml/adapters/multi_image_input.py
```diff
@@ -39,59 +50,96 @@ class MultiImageInput(BaseInputAdapter):
...
     def handle_aws_lambda_event(self, event, func):
         """Handles a Lambda event, convert event dict into corresponding
         data format that user API function is expecting, and use API
         function result as response
-
         :param event: AWS lambda event data of the python `dict` type
         :param func: user API function
         """
-        raise NotImplementedError
+        content_type = event.headers['content-type']
```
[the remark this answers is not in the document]
bojiang: cc @yubozhao

bentoml/adapters/multi_image_input.py
bojiang: To decode multipart requests, we can use werkzeug Request like this: https://github.com/bentoml/BentoML/blob/master/bentoml/adapters/image_input.py#L175

bentoml/adapters/multi_image_input.py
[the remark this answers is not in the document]
yubozhao: The request is coming from API gateway to the backend Lambda, AWS wraps the entire client request to the input event parameter of the backend Lambda function.  For headers, it is still in the format of the income client request header.

I found the documentation about this topic here: https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-input-format

akainth015:
All of the changes requested above should have been made. I will write tests for the adapter to make sure it checks out (and that code coverage does not go down)

bentoml/adapters/multi_image_input.py
```diff
@@ -39,61 +50,110 @@ class MultiImageInput(BaseInputAdapter):
...
-        class MyService(BentoService):
-            @bentoml.api(input=MultiImageInput(input_names=('imageX', 'imageY')))
-            def predict(self, image_groups):
-                for image_group in image_groups:
-                    image_array_x = image_group['imageX']
-                    image_array_y = image_group['imageY']
-        ```
+
+    >>> class MyService(BentoService):
+    >>> @bentoml.api(input=MultiImageInput(input_names=('imageX', 'imageY')))
```
yubozhao, line 55: This is great to have a code example in the docstring.  Let's also add accepted_image_formats to this.

yubozhao:
@akainth015 Everything looks good to me.  Let's add those tests and then I think we can merge it

bentoml/adapters/multi_image_input.py
parano: not many user will need the accepted_image_formats option, I think it's ok to keep it out

bentoml/adapters/multi_image_input.py
```diff
@@ -39,61 +50,110 @@ class MultiImageInput(BaseInputAdapter):
...
         input_names=("image",),
-        accept_image_formats=None,
+        accepted_image_formats=None,
         pilmode="RGB",
         is_batch_input=False,
         **base_kwargs,
     ):
         if is_batch_input:
-            raise ValueError('ImageInput can not accpept batch inputs')
+            raise ValueError('ImageInput can not accept batch inputs')
```
bojiang, line 71: MultiImageIput

bentoml/adapters/multi_image_input.py
```diff
@@ -39,61 +50,110 @@ class MultiImageInput(BaseInputAdapter):
         ImportError: imageio package is required to use ImageInput
```
bojiang, line 50: MultiImageInput

bojiang:
https://github.com/bentoml/BentoML/blob/master/bentoml/adapters/image_input.py#L180
Better to catch known exceptions, otherwise one request with invalid input will break other requests in the same batch.

bentoml/adapters/multi_image_input.py
```diff
@@ -39,61 +50,110 @@ class MultiImageInput(BaseInputAdapter):
...
+
+    def handle_batch_request(
+        self, requests: Iterable[SimpleRequest], func
+    ) -> Iterable[SimpleResponse]:
+        inputs = []
+        slices = []
+        for i, req in enumerate(requests):
+            content_type = req.headers['Content-Type']
+
+            if "multipart/form-data" not in content_type:
```
akainth015, line 102: > https://github.com/bentoml/BentoML/blob/master/bentoml/adapters/image_input.py#L180
> Better to catch known exceptions, otherwise one request with invalid input will break other requests in the same batch.

This is handled here. Neither the `ImageInput` nor the `MultiImageInput` handle the file body being an invalid image payload (possible if someone manages to submit an arbitrary file masquerading as a valid image file)

dev/multipart-unit-test-generator.js
```diff
@@ -0,0 +1,18 @@
+const http = require("http");
```
parano, line 1: Hi @akainth015 - looks like this is not really required by the unit test?  Is it only used for running manual tests?

bentoml/adapters/multi_image_input.py
```diff
@@ -11,21 +11,32 @@
...
-from bentoml.adapters.image_input import get_default_accept_image_formats
+from bentoml.adapters.image_input import (
+    get_default_accept_image_formats,
+    verify_image_format_or_raise,
+    _import_imageio_imread,
+)
+from bentoml.exceptions import BadInput
+from bentoml.marshal.utils import SimpleRequest, SimpleResponse
+
+imread = _import_imageio_imread()
```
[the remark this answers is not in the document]
parano: This will raise a MissingDependencyException for BentoML users that does not have `imageio` installed. The existing pattern was for avoiding that issue. `imageio` to BentoML is an optional dependency, only users who need ImageInput or MultiImageInput are required to install it.

bentoml/adapters/multi_image_input.py
```diff
@@ -36,64 +47,115 @@ class MultiImageInput(BaseInputAdapter):
...
+
+    >>> from bentoml import BentoService
+    >>> import bentoml
+    >>>
+    >>> class MyService(BentoService):
+    >>>     @bentoml.api(input=MultiImageInput(input_names=('imageX', 'imageY')))
+    >>>     def predict(self, image_groups):
+    >>>         for image_group in image_groups:
+    >>>             image_array_x = image_group['imageX']
+    >>>             image_array_y = image_group['imageY']
```
parano, line 62: Could you also add example client code here to demo how to submit a request with multiple images to this endpoint?

dev/multipart-unit-test-generator.js
akainth015: Not for manual tests, it generates `tests/multipart` for the `MultiImageInput` test, and I have put it into Git so theoretically we never need to run it again. I have included it just in case something happens and we end up needing to regenerate the file.

parano: if it's for automated unit tests, ideally this file should be generated from your python test code and only saved to a temporary directory

bentoml/adapters/multi_image_input.py
```diff
@@ -36,64 +47,130 @@ class MultiImageInput(BaseInputAdapter):
...
+
+    The endpoint could then be used with an HTML form that sends multipart data, like
+    the example below
+
+
+    >>> <form action="http://localhost:8000" method="POST"
```
parano, line 68: could you also add an example with curl command? I think this should be equivalent to: `curl -F imageX=@image_file_x.png -F imageY=@image_file_y.jpg http://localhost:8000`

tests/handlers/test_multi_image_input_adapter.py
```diff
@@ -0,0 +1,101 @@
...
+import string
+import random
+
+from werkzeug import Request
+
+from bentoml.adapters.multi_image_input import MultiImageInput
+from bentoml.marshal.utils import SimpleRequest
+
+
+def generate_multipart_body(image_file):
```
parano, line 11: Might be best to just use requests' multipart encode feature:

https://requests.readthedocs.io/en/master/user/advanced/#post-multiple-multipart-encoded-files

bentoml/adapters/multi_image_input.py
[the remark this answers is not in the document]
parano: Surfacing this comment again, I think this still needs to change, otherwise, it will break for users who do not have imageio installed in their environment.  Too bad there isn't a test to cover this yet, because `imageio` is in the test dependencies list. But one thing you can try is first uninstall `imageio`, and then run `import bentoml`, you should see an import error.

bentoml/adapters/multi_image_input.py
```diff
@@ -36,64 +47,133 @@ class MultiImageInput(BaseInputAdapter):
...
+     enctype="multipart/form-data">
```
parano, line 69: Looks like this line will break the formatting? this docstring will be used to generate the API docs page, would be nice if you could build the docs locally and check the docs formatting, here are related docs for doing that: https://github.com/bentoml/BentoML/blob/master/DEVELOPMENT.md#how-to-edit-run-build-documentation-site

bentoml/adapters/multi_image_input.py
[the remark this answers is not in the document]
akainth015: Currently, I am just using the same pattern (and function) as `ImageInput`, which is to raise a `bentoml.exceptions.MissingDepedencyException` if `imageio` is not present. Would you prefer different behaviour if the module is missing?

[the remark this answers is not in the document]
parano: The change I was requesting was to move this line to inside the `__init__` method of the MultiImageInput adapter class. The reason as I explained above, with the current implementation, users who did not have imageio installed, will run into an error when `import bentoml`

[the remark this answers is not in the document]
parano: actually - I just submitted a PR that added lazy loading, this won't be a problem anymore. Let's keep it this way as you have implemented and we can even migrate the others away from the `self.imageio` pattern later on  https://github.com/bentoml/BentoML/pull/878

parano:
@akainth015 there are still some linter errors you need to fix, everything else looks great!

parano:
Thanks for updating the PR @akainth015, merging now!

PR #830, opened 2020-06-23
Docker base image rename alpine -> slim

jackyzha0:
Small naming fix to change Alpine to Slim as per https://bentoml.slack.com/archives/CKC2JP3TM/p1592866038062500


parano:
Thanks @jackyzha0!

PR #839, opened 2020-06-24
Custom Web UI support for API Server

### LINKED ISSUE #739: Custom service web UI to replace default swagger UI
* Currently when starting a BentoML API model server, it starts a simple web UI in addition to the API endpoints. This Web UI is generated with swagger API schema, it can be used for data scientists to test the API endpoint from a browser. But not ideal for a non-technical user to use.

* This proposal is to allow BentoML user to provide a tarfile when building a BentoService saved bundle, which contains static content being hosted web page code, including a `index.html` and optionally other JS code.

* User javascript code should be able to send prediction request to the API model server for prediction

* This makes it possible for users to customize the web UI of their ML application for end-user to consume directly. It can be very powerful for building internal tools.

API design draft:
```python
@env(..)
@artifacts([...])
@web_static_files("./web_ui/*")
class MyPredictionService(BentoService)
     
         @api(...)
         def predict(...)
              pass
```

BentoML should provide a few template frontend projects for a user to get started with building a web UI easily. e.g. a basic HTML form submitting input data, a file upload window that can be used to drag-n-drop a image or audio file, a d3 visualization that visualized the prediction result, etc

An advanced version of this should allow user to declared UI elements right from their Python code. similar to how R Shinny works.

Korusuke:
## Description

Closes #739 

Requires a directory to be passed containing `index.html` and `static` dir.
**Note**: Any files not inside `static` will not be accessible. 
Example deployment: https://bentoml-her0ku-mtu5mtkwnjq0nao.herokuapp.com/

The example deployment is still WIP and has no loading indicator/error handling.
Please recommend changes for the demo deployment so the same can be used for documentation.

## Motivation and Context

Implements #739 

## How Has This Been Tested?

Tested locally and deployed on Heroku

## Types of changes

- [ ] Documentation
- [ ] Test, CI, or build

## Component(s) if applicable
- [ ] BentoML CLI

## Checklist:


bentoml/saved_bundle/loader.py
```diff
@@ -177,6 +177,7 @@ def load(bundle_path):
 
     svc_cls = load_bento_service_class(bundle_path)
     svc = svc_cls()
+    svc._path = bundle_path
```
Korusuke, line 180: I am not sure about this, there may be a better way of getting the bundle path.

bentoml/server/bento_api_server.py
```diff
@@ -106,16 +116,18 @@ def start(self):
...
         """
         The index route for BentoML API server
         """
-        return Response(
-            response=INDEX_HTML.format(url='/docs.json'),
-            status=200,
-            mimetype="text/html",
-        )
+        if hasattr(self.bento_service, '_static_files'):
+            return render_template('index.html')
```
Korusuke, line 124: In case custom UI is enabled, do we want to also have `swagger` UI on some different endpoint?

bentoml/saved_bundle/bundler.py
```diff
@@ -142,6 +142,13 @@ def save_to_dir(bento_service, path, version=None, silent=False):
             )
         )
 
+    # copy custom webui if enabled
```
flosincapite, line 145: My instinct is to add a property to the `bento_service`, e.g.

```
@property
def static_files(self):
    """Returns static files if present.

    Raises:
        AttributeError if no static files
    """
    return self._static_files
```

but the maintainers can advise you more.

Motivation: underscore-prefixed attributes are "private", so it doesn't make sense to access them externally like this.

bentoml/server/bento_api_server.py
```diff
@@ -83,13 +83,23 @@ def __init__(self, bento_service, port=_DEFAULT_PORT, app_name=None):
...
         self.port = port
         self.bento_service = bento_service
-
-        self.app = Flask(
-            app_name,
-            static_folder=os.path.join(
-                os.path.dirname(os.path.abspath(__file__)), 'static'
-            ),
-        )
+        if hasattr(bento_service, '_static_files'):
```
flosincapite, line 86: Ditto the above.

bentoml/server/bento_api_server.py
```diff
@@ -106,16 +116,18 @@ def start(self):
         # Flask dev server enabled threaded by default, disable it.
         self.app.run(port=self.port, threaded=False)
 
-    @staticmethod
-    def index_view_func():
+    def index_view_func(self):
```
flosincapite, line 119: The best replacement for `@staticmethod` is `@classmethod`:

```
@classmethod
def index_view_func(cls):
    [...]
```

`cls` refers to the class, just like `self` refers to the instance.

bentoml/server/bento_api_server.py
```diff
@@ -106,16 +116,18 @@ def start(self):
...
         """
         The index route for BentoML API server
         """
-        return Response(
-            response=INDEX_HTML.format(url='/docs.json'),
-            status=200,
-            mimetype="text/html",
-        )
+        if hasattr(self.bento_service, '_static_files'):
```
flosincapite, line 123: Ditto the above.

bentoml/service.py
```diff
@@ -299,6 +299,24 @@ def decorator(func):
...
+
+    Args:
+        static_files: path to directory containg index.html and static dir
+
+    >>>  @webui('./ui/')
+    >>>  class MyMLService(BentoService):
+    >>>     pass
+    """
+
+    def decorator(bento_service_cls):
```
flosincapite, line 313: So I think BentoService needs both

```
@property
def static_files(self):
  [...]
```

as defined above and

```
@static_files.setter
def static_files(self, static_files):
  [...]
```

Obviously, you can do what you're doing just fine--it's Python--but it's nice to define a more solid "contract" in cases like this.

bentoml/saved_bundle/bundler.py
Korusuke: Done!

bentoml/server/bento_api_server.py
Korusuke: I was not sure how to use `@classmethod` so I just converted it back to `@staticmethod` and used it in same way as done for `docs_view_func`

bentoml/service.py
Korusuke: > @static_files.setter
> def static_files(self, static_files):
>   [...]
> ```

I am not sure what you mean here, can you please explain a bit more😅

bentoml/server/bento_api_server.py
flosincapite: If `staticmethod` is an idiom in the codebase, then np.

But for the record, `classmethod` works as in my example above :). You just add the `@classmethod` decorator and then provide `cls` as the first argument (instead of `self` as for a normal member function).

Korusuke: I tried that, but then is it possible to access `bentoservice` object without passing it?

bentoml/service.py
flosincapite: https://stackoverflow.com/questions/17330160/how-does-the-property-decorator-work

You have a call somewhere like

```
service._static_files = foo
```

What I am suggesting is two changes: 

- in the class definition of BentoService, add

```
@static_files.setter
def static_files(self, static_files):
  self._static_files = static_files
```

and 

- instead of calling

```
service._static_files = foo
```

call

```
service.static_files = foo
```

Korusuke: Ah, ok. I get what you are saying now, but since all other functions are setting in the same way I guess we should be fine. And also it's only sets `_static_files` in one place. 
I'll change it to what you have suggested if needed once maintainers comments come in.

Thanks for the link! I understand decorators much better now!

bentoml/service.py
```diff
@@ -553,6 +574,10 @@ def artifacts(self):
     def env(self):
         return self._env
 
+    @property
+    def webui(self):
```
parano, line 578: web_

parano, line 578: `webui` and `_static_files` here seems to be referencing the same thing, maybe rename both to just `web_static_content`? This API technically can also be used to host other static files right?  e.g. it can be a just a markdown documentation page for using this API and allow end-user to download example input json/image files

bentoml/server/bento_api_server.py
parano: Yea, that'd be great for debugging, I think we can add a `/swagger.html` endpoint which always renders the swagger UI

bentoml/server/bento_api_server.py
```diff
@@ -107,15 +117,18 @@ def start(self):
         self.app.run(port=self.port, threaded=False)
 
     @staticmethod
-    def index_view_func():
+    def index_view_func(bento_service):
```
parano, line 120: consider refactor this to two different functions? `swagger_ui_func` and `user_defined_ui_func`, and decide which function to use in the `setup_routes` method?

bentoml/server/bento_api_server.py
parano: Does it actually need the `redner_template` method? the user can not really use the templating functionality right?  I'm more inclined to simply put all the static files under the `static` directory if they are just static files

bentoml/saved_bundle/loader.py
parano: you can use `svc._bento_service_bundle_path`

bentoml/server/bento_api_server.py
Korusuke: Makes sense!

so now the dir structure will be:
```
webui_dir/
     |-- js/
     |--css/
     |--imgs/
     |--index.html
```

instead of `render_template` now it uses `send_from_directory`

Korusuke: > Yea, that'd be great for debugging, I think we can add a `/swagger.html` endpoint which always renders the swagger UI

In my opinion, `/docs` endpoint would be better for `swagger UI` since this is the same as what `fastapi` uses. And also if in future we change from `swagger` to something else, the path can remain the same. 

So `/docs` for the HTML page and `/docs.json` for openapi json.

What do you think?

parano: @Korusuke Yea I think `/docs` for the UI page and `/docs.json`for the json sounds great!

parano: What's the reason for `send_from_directory` instead of just the `static_folder` approach?

Korusuke: > What's the reason for `send_from_directory` instead of just the `static_folder` parameter?

For serving `index.html` on `/` endpoint. if we don't use `send_from_directory` then only `/index.html` will work

parano: I believe it should serve all files under the static_folder directory not just index.html, you may also need to specify static_url_path: `app = Flask(__name__, static_folder='../build', static_url_path='/')`
and in your index view function do something like ```app.send_static_file('index.html')```

parano: I think this way it is more friendly to the web frontend developer, for example, if the user creates the frontend project with something like https://github.com/facebook/create-react-app or use webpack to build their js project, they can simply supply the built `dist` directory for BentoML to serve as static files, and everything should just work.

bentoml/server/bento_api_server.py
flosincapite: Aha, sorry, I hadn't seen that. Yeah, you did the right thing :)

bentoml/server/bento_api_server.py
Korusuke: > I believe it should serve all files under the static_folder directory not just index.html, you may also need to specify static_url_path: `app = Flask(__name__, static_folder='../web_client', static_url_path='/')`
> and in your index view function do something like `app.send_static_file('index.html')`

yeh, this is working currently!

> Adding static_url_path might break the swagger js/css files tho, you may need to adjust accordingly

Fixed that using send_from_directory and changing swagger static dir name.

> I think this way it is more friendly to the web frontend developer, for example, if the user creates the frontend project with something like https://github.com/facebook/create-react-app or use webpack to build their js project, they can simply supply the built `dist` directory for BentoML to serve as static files, and everything should just work.

Yep, I am trying to achieve this...currently the only issue is serving `any_path/index.html` for  `any_path/`, `any_path/index.html` and also `any_path/index`.

trying to find some better solution instead of just writing down all the cases.

Korusuke: One another doubt is - should we allow user-defined routes to override the system defined (predefined by bento)?

currently, if the user-defined function name is `index` then the server doesn't start, so should we allow to override that?
this may provide some extra flexibility but may even be a bit confusing at times.

Korusuke:
There are a bunch of new changes to try and support everything. Below are some reasons for doing what I have done!

`send_from_directory`: it's just a more secure version of `send_static_file` which checks if the file actually belongs to the specified `dir`

`static_folder = None`: Without this, it was just not working as the inbuilt function interfered with `/<path:file_path>` route.

and lastly, the reason for setting up so many routes is to support almost all possible cases!
Supported paths for any type of static files:
`/styles.css`
`/styles.css`

Korusuke:
I think the tests failed since `bento_service._bento_service_bundle_path` is undefined in the test environment.

So should I add a check? or define `_bento_service_bundle_path` in the tests?

bentoml/server/bento_api_server.py
parano: @Korusuke  - I think any of the `@api` routes and bentoml routes such as `/healthz` and `/metrics` should take place before the static folder.  I thought that's already the case for the `static_folder` parameter with Flask?

> currently, if the user-defined function name is index then the server doesn't start, so should we allow to override that?

`.. the server doesn't start ..`  is that because it has added two rules handling the same path? Maybe we should validate the `@api` name and make sure it is not in the reserved name list, such as `index`, `healthz`, `metrics` etc.  Currently there's only a `isidentifier` check, you can add this validation here: https://github.com/bentoml/BentoML/blob/v0.8.2/bentoml/service.py#L311

Korusuke: Okey great!
should I include this in the same PR or create another issue/PR for this?

bentoml/server/bento_api_server.py
```diff
@@ -83,12 +84,16 @@ def __init__(self, bento_service, port=_DEFAULT_PORT, app_name=None):
 
         self.port = port
         self.bento_service = bento_service
+        self.app = Flask(app_name, static_folder=None)
 
-        self.app = Flask(
+        self.static_path = os.path.join(
+            self.bento_service._bento_service_bundle_path,
```
parano, line 90: `_bento_service_bundle_path` is the only set when loading from a saved BentoService bundle.

Maybe you could add a method `get_web_static_content_path`:

```
def get_web_static_content_path(self):
   if self.bento_service._bento_service_bundle_path:
      return os.path.join(
            self.bento_service._bento_service_bundle_path, 'web_static_content'
      )
    elif: bento_service.web_static_content:
        return os.path.join(os.getcwd(), bento_service.web_static_content)
    else:
        return None / raise ?

```

parano: This should fix the error you see in the unit test, it will also allow the user to test the BentoAPIServer without saving and loading, this can be super useful for debugging in an interactive session or Jupyter notebook

bentoml/server/bento_api_server.py
parano: @Korusuke a smaller PR that adds the validation would be nice, but it's ok to just include it in this PR as well, whichever is easier for you

bentoml/server/bento_api_server.py
Korusuke: This is perfect!!
I wanted something like this but was not sure how to implement it😅

bentoml/server/bento_api_server.py
```diff
@@ -172,8 +200,29 @@ def setup_routes(self):
         /classify
         /predict
         """
+        if self.static_path:
+            self.app.add_url_rule(
```
parano, line 204: Could you add some inline comments about these three `url_rule` here? reading this code it is not immediately clear to me what it does

bentoml/service.py
```diff
@@ -553,6 +574,21 @@ def artifacts(self):
     def env(self):
         return self._env
 
+    @property
+    def web_static_content(self):
+        return self._web_static_content
+
+    @property
+    def get_web_static_content_path(self):
+        if self._bento_service_bundle_path and self.name:
```
parano, line 583: This should be `if self.web_static_content and self._bento_service_bundle_path` to make BentoService that does not have `web_static_content` to show an index page properly right?

self.name should already be presented

bentoml/service.py
```diff
@@ -299,6 +299,24 @@ def decorator(func):
     return decorator
 
 
+def webui_decorator(web_static_content):
+    """Define web UI static files required to be bundled with a BentoService
+
+    Args:
+        web_static_content: path to directory containg index.html and static dir
+
+    >>>  @webui('./ui/')
```
parano, line 308: I'm a bit inclined to just name it `@web_static_content('./ui/')` as the user-facing API, thoughts?

bentoml/server/bento_api_server.py
Korusuke: Ah, sorry...this is a neccessary evil😅 as flask doesn't handle it automagically.
```python
            # serve static files for any given path
            # this will also serve index.html from directory /any_path/
            # for path as /any_path/
            self.app.add_url_rule(
                "/<path:file_path>",
                "static_proxy",
                partial(self.static_serve, self.static_path),
            )
            # serve index.html from the directory /any_path
            # for path as /any_path/index
            self.app.add_url_rule(
                "/<path:file_path>/index",
                "static_proxy2",
                partial(self.static_serve, self.static_path),
            )
            # serve index.html from root directory for path as /
            self.app.add_url_rule(
                "/", "index", partial(self.index_view_func, self.static_path)
            )
```
Any changes to this? Sorry but I am still trying to improve my commenting and documentation skills.

bentoml/service.py
Korusuke: I am fine either way, will change it to `web_static_content`

parano: sounds good! I was thinking that users can potentially use this to host other files other than web UI files, e.g., using this to host test input files, JSON config file(some use cases require exposing server-side config to the client), or even pdf documentation, etc.

Korusuke: Makes sense, the only downside is users currently cannot specify custom `/` path. So by default `bento` will look for index.html.
So for example, if someone is trying to serve `index.json` at `/`, then this won't work but `/index.json` will work.

We can add option to customize this, but I am more inclined not to as for production deployment it is better to just use `nginx` or something as serving static files via flask is both slow and processor intensive.

bentoml/service.py
Korusuke: > This should be `if self.web_static_content and self._bento_service_bundle_path` to make BentoService that does not have `web_static_content` to show an index page properly right?

Yeh, sorry missed that in the final commit

> self.name should always be presented

I swear previously I was getting an error on `self.name` but now it's working😅
I think I may have used `_bento_service_name` while testing and cause that was giving error I replaced everything with `self.name` and then never checked it properly.

bentoml/server/bento_api_server.py
parano: @Korusuke that looks great, adding a little context here helps a lot!

bentoml/saved_bundle/bundler.py
```diff
@@ -142,6 +142,17 @@ def save_to_dir(bento_service, path, version=None, silent=False):
...
 
+    # copy custom web_static_content if enabled
+    if bento_service.web_static_content:
+        src_web_static_content_dir = os.path.join(
+            os.getcwd(), bento_service.web_static_content
+        )
+        dest_web_static_content_dir = os.path.join(
+            module_base_path, 'web_static_content'
+        )
+        # os.mkdir(dest_web_dir)
```
parano, line 153: remove?

bentoml/saved_bundle/bundler.py
```diff
@@ -142,6 +142,17 @@ def save_to_dir(bento_service, path, version=None, silent=False):
             )
         )
 
+    # copy custom web_static_content if enabled
+    if bento_service.web_static_content:
+        src_web_static_content_dir = os.path.join(
+            os.getcwd(), bento_service.web_static_content
```
parano, line 148: add a `os.path.isdir(src_web_static_content_dir)` check here? , would be great to also support absolute path

bentoml/service.py
```diff
@@ -553,6 +574,21 @@ def artifacts(self):
     def env(self):
         return self._env
 
+    @property
+    def web_static_content(self):
+        return self._web_static_content
+
+    @property
+    def get_web_static_content_path(self):
```
parano, line 582: usually `property` won't have the verb `get_`,  I feel making this a method instead of a property would be better

parano:
@Korusuke finished a pass with a few minor comments, otherwise this looks great!  It would be great if you could add your web UI demo to the bentoml/gallery repository, it looks super cool!

bentoml/saved_bundle/bundler.py
Korusuke: Added, it already supports absolute path. `os.path.join` is very smart!

parano:
@Korusuke Thanks again for the awesome contribution, can't wait to try out building some ML-powered web apps with this feature. Merging now!

parano:
Would be great to add some integration test for this later, currently, the related test infra are still working-in-progress https://github.com/bentoml/BentoML/pull/855

PR #831, opened 2020-06-23
[CLI] Refactor CLI exception handling and usage tracking

yubozhao:
* make sure all CLI command exit with proper exit code
* Move track cli to BentoMLCmdGroup

## Description

* Make sure all commands exit with proper exit code
* Refactor track cli to BentoMLCommandGroup

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation
- [ ] Test, CI, or build
- [ ] None

## Components (if applicable)
- [ ] BentoService (model packaging, dependency management, handler definition)
- [ ] Model Artifact (model serialization, multi-framework support)
- [ ] Model Server (mico-batching, logging, metrics, tracing, benchmark, OpenAPI)
- [ ] YataiService (model management, deployment automation)
- [ ] Documentation

## Checklist:

- [ ] My code follows the bentoml code style, both `./dev/format.sh` and
  `./dev/lint.sh` script have passed
  ([instructions](https://github.com/bentoml/BentoML/blob/master/DEVELOPMENT.md#style-check-and-auto-formatting-your-code)).
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] My change requires a change in [bentoml/gallery](https://github.com/bentoml/gallery) example notebooks
- [ ] I have sent a pull request to [bentoml/gallery](https://github.com/bentoml/gallery) to make that change

PR #847, opened 2020-06-26
Containerize command

### LINKED ISSUE #696: Add a command to containerize the model API server
BentoML supports various serving clusters now e.g. K8s, KNative, Kubeflow and KFServing. However we should execute some Docker commands to containerize our model API server. ([Docs](https://docs.bentoml.org/en/latest/deployment/kubernetes.html#deploy-bentoservice-to-kubernetes)) How about adding a command to containerize our model API server to the BentoML instead of it?

```bash
$ bentoml containerize [--docker-username ID] [--push] IrisClassifier:latest
```

N/A

**Additional context**
N/A

jackyzha0:
## Description
* add implementation of `bentoml containerize` with `--push` and `--tag` flags

## Motivation and Context
* closes #696 
* make it easier to build and push Docker images from within BentoML

## How Has This Been Tested?
* locally

## Types of changes

- [ ] Documentation
- [ ] Test, CI, or build

## Component(s) if applicable
- [x] BentoML CLI

## Checklist:


bentoml/cli/bento.py
```diff
@@ -260,3 +270,100 @@ def retrieve(bento, target_dir):
         safe_retrieve(bento_service_bundle_path, target_dir)
 
         click.echo('Service %s artifact directory => %s' % (name, target_dir))
+
+    @cli.command(
+        help='Containerize given Bento into a Docker image',
+        short_help="Containerize given Bento into a Docker image",
+    )
+    @click.argument("bento", type=click.STRING)
```
yubozhao, line 278: We can utilize the callback option in Click.  It can be almost identical to the callback for bento as an option, you can find an example of it at https://github.com/bentoml/BentoML/blob/08defd2c461c614df8fe156c5b6c81aac354c969/bentoml/cli/azure_functions.py#L74

bentoml/cli/bento.py
```diff
@@ -260,3 +270,100 @@ def retrieve(bento, target_dir):
...
+
+        `bentoml containerize` command also supports the use of the `latest` tag
+        and will automatically use the last built version of your Bento.
+
+        // add stuff about --push
+        // add stuff about --docker-repository
+        """
+        if ':' not in bento:
+            _echo(f'BentoService {bento} invalid - specify name:version')
+            return
```
yubozhao, line 298: Remove this, if using the callback option mentioned above

bentoml/cli/bento.py
```diff
@@ -260,3 +270,100 @@ def retrieve(bento, target_dir):
...
+        name, version = bento.split(':')
+
+        yatai_client = YataiClient()
+
+        track_cli('bento-containerize')
```
yubozhao, line 303: Remove this.  #831 will centralize track at the end.

bentoml/cli/bento.py
```diff
@@ -260,3 +270,100 @@ def retrieve(bento, target_dir):
...
+        get_bento_result = yatai_client.repository.get(name, version)
+        if get_bento_result.status.status_code != status_pb2.Status.OK:
+            error_code, error_message = status_pb_to_error_code_and_message(
+                get_bento_result.status
+            )
+            _echo(
```
yubozhao, line 309: For now, print a message is fine. After #831 is merged, we want to raise a `BentoMLExcpetion` here, so we can properly catch it and raise ClickException and make the CLI exit with a proper return code.

bentoml/cli/bento.py
```diff
@@ -260,3 +270,100 @@ def retrieve(bento, target_dir):
...
+                f'BentoService {name}:{version} not found - '
+                f'{error_code}:{error_message}',
+                CLI_COLOR_ERROR,
+            )
+            return
```
yubozhao, line 314: For now, print a message is fine. After #831 is merged, we want to raise a BentoMLExcpetion here, so we can properly catch it and raise ClickException and make the CLI exit with a proper return code.

bentoml/cli/bento.py
```diff
@@ -260,3 +270,100 @@ def retrieve(bento, target_dir):
...
+        if tag != bento:
+            _echo(
+                f'Bento name or tag was changed to be Docker compatible. \n'
+                f'"{bento}" -> "{tag}"',
+                CLI_COLOR_WARNING,
+            )
+
+        try:
+            docker_api = docker.APIClient()
+            ensure_docker_available_or_raise()
```
yubozhao, line 337: Let's move this all the way to the top

bentoml/cli/bento.py
```diff
@@ -260,3 +270,100 @@ def retrieve(bento, target_dir):
...
+                    )
+                )
+        except (FileNotFoundError, subprocess.CalledProcessError) as error:
+            _echo(
+                f'Could not build Docker image: {error}', CLI_COLOR_ERROR,
+            )
+            return
+
+        _echo(
+            f'Built {tag}', CLI_COLOR_SUCCESS,
```
yubozhao, line 351: Let's provide better description. Maybe `Finished built {tag} for BentoService {bento}`

bentoml/cli/bento.py
```diff
@@ -260,3 +270,100 @@ def retrieve(bento, target_dir):
...
+                )
+                return
+            with Spinner(f"Pushing docker image to {tag} "):
+                _echo_docker_api_result(
+                    docker_api.push(
+                        repository=name, tag=version, stream=True, decode=True,
+                    )
+                )
+            _echo(
+                f'Pushed {tag}', CLI_COLOR_SUCCESS,
```
yubozhao, line 368: Similar to the built message, let's provide more detail

bentoml/cli/utils.py
```diff
@@ -126,6 +155,14 @@ def _format_deployment_age_for_print(deployment_pb):
         return humanfriendly_age_from_datetime(deployment_pb.created_at.ToDatetime())
 
 
+def humanfriendly_bytes(num):
+    step_unit = 1000  # apparently docker uses 1000?
+    for x in ['B', 'KB', 'MB', 'GB', 'TB']:
+        if num < step_unit:
+            return "%3.2f %s" % (num, x)
+        num /= step_unit
```
yubozhao, line 163: You can try use the package humanfriendly. It is already part of the required modules for BentoML.

https://github.com/xolox/python-humanfriendly#a-note-about-size-units

bentoml/cli/bento.py
jackyzha0: Looks like Python Docker SDK will give an error if Docker isn't installed anyways, so I removed it because it was redundant

bentoml/cli/utils.py
```diff
@@ -92,6 +93,43 @@ def parse_key_value_pairs(key_value_pairs_str):
...
+def _echo_docker_api_result(docker_generator):
+    layers = {}
+    for line in docker_generator:
+        if "stream" in line:
+            cleaned = line['stream'].rstrip("\n")
+            if cleaned != "":
+                _echo(cleaned)
+        if "status" in line and line["status"] == "Pushing":
+            progress = line["progressDetail"]
+            layers[line["id"]] = progress["current"], progress["total"]
```
flosincapite, line 105: When you get multiple lines with the same `"id"` key, does `progress["total"]` ever change?

bentoml/cli/bento.py
```diff
@@ -237,3 +245,112 @@ def retrieve(bento, target_dir):
...
+        tag = f"{name}:{version}"
+        if tag != bento:
+            _echo(
+                f'Bento name or tag was changed to be Docker compatible. \n'
+                f'"{bento}" -> "{tag}"',
+                CLI_COLOR_WARNING,
+            )
+
+        try:
+            docker_api = docker.APIClient()
```
flosincapite, line 317: Nitpick: I would initialize the client outside of the `try` block--makes it more clear where the error can be raised. 

If initializing the client can raise an APIError, it makes sense for that to have its own try block, since the error source is different (and you'd want to say so when you raise the CLIException).

bentoml/cli/bento.py
```diff
@@ -237,3 +245,112 @@ def retrieve(bento, target_dir):
...
+
+        You can also optionally provide a `--push` flag, which will push the built
+        image to the Docker repository specified by the `--docker-repository`.
+
+        If you would like to push to Docker Hub, `--docker-repository` can just be
+        your Docker Hub username. Otherwise, the tag should include the hostname as
+        well. For example, for Google Container Registry, the `--docker-repository`
+        would be `[HOSTNAME]/[PROJECT-ID]`
+        """
+        name, version = bento.split(':')
```
flosincapite, line 284: Nitpick: this should live with the other docker-related utility functions in `cli/utils.py`. Rationale: even though this is really simple logic, if some other piece of code wants to split up Docker images in `name:tag` format, there should be a single function that does that work.

flosincapite: Another nice thing about a single function: if I just provide the name without a tag, that function could provide a reasonable default tag (e.g. "latest"), raise an exception, or do some other sensible thing.

bentoml/cli/utils.py
jackyzha0: No, but each 'loop' may not have more layers than the previous so we need to recompute the sum anyways. There could technically be some caching done here, but I don't think the performance gain is worth the sacrifice in readability

bentoml/cli/utils.py
```diff
@@ -92,6 +93,43 @@ def parse_key_value_pairs(key_value_pairs_str):
     return result
 
 
+def _echo_docker_api_result(docker_generator):
```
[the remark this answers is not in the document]
jackyzha0: From what I've seen, all `_echo` is commands are prefixed with _ so I'll keep this function in that form, but I'll change `_make_bento_name_docker_compatible` to `make_bento_name_docker_compatible`

bentoml/cli/bento.py
jackyzha0: Looks like a lot of other CLI functions do similar logic so I might make another issue and PR to refactor that later. Will keep it as is in current PR though

bentoml/cli/bento.py
```diff
@@ -237,3 +245,112 @@ def retrieve(bento, target_dir):
         safe_retrieve(bento_service_bundle_path, target_dir)
 
         click.echo('Service %s artifact directory => %s' % (name, target_dir))
+
+    @cli.command(
+        help='Containerize given Bento into a Docker image',
+        short_help="Containerize given Bento into a Docker image",
+    )
+    @click.argument(
+        "bento", type=click.STRING, callback=parse_bento_tag_callback,
```
yubozhao, line 254: Let's add the required flag since we need it to make docker image anyway

bentoml/cli/bento.py
```diff
@@ -237,3 +245,112 @@ def retrieve(bento, target_dir):
         safe_retrieve(bento_service_bundle_path, target_dir)
 
         click.echo('Service %s artifact directory => %s' % (name, target_dir))
+
+    @cli.command(
+        help='Containerize given Bento into a Docker image',
```
yubozhao, line 250: Better help message, maybe something along the line of `Containerize the given BentoService into a BentoML model service docker image.`? You probably can find a better way to phrase it

bentoml/cli/bento.py
jackyzha0: If it's an argument, it should already be required right?
image

bentoml/cli/bento.py
yubozhao: One thing we want to have a good handle on is the special tag `latest`.  

Right now, you can use it to get the latest BentoService, for example `,bentoml get IrisClassifier:latest`.  For this command,  I think we should be the image and also make sure to tag the image with both the `latest` tag and the actual version tag.  If a user decided to push to a docker registry, we should push both tags to the registry as well.

This way, the user can use the `latest` tag in their application and still have reference to the previous version of Docker image, if the `latest` tag had been overwritten.

bentoml/cli/bento.py
yubozhao: Oh it is an argument. Good catch!

bentoml/cli/bento.py
```diff
@@ -237,3 +245,112 @@ def retrieve(bento, target_dir):
         safe_retrieve(bento_service_bundle_path, target_dir)
 
         click.echo('Service %s artifact directory => %s' % (name, target_dir))
+
+    @cli.command(
```
yubozhao, line 249: Maybe we should rename the command to `create-docker-image` using the `name=`? I want something very simple and direct.  What do you think?

bentoml/cli/bento.py
```diff
@@ -237,3 +245,112 @@ def retrieve(bento, target_dir):
...
+                f'"{bento}" -> "{tag}"',
+                CLI_COLOR_WARNING,
+            )
+
+        docker_api = docker.APIClient()
+        try:
+            with Spinner(f"Building Docker image: {name}\n"):
+                _echo_docker_api_result(
+                    docker_api.build(
+                        path=bento_service_bundle_path, tag=tag, decode=True,
```
yubozhao, line 321: Can you write a test for building from the local path and remote s3 URL?

bentoml/cli/bento.py
jackyzha0: IMO `containerize` is pretty self-explanatory and can be left as-is. Thoughts, @parano?

bentoml/cli/bento.py
jackyzha0: I've tested this locally with both local and remote s3 URL but I have no clue how to go about writing a unit test for this ://

bentoml/cli/bento.py
```diff
@@ -237,3 +245,112 @@ def retrieve(bento, target_dir):
...
+
+            try:
+                with Spinner(f"Pushing docker image to {tag}\n"):
+                    _echo_docker_api_result(
+                        docker_api.push(
+                            repository=name,
+                            tag=version,
+                            stream=True,
+                            decode=True,
+                            auth_config=auth_config_payload,
```
parano, line 349: when username and password are not provided via the CLI, does `docker_api.push` pick up the local docker authentication information?

jackyzha0: yes

bentoml/cli/bento.py
```diff
@@ -237,3 +245,112 @@ def retrieve(bento, target_dir):
...
+        except docker.errors.APIError as error:
+            raise CLIException(f'Could not build Docker image: {error}')
+
+        _echo(
+            f'Finished building {tag} from {bento}', CLI_COLOR_SUCCESS,
+        )
+
+        if push:
+            if not docker_repository:
+                raise CLIException('Docker Registry must be specified when pushing.')
```
parano, line 333: consider moving this check to the beginning of this function and raise a click.BadParameter error there?

bentoml/cli/bento.py
```diff
@@ -237,3 +245,112 @@ def retrieve(bento, target_dir):
...
+    @click.option(
+        '-u', '--username', type=click.STRING, required=False,
+    )
+    @click.option(
+        '-p', '--password', type=click.STRING, required=False,
+    )
+    def containerize(bento, push, docker_repository, username, password):
+        """Containerize specified BentoService.
+
+        BENTO is the target BentoService to be containerized, referenced by its name
```
parano, line 270: This might be a bit inconsistent in the codebase right now, the `BENTO` parameter in CLI commands in `create_bento_service_cli` actually supports both `name:version` tag and a local file path to the saved bundle. we should probably also support that here.

Some users use the `BentoService#save_to_dir` method to skip BentoML's model registry and build their own model management workflow, for those users, they will need to run `containerize` by giving it the saved path

bentoml/cli/bento.py
parano: @flosincapite great suggestion, and indeed we should clean up lots of other places in the codebase that does this

parano: @yubozhao docker registry manages `latest` automatically, I don't think you need to tag or push to `latest` tag separately

bentoml/cli/click_utils.py
```diff
@@ -139,9 +139,8 @@ def wrapper(*args, **kwargs):
             try:
                 return func(*args, **kwargs)
             except BentoMLException as e:
-                raise ClickException(
-                    f'{cmd_group.name} {command_name} failed: {str(e)}'
-                )
+                msg = f'{cmd_group.name} {command_name} failed: {str(e)}'
+                raise ClickException(click.style(msg, fg='red'))
```
parano, line 143: 👍

bentoml/cli/utils.py
```diff
@@ -92,6 +93,43 @@ def parse_key_value_pairs(key_value_pairs_str):
...
+                humanfriendly.format_size(sum(cur)),
+                humanfriendly.format_size(sum(total)),
+            )
+            _echo(f"Pushed {cur} / {total}")
+        if "errorDetail" in line:
+            error = line["errorDetail"]
+            raise BentoMLException(f'Could not push Docker image: {error["message"]}')
+
+
+def make_bento_name_docker_compatible(name, tag):
```
parano, line 117: nit: the parameter here should be `(name, version)` right?

bentoml/cli/bento.py
```diff
@@ -237,3 +245,112 @@ def retrieve(bento, target_dir):
...
+        `bentoml containerize` command also supports the use of the `latest` tag
+        and will automatically use the last built version of your Bento.
+
+        You can also optionally provide a `--push` flag, which will push the built
+        image to the Docker repository specified by the `--docker-repository`.
+
+        If you would like to push to Docker Hub, `--docker-repository` can just be
+        your Docker Hub username. Otherwise, the tag should include the hostname as
+        well. For example, for Google Container Registry, the `--docker-repository`
+        would be `[HOSTNAME]/[PROJECT-ID]`
```
parano, line 282: Add to the doc: `for private docker registry, this would be 'myregistrydomain.com:5000/'`

bentoml/cli/bento.py
```diff
@@ -237,3 +245,112 @@ def retrieve(bento, target_dir):
...
+
+        if get_bento_result.bento.uri.s3_presigned_url:
+            bento_service_bundle_path = get_bento_result.bento.uri.s3_presigned_url
+        else:
+            bento_service_bundle_path = get_bento_result.bento.uri.uri
+
+        _echo(f"Found Bento: {bento_service_bundle_path}")
+
+        if docker_repository is not None:
+            name = f'{docker_repository}/{name}'
```
parano, line 305: we might need  `docker_repository.rstrip('/')` here?

bentoml/cli/bento.py
parano: This parameter is a little bit confusing because the `--docker-repository` parameter is not really the `repository` concept in docker registry.  I'm thinking just calling this `--docker-image-tag-prefix` instead?  

Alternatively we can make it two separate parameters: `--docker-image-tag` and `--docker-registry-hostname`. For docker hub user, the --docker-image-tag here can be `{dockerhub_username}/IrisClassifier`, we will use this tag name instead of the BentoService name. For private docker registry, this can be `--docker--registry-hostname=mydockerregistry.com:5000/` and for google cloud user: `--docker--registry-hostname=HOSTNAME  --docker-image-tag=PROJECT-ID/IrisClassifier`.

bentoml/cli/bento.py
```diff
@@ -237,3 +245,112 @@ def retrieve(bento, target_dir):
...
+        name, version = bento.split(':')
+        yatai_client = YataiClient()
+
+        get_bento_result = yatai_client.repository.get(name, version)
+        if get_bento_result.status.status_code != status_pb2.Status.OK:
+            error_code, error_message = status_pb_to_error_code_and_message(
+                get_bento_result.status
+            )
+            raise CLIException(
+                f'BentoService {name}:{version} not found - '
```
parano, line 293: Nit: `f'Failed to access BentoService {name}:{version}': `  - because the error_code here can be `internal error` instead of `not found` in some cases.

bentoml/cli/bento.py
parano: A third option is to just provide one `--docker-image-tag` option, and for users using `push`, they are responsible for prefixing the tag with a hostname. It seems easier to understand and matches the experience of using docker build from CLI. thoughts?

bentoml/cli/utils.py
```diff
@@ -92,6 +93,43 @@ def parse_key_value_pairs(key_value_pairs_str):
...
+            layers[line["id"]] = progress["current"], progress["total"]
+            cur, total = zip(*layers.values())
+            cur, total = (
+                humanfriendly.format_size(sum(cur)),
+                humanfriendly.format_size(sum(total)),
+            )
+            _echo(f"Pushed {cur} / {total}")
+        if "errorDetail" in line:
+            error = line["errorDetail"]
+            raise BentoMLException(f'Could not push Docker image: {error["message"]}')
```
yubozhao, line 114: Is there a similar error for building the docker image?

jackyzha0: `.build` seems to already raise a `docker.errors.APIError` but `.push` will instead stream `"errorDetail"`, so we should catch and raise an error here.

bentoml/cli/bento.py
jackyzha0: I think the third option here makes sense! I'll work on making that change

bentoml/cli/bento.py
parano: @jackyzha0 yea I think `containerize` works

bentoml/cli/bento.py
jackyzha0: support added :)

bentoml/cli/__init__.py
```diff
@@ -12,405 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import re
```
flosincapite, line 15: FYI: it looks like you moved this stuff to `bento_service.py`. You can use `git mv` or `git cp` when you migrate a lot of code from one file into a new file--that way, the new file will retain the old file's version history (and diffs will be more reasonable).

bentoml/cli/bento_service.py
```diff
@@ -0,0 +1,536 @@
...
+def escape_shell_params(param):
+    k, v = param.split('=')
+    v = re.sub(r'([^a-zA-Z0-9])', r'\\\1', v)
+    return '{}={}'.format(k, v)
+
+
+def get_name_version_from_tag(bento):
+    # returns name:version
+    if ":" in bento:
+        return bento.split(":")[:2]
```
flosincapite, line 64: Would be good to log a warning or something if there is more than one colon in the string--[:2] is a nice way to make things run smoothly, but doing this cleanup silently might have surprising results for a naive user.

bentoml/cli/bento_service.py
```diff
@@ -0,0 +1,536 @@
...
+    )
+    return
+
+
+def make_bento_name_docker_compatible(name, tag):
+    """
+    Name components may contain lowercase letters, digits and separators.
+    A separator is defined as a period, one or two underscores, or one or more dashes.
+
+    A tag name must be valid ASCII and may contain lowercase and uppercase letters,
```
flosincapite, line 103: Are name and tag validated somewhere beforehand?

tests/cli/test_cli.py
```diff
@@ -21,6 +27,34 @@ def generate_test_input_file():
     return file_path
 
 
+def assert_equal_lists(res, expected):
```
flosincapite, line 30: Hmm, my understanding is that

```
assert res == expected
```

will work fine for lists. If you're concerned about arbitrary types of iterable, you can do

```
assert list(res) == list(expected)
```

tests/cli/test_cli.py
```diff
@@ -21,6 +27,34 @@ def generate_test_input_file():
...
+    assert len(expected) == len(res)
+    assert all([a == b for a, b in zip(expected, res)])
+
+
+def test_make_bento_name_docker_compatible():
```
flosincapite, line 35: This is a good use case for [parametrized tests](https://docs.pytest.org/en/stable/parametrize.html#:~:text=%40pytest.,custom%20parametrization%20schemes%20or%20extensions). Internally, `pytest` will run a separate fixture with each set of parameters, so if there's a failure, you'll see exactly which set of parameters failed.

tests/cli/test_cli.py
```diff
@@ -60,6 +94,96 @@ def test_run_command_with_input_file(bento_bundle_path):
...
+        {'status': 'Waiting', 'progressDetail': {}, 'id': 'fcd8d39597dd'},
+        {
+            'errorDetail': {
+                'message': 'denied: requested access to the resource is denied'
+            },
+            'error': 'denied: requested access to the resource is denied',
+        },
+    ]
+
+    with pytest.raises(BentoMLException) as e:
```
flosincapite, line 140: Excellent :)

bentoml/cli/utils.py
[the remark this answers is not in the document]
parano: the `_echo` probably should be renamed as @flosincapite suggested, feel free to do that in follow up PR tho

bentoml/cli/bento_service.py
jackyzha0: no but we decided to just reformat it to be compatible and notify the user as per https://bentoml.slack.com/archives/CKC2JP3TM/p1593107935076500

bentoml/cli/bento_service.py
jackyzha0: will log a warning here

jackyzha0: oh wait actually it seems like the `resolve_bundle_path` function above this should catch this and raise an error

bentoml/cli/bento_service.py
```diff
@@ -0,0 +1,536 @@
...
+        By default, the `containerize` command will use the credentials provided by
+        Docker. You may provide your own through `--username` and `--password`.
+        """
+        bento_service_bundle_path = resolve_bundle_path(
+            bento, pip_installed_bundle_path
+        )
+
+        _echo(f"Found Bento: {bento_service_bundle_path}")
+
+        name, version = get_name_version_from_tag(bento)
```
parano, line 477: This will break when the `bento` parameter is a file path. You can get the name and version with `load_bento_service_metadata(bento_service_bundle_path)`

parano: This will allow us to use the containerize by giving it the saved_path:

`bentoml containerize $SAVED_PATH -t test_image`

parano: Currently it breaks with the follow error:
```
 $  bentoml containerize /Users/chaoyu/bentoml/repository/IrisClassifier/20200701222443_EC5EAB -t test_image
Found Bento: /Users/chaoyu/bentoml/repository/IrisClassifier/20200701222443_EC5EAB
  File "/Users/chaoyu/opt/miniconda3/envs/btml-dev/bin/bentoml", line 33, in <module>
    sys.exit(load_entry_point('BentoML', 'console_scripts', 'bentoml')())
  File "/Users/chaoyu/opt/miniconda3/envs/btml-dev/lib/python3.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/Users/chaoyu/opt/miniconda3/envs/btml-dev/lib/python3.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/Users/chaoyu/opt/miniconda3/envs/btml-dev/lib/python3.7/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/chaoyu/opt/miniconda3/envs/btml-dev/lib/python3.7/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/chaoyu/opt/miniconda3/envs/btml-dev/lib/python3.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/Users/chaoyu/workspace/BentoML/bentoml/cli/click_utils.py", line 140, in wrapper
    return func(*args, **kwargs)
  File "/Users/chaoyu/workspace/BentoML/bentoml/cli/click_utils.py", line 117, in wrapper
    return_value = func(*args, **kwargs)
  File "/Users/chaoyu/workspace/BentoML/bentoml/cli/click_utils.py", line 101, in wrapper
    return func(*args, **kwargs)
  File "/Users/chaoyu/workspace/BentoML/bentoml/cli/bento_service.py", line 484, in containerize
    name, version = make_bento_name_docker_compatible(name, version)
  File "/Users/chaoyu/workspace/BentoML/bentoml/cli/bento_service.py", line 110, in make_bento_name_docker_compatible
    tag = tag.lstrip(".-")[:128]
AttributeError: 'NoneType' object has no attribute 'lstrip'
```

bentoml/cli/bento_service.py
```diff
@@ -0,0 +1,536 @@
...
+        '{{ [ -f {pip_req} ] && pip install -r {pip_req} || echo "no pip '
+        'dependencies."; }} && {cmd}'.format(
+            env_name=env_name, env_file=env_path, pip_req=pip_req, cmd=command,
+        ),
+        shell=True,
+    )
+    return
+
+
+def make_bento_name_docker_compatible(name, tag):
```
parano, line 98: what's the difference between `name` and  `tag` here? It wasn't clear to me.

tag parameter here is the docker image version, not the entire docker image tag, right?

bentoml/cli/bento_service.py
```diff
@@ -0,0 +1,536 @@
...
+        short_help="Containerizes given Bento into a ready-to-use Docker image",
+    )
+    @click.argument("bento", type=click.STRING)
+    @click.option('-p', '--push', is_flag=True)
+    @click.option(
+        '-t',
+        '--tag',
+        help="Optional image tag. If not specified, Bento will generate one from "
+        "the name of the Bento.",
+        required=False,
```
parano, line 440: I feel if you could valid this `tag` parameter here with a validation callback here, the rest of the code could be simplified https://click.palletsprojects.com/en/7.x/options/#callbacks-for-validation

bentoml/cli/bento_service.py
```diff
@@ -0,0 +1,536 @@
...
+        )
+
+        _echo(f"Found Bento: {bento_service_bundle_path}")
+
+        name, version = get_name_version_from_tag(bento)
+        # use tag name and version where applicable
+        if tag is not None:
+            name, v = get_name_version_from_tag(tag)
+            if v is not None:
+                version = v
```
parano, line 482: add a log message here? `f"Tag {tag} does not specify an image version, using BentoService version {bento_service_version} as docker image version."`

bentoml/cli/bento_service.py
```diff
@@ -0,0 +1,536 @@
...
+
+        name, version = make_bento_name_docker_compatible(name, version)
+
+        full_tag = f"{name}:{version}"
+        if full_tag != bento:
+            _echo(
+                f'Bento tag was changed to be Docker compatible. \n'
+                f'"{bento}" -> "{full_tag}"',
+                CLI_COLOR_WARNING,
+            )
```
parano, line 492: Assuming you added some validation for the tag parameter, this part could be rewritten to something like this:
```
bs_metadata = load_bento_service_metadata(bento_service_bundle_path)

if ":" not in tag:
   _echo(f"Tag {tag} does not specify an image version, using BentoService version {bs_metadata.version} as docker image version.")
   version= to_docker_tag_compatible_string(bs_metadata.version)
   tag = f"{tag}:{version}"

_echo("Containerizing BentoService {bs_metadata.name}:{bs_metadata.version}, docker image tag: {tag}")

docker_api = docker.APIClient()
...



```

parano:
Great work @jackyzha0! I tried it out locally and it works great and I really love the UX.

I just finished a pass with a few minor comments. I think the most important one being fixing running containerize with a file path to saved bundle. As the PR is getting really large, I will go ahead and merge it, feel free to address those comments in a follow-up PR.

Issue #721, opened 2020-05-27
[TEST/CI] Add tests covering TensorflowSavedModelArtifact


parano:
Currently, there isn't any integration test covering the `TensorflowSavedModelArtifact`:

* Similar to PR #726, add a unit test assuming tensorflow pypi package installed, and create a BentoService with TensorflowSavedModelArtifact, test the major functionalities including save, load, run inference
* Build a docker container with the saved bundle and test the API server endpoint, see related PR #855 and #871 which have some utility function code that can be re-used for this
* Add a script `./travis/run-tensorflow-v2.2-tests.sh` that runs `pip install tensorflow==2.2` and `pytest tests/test_tensorflow_v2.2_model_artifact.py ...`
* Add another script `./travis/run-tensorflow-v1.14-tests.sh` that runs `pip install tensorflow==1.14` and `pytest tests/test_tensorflow_v1.14_model_artifact.py ...`

* Add two separate task in travis.yml that launches the two test scripts above https://github.com/bentoml/BentoML/blob/master/.travis.yml

joshuacwnewton:
Hi @parano. I'm a student in the MLH Fellowship program. I'd like to begin working on this issue. Thanks! :)

joshuacwnewton:
It looks like the requirements have updated, and #876 no longer fixes this. I will make the new changes to address the updated issue. :)

joshuacwnewton:
> Add another script ./travis/run-tensorflow-v1.14-tests.sh that runs pip install tensorflow==1.14 and pytest tests/test_tensorflow_v1.14_model_artifact.py ...

Two small notes:
* 1.15 is the last TF1 revision, I think
* Python 3.8 breaks tests using `pip install tensorflow==1.15`, see: https://github.com/tensorflow/tensorflow/issues/39768#issuecomment-632799481. I can stick to developing with 3.7 for now, but it might be worth keeping in mind for the future.

joshuacwnewton:
I am having some difficulty developing a simple TF 1.14 example. I have only used TF2, so it is a bit challenging to search through old TF1 documentation! :stuck_out_tongue: 

Because the test I am writing involves saving and loading a TF1 model artifact, I have run into this warning from Bento:

```bash
                WARNING - Importing SavedModels from TensorFlow 1.x.
                `outputs = imported(inputs)` is not supported in bento service due to
                tensorflow API.

                Recommended usage:

                ```python
                from tensorflow.python.saved_model import signature_constants

[...]

PR #762, opened 2020-06-05
Integration tests for Pytorch artifacts

jjmachan:
I have added the integration tests for PyTorch artifacts based on the commits form #726.

## Description

I've added a unit test for the Pytorch artefact were I create a simple model where the model is initialized with all *ones*. Hence the forward run of the model will give always give deterministic results. I also created a `run-keras-tests.py` script that installs tensorflow and performs the tests.

## Motivation and Context

This PR is based on #723 

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation
- [x] Test, CI, or build
- [ ] None

## Components (if applicable)
- [ ] BentoService (model packaging, dependency management, handler definition)
- [x] Model Artifact (model serialization, multi-framework support)
- [ ] Model Server (mico-batching, logging, metrics, tracing, benchmark, OpenAPI)
- [ ] YataiService (model management, deployment automation)
- [ ] Documentation

## Checklist:

- [x] My code follows the bentoml code style, both `./dev/format.sh` and
  `./dev/lint.sh` script have passed
  ([instructions](https://github.com/bentoml/BentoML/blob/master/DEVELOPMENT.md#style-check-and-auto-formatting-your-code)).
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] My change requires a change in [bentoml/gallery](https://github.com/bentoml/gallery) example notebooks
- [ ] I have sent a pull request to [bentoml/gallery](https://github.com/bentoml/gallery) to make that change

PR #726, opened 2020-05-27
[Tests]Include onnx artifact test on travis

yubozhao:
## Description

Include onnx artifact tests with python 3.6 on travis ci
Add travis directory

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation
- [x] Test, CI, or build
- [ ] None

## Components (if applicable)
- [ ] BentoService (model packaging, dependency management, handler definition)
- [ ] Model Artifact (model serialization, multi-framework support)
- [ ] Model Server (mico-batching, logging, metrics, tracing, benchmark, OpenAPI)
- [ ] YataiService (model management, deployment automation)
- [ ] Documentation

## Checklist:

- [ ] My code follows the bentoml code style, both `./dev/format.sh` and
  `./dev/lint.sh` script have passed
  ([instructions](https://github.com/bentoml/BentoML/blob/master/DEVELOPMENT.md#style-check-and-auto-formatting-your-code)).
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] My change requires a change in [bentoml/gallery](https://github.com/bentoml/gallery) example notebooks
- [ ] I have sent a pull request to [bentoml/gallery](https://github.com/bentoml/gallery) to make that change

PR #871, opened 2020-07-08
[CI] Refactor API Server integration tests

bojiang:
## Description

Employ pytest session to start/stop server container.

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature and improvements (non-breaking change which adds/improves functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Code Refactoring (internal change which is not user facing)
- [ ] Documentation
- [x] Test, CI, or build

## Component(s) if applicable
- [ ] BentoService (service definition, dependency management, API input/output adapters)
- [ ] Model Artifact (model serialization, multi-framework support)
- [x] Model Server (mico-batching, dockerisation, logging, OpenAPI, instruments)
- [ ] YataiService gRPC server (model registry, cloud deployment automation)
- [ ] YataiService web server (nodejs HTTP server and web UI)
- [ ] Internal (BentoML's own configuration, logging, utility, exception handling)
- [ ] BentoML CLI

## Checklist:

- [x] My code follows the bentoml code style, both `./dev/format.sh` and
  `./dev/lint.sh` script have passed
  ([instructions](https://github.com/bentoml/BentoML/blob/master/DEVELOPMENT.md#style-check-and-auto-formatting-your-code)).
- [ ] My change reduces project test coverage and requires unit tests to be added
- [ ] I have added unit tests covering my code change
- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly

PR #876, opened 2020-07-09
Add integration test for Tensorflow SavedModel artifact

joshuacwnewton:
## Description
 * Create example BentoService in `tests/bento_service_examples/` that uses TensorFlow SavedModel artifact
 * Create test in `tests/integration_tests/` which integrates sample TF model with TF BentoService
 * Create shell script in `travis/` to import TF and run test
 * Add shell script to `.travis.yml`

I am uncertain about this line in issue #721, though:

> Make sure this `tests/test_tensorflow_savedmodel_artifact.py` is ignored in regular CI build

## Motivation and Context

 * Fixes issue #721.
 * Changes based off of #726 and #762.

## How Has This Been Tested?
Ran `pytest tests/integration` locally.

## Types of changes

- [ ] Documentation
- [X] Test, CI, or build

## Component(s) if applicable
- [X] BentoService (service definition, dependency management, API input/output adapters)
- [X] Model Artifact (model serialization, multi-framework support)
- [ ] BentoML CLI

## Checklist:

- [X] My code follows the bentoml code style, both `./dev/format.sh` and


parano:
Sorry about the confusion, I think by `Make sure this tests/test_tensorflow_savedmodel_artifact.py is ignored in regular CI build`, I meant not including it in the unit test runs, I think we already do that now:  https://github.com/bentoml/BentoML/blob/v0.8.3/travis/unit_tests.sh#L12

travis/tensorflow_integration_tests.sh
```diff
@@ -0,0 +1,15 @@
...
+
+# https://stackoverflow.com/questions/42218009/how-to-tell-if-any-command-in-bash-script-failed-non-zero-exit-status/42219754#42219754
+# Set err to 1 if pytest failed.
+error=0
+trap 'error=1' ERR
+
+GIT_ROOT=$(git rev-parse --show-toplevel)
+cd "$GIT_ROOT" || exit
+
+pip install tensorflow
```
parano, line 12: should we fix the TensorFlow version here?

parano:
The test you added is failing due to a permission issue running the `.sh` script, you may need to run `chmod +x` on this file and commit the change

bojiang:
Hi. After https://github.com/bentoml/BentoML/pull/871 , I recommend to implement artifact integration tests here: https://github.com/bentoml/BentoML/blob/master/tests/integration/api_server/test_artifact.py 

cc @parano

joshuacwnewton:
Thank you much for the clarification and advice, @parano. :)

@bojiang Would you prefer all of the tests be written/moved together in one PR? (i.e. including [`test_onnx_model_artifact.py`](https://github.com/bentoml/BentoML/blob/master/tests/integration/test_onnx_model_artifact.py) and [`test_pytorch_model_artifact.py`](https://github.com/bentoml/BentoML/blob/master/tests/integration/test_pytorch_model_artifact.py))

I'm considering also doing #722 and #724 at the same time because the issue descriptions would become outdated, and so would be less useful for newcomers to the project.

travis/tensorflow_integration_tests.sh
joshuacwnewton: This is an interesting question. Just for comparison, the versions were indeed fixed for [`pytorch_integration_tests.sh`](https://github.com/bentoml/BentoML/blob/master/travis/pytorch_integration_tests.sh), but not for [`onnx_integration_tests.sh`](https://github.com/bentoml/BentoML/blob/master/travis/onnx_integration_tests.sh).

Is the worry that future API changes to TensorFlow might break the tests?

parano: @joshuacwnewton yes I think that is the concern, usually, it is best practice to pin the version so that the tests do not break due to external changes.   Although in current stage of the bentoml project, it is nice for us to get notified when a newer version of TensorFlow breaks our integration. Ideally, we should run a compatibility matrix to determine which are the versions of Tensorflow APIs that BentoML can support.

But before we get there, would be great if you could pin the version here to the current TensorFlow stable release. I think the onnx test will need to do the same as well. cc @yubozhao

parano:
@joshuacwnewton indeed those issue descriptions are now outdated. I think @bojiang made a good point, I think the test structure introduced in #871 is preferred over this older pattern.  We should migrate ONNX and PyTorch integration tests to follow #871's structure.    sorry about the confusion I will update the PR descriptions accordingly and add a new one for migrating existing tests.

joshuacwnewton:
@parano No worries whatsoever! I appreciate the quick updates very much, so thank you for your communication. :)

Feel free to tag me on the issue for migrating the tests if you like, and I will be happy to make the necessary changes.

travis/tensorflow_integration_tests.sh
joshuacwnewton: Thank you for explaining things so clearly. The reasoning makes a lot of sense! :)

parano:
@bojiang @joshuacwnewton Looking at the integration test code introduced in  #855 #871 again, I think it might be a bad idea for us to add all the artifact tests into the existing BentoService example here: https://github.com/bentoml/BentoML/blob/master/tests/integration/api_server/example_service.py and test endpoints in https://github.com/bentoml/BentoML/blob/master/tests/integration/api_server/test_artifact.py

The reason is that we may want to test integration with different versions of the same library. e.g. one test for Tensorflow 1.5 and one for 2.2. And just by adding new artifacts to the current example service won't allow us to do that. It also makes building that one docker image very slow, due to the fact that it will need to install a lot of libraries once we've added all the artifact tests to it. Separating them also allow us to run all those tests in parallel locally with tools like pytest-xdist.

That being said, I think we should still have individual Travis CI tasks for integration tests on each framework like what @joshuacwnewton has implemented in this PR, as well as what we have for test_onnx_model_artifact.py and test_pytorch_model_artifact.py. But in addition to that, where we only test loading the saved artifact from Python session, we should use the saved bundle created to build a docker image and verify that the model artifact can be loaded properly when deploying with docker.

@joshuacwnewton If you could fix the permission error in this PR, we can get this one merged. And we can add the docker build&run part in follow-up PRs.

bojiang:
@parano Yeah, I agree with the `with different versions of the same library` part. I think we should have more example services rather than all in one. But I think we should still have them in one travis task(also session). 

* travis allows max five concurrent tasks at same time. Too many tasks made CI too slow.
* `docker build .` command would reuse intermediate images before the step of installing artifact/adapter specific dependencies. (If it didn't, we should optimize it to achieve this).

parano:
@bojiang  yea I think as long as we do not put all artifact tests in one example BentoService, we can easily optimize how they are scheduled on Travis CI.

For example, we can easily combine the existing `tests/integration/test_pytorch_model_artifact.py`  and `tests/integration/test_onnx_model_artifact.py ` to run in one Travis CI session instead of two separate ones.  Let's start with putting them in separate sessions for now and optimize this later?

parano:
Thanks, @joshuacwnewton!

PR #884, opened 2020-07-11
address comments for containerize command

jackyzha0:
* rename _echo -> echo
* refactor to use load_bento_service_metadata to support file paths
* callback tag validation to cleanup containerize
* parameterized testing

### DIFF
--- a/bentoml/cli/bento_service.py
+++ b/bentoml/cli/bento_service.py
@@ -28,7 +28,7 @@ from bentoml.cli.click_utils import (
     conditional_argument,
 )
 from bentoml.cli.utils import (
-    _echo_docker_api_result,
+    echo_docker_api_result,
     Spinner,
     get_default_yatai_client,
 )
@@ -57,14 +57,6 @@ def escape_shell_params(param):
     k, v = param.split('=')
     v = re.sub(r'([^a-zA-Z0-9])', r'\\\1', v)
     return '{}={}'.format(k, v)
-
-
-def get_name_version_from_tag(bento):
-    # returns name:version
-    if ":" in bento:
-        return bento.split(":")[:2]
-    else:
-        return bento, None
 
 
 def run_with_conda_env(bundle_path, command):
@@ -96,20 +88,33 @@ def run_with_conda_env(bundle_path, command):
     return
 
 
-def make_bento_name_docker_compatible(name, tag):
+def make_bento_name_docker_compatible(name, version):
     """
     Name components may contain lowercase letters, digits and separators.
     A separator is defined as a period, one or two underscores, or one or more dashes.
 
-    A tag name must be valid ASCII and may contain lowercase and uppercase letters,
-    digits, underscores, periods and dashes. A tag name may not start with a period
+    The version must be valid ASCII and may contain lowercase and uppercase letters,
+    digits, underscores, periods and dashes. Version may not start with a period
     or a dash and may contain a maximum of 128 characters.
 
     https://docs.docker.com/engine/reference/commandline/tag/#extended-description
     """
     name = name.lower().strip("._-")
-    tag = tag.lstrip(".-")[:128]
-    return name, tag
+    version = tag.lstrip(".-")[:128]
+    return name, version
+
+
+def validate_tag(ctx, param, tag):  # pylint: disable=unused-argument
+    if ":" in tag:
+        tag = tag.split(":")[:2]
+    else:
+        _echo(
+            f"Tag {tag} does not specify an image version, " f"using 'latest'",
+            CLI_COLOR_WARNING,
+        )
+        tag = tag, "latest"
+    name, version = make_bento_name_docker_compatible(*tag)
+    return f"{name}:{version}"
 
 
 def resolve_bundle_path(bento, pip_installed_bundle_path):
@@ -441,6 +446,7 @@ def create_bento_service_cli(pip_installed_bundle_path=None):
         help="Optional image tag. If not specified, Bento will generate one from "
         "the name of the Bento.",
         required=False,
+        callback=validate_tag,
     )
     @click.option(
         '-u', '--username', type=click.STRING, required=False,
@@ -477,20 +483,20 @@ def create_bento_service_cli(pip_installed_bundle_path=None):
 
         _echo(f"Found Bento: {bento_service_bundle_path}")
 
-        name, version = get_name_version_from_tag(bento)
-        # use tag name and version where applicable
-        if tag is not None:
-            name, v = get_name_version_from_tag(tag)
-            if v is not None:
-                version = v
+        bento_service_metadata_pb = load_bento_service_metadata(
+            bento_service_bundle_path
+        )
+        name, version = make_bento_name_docker_compatible(
+            bento_service_metadata_pb.name, bento_service_metadata_pb.version,
+        )
 
-        name, version = make_bento_name_docker_compatible(name, version)
-
-        full_tag = f"{name}:{version}"
-        if full_tag != bento:
+        # build docker compatible tag if one isnt provided
+        if tag is None:
+            tag = f"{name}:{version}"
+        if tag != bento:
             _echo(
                 f'Bento tag was changed to be Docker compatible. \n'
-                f'"{bento}" -> "{full_tag}"',
+                f'"{bento}"" -> "{tag}"'
                 CLI_COLOR_WARNING,
             )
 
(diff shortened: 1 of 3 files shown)


bentoml/cli/bento_service.py
```diff
@@ -96,20 +88,33 @@ def run_with_conda_env(bundle_path, command):
...
     https://docs.docker.com/engine/reference/commandline/tag/#extended-description
     """
     name = name.lower().strip("._-")
-    tag = tag.lstrip(".-")[:128]
-    return name, tag
+    version = version.lstrip(".-")[:128]
+    return name, version
+
+
+def validate_tag(ctx, param, tag):  # pylint: disable=unused-argument
```
parano, line 107: I think the ideal behavior of `validate_tag` is telling the user that they are using a wrong format for docker tag instead of converting it silently. This matches `docker build` cli's behavior as well:

Screen Shot 2020-07-11 at 3 11 46 PM

bentoml/cli/bento_service.py
```diff
@@ -96,20 +88,33 @@ def run_with_conda_env(bundle_path, command):
...
+    if ":" in tag:
+        tag = tag.split(":")[:2]
+    else:
+        _echo(
+            f"Tag {tag} does not specify an image version, " f"using 'latest'",
```
parano, line 112: We should probably not do this either, "latest" is the default for docker, it is also a bad decision on docker side.  And I think the behavior we want, is to use the BentoService version as the default if the user did not specify a version right?

bentoml/cli/bento_service.py
```diff
@@ -96,20 +88,33 @@ def run_with_conda_env(bundle_path, command):
     return
 
 
-def make_bento_name_docker_compatible(name, tag):
+def make_bento_name_docker_compatible(name, version):
```
parano, line 91: let's separate this to two separate functions? `to_valid_docker_image_name` and `to_valid_docker_image_tag`?  Since the logic itself has nothing to do with BentoML

bentoml/cli/bento_service.py
```diff
@@ -477,20 +483,20 @@ def containerize(bento, push, tag, username, password):
...
-        full_tag = f"{name}:{version}"
-        if full_tag != bento:
+        # build docker compatible tag if one isnt provided
+        if tag is None:
+            tag = f"{name}:{version}"
+        if tag != bento:
             _echo(
                 f'Bento tag was changed to be Docker compatible. \n'
-                f'"{bento}" -> "{full_tag}"',
+                f'"{bento}"" -> "{tag}"',
```
parano, line 499: `bento` here can be a local file path, you will need to update this message

bentoml/cli/bento_service.py
```diff
@@ -477,20 +483,20 @@ def containerize(bento, push, tag, username, password):
...
-        if tag is not None:
-            name, v = get_name_version_from_tag(tag)
-            if v is not None:
-                version = v
-
-        name, version = make_bento_name_docker_compatible(name, version)
+        bento_service_metadata_pb = load_bento_service_metadata(
+            bento_service_bundle_path
+        )
+        name, version = make_bento_name_docker_compatible(
```
parano, line 489: This behavior might be confusing to the end-user, I'd prefer to avoid implicitly converting the command argument they provide. For example, I write the following bash script to automate my deployment process:

```bash
saved_path="/Users/chaoyu/bentoml/repository/IrisClassifier/20200701222443_EC5EAB"
image_name="my-image-something-that-is-very-long..........."
bentoml containerize $saved_path -t $image_name
docker run $image_name
``` 
I will get an `Unable to find image` error due to the fact that the image name was modified implicitly.

I'm inclined to go with a simpler approach and only do the conversion if the value is automatically set from the BentoService metadata:
```python
...
bs_metadata = load_bento_service_metadata(bento_service_bundle_path)

_echo(f"Building docker image for BentoService {bs_metadata.name}:{bs_metadata.version}")

if not tag:
   tag = "{to_valid_docker_image_name(bs_metadata.name)}:{to_valid_docker_image_tag(bs_metadata.version)}"

if ":" not in tag:  # user have specified image name but not image tag in -t
   version= to_valid_docker_image_tag(bs_metadata.version)
   tag = f"{tag}:{version}"

_echo("Building docker image with tag: {tag}")

docker_api = docker.APIClient()
...
```

bentoml/cli/bento_service.py
```diff
@@ -96,20 +88,41 @@ def run_with_conda_env(bundle_path, command):
...
+    return version.lstrip(".-")[:128]
+
+
+def validate_tag(ctx, param, tag):  # pylint: disable=unused-argument
+    if ":" in tag:
+        name, version = tag.split(":")[:2]
+    else:
+        name, version = tag, ""
+    if to_valid_docker_image_name(name) != name:
+        raise click.BadParameter(
```
parano, line 107: 👍

bentoml/cli/bento_service.py
```diff
@@ -96,20 +88,41 @@ def run_with_conda_env(bundle_path, command):
...
+    # https://docs.docker.com/engine/reference/commandline/tag/#extended-description
+    return version.lstrip(".-")[:128]
+
+
+def validate_tag(ctx, param, tag):  # pylint: disable=unused-argument
+    if ":" in tag:
+        name, version = tag.split(":")[:2]
+    else:
+        name, version = tag, ""
+    if to_valid_docker_image_name(name) != name:
```
parano, line 106: Could we use a regex to actually validate the input?  This can break if the user provides an upper case character or a non-ASCII character. see https://docs.docker.com/engine/reference/commandline/tag/#extended-description

> Name components may contain lowercase letters, digits and separators. A separator is defined as a period, one or two underscores, or one or more dashes. A name component may not start or end with a separator.

> A tag name must be valid ASCII and may contain lowercase and uppercase letters, digits, underscores, periods, and dashes. A tag name may not start with a period or a dash and may contain a maximum of 128 characters.

If it's too tricky to get this validation to match docker's rules, we can also just try to catch the `invalid argument` error when running `docker build` command, and show a proper error message to the user. Currently it shows a confusing `Broken pip` error when using an invalid docker tag:

```bash
> bentoml containerize IrisClassifier:latest -t abc___def

[2020-07-12 18:06:02,231] INFO - Getting latest version IrisClassifier:20200701222443_EC5EAB
Found Bento: /Users/chaoyu/bentoml/repository/IrisClassifier/20200701222443_EC5EAB
Bento tag was changed to be Docker compatible.
"IrisClassifier:latest" -> "abc___def:latest"
Building Docker image: abc___def
  File "/Users/chaoyu/opt/miniconda3/envs/btml-dev/lib/python3.7/site-packages/urllib3/connectionpool.py", line 600, in urlopen
    chunked=chunked)
  File "/Users/chaoyu/opt/miniconda3/envs/btml-dev/lib/python3.7/site-packages/urllib3/connectionpool.py", line 354, in _make_request
    conn.request(method, url, **httplib_request_kw)
  File "/Users/chaoyu/opt/miniconda3/envs/btml-dev/lib/python3.7/http/client.py", line 1252, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/Users/chaoyu/opt/miniconda3/envs/btml-dev/lib/python3.7/http/client.py", line 1298, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
  File "/Users/chaoyu/opt/miniconda3/envs/btml-dev/lib/python3.7/http/client.py", line 1247, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/Users/chaoyu/opt/miniconda3/envs/btml-dev/lib/python3.7/http/client.py", line 1065, in _send_output
    self.send(chunk)
  File "/Users/chaoyu/opt/miniconda3/envs/btml-dev/lib/python3.7/http/client.py", line 987, in send
    self.sock.sendall(data)
BrokenPipeError: [Errno 32] Broken pipe
```

bentoml/cli/bento_service.py
```diff
@@ -96,20 +88,61 @@ def run_with_conda_env(bundle_path, command):
...
+def validate_tag(ctx, param, tag):  # pylint: disable=unused-argument
+    if ":" in tag:
+        name, version = tag.split(":")[:2]
+    else:
+        name, version = tag, None
+
+    valid_name_pattern = re.compile(
+        r"""
+        ^(
+        [a-z0-9]+      # alphanumeric
```
parano, line 110: 👍

parano: love the regex inline comments here!

PR #916, opened 2020-07-20
Do not cache pip requirements in Dockerfile

co42:
## Description

Add `--no-cache-dir` option to `pip install`.

## Motivation and Context

Lightens Docker images.
This removes 750Mb of pip cache for Torch for example.


parano:
Thanks @co42!

Issue #926, opened 2020-07-22
CoreML Support / can I extend `BentoServiceArtifact` and serve it?


guy4261:
**Is your feature request related to a problem? Please describe.**
I'm running BentoML directly on a Mac. To gain speedups, I would like to convert my PyTorch model (which you support) to CoreML (which you do not support). To do that, I would like to extend `BentoServiceArtifact`. Is that even allowed? Any caveats I should expect?

**Describe the solution you'd like**
Seems like `BentoServiceArtifact` is just a class saying how to save and load an object. So I'd like to write one for a CoreML model (probably one I converted via https://coremltools.readme.io/docs/pytorch-conversion).

**Describe alternatives you've considered**
Convert my model on the first time I load it. Just seems cheesy.

**Additional context**
Supporting CoreML = supporting faster inference for people running the server directly (`$ bentoml serve Foo:latest`) on a Mac.

yubozhao:
Hi, @guy4261. Good suggestion! You can definitely write a new `BentoServiceArtifact` for CoreML.  I don't think there are any caveats, from looking at the `coremltools` package.

I think other people can benefit from this, feel free to submit a PR.

guy4261:
Thanks @yubozhao ! How do I test it? `pip install -e` the BentoML repo, edit the code and do it? This is why I asked - to verify there are no apparent caveats before I begin.

Thanks!

yubozhao:
Yes, for developing you install from the source, install the local version with `pip install -e` and go from there. 

For testing, you can use the ONNX artifact as an example.  We create an [integration test](https://github.com/bentoml/BentoML/blob/master/tests/integration/test_onnx_model_artifact.py) and add additional test suit for Travis by adding [test script](https://github.com/bentoml/BentoML/blob/master/travis/onnx_integration_tests.sh) and updating the [Travis configuration](https://github.com/bentoml/BentoML/blob/master/.travis.yml)

I am pretty excited about this and looking forward to seeing it

parano:
@guy4261 be sure to build base off the latest master branch, there were some recent refactoring on that code path that drastically changed how new artifact can be implemented. See also #758 #925.

guy4261:
Thanks!

Questions begin:

Seems like I can't use `cloudpickle` for the CoreML model. Can I use the library's standard tools, i.e.
```
self._model.save(self._file_path(dst))
```

or does it have to be `cloudpickle`-able?

Log below:

```python
[...]

PR #939, opened 2020-07-26
Added artifact for CoreML

guy4261:
## Description

Added `bentoml.artifact.CoreMLModelArtifact`, including unit tests based on the PyTorch ones and `.rst` references.

## Motivation and Context

Relevant issue: https://github.com/bentoml/BentoML/issues/926

For use cases involving running on MacOS, running directly on the Mac rather than via Docker can allow using Apple's ML framework - CoreML. Running on CoreML (rather than on dockerized or even bare-metal tensorflow/torch/etc) would allow the best utiliziation of resources during inference.

## How Has This Been Tested?

My changes in this PR involved adding a new type of artifact and registering it on the `bentoml/artifacts/__init__.py`, so no other code was affected.

In CoreML you don't train models but rather convert existing (trained) models to run on CoreML. So My unit test borrows model and data from the existing PyTorch unit test and runs them on CoreML using my new type of artifact (and a matching classifier).

## Types of changes

- [x] Documentation
- [x] Test, CI, or build

## Component(s) if applicable
- [x] Model Artifact (model serialization, multi-framework support)
- [ ] BentoML CLI

## Checklist:


bentoml/artifact/coreml_model_artifact.py
```diff
@@ -0,0 +1,129 @@
...
+    def _file_path(self, base_path):
+        return os.path.join(base_path, self.name + self._file_extension)
+
+    def pack(self, model):  # pylint:disable=arguments-differ
+        try:
+            import coremltools
+        except ImportError:
+            raise MissingDependencyException(
+                "coremltools>=4.0 package is required to use CoreMLModelArtifact"
+            )
```
yubozhao, line 92: Can we also check for the `CoreML` version, since it is one of the requirements?

yubozhao:
@guy4261 great PR. It looks good, just one small nitpick mentioned above.

I think Travis-CI is not picking up your integration test because we didn't specify it.  Can you write the Travis test script similar to [the ONNX one](https://github.com/bentoml/BentoML/blob/master/travis/onnx_integration_tests.sh) and update the [`.travis.yml`](https://github.com/bentoml/BentoML/blob/2b6d6778bdc8fd9195d4f76475e8bc7264487105/.travis.yml#L37)?

Also, the format/linter test is failing for your integration test. Could you fix that?

Thanks for the awesome pr!

guy4261:
Will do!

Also, formatter and linter pass one my machine (even made some commits that addressed this) - will check again 👍🏻

bentoml/artifact/coreml_model_artifact.py
guy4261: Added the requirement to the artifact:

```
class CoreMLModelArtifact(BentoServiceArtifact):
    ...
    def set_dependencies(self, env: BentoServiceEnv):
        pip_deps = ['coremltools>=4.0b1']
        env.add_pip_dependencies_if_missing(pip_deps)
```

guy4261: Is that enough? Or should I check in the condition as well?

yubozhao:
Awesome, thanks!  I totally forgot to mention about specific os type for the Travis-CI(You can find [the documentation here](https://docs.travis-ci.com/user/reference/osx#using-macos)).  Sorry about that, I totally take Mac for grant, and didn't even think about it until too late.

guy4261:
@yubozhao ugh that's not good: the travis itself gets 404 for
`$ curl -sSf --retry 5 -o python-3.6.tar.bz2 https://storage.googleapis.com/travis-ci-language-archives/python/binaries/osx/10.13/x86_64/python-3.6.tar.bz2`

I'll figure this one out later today. Also, looks like the default osx is pretty dated (10.13); I might target newer releases.

parano:
@guy4261 For the linter error, you can add a `pylint: disable=import-error` to the import line that's breaking the linter test on Travis: `tests/bento_service_examples/coreml_classifier.py:3:0: E0401: Unable to import '\''coremltools.models'\'' (import-error)`,  it is due to missing this library on the linter test environment.

guy4261:
So after some cheating I get this https://travis-ci.org/github/bentoml/BentoML/jobs/712907548#L915-L926 :

```python
ImportError while loading conftest '/Users/travis/build/bentoml/BentoML/tests/conftest.py'.
tests/conftest.py:6: in <module>
    from bentoml.yatai.client import YataiClient
bentoml/__init__.py:27: in <module>
    from bentoml.saved_bundle import load, save_to_dir
bentoml/saved_bundle/__init__.py:15: in <module>
    from bentoml.saved_bundle.bundler import save_to_dir
bentoml/saved_bundle/bundler.py:31: in <module>
    from bentoml.utils.usage_stats import track_save
bentoml/utils/usage_stats.py:22: in <module>
    from ruamel.yaml import YAML
E   ModuleNotFoundError: No module named 'ruamel'
```

Before I run yet another 10 Travis CI pipelines 😰 did you ever see something like this? `ruamel` seems to already have been installed https://travis-ci.org/github/bentoml/BentoML/jobs/712907548#L464 :
```
Requirement already satisfied: ruamel.yaml>=0.15.0 in /Users/travis/miniconda/lib/python3.6/site-packages (from BentoML==0.8.3+55.g8848892) (0.15.80)
```

guy4261:
Really weird issue - see here: https://travis-ci.org/github/bentoml/BentoML/builds/713025643

Seems like I'm getting some infinite recursion of infinite imports due to 

```python
bentoml/service.py:758: in save
    return save(self, base_path, version)
bentoml/service.py:470: in save
    return yatai_client.repository.upload(bento_service, version)
bentoml/yatai/client/bento_repository_api.py:59: in upload
    save_to_dir(bento_service, tmpdir, version, silent=True)
bentoml/saved_bundle/bundler.py:111: in save_to_dir
    bento_service.__class__.__module__, os.path.join(path, bento_service.name)
bentoml/saved_bundle/py_module_utils.py:141: in copy_used_py_modules
    finder.run_script(target_module_file)
```

This happens when `target_module_file` is (the full path of) `"BentoML/tests/bento_service_examples/coreml_classifier.py"`.

I have no idea why this happens!

I tried two things:
* Cancelling out `@bentoml.env(auto_pip_dependencies=True)` from `coreml_classifier.py`, 
* Instead of importing Pytorch model from the unit tests, copy+paste this code in my unit test.

But it didn't work. Any ideas @yubozhao @parano ?

guy4261:
It's getting weirder and weirder:

* On `master`, the **PyTorch** unit test passes (`tests/integration/test_pytorch_model_artifact.py`).
* On my branch, it fails with this recursion error...
* But if I remove `from bentoml.artifact.coreml_model_artifact import CoreMLModelArtifact` from `bentoml/artifact/__init__.py` (and also remove the class from the `___all__` list - the PyTorch unit test runs again!

So merely importing the `CoreMLModelArtifact` creates havoc! Nothing runs there, and it only has the imports below (no `coremltools` or anything). Do you have any ideas why this happens?

Imports made in `bentoml/artifact/coreml_model_artifact.py`:

```python
import os
import logging

from bentoml.artifact import BentoServiceArtifact
from bentoml.service_env import BentoServiceEnv
from bentoml.exceptions import MissingDependencyException, InvalidArgument
```

parano:
@guy4261 that seems really weird, I just tried pulling down your branch and running both `./travis/coreml_integration_tests.sh` and `tests/integration/test_pytorch_model_artifact.py`, both tests are passing.  I'm on Mac OS using conda to install all the dependencies.  I will look a bit more into it and update if I found anything.

tests/bento_service_examples/coreml_classifier.py
```diff
@@ -0,0 +1,18 @@
+import numpy
+import pandas as pd
+from coremltools.models import MLModel  # pylint: disable=import-error
```
parano, line 3: I can reproduce this recursion error on my local machine in docker(linux environment with a similar setup with conda, py36 and ran the setup script in this PR). The error went away after I removed this import here and the type annotation below. 

Not sure why that is the case, I will look a bit more into it.

parano: I also removed the `predict` call in the test, so I can test the `save` under non-mac enviornment

parano: @guy4261 hope you don't mind, I will try pushing some changes to your branch to trigger some test runs here

parano: Update: just tried running the quick start guide example and it also fails in this environment. I guess it has to do with the python import path and how BentoML is installed.

parano:
@guy4261  sharing a quick update here:

The root cause of this issue is that the `modulefinder#run_script` call reached the max recursion depth defined in the python environment.  BentoML uses the `modulefinder` system module to find all local python dependencies and the recursion depth increases as the number of imports increases in the BentoService class file.  

We already had a new implementation for parsing BentoService class' dependencies for `auto_pip_dependencies` here: https://github.com/bentoml/BentoML/blob/master/bentoml/saved_bundle/pip_pkg.py which allow us to migrate `py_modules_util` away from using `modulefinder` and avoid this problem. But before we do that, a workaround is set a larger recursion depth when hitting this problem:

```python
import sys
print(sys.getrecursionlimit())  # view current setting
sys.setrecursionlimit(sys.getrecursionlimit() * 2) # increase the limit


```

tests/bento_service_examples/coreml_classifier.py
guy4261: Hi @parano , I see the unit tests passed (on Travis). So is the code broken or not? I got confused 😰

guy4261:
@yubozhao @parano

Thanks for the pushes!

How can I improve the `codecov`? The report is not  very clear I'm afraid :(

parano:
Hey @guy4261, don't worry about the codecov, I think there are some issues with how it is currently configured. Is there anything else you'd like to add to the PR or to clean up?  If not, I think we can go ahead and merge this PR now.

guy4261:
Looks good to me 👍🏻  
The only thing I am sorry about is that I'm downloading the Miniconda installation and run it on OSX every time. Will that cause a problem? If you are OK with that, let's merge! 🎉   
Thank you both!!! 🙏

parano:
@guy4261 I think that is ok, testing with the default python on Mac OS is a bit too restricted, and this way we can add new tests for other python versions on Mac OS more easily in the future.  

Merging now!

Issue #725, opened 2020-05-27
Support passing a conda environment.yml file to `@env`


parano:
Currently, the user will need to pass in conda environment parameters individually, e.g.:

```python
@bentoml.artifacts([H2oModelArtifact('model')])
@bentoml.env(
  pip_dependencies=['pandas', 'h2o==3.24.0.2'],
  conda_channels=['h2oai'],
  conda_dependencies=['h2o==3.24.0.2']
)
class ExamplePredictionService(bentoml.BentoService):

   ...
```

This can be tedious If the user already has a conda enviornment.yml file as the source of truth. BentoML should support taking a conda environment.yaml file directly, e.g.:
```python
@env(conda_env_yaml="./environment.yml")
```

Feature suggested by Cameron T in the community slack channel.

yashika51:
Hi @parano, are we looking to make changes in [this class](https://github.com/bentoml/BentoML/blob/7af5a8f4ebe4e053bb65638ad8b9ab59555a411b/bentoml/service_env.py#L54-L83)?

parano:
@yashika51 yes that's correct, you can add a class method for initializing a CondaEnv with a yaml file, and you will also need to modify the `env` decorator implementation

yashika51:
Okay, I have made a draft PR just to show you the method before moving forward. I've only added the method for your approval first.

parano:
This is now closed in https://github.com/bentoml/BentoML/pull/1084

PR #950, opened 2020-08-04
[Draft] Method for yaml file added

yashika51:
## Description

I have just added the method `conda_env_yaml` to confirm I have understood @parano's point. The rest of the changes along with changing the decorator is yet to be done.

## Motivation and Context

In reference to #725. This will allow user to put the path to the `yaml` file instead of writing dependencies on their own.

## Types of changes

- [ ] Documentation
- [ ] Test, CI, or build

## Component(s) if applicable
- [ ] BentoML CLI

## Checklist:

### DIFF
--- a/bentoml/service_env.py
+++ b/bentoml/service_env.py
@@ -84,15 +84,15 @@ class CondaEnv(object):
 
     def conda_env_yaml(self, filepath):
         with open(filepath) as file:
-            data=self._yaml.load(file)
-        for k,v in data.items():
-            if k=="name":
+            data = self._yaml.load(file)
+        for k, v in data.items():
+            if k == "name":
                 self._conda_env["name"] = v
-            elif k=="channels":
+            elif k == "channels":
                 self._conda_env["channels"] = v
             else:
                 self._conda_env["dependencies"] = v
- 
+
 
 class BentoServiceEnv(object):
     """Defines all aspect of the system environment requirements for a custom
--- a/bentoml/service_env.py
+++ b/bentoml/service_env.py
@@ -82,6 +82,17 @@ class CondaEnv(object):
         with open(filepath, 'wb') as output_yaml:
             self._yaml.dump(self._conda_env, output_yaml)
 
+    def conda_env_yaml(self, filepath):
+        with open(filepath) as file:
+            data=self._yaml.load(file)
+        for k,v in data.items():
+            if k=="name":
+                self._conda_env["name"] = v
+            elif k=="channels":
+                self._conda_env["channels"] = v
+            else:
+                self._conda_env["dependencies"] = v
+ 
 
 class BentoServiceEnv(object):
     """Defines all aspect of the system environment requirements for a custom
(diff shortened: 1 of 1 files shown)


bentoml/service_env.py
```diff
@@ -82,6 +82,17 @@ def write_to_yaml_file(self, filepath):
         with open(filepath, 'wb') as output_yaml:
             self._yaml.dump(self._conda_env, output_yaml)
 
+    def conda_env_yaml(self, filepath):
```
parano, line 85: simply do `self._conda_env = self._yaml.load(filepath)` here and then 1) verify the python version matches current python session's python versions, 2) modify the env name as it is used for activating the env in docker image

yashika51: okay 👍

parano: Actually, that approach may lead to losing other dependencies specified in `self._conda_env`. If you could merge two `yaml` that'd be ideal, you may refer to the official conda-merge's code here; https://github.com/amitbeka/conda-merge/blob/master/conda_merge.py

yashika51: Oh, I didn't realize that. Thanks for pointing it out :). 
I'll update the draft for your review again.

yashika51: Does doing something like this will serve the purpose?

` def conda_env_yaml(self, filepath):
        dep=set()
        conda_yaml = self._yaml.load(filepath)
        #check python version
        if self._conda_env['python_version']==conda_yaml['python_version']:

                #merge conda_yaml and _conda_env
                dep.add(self._conda_env['dependencies'])
                dep.add(conda_yaml['dependencies'])
                conda_yaml['dependencies']=dep
        else:
            raise "some error"`

parano: I think support merging two conda environment file is just way too error-prone and hard to implement.

Could we just add a class method for initializing a CondaEnv instance with an env.yml file? that should be a lot easier to implement.

yashika51: Hi, apologies for being away, I was caught up with university applications.
After adding this parameter we would take the config_file path as an input and then? How are we going to merge the dependencies? 

 ```
 def write_to_yaml_file(self, filepath):
        with open(filepath, 'wb') as output_yaml:
            self._yaml.dump(self._conda_env, output_yaml)
        if self.env_config_file:
            config_file_env=self._yaml.load(self.env_config_file)
            self._yaml['dependencies'].append(config_file_env['dependencies'])
            print(_yaml['dependencies'])
```

As you mentioned earlier I also think we might have to check the python version before doing this ^^

thoo:
I am also interested in this feature since this will make it easier for me to migrate from Azureml.  
Can we replace `CONDA_ENV_BASE_YAML`  with the user provided conda yaml file if there is one?

parano:
@yashika51 looks like this PR is now conflicting with the master branch, I will try revisit this issue in a following up PR. Thanks again for making the effort to experiment with this fix

PR #972, opened 2020-08-07
[Internal] Refactor Input Adapters

akainth015:
## Description
This pull request will add a lot more documentation and IDE auto-completion support to the input and output adapters. It will make it much simpler to implement input adapters by making specialized (optimized) batch request processing, and custom Lambda event processing opt-in instead of the default behavior.

This pull request aims not to have any breaking changes, but users with custom adapters may need to update parts of their code to reflect new method signatures.

## Motivation and Context
Right now, you have to implement 4 methods when writing an input adapter, `handle_request`, `handle_batch_request`, `handle_cli`, and `handle_aws_lambda_event`. This often results in a lot of duplicated code between `handle_request`, `handle_batch_request`, and `handle_aws_lambda_event`. To resolve this issue, [this new architecture](https://app.lucidchart.com/documents/edit/74d25a51-2b73-448a-aac0-c9b814b5b618/0_0) was proposed. 

It was noted that certain input adapters, such as the `DataframeInputAdapter` take advantage of optimized processing, and this yields significant performance boosts. As such, a provision was made to have batch processing opt-in.

This solution is not without its faults, however. The post-batch processing means we cannot guarantee that all requests in a batch are valid, meaning that some optimizations will not be as consistent as they might otherwise be. This is also visible as part of  #940. Ideally, this will be supported by adding a `validate` parameter to the `@api` decorator, which will be a function that can run a lightweight validation of the input.

## How Has This Been Tested?
The changes in this pull request will be unit tested and any existing code samples should function as currently written.

## Types of changes

- [x] Code Refactoring (an internal change which is not user-facing)
- [ ] Documentation
- [ ] Test, CI, or build

## Component(s) if applicable
- [ ] BentoML CLI

## Checklist:

### DIFF
--- a/bentoml/adapters/base_input.py
+++ b/bentoml/adapters/base_input.py
@@ -12,14 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Iterable
-
+from typing import Iterable, TypeVar, Generic, Dict
 
 from bentoml.marshal.utils import SimpleResponse, SimpleRequest, BATCH_REQUEST_HEADER
 
+I = TypeVar("I")
 
-class BaseInputAdapter:
-    """InputAdapter is an abstraction layer between user defined API callback function
+
+class BaseInputAdapter(Generic[I]):
+    """
+    InputAdapter is an abstraction layer between user defined API callback function
     and prediction request input in a variety of different forms, such as HTTP request
     body, command line arguments or AWS Lambda event object.
     """
@@ -29,10 +31,10 @@ class BaseInputAdapter:
     BATCH_MODE_SUPPORTED = False
 
     def __init__(self, output_adapter=None, http_input_example=None, **base_config):
-        '''
+        """
         base_configs:
             - is_batch_input
-        '''
+        """
         self._config = base_config
         self._output_adapter = output_adapter
         self._http_input_example = http_input_example
@@ -56,19 +58,19 @@ class BaseInputAdapter:
             self._output_adapter = DefaultOutput()
         return self._output_adapter
 
-    def handle_request(self, request, func):
+    def handle_request(self, headers: Dict[str, str], body) -> I:
         """Handles an HTTP request, convert it into corresponding data
         format that user API function is expecting, and return API
         function result as the HTTP response to client
 
-        :param request: Flask request object
-        :param func: user API function
+        :param headers: The request's HTTP headers
+        :param body: The request's body
         """
         raise NotImplementedError
 
     def handle_batch_request(
         self, requests: Iterable[SimpleRequest], func
-    ) -> Iterable[SimpleResponse]:
+    ) -> Iterable[I]:
         """Handles an HTTP request, convert it into corresponding data
         format that user API function is expecting, and return API
         function result as the HTTP response to client
--- a/bentoml/adapters/base_input.py
+++ b/bentoml/adapters/base_input.py
@@ -16,10 +16,10 @@ from typing import Iterable, TypeVar, Generic, Dict
 
 from bentoml.marshal.utils import SimpleResponse, SimpleRequest, BATCH_REQUEST_HEADER
 
-I = TypeVar("I")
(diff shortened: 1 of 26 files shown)


bentoml/server/api_server.py
```diff
@@ -299,22 +275,17 @@ def bento_service_api_func_wrapper(self, api):
         service_version = self.bento_service.version
 
         def api_func():
-            # Log image files in request if there is any
-            image_paths = self.log_image(request, request_id)
-
-            # _request_to_json parses request as JSON; in case errors, it raises
-            # a 400 exception. (consider 4xx before 5xx.)
-            request_for_log = _request_to_json(request)
```
parano, line 307: is the plan to move prediction logging out of here and make it work with `Input` directly?

parano:
Love that we are finally moving `func` and output adapter out of the input adapter! This is so much cleaner now.

The `Input` type added in this PR is overlapped with the `JsonInferenceTask` and `extract` method proposed here in this draft PR: https://github.com/bentoml/BentoML/pull/961/  I think we can get this PR merged first and @bojiang can rebase on this PR in #961.  We still need to update all the existing input adapters and output adapters to this new pattern right? @akainth015

parano:
> This pull request aims not to have any breaking changes, but users with custom adapters may need to update parts of their code to reflect new method signatures.

We likely will break custom input adapter in the next release giving all the changes happening around adapters API, I will make sure to document this well for users to upgrade to the new version.

bentoml/server/api_server.py
akainth015: Yeah, that is the idea since there was specialized logging for both JSON and image inputs, which made more sense to have in their respective input adapters. I think we'll need to come up with some sort of interface to let input adapters cleanly document their inputs.

akainth015:
> We still need to update all the existing input adapters and output adapters to this new pattern right?

Yes, I will work on updating all of them since they need to have the new method signature. This pull request still has a ways to go, but I hope it will be done soon.

parano:
Sounds good, @bojiang could you also take a look and advise how this could best work with #961 to avoid conflicts?

bentoml/server/api_server.py
parano: The problem with that is now the input adapter does not have access to the output of the API request. Usually, the user would expect one prediction log item to have both the inference input and output. We probably want to move the logging logic to `InferenceAPI` and allow adapters to define how to turn a `InferenceTask` and `InferenceResult` into a log record.

I think it might be easier to leave this change on prediction logging for later PR to reduce the scope of this pull request. The sooner we merge the refactoring in this PR, the easier would be for @bojiang to build the adapter internal refactoring base upon this.

bentoml/adapters/base_input.py
```diff
@@ -12,14 +12,16 @@
...
 
+from bentoml.marshal.utils import SimpleRequest, BATCH_REQUEST_HEADER
 
-from bentoml.marshal.utils import SimpleResponse, SimpleRequest, BATCH_REQUEST_HEADER
+Input = TypeVar("Input")
 
 
-class BaseInputAdapter:
-    """InputAdapter is an abstraction layer between user defined API callback function
+class BaseInputAdapter(Generic[Input]):
```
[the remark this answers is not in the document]
akainth015: Ok, so I should omit the return types from the type definitions on `BaseInputAdapter` and only include them in the concrete subclasses like `ImageInput`?

bentoml/service.py
```diff
@@ -231,52 +242,38 @@ def api_decorator(
     >>>     @api(input=DataframeInput(input_json_orient='records'))
     >>>     def identity(self, df):
     >>>         # user-defined callback function that process inference requests
-
-
-    Deprecated syntax before version 0.8.0:
```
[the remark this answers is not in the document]
akainth015: Yes, I figured if it was deprecated it should be safe to remove since we're touching the code anywways.

PR #982, opened 2020-08-11
Refactor(adapter): generic logic

akainth015:
See description for #972


bentoml/adapters/base_output.py
```diff
@@ -32,46 +37,39 @@ def __init__(self, cors='*'):
     def config(self):
         return dict(cors=self.cors,)
 
-    def to_response(self, result, request: flask.Request) -> flask.Response:
-        """Converts corresponding data into an HTTP response
-
-        :param result: result of user API function
-        :param request: request object
+    @property
+    def pip_dependencies(self):
```
parano, line 41: would be great to also adopt the `set_dependencies(self, env)` approach in the new Artifacts API, e.g.: https://github.com/bentoml/BentoML/blob/master/bentoml/artifact/coreml_model_artifact.py#L120

bojiang:
### TODO
cli:
- [ ] better cli error message (now just print err_msg)
- [ ] system cli args encoding?
- [ ] multiple cli input

adapters:
- [ ] other input adapters
- [ ] other output adapters

chore:
- [ ] apply `set_dependencies`

bentoml/adapters/base_input.py
```diff
@@ -91,7 +91,7 @@ def parse_cli_input(cli_args: Iterable[str]) -> Iterator[bytes]:
     input_g.add_argument('--input', nargs="+", type=str)
     input_g.add_argument('--input-file', nargs="+")
 
-    parsed_args, _ = parser.parse_known_args(cli_args)
+    parsed_args, _ = parser.parse_known_args(list(cli_args))
```
bojiang, line 94: @akainth015 why list here?

bentoml/adapters/default_output.py
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Iterable, Tuple
+from typing import Iterable, Tuple, Type, Union
```
bojiang, line 15: Is Union here used?

bentoml/adapters/image_input.py
```diff
@@ -100,16 +99,16 @@ class ImageInput(BaseInputAdapter):
     BATCH_MODE_SUPPORTED = True
 
     def __init__(
-        self,
-        accept_image_formats=None,
-        pilmode="RGB",
-        is_batch_input=False,
-        **base_kwargs,
+            self,
```
bojiang: @akainth015 Didn't get it

bentoml/adapters/json_input.py
```diff
@@ -110,4 +110,4 @@ def extract_user_func_args(
             except Exception:  # pylint: disable=broad-except
                 err = traceback.format_exc()
                 task.discard(http_status=500, err_msg=f"Internal Server Error: {err}")
-        return (json_inputs,)
+        return json_inputs,
```
bojiang: @akainth015 The formatter we are using is black. It gives different advice than this.

bentoml/yatai/deployment/sagemaker/serve
```diff
@@ -56,7 +56,7 @@ def _serve():
         signal.SIGTERM, lambda a, b: sigterm_handler(nginx.pid, gunicorn_app.pid)
     )
 
-    pids = set([nginx.pid, gunicorn_app.pid])
+    pids = {nginx.pid, gunicorn_app.pid}
```
bojiang, line 59: 👍

bojiang:
@akainth015 
It would be better if we format it after major modifications are completed.

akainth015:
Yeah my IDEs formatter doesn't conform with black and that's causing
issues, and the pep8 bot reports issues with the reformatting that the
format script performs since it removes spaces after a trailing comma.

bentoml/adapters/default_output.py
akainth015: You're right I'll remove it

bentoml/adapters/base_input.py
akainth015: The function signature takes an Optional Sequence, but the variable is an iterator. Maybe the issue is with the function signature.

akainth015:
Ok, I was just going through the issues that my IDE reported to get a better understanding of where we stand in the refactoring and I've got an idea of what's remaining to do.

bentoml/adapters/json_input.py
```diff
@@ -29,7 +28,7 @@
 
 
 ApiFuncArgs = Tuple[
-    Tuple[JsonSerializable],
+    Sequence[JsonSerializable],
```
bojiang, line 31: Make sense.

bentoml/service.py
```diff
@@ -199,16 +199,16 @@ def _validation(inf_tasks):
...
 
         # call user function
         user_return = self.user_func(*user_args, contexts=contexts)
 
         if (
-            isinstance(user_return, (tuple, list))
+            isinstance(user_return, (Sequence, Iterable))
             and user_return
-            and isinstance(user_return[0], InferenceResult)
+            and all(map(lambda x: isinstance(x, InferenceResult), user_return))
```
bojiang, line 211: What is the point of more type checking here?

bentoml/service.py
```diff
@@ -199,16 +199,16 @@ def _validation(inf_tasks):
...
-        user_args = self.input_adapter.extract_user_func_args(_validation(inf_tasks))
+        user_args = self.input_adapter.extract_user_func_args(valid_tasks(inf_tasks))
         contexts = tuple(t.context for t in inf_tasks if not t.is_discarded)
 
         # call user function
         user_return = self.user_func(*user_args, contexts=contexts)
 
         if (
-            isinstance(user_return, (tuple, list))
+            isinstance(user_return, (Sequence, Iterable))
```
bojiang, line 209: Generator is also a Iterable, but shouldn't be allowed. Otherwise the type checking above would consume some elements.

bentoml/types.py
```diff
@@ -6,13 +6,14 @@
 
 # http://www.apache.org/licenses/LICENSE-2.0
 
+import functools
```
bojiang, line 9: ?

bentoml/types.py
```diff
@@ -123,7 +121,7 @@ class DefaultErrorContext(InferenceContext):
 class InferenceTask(Generic[Input]):
     def __init__(self, data: Input, context: InferenceContext = None):
         self.data = data
-        self.context = context or InferenceContext()
+        self.context: InferenceContext = context or InferenceContext()
```
bojiang, line 124: ?

bentoml/types.py
akainth015: My IDE moved the import 2 lines up, I'm not sure why

bentoml/service.py
bojiang: It takes more time to check each one, but it does not bring benefits.

bentoml/types.py
akainth015: Alright, that makes sense. I only added the type annotation to this one, because PyCharm wasn't able to infer the type of `inf_task.context`, probably because there is an `or` there.

bentoml/service.py
akainth015: I will revert that change, I made it since checking only the first element makes my hair raise :sweat_smile: I suppose that since the `user_return` shouldn't be empty it is okay to assume that there is always one element in the array.

bentoml/types.py
bojiang: @akainth015 I don't think you should add them all. It looks redundant. How about replace the `context or InferenceContext()` with `context if context else InferenceContext()`

bentoml/service.py
bojiang: The `and user_return` make sure it couldn't be an empty list.

Issue #1002, opened 2020-08-21
Refactor adapters


bojiang:
## Is your feature request related to a problem? Please describe.
- Allowed users to customize HTTP response/ AWS lambda event like headers, etc
- Made contribution of adapters easier.
- Drier adapter code
- A more handy CLI UX for user defined API function

## Describe the solution you'd like

- Introduce InferenceTask & InferenceResult types
- Move micro-batching related logic out of adapters

### For adapter developers
Now a basic JSON input adapter would be very simple:
```python

ApiFuncArgs = Tuple[Sequence[JsonSerializable],]

class JsonInput(BaseInputAdapter):
    def from_http_request(self, req: HTTPRequest) -> InferenceTask:
        return  InferenceTask(http_headers=r.parsed_headers, data=req.body)

    def from_aws_lambda_event(self, event: AwsLambdaEvent) -> InferenceTask:
        return InferenceTask(aws_lambda_event=event, data=event['body'].encode(JSON_CHARSET))

    def from_cli(self, cli_args: Sequence[str]) -> Iterator[InferenceTask]:
        for json_file in parse_cli_input(cli_args):
            yield InferenceTask(cli_args=cli_args, data=json_file.read())

    def extract_user_func_args(self, tasks: Sequence[InferenceTask]) -> ApiFuncArgs:
        json_inputs = []
        for task in tasks:
            try:
                json_str = task.data.decode(JSON_CHARSET)
                parsed_json = json.loads(json_str)
                json_inputs.append(parsed_json)
            except UnicodeDecodeError:
                task.discard(
                    http_status=400, err_msg=f"JSON must be encoded in {JSON_CHARSET}"
                )
            except json.JSONDecodeError:
                task.discard(http_status=400, err_msg="Not a valid JSON format")
            except Exception:  # pylint: disable=broad-except
                err = traceback.format_exc()
                task.discard(http_status=500, err_msg=f"Internal Server Error: {err}")
        return (json_inputs,)
```

And it allows user to query:

**by CLI**
```shell
bentoml run <yourservice>:latest --input {"height": 180, "weight": 80} {"height": 190, "weight": 100}
```

```shell
bentoml run <yourservice>:latest --input-file path/to/your/*.json --max-batch-size 10
```

**by HTTP**
```shell
curl -i \
    --header "Content-Type: application/json" \
    --request POST \
    --data '{"height": 190, "weight": 100}' \
    localhost:5000/predict
```

### For Users
[...]

PR #1025, opened 2020-08-25
Refine MultiFileInput; Refactor AnnotatedImageInput

bojiang:


bojiang:
@ecrows 
Hi.
In this branch under development, the logic of InferenceAPI was redesigned, and several internal types such as InferenceTask were added to optimize adapters. See https://github.com/bentoml/BentoML/issues/1002 for details.

A new adapter, [MultiFileInput](https://github.com/bentoml/BentoML/blob/refactor-adapters/bentoml/adapters/multi_file_input.py), was also added. For a drier code, I just refactored the AnnotationImageInput by inheriting MultiFileInput.

But it take some breaking changes:
* AnnotationImageInput now relies more on input names rather than the extension of files. For HTTP & AWS lambda, it means multipart form input names. For CLI, it means CLI args `--input-image --input-annotations` or `--input-file-image --input-file-annotations`
* AnnotationImageInput now accepts HTTP requests & AWS lambda events without annotations file. But for CLI, both `--input-image` and `--input-annotations` are required and should have the same number of inputs.
*

ecrows:
Awesome changes, I like the direction that InputAdapters are going.  Thanks for the heads up too.

PR #1050, opened 2020-09-01
Refine imports and type annotations

akainth015:
Signed-off-by: Aanand Kainth <aanand@akainth.me>

### DIFF
--- a/bentoml/adapters/base_input.py
+++ b/bentoml/adapters/base_input.py
@@ -6,17 +6,17 @@
 
 # http://www.apache.org/licenses/LICENSE-2.0
 
+import argparse
+import contextlib
+import functools
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import io
-import sys
-import argparse
 import itertools
-import functools
-import contextlib
+import sys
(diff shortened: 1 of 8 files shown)


bentoml/adapters/dataframe_input.py
```diff
@@ -55,6 +58,92 @@ class DataframeInput(BaseInputAdapter):
         ValueError: Incoming data format can not be handled. Only json and csv
     """
 
+    def from_cli(self, cli_args: Tuple[str, ...]) -> Iterator[InferenceTask]:
+        parser = argparse.ArgumentParser()
```
bojiang, line 62: Why not call `parse_cli_input` here?

akainth015: I don't think I can add the `json` and `csv` options then, right?

bojiang: We could get the file ext ".csv" or ".json" like this:

for bin_io in parse_cli_input(cli_args):
    file_ext = bin_io.name

akainth015: Oh ok, what would I do for the string input?

bentoml/adapters/base_input.py
```diff
@@ -6,17 +6,17 @@
 
 # http://www.apache.org/licenses/LICENSE-2.0
 
+import argparse
+import contextlib
+import functools
 # Unless required by applicable law or agreed to in writing, software
```
bojiang, line 12: ?

akainth015: I've got no idea what happened there, I couldn't reproduce it with either `dev/format` nor PyCharm reformatting.

bojiang:
LGTM & Merge

PR #1081, opened 2020-09-12
Optimize image build time

ZeyadYasser:
## Description
I split the copying of the service files in the Dockerfile into two parts:
1. Files required for bento-init.sh
2. The rest of the service files

This allows docker to cache the service dependencies effectively (Copying the whole directory at the beginning invalidates cache each time), thus the same or other services would now be able to reuse this cached layer if they have the same dependencies.

NOTE: I had to keep the conda environment name fixed to avoid cache invalidation when copying it over. (I don't think this is the best way, maybe just offer it as the default?? and allow changing it)

## Motivation and Context
I have a use case where I run celery tasks to build custom bentoml service images very frequently, and push them to a private docker repository. I found that each build was taking too much time, nearly 7 minutes, and was pushing all this to the private repository. I tracked the slowdown to cache invalidation at the very start of the image build (Copying the whole service directory). As a workaround I built a custom base image with all my dependencies.

The build and push time decreased drastically down to ~50 seconds, and also the private repository was using much less storage as it was just saving the changed layers.

If this was built in, I would have got this performance boost out of the box, this would have saved a lot of debugging time and allow changing the dependencies easily as I won't have to rebuild the base image each time, it will just invalidate the cache when the dependencies change. 

## Types of changes

- [ ] Documentation
- [x] Test, CI, or build

## Component(s) if applicable
- [ ] BentoML CLI

## Checklist:

### DIFF
--- a/bentoml/saved_bundle/templates.py
+++ b/bentoml/saved_bundle/templates.py
@@ -71,21 +71,27 @@ graft {service_name}/artifacts
 MODEL_SERVER_DOCKERFILE_CPU = """\
 FROM {docker_base_image}
 
-# copy over model files
-COPY . /bento
-WORKDIR /bento
-
 # Configuring PyPI index
 ARG PIP_INDEX_URL=https://pypi.python.org/simple/
 ARG PIP_TRUSTED_HOST=pypi.python.org
 ENV PIP_INDEX_URL $PIP_INDEX_URL
 ENV PIP_TRUSTED_HOST $PIP_TRUSTED_HOST
 
+# copy over files needed for init script
+COPY environment.yml requirements.txt setup.sh* bentoml-init.sh /bento/
+WORKDIR /bento
+
 # Execute permission for bentoml-init.sh
 RUN chmod +x /bento/bentoml-init.sh
 
 # Install conda, pip dependencies and run user defined setup script
 RUN if [ -f /bento/bentoml-init.sh ]; then bash -c /bento/bentoml-init.sh; fi
+
+# copy over model files
+COPY . /bento
+
+# Install bundled bentoml if it exists (used for development)
+RUN if [ -d /bento/bundled_pip_dependencies ]; then pip install -U bundled_pip_dependencies/* ;fi
 
 # the env var $PORT is required by heroku container runtime
 ENV PORT 5000
--- a/bentoml/service_env.py
+++ b/bentoml/service_env.py
@@ -113,7 +113,6 @@ class BentoServiceEnv(object):
 
 
     Args:
-        bento_service_name: name of the BentoService name bundled with this Env
         pip_dependencies: list of pip_dependencies required, specified by package name
             or with specified version `{package_name}=={package_version}`
         auto_pip_dependencies: (Beta) whether to automatically find all the required
@@ -129,7 +128,6 @@ class BentoServiceEnv(object):
 
     def __init__(
         self,
-        bento_service_name: str,
         pip_dependencies: List[str] = None,
         auto_pip_dependencies: bool = False,
         requirements_txt_file: str = None,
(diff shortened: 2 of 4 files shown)


parano:
Thanks for the PR @ZeyadYasser! I've wanted to get to work on this for a while, this is huge for users!

bentoml/saved_bundle/templates.py
```diff
@@ -68,25 +68,33 @@
...
-WORKDIR /bento
-
 # Configuring PyPI index
 ARG PIP_INDEX_URL=https://pypi.python.org/simple/
 ARG PIP_TRUSTED_HOST=pypi.python.org
 ENV PIP_INDEX_URL $PIP_INDEX_URL
 ENV PIP_TRUSTED_HOST $PIP_TRUSTED_HOST
 
+# copy over files needed for init script
+COPY environment.yml requirements.txt setup.sh* bentoml-init.sh /bento/
```
parano, line 85: what's the reason for the `*` in `setup.sh*`?

parano: It seems to me we could just do the following here:
```
COPY environment.yml requirements.txt setup.sh bentoml-init.sh bundled_pip_dependencies* /bento/
```

And avoid the changes around `bundled_pip_dependencies_saved` and make the code path a little cleaner. Is there any reason that this may not work or it will invalidate the build cache?  For users, the `bundled_pip_dependencies` directory should be empty. This directory is only for BentoML developers to bundle their local fork of BentoML in the docker image.

bentoml/service_env.py
```diff
@@ -141,7 +139,7 @@ def __init__(
         self._python_version = PYTHON_VERSION
 
         self._conda_env = CondaEnv(
-            "bentoml-" + bento_service_name,
+            CONDA_ENV_DEFAULT_NAME,
```
parano, line 142: that's a great call! the conda env name was not used anyway. thank you for fixing this!

bentoml/saved_bundle/templates.py
ZeyadYasser: @parano Thank you for such a quick response.

I tried a lot to make it something similar but I faced some limitations in docker unfortunately.

Since the ```setup.sh``` file is optional, it is not always included in the service directory, so docker gives an error when it tries to copy it and it does not exist.
The ```*``` is a workaround to allow docker to skip it if it does not exist.

For the ```bundled_pip_dependencies```, I agree the changes are not clean at all.
The problem I faced when it is copied this way is that it is flattened in the ```/bento/``` directory not in its own directory (i.e. ```/bento/bundle.tar.gz``` not ```/bento/bundled_pip_dependencies/bundle.tar.gz```), which is not compatible with ```bento-init.sh``` script.

I am thinking of two ways to make it cleaner and avoid changes around ```bundled_pip_dependencies_saved``` altogether:
1. Change ```bento-init.sh``` to loop over the flattened files ```for filename in ./*.tar.gz``` but I am afraid this could break the script if someone is using it locally not in the context of docker but it could just be changed try both directories ```./*.tar.gz``` and ```./bundled_pip_dependencies/*.tar.gz``` and that would make it work in both contexts.
2. After copying the ```*.tar.gz``` into  ```/bento/``` copy them into the correct directory ```RUN mkdir bundled_pip_dependencies && mv /bento/*.tar.gz /bento/bundled_pip_dependencies/```

What do you think? your feedback would be very helpful.

parano: Got it, that makes sense about `setup.sh`

As for `bundled_pip_dependencies`, maybe try add the following line after the `COPY . /bento` line, and remove related code from `bentoml-init.sh`?

```
RUN if [ -d /bento/bundled_pip_dependencies ]; then pip install -U bundled_pip_dependencies/* ;fi
```

ZeyadYasser: Nice, This is very elegant.

parano:
@ZeyadYasser thanks for updating the PR! Looking great!

ZeyadYasser:
@parano Thank you!

bojiang:
@ZeyadYasser Great work!

Issue #1098, opened 2020-09-17
Cli get command crash if bento name is not found bento meatadata store


mayurnewase:
**Describe the bug**
if we run command bento get InvalidBentoName:latest,cli throws
`[2020-09-17 19:15:39,725] ERROR - RPC ERROR GetBento: 'NoneType' object has no attribute 'version'`

**Expected behavior**
should give `Error: bentoml-cli get failed: NOT_FOUND:BentoService `InvalidBentoName:latest` is not found`

parano:
hey @mayurnewase, could you share more detail about your setup and what `get` command did you run?

yubozhao:
Thanks for the PR @mayurnewase 
Close with this issue with the PR merged

PR #1099, opened 2020-09-17
If bento name is not found,cli crashes as it tries to access its version

mayurnewase:
# #1098

### DIFF
--- a/bentoml/yatai/yatai_service_impl.py
+++ b/bentoml/yatai/yatai_service_impl.py
@@ -389,14 +389,13 @@ class YataiService(YataiServicer):
             bento_pb = self.bento_metadata_store.get(
                 request.bento_name, request.bento_version
             )
-            if request.bento_version.lower() == 'latest':
-                logger.info(
-                    'Getting latest version %s:%s',
-                    request.bento_name,
-                    request.bento_version,
-                )
-
             if bento_pb:
+                if request.bento_version.lower() == 'latest':
+                    logger.info(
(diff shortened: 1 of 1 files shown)


yubozhao (dismissed):

bentoml/yatai/yatai_service_impl.py
```diff
@@ -393,7 +393,7 @@ def GetBento(self, request, context=None):
                 logger.info(
                     'Getting latest version %s:%s',
                     request.bento_name,
-                    bento_pb.version,
+                    request.bento_version,
                 )
```
parano, line 397: I think `bento_pb.version` here is intentional, so that users using 'latest' tag, gets to know the exact version got fetched.

Can we fix the issue by moving L392-397 to the `if bento_pb` block below?

```
     if bento_pb:
            if request.bento_version.lower() == 'latest':
                logger.info(
                    'Getting latest version %s:%s',
                    request.bento_name,
                    bento_pb.version,
                )

        if bento_pb.uri.type == BentoUri.S3:
             ...
```

mayurnewase: yeah that could work,but it won't log if bento is not found.

parano: On L406 it will actually return a proper error message when the bento is not found

mayurnewase: ok,changing accordingly

PR #1184, opened 2020-10-18
Add location prefix to the URL being proxied (Fixes #1063)

### LINKED ISSUE #1063: Support URL location prefix in yatai-service gRPC/Web server

It is common to run YataiService behind a reverse proxy server, and the user may want to add a location prefix to the URL being proxied. Currently, the YataiService Web UI does not work properly when a location prefix is specified, and we should support it.

Add a parameter to YataiService entry point to set a custom URL prefix, e.g.:

```
bentoml yatai-service-start --base-url="/yatai"
# or
docker run -p 50051:50051 -p 3000:3000 bentoml/yatai-service:latest --base-url="/yatai"
```

n/a

**Additional context**

Related discussion #1057 https://github.com/bentoml/BentoML/discussions/1057

pncnmnp:
## Description
Add support for URL location prefix in yatai-service gRPC/Web server (#1063)
Screenshot for the command line --base-url="path"
Screenshot for the .conf file - nginx configuration
Screenshot for the nginx port /baseURL/

## Motivation and Context
* User can run YataiService behind a reverse proxy server and add a location prefix to the URL being proxied using command line.
* Fixes the Issue #1063 with relevant discussion on #1057

## How Has This Been Tested?
Local deployment

## Types of changes

- [ ] Documentation
- [ ] Test, CI, or build

## Component(s) if applicable
- [x] YataiService web server (nodejs HTTP server and web UI)
- [x] BentoML CLI

## Checklist:

### DIFF
--- a/bentoml/cli/yatai_service.py
+++ b/bentoml/cli/yatai_service.py
@@ -71,7 +71,7 @@ def add_yatai_service_sub_command(cli):
     @click.option(
         '--base-url',
         type=click.STRING,
-        default='/',
+        default='',
         help='Add a location prefix to the URL when running YataiService behind a reverse proxy server',
         envvar='BENTOML_BASE_URL'
     )
--- a/bentoml/cli/yatai_service.py
+++ b/bentoml/cli/yatai_service.py
@@ -72,8 +72,9 @@ def add_yatai_service_sub_command(cli):
         '--base-url',
         type=click.STRING,
         default='.',
-        help='Add a location prefix to the URL when running YataiService behind a reverse proxy server',
-        envvar='BENTOML_BASE_URL'
+        help='Add a location prefix to the URL when running YataiService'
+        'behind a reverse proxy server',
+        envvar='BENTOML_BASE_URL',
     )
     def yatai_service_start(
         db_url, repo_base_url, grpc_port, ui_port, ui, s3_endpoint_url, base_url
--- a/bentoml/cli/yatai_service.py
+++ b/bentoml/cli/yatai_service.py
@@ -71,7 +71,7 @@ def add_yatai_service_sub_command(cli):
     @click.option(
         '--base-url',
         type=click.STRING,
-        default='',
+        default='.',
         help='Add a location prefix to the URL when running YataiService behind a reverse proxy server',
         envvar='BENTOML_BASE_URL'
     )
--- a/bentoml/cli/yatai_service.py
+++ b/bentoml/cli/yatai_service.py
@@ -68,11 +68,18 @@ def add_yatai_service_sub_command(cli):
         'compatible with Amazon S3, such as MinIO',
         envvar='BENTOML_S3_ENDPOINT_URL',
     )
+    @click.option(
+        '--base-url',
+        type=click.STRING,
+        default='/',
+        help='Add a location prefix to the URL when running YataiService behind a reverse proxy server',
+        envvar='BENTOML_BASE_URL'
+    )
     def yatai_service_start(
-        db_url, repo_base_url, grpc_port, ui_port, ui, s3_endpoint_url
+        db_url, repo_base_url, grpc_port, ui_port, ui, s3_endpoint_url, base_url
     ):
         from bentoml.yatai.yatai_service import start_yatai_service_grpc_server
 
         start_yatai_service_grpc_server(
-            db_url, repo_base_url, grpc_port, ui_port, ui, s3_endpoint_url
+            db_url, repo_base_url, grpc_port, ui_port, ui, s3_endpoint_url, base_url
         )
--- a/bentoml/cli/yatai_service.py
+++ b/bentoml/cli/yatai_service.py
@@ -82,5 +82,11 @@ def add_yatai_service_sub_command(cli):
         from bentoml.yatai.yatai_service import start_yatai_service_grpc_server
 
         start_yatai_service_grpc_server(
-            db_url, repo_base_url, grpc_port, ui_port, ui, s3_endpoint_url, web_prefix_path
+            db_url,
+            repo_base_url,
+            grpc_port,
+            ui_port,
+            ui,
+            s3_endpoint_url,
+            web_prefix_path,
         )
--- a/bentoml/cli/yatai_service.py
+++ b/bentoml/cli/yatai_service.py
@@ -69,18 +69,18 @@ def add_yatai_service_sub_command(cli):
         envvar='BENTOML_S3_ENDPOINT_URL',
     )
     @click.option(
-        '--base-url',
+        '--web-prefix-path',
         type=click.STRING,
         default='.',
         help='Add a location prefix to the URL when running YataiService'
         'behind a reverse proxy server',
-        envvar='BENTOML_BASE_URL',
+        envvar='BENTOML_YATAI_WEB_PREFIX_PATH',
     )
     def yatai_service_start(
-        db_url, repo_base_url, grpc_port, ui_port, ui, s3_endpoint_url, base_url
+        db_url, repo_base_url, grpc_port, ui_port, ui, s3_endpoint_url, web_prefix_path
     ):
         from bentoml.yatai.yatai_service import start_yatai_service_grpc_server
 
         start_yatai_service_grpc_server(
-            db_url, repo_base_url, grpc_port, ui_port, ui, s3_endpoint_url, base_url
+            db_url, repo_base_url, grpc_port, ui_port, ui, s3_endpoint_url, web_prefix_path
         )
--- a/bentoml/yatai/web/client/src/components/Breadcrumbs.tsx
+++ b/bentoml/yatai/web/client/src/components/Breadcrumbs.tsx
@@ -8,8 +8,11 @@ import {
 } from "@blueprintjs/core";
 import { Section } from "../ui/Layout";
 
+export interface IBreadcrumbsProps {
+  baseURL: string;
+}
 
-const Breadcrumbs: React.FC = (props) => {
+const Breadcrumbs: React.FC<IBreadcrumbsProps> = (props) => {
   const [breadcrumbs, setBreadcrumbs] = React.useState<Array<IBreadcrumbProps>>(
     []
   );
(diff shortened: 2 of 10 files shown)


yubozhao:
Good job! This looks great so far. I think with few updates, we can merge it in.

Let's make sure we update the CLI messages as well, when the user use the `--base-url` option
https://github.com/bentoml/BentoML/blob/5f21fc5f089e803d2c29479405e649520a798ae3/bentoml/yatai/yatai_service.py#L111

bentoml/cli/yatai_service.py
```diff
@@ -68,11 +68,19 @@ def add_yatai_service_sub_command(cli):
         'compatible with Amazon S3, such as MinIO',
         envvar='BENTOML_S3_ENDPOINT_URL',
     )
+    @click.option(
+        '--base-url',
+        type=click.STRING,
+        default='.',
+        help='Add a location prefix to the URL when running YataiService'
+        'behind a reverse proxy server',
+        envvar='BENTOML_BASE_URL',
```
yubozhao, line 77: Let's be more descriptive, maybe `BENTOML_YATAI_WEB_BASE_URL`?

bentoml/cli/yatai_service.py
```diff
@@ -68,11 +68,19 @@ def add_yatai_service_sub_command(cli):
         'compatible with Amazon S3, such as MinIO',
         envvar='BENTOML_S3_ENDPOINT_URL',
     )
+    @click.option(
+        '--base-url',
+        type=click.STRING,
+        default='.',
```
yubozhao, line 74: Would this work with more complex path like `yatai/another_path`?

Lets add validation as well

yubozhao: This option name might be confusing to users. Let's be more explict and name it something along the line of `--web-prefix-path`

bentoml/yatai/web/src/server.ts
```diff
@@ -195,26 +195,45 @@ const createRoutes = (app, yataiClient) => {
...
 
-export const getExpressApp = (grpcAddress: string | null) => {
+export const getExpressApp = (grpcAddress: string | null, baseURL: string) => {
   const app = express();
+  const cookieParser = require('cookie-parser');
+  app.use(cookieParser());
+  app.use(function (req, res, next) {
+    var cookie = req.cookies.cookieName;
+    if (cookie === undefined) {
+      //var baseURL="/yatai/";
```
flosincapite, line 205: Remove before submitting.

bentoml/yatai/web/src/server.ts
```diff
@@ -195,26 +195,45 @@ const createRoutes = (app, yataiClient) => {
...
+      res.cookie('baseURLCookie',baseURL, { maxAge: 900000, httpOnly: false });
+      console.log('cookie created successfully');
+    } else {
+      console.log('cookie exists', cookie);
+    } 
+    next(); // <-- important!
```
flosincapite, line 211: Why is it important? Please make the comment more descriptive.

bentoml/yatai/web/client/src/components/Breadcrumbs.tsx
```diff
@@ -8,19 +8,24 @@ import {
...
+  const {baseURL} = props;
+  console.log(baseURL)
+  const HOME_CRUMB: IBreadcrumbProps = {
+    text: "Home",
+    href: baseURL + '/'
+  };
 
   React.useEffect(() => {
-    const parsedBreadcrumbs = getBreadcrumbs(location.pathname);
+    const parsedBreadcrumbs = getBreadcrumbs(baseURL,location.pathname);
```
flosincapite, line 28: Nitpick: add space after the comma

pncnmnp:
> After I reinstall the NPM modules and rebuild the Yatai web, I tried to run the command bentoml yatai-service-start --base-url='yatai' --debug and it didn't seem working.
> Any idea why?

@yubozhao You are right, there seems to be an issue when using `--debug`. However, the issue persists on our master branch which has the following **HEAD**:
> commit 25c319d8161629d695eaecf6418e50fea6535d6e
Author: Bozhao <yubz86@gmail.com>
Date:   Thu Oct 15 09:08:27 2020 -0700
[Build]Lock web pack version to be compatible with react-scripts (#1178)

Also, regarding this -
> Let's make sure we update the CLI messages as well, when the user use the --base-url option

For now we can show the `prefix` but not the `port number`. 
The issue with implementing this feature is the lack of clarity regarding `.conf` file's destination. Presently, the only way to add the `port number` would be to fetch the `.conf` file and parse it, or let user provide the port number in an ENV variable. We are not sure if any other method exists.

bentoml/cli/yatai_service.py
kishore-ganesh: I tested with more complex paths, it does work

pncnmnp:
@yubozhao Can you check the changes? We updated the variable/env names, and made minor structural changes suggested by @flosincapite

yubozhao:
@pncnmnp Yeah, I saw the changes thanks.

I don't think the current implementation is working, because we didn't pass the `web-prefix-path` to the API routes.  If you see my screenshots from the previous comment, with the prefix, I don't see the list of the deployments and repositories on the home page.  
To dig a little bit deeper, you can see the API requests are sending to `HOST/PREFIX_PATH/api_endpoints`. The node server didn't update them, and those APIs are still in the routes like `api/listDeployments` and etc.

You will need to update all of the API routes in the node server

yubozhao (changes requested):
Need to update the base routes to APIs as well

kishore-ganesh:
@yubozhao If I'm not wrong, you tested the routes on localhost:3000/yatai right. Once the prefix is there, it would no longer works on localhost:3000, and it needs to be accessed through a reverse proxy with the given prefix. Then it does load the given configuration. We did not change the API server routes because the reverse proxy you use will remove the prefix. For example you request to localhost:5026/yatai/123, then it'll redirect the request to localhost:3000/123, so the change to the APIs wouldn't have worked. 

We can however also assume that the reverse proxy config should always be configured in such a way that it does not remove the prefix, in which case we can make the changes to the API side also. Should we do that?

yubozhao:
@kishore-ganesh Yep. I tested without a reverse proxy in front of it.  I think we should make it available for those users who just want it with the prefix path but didn't setup reverse proxy.

kishore-ganesh (commented):
The API routes now are prefixed correctly.

yubozhao:
@pncnmnp sorry about the late reply. 

This changes is good. I think we are very close. Need few more fixes. 

I ran the command with `abc` as my prefix
Screen Shot 2020-10-28 at 6 52 58 PM

And while the pages load the content correctly, it didn't seem the prefix path is passed to the server correctly and resulted in `undefined` as the prefix path
Screen Shot 2020-10-28 at 6 52 51 PM

bentoml/yatai/yatai_service.py
```diff
@@ -101,14 +101,15 @@ def start_yatai_service_grpc_server(
...
-            yatai_grpc_server_address, ui_port, web_ui_log_path, debug_mode
+            yatai_grpc_server_address, ui_port, web_ui_log_path, debug_mode, base_url
         )
 
     # We don't import _echo function from click_utils because of circular dep
     click.echo(
         f'* Starting BentoML YataiService gRPC Server\n'
         f'* Debug mode: { "on" if debug_mode else "off"}\n'
         f'* Web UI: {f"running on http://127.0.0.1:{ui_port}" if with_ui else "off"}\n'
+        f'* Web Prefix Path: {f"expecting to be hosted with prefix: {base_url}"}\n'
```
yubozhao, line 112: I think this is a very confusing message to users.  If users didn't pass in a prefix path, the additional message on prefix is going to confuse them even more.

Let's just construct the web UI messages outside of the echo command and pass in.

kishore-ganesh:
@yubozhao I think I understand why it's failing on debug. The current flow is that the server sets a cookie which the app root reads and communicates to its children. During debug, since a dev server runs (which does not provide the cookie functionality), and so the frontend doesn't know what's the correct url to send requests to. We'll think a bit about how to solve this.

yubozhao:
@kishore-ganesh Awesome. I think we are almost there. Few more changes and we should be good to merge

bentoml/adapters/annotated_image_input.py
```diff
@@ -113,6 +113,40 @@ def predict(
             <input type="submit">
         </form>
 
+    OR by python requests:
```
parano, line 116: Are these changes from another PR?

bentoml/yatai/yatai_service.py
yubozhao: Let's make sure this is changed before we merge it.

kishore-ganesh:
@yubozhao Sorry, couldn't work on this PR for a bit. We had one query before we proceeded further. So a bit of background, before the webpack merge, things worked on dev since there wasn't a dev server. Now that there's a dev server bundled in with webpack, it causes a port conflict with our actual server and so dev mode wasn't working correctly, since only one of the dev server or the web server could actually run. Now the obvious solution is to change the port of the dev server, which we have done and proxied the requests to our actual server. The problem is that this would break if the user specifies another port when running on dev (since the proxy is static). 

So create-react-app allows the option to pass in environment variables, which we were thinking of using (so if on debug, it'll not request relatively but absolutely to the url from the env variable). Another option would be to change the package.json file's proxy option itself before run, but that's a bit ugly. 

How do you think we should proceed?
Reference: [Create React App Docs](https://create-react-app.dev/docs/proxying-api-requests-in-development/)

yubozhao:
@kishore-ganesh Thanks for following up.

I think for the use-cases that need to have `--debug` flag on, your current solution is ok for now.

To make using `--debug` less brutal, I think you guys on the right approach with envvar.

We can pass in the port via environment var(https://create-react-app.dev/docs/adding-custom-environment-variables/#adding-temporary-environment-variables-in-your-shell)  and setup the proxy with `http-proxy-middleware` (https://create-react-app.dev/docs/proxying-api-requests-in-development/#configuring-the-proxy-manually).

Issue #1210, opened 2020-10-27
Log errors and stack trace inside inference API function user code


parano:
**Is your feature request related to a problem? Please describe.**

Currently, the BentoML API model server does not print the errors and stack trace when the exception was raised within the user's inference API function code. This makes it hard for users to debug issues in their code.

**Describe the solution you'd like**

Pieter Marsman suggested the following workaround in bentoml slack group:
```python
def log_traceback(f):
    """Log traceback before reraising error
    BentoML does not show traceback by default
    """
    @wraps(f)
    def wrapper(*args, **kwargs):
        try:
            return f(*args, **kwargs)
        except Exception as e:
            logger.exception('Something went wrong')
            raise e
    return wrapper
```

This should be the default behavior in BentoML

**Describe alternatives you've considered**

* always print the error type and error message, but only print stack trace if it’s in debug mode
* vs. always print the error message and stack trace

**Additional context**
\

Discussion in slack channel: https://bentoml.slack.com/archives/CKRANBHPH/p1603730524277500

AyoubDali:
Hey, 
this issue got my attention, can I work on it?

parano:
@AyoubDali sure thing!  I'm always available in slack if you have questions.

parano:
Hi @AyoubDali, just checking in, any update on this?

ramanaditya:
I have considered creating a decorator and using them by wrapping over all the functions.
Will it solve the problem?

parano:
@ramanaditya a decorator API for the user? I don't think that is necessary, this should be BentoML internal and default behavior, user does not need to be aware of it or write any code to invoke it.

ramanaditya:
@parano, not the decorator API for the user, it was meant to be internally used. Or, shall I script down the logger inside each of the function?

parano:
@ramanaditya oh if it's internal use then definitely we can do that to make the code changes cleaner!

PR #1331, opened 2020-12-16
Log traceback

ramanaditya:
## Description
Based upon the workaround suggested by Pieter Marsman
Created a `@log_traceback` decorator, to be internally used to print the stack trace, log the errors and raise the errors.

## Motivation and Context

**issue:** #1210 

## Types of changes

- [ ] Documentation
- [ ] Test, CI, or build

## Component(s) if applicable
- [ ] BentoML CLI

## Checklist:


bentoml/server/api_server.py
```diff
@@ -60,6 +61,28 @@
...
+            return func(*args, **kwargs)
+        except Exception:
+            exception_type, exception_instance, exception_traceback = sys.exc_info()
+            formatted_traceback = "".join(traceback.format_tb(exception_traceback))
+            exception_message = (
+                f"{formatted_traceback}\n{exception_type.__name__}: "
+                f"{exception_instance}"
+            )
+            logger.error(exception_message)
+            raise exception_type(exception_message)
```
flosincapite, line 81: If you just want to re-raise the original exception, you can do

`except Exception as exc`

at line 73, then, here, just do

`raise exc`

bentoml/server/api_server.py
```diff
@@ -60,6 +61,28 @@
...
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except Exception:
+            exception_type, exception_instance, exception_traceback = sys.exc_info()
+            formatted_traceback = "".join(traceback.format_tb(exception_traceback))
+            exception_message = (
+                f"{formatted_traceback}\n{exception_type.__name__}: "
```
flosincapite, line 77: Please add a test which exercises this code path.

bojiang:
IMO the main purpose of the issue is to log errors that occurred in the API function user code. It helps users to debug.
For the statics endpoint and swagger endpoint, all error happened should be the BentoML's internal error.

bojiang:
I just drafted a PR to demonstrate the solution in my mind.

ramanaditya:
> I just drafted a PR to demonstrate the solution in my mind.
> Do you think it makes sense?

The solution we created are pretty similar, the difference is I created the decorator, I thought it would make the code cleaner, 
about the implementation, we can directly wrap any function with this decorator.

If it was meant to be used in one place, then the draft PR you created looks nice. But if we want to use it at multiple places then I feel like we should go ahead with the function.


> For the statics and swagger endpoint, all error happened should be the BentoML's internal error, and raised directly.

I was not sure which all function to wrap, so I was just making the test over them. Please let me know the functions which we need to print traceback so that I could wrap them.

bentoml/server/api_server.py
```diff
@@ -141,6 +165,7 @@ def static_serve(static_path, file_path):
             )
 
     @staticmethod
+    @log_traceback
```
parano, line 168: this is probably not necessary, it is only meant for wrapping around user-defined API callback function.

parano: no need to wrap it around static_serve nor index_view_func, I think @bojiang's suggestion makes sense. Using a decorator does not seem to make the code cleaner in this case

yubozhao:
Close this with #1340 merged.

PR #1360, opened 2020-12-24
[Fix] Include conda environment file in the service module directory.

### LINKED ISSUE #1344: Bentoml is looking for environment.yml in the Service directory
I'm using a custom environment.yml and for some reason, while the environment.yml file is saved in the root directory, `bentoml serve` is looking for it in the actual Service directory.

```
Error: bentoml-cli serve failed: Can not find conda environment config yaml file at: `/home/lminer/bentoml/repository/MyService/20201218133541_A5C726/MyService/environment.yml`
```

It's saved instead to: `/home/lminer/bentoml/repository/MyService/20201218133541_A5C726/environment.yml`

works fine if I copy it over.

Bentoml version 0.10.1

yubozhao:
## Description

fix #1344 

Issue:

When users use a custom conda environment file in the BentoService config. It will raise `conda environment config yaml file not found` exception when loading the BentoService from the saved bundle.  This issue is caused by BentoService is looking for the environment.yml in the current module directory rather than the bundle directory one level up.

Solution:
Similar to `bentoml.yml` create an extra copy in the module directory.

## Types of changes

- [ ] Documentation
- [ ] Test, CI, or build

## Component(s) if applicable
- [ ] BentoML CLI

## Checklist:

### DIFF
--- a/bentoml/saved_bundle/bundler.py
+++ b/bentoml/saved_bundle/bundler.py
@@ -93,6 +93,12 @@ def _write_bento_content_to_dir(bento_service, path):
     # write conda environment, requirement.txt
     bento_service.env.infer_pip_packages(bento_service)
     bento_service.env.save(path)
+    # Also write conda environment.yml to module base path to make it accessible when
+    # users specificed custom conda environment.
+    shutil.copyfile(
+        os.path.join(path, 'environment.yml'),
+        os.path.join(path, bento_service.name, 'environment.yml'),
+    )
 
     # Copy all local python modules used by the module containing the `bento_service`'s
     # class definition to saved bundle directory
--- a/tests/test_service_env.py
+++ b/tests/test_service_env.py
@@ -247,3 +247,7 @@ dependencies:
 
     assert 'test-dep-1' in env_yml['dependencies']
     assert 'bentoml-test-lib' in env_yml['dependencies']
+
+    assert os.path.isfile(
+        Path(os.path.join(tmpdir, service_with_string.name, 'environment.yml'))
+    )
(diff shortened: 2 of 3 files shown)


bentoml/saved_bundle/bundler.py
```diff
@@ -93,6 +93,12 @@ def _write_bento_content_to_dir(bento_service, path):
     # write conda environment, requirement.txt
     bento_service.env.infer_pip_packages(bento_service)
     bento_service.env.save(path)
+    # Also write conda environment.yml to module base path to make it accessible when
+    # users specificed custom conda environment.
+    shutil.copyfile(
```
parano, line 98: why is this fixing the issue?

parano:
This only fixes when the user-specified `environment.yml` file is located in the same directory as the python module file that defines the BentoService class, is that right? If the user provides any path other than that, such as an absolute path or a relative path, this will fail, right? Can we come up with a solution that works for all those scenarios?

tests/test_service_env.py
```diff
@@ -247,3 +247,7 @@ def predict(self, df):
 
     assert 'test-dep-1' in env_yml['dependencies']
     assert 'bentoml-test-lib' in env_yml['dependencies']
+
+    assert os.path.isfile(
```
parano, line 251: Always test the actual behavior you are implementing or fixing - in this case, add a new example BentoService that uses the conda env file argument and test if it works as expected.

bentoml/saved_bundle/bundler.py
parano: the `bentoml.yml` was duplicated in that directory for a completely different reason, we should not duplicate the env yml file here - see comment below

Issue #998, opened 2020-08-19
Unexpected / undesirable behaviour with conda env settings


gregd33:
**Describe the bug**
Essentially, an `environment.yml` file will be produced no matter what (even if `conda_dependencies=None`). If this is used in conjunction with a base docker image that has conda installed + some packages, this can make dockerizing much slower (10x). 

Further, even if conda packages are specified, the inclusion of `python` and `pip` in the specification can have similar issues

**To Reproduce**
1. Create a base docker image on top of bentoml/model-server that has conda installed some packages (e.g. tensorflow)
2. Create a bentoml service that uses this base docker image via `docker_base_image` and set `conda_dependencies=None`
3. `make` the service and build the resulting docker image

**Expected behavior**
When the docker build gets to the conda step, it will have conflicts. It will eventually resolve and move on, but will greatly slow down the build time (from seconds to many minutes)

**Screenshots/Logs**
Example of the docker output 
![image](https://user-images.githubusercontent.com/62316076/90665325-e1e28500-e219-11ea-8de2-3aefe18167ff.png)

**Environment:**
 - CentOS 8:
 - Python/BentoML Version [e.g. Python 3.7.6, BentoML-0.8.3]

**Additional Context**

It also seems that `CONDA_ENV_DEFAULT_NAME` is unused.  I don't see the need for the name of the environment anyways since it is installed into `base`

**Potential Fixes**

For one, if no `conda_dependencies` are passed then don't create the YAML file. I've tested this quickly and it will work - it will skip the step in the same way it would if there is no conda environment. Or, don't include python in the file as that is what causes the conflict (having just pip is fine). 

Even when we do specify  `conda_dependencies`, including python in the environment is not ideal and cause the same issues. Presumably if someone is using a specific base image, they have chosen the version of python and overriding it could be problematic. Or if the version of python is important, they could specify it in the `conda_dependencies`

In short, excluding `python` from the environment.yml would be a reasonable fix. The downside is no explicit python version a) users can put that in their `conda_dependencies` if its important.

gregd33:
A couple other comments about this: 

1.  If the user does specify a specific version of python in the `conda_dependencies` this will cause issues since it will have now have python specified twice. So even more of a reason to remove python from this list.

2. I also think that the `default` channel should be removable or removed if there is another channel present. The context for this if images are being built without access to the default conda channel, this will fail even if reachable repos are given as channels.

gregd33:
[...]

PR #1375, opened 2021-01-04
[Fix] ensure Python version; exclude conda-forge from default conda channels

bojiang:
## Description

- always call `conda install python="{version}"`
- locked the channel of `python` and `pip`

## Motivation and Context

In the past two weeks, our integration tests on TensorFlow, Keras with docker failed at some point.
Here is an example: https://github.com/bentoml/BentoML/pull/1375/checks?check_run_id=1644036917

Two issues related to it in the `bento-init.sh`:
- If the Python version in the base image was the same as required, `conda install python={version}` would be skipped. The `python` package wasn't added to the specs of conda. Thus conda will also try to install the newest version of python, 3.9, in the next call.
- The channels of conda are set as ['conda-forge', 'defaults']. Plus we call `conda install python=3.7` to update Python. Sometimes this version code matches "python_pypy=3.7" in the conda-forge channel better. Thus conda will install PyPy instead of CPython.

There is no evidence that this problem will not occur in production.

## Types of changes

- [ ] Documentation
- [ ] Test, CI, or build

## Component(s) if applicable
- [ ] BentoML CLI

## Checklist:

### DIFF
--- a/bentoml/saved_bundle/bentoml-init.sh
+++ b/bentoml/saved_bundle/bentoml-init.sh
@@ -20,7 +20,7 @@ if [ -f ./python_version ]; then
 
   if command -v conda >/dev/null 2>&1; then
     echo "Installing python=$DESIRED_PY_VERSION with conda:"
-    conda install -y -n base conda-forge::python=$DESIRED_PY_VERSION
+    conda install -y -n base defaults::python=$DESIRED_PY_VERSION
   else
     if [[ "$DESIRED_PY_VERSION" != "$CURRENT_PY_VERSION" ]]; then
       echo "WARNING: Python Version $DESIRED_PY_VERSION is required, but $CURRENT_PY_VERSION was found."
--- a/bentoml/saved_bundle/bentoml-init.sh
+++ b/bentoml/saved_bundle/bentoml-init.sh
@@ -18,13 +18,11 @@ if [ -f ./python_version ]; then
   DESIRED_PY_VERSION=${PY_VERSION_SAVED:0:3} # returns 3.6, 3.7 or 3.8
   CURRENT_PY_VERSION=$(python -c 'import sys; print(f"{sys.version_info.major}.{sys.version_info.minor}")')
 
-  if [[ "$DESIRED_PY_VERSION" == "$CURRENT_PY_VERSION" ]]; then
-    echo "Python Version in docker base image $CURRENT_PY_VERSION matches requirement python=$DESIRED_PY_VERSION. Skipping."
+  if command -v conda >/dev/null 2>&1; then
+    echo "Installing python=$DESIRED_PY_VERSION with conda:"
+    conda install -y -n base python=$DESIRED_PY_VERSION
   else
-    if command -v conda >/dev/null 2>&1; then
-      echo "Installing python=$DESIRED_PY_VERSION with conda:"
-      conda install -y -n base python=$DESIRED_PY_VERSION
-    else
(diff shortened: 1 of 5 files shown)


bentoml/saved_bundle/bentoml-init.sh
```diff
@@ -18,13 +18,11 @@ if [ -f ./python_version ]; then
   DESIRED_PY_VERSION=${PY_VERSION_SAVED:0:3} # returns 3.6, 3.7 or 3.8
   CURRENT_PY_VERSION=$(python -c 'import sys; print(f"{sys.version_info.major}.{sys.version_info.minor}")')
 
-  if [[ "$DESIRED_PY_VERSION" == "$CURRENT_PY_VERSION" ]]; then
-    echo "Python Version in docker base image $CURRENT_PY_VERSION matches requirement python=$DESIRED_PY_VERSION. Skipping."
+  if command -v conda >/dev/null 2>&1; then
```
parano, line 21: This was meant to avoid installing Python - when the user uses a base image with Python pre-installed.  Is the idea that if `conda` command is presented, then re-installing the same python version should be fast?  Does a different `patch` version affects this behavior? For example, if the base image has 3.7.1 but BentoService requires 3.7.4?

bojiang: As I said in the context, we have to call `conda install` to lock the Python version, or it will try to install Python3.9 once `conda install pip` called.

bojiang:
@parano ready

parano:
Looks great, thanks @bojiang!

gregd33:
This seems to re-break the issues I mentioned in https://github.com/bentoml/BentoML/issues/998 by needlessly installing python via conda even if no conda installs are being done. 

If understand, the issue about it trying to install python=3.9 will only occur if there is a non-empty conda install. So could the logic be changed to modify the `env.py` module to add the python dependency iff there is a conda dependency? So change https://github.com/bojiang/BentoML/blob/22d8c0702cfcdf9da0628a39e97fb86d96f51b4d/bentoml/service/env.py#L99 to add PYTHON_MINOR_VERSION  if it's not in there, or something like that.

Unless I'm missing another case where there would be a conda instlal being done

parano:
Hey @gregd33  - did you experience slow down when building docker image with the new release? I believe even tho it runs the `conda install` command to install python here, the required python version should be already installed in the base image and conda should just skip the installation. It only tries to install python iff the user is using a custom docker base image, the base image has conda installed and the conda env does not have the required python version.

gregd33:
"the conda env does not have the required python version"

You are right that it will only do the install in this case. But it will try in all cases (i.e. it will run the `conda install` in all cases that conda is present). There are a couple of potential problems:

- As discussed in #998 this can cause a significant slow down if the existing environment is complex as it checks for all conflicts
- If the container is built in an environment without easy conda access (either no access or custom urls) then this will break/require extra config

The latter was the case for me.

It's not the worst thing, since there are workarounds and I've been trying to move towards the `slim` images where possible. 

But I think I've seen this premise elsewhere - that conda installing python is a trivial step that will essentially be skipped if it's in place. And I think that premise should be reconsidered for future updates. Although I guess changes made to the services now to accommodate this update would also work for other things that rely on this assumption.

bojiang:
@gregd33 
I think it makes sense.
I'll draft a new PR to revert that part.
Just removing "conda-forge" from the channels already resolved my issue.
Thank you!

gregd33:
Great, thanks a lot!

PR #1405, opened 2021-01-14
opentracing implementation with jaeger

vgeshiktor:
## Description
Adds [opentracing](https://opentracing.io/) implementation with [jaeger](https://www.jaegertracing.io/).
opentracing is vendor-neutral APIs and instrumentation for distributed tracing

Output example:
Screen Shot 2021-01-12 at 20 46 25

By default opentracing is off, like zipkin, 
can be turned on like zipkin by providing server address and port default_bentoml.cfg:
`
[tracing]
# example: http://127.0.0.1:9411/api/v1/spans
zipkin_api_url =
# example: localhost
opentracing_server_address = localhost
# example: 5775
opentracing_server_port = 5775
`

## Motivation and Context
Feature of opentracing provides additional functionality and methods of tracing calls, will ease debugging and performance checks kind of problems, provides more standard way of reporting and using traces and also eliminates vendor dependencies. Supports also zipkin trace format, so can totally replace it.

## How Has This Been Tested?
Unit tests in tests/server - passed on following scenarios:
1. No tracing at all
2. zipkin tracing is on
3. opentracing is on
Tests run locally on macos (m1) python 3.8

## Types of changes
- [ ] Documentation
- [ ] Test, CI, or build

## Component(s) if applicable
- [ ] BentoML CLI

## Checklist:

### DIFF
--- a/bentoml/server/opentrace.py
+++ b/bentoml/server/opentrace.py
@@ -17,13 +17,10 @@ def initialize_tracer(service_name, log_level=logging.DEBUG):
     logging.basicConfig(level=log_level)
 
     config = Config(
-        config={
-            'sampler': {'type': 'const', 'param': 1},
-            'logging': True
-        },
+        config={'sampler': {'type': 'const', 'param': 1}, 'logging': True},
         service_name=service_name,
         validate=True,
-        scope_manager=AsyncioScopeManager()
+        scope_manager=AsyncioScopeManager(),
     )
 
     return config.initialize_tracer()
@@ -31,15 +28,15 @@ def initialize_tracer(service_name, log_level=logging.DEBUG):
 
 @contextmanager
 def trace(
-        server_url=None,
-        request_headers=None,
-        async_transport=False,
-        sample_rate=1.0,
-        standalone=False,
-        is_root=False,
-        service_name="some service",
-        span_name="service procedure",
-        port=0,
+    server_url=None,
+    request_headers=None,
+    async_transport=False,
+    sample_rate=1.0,
+    standalone=False,
+    is_root=False,
+    service_name="some service",
+    span_name="service procedure",
+    port=0,
 ):
     """
     Opentracing tracer function
@@ -59,8 +56,9 @@ def trace(
     if span_context is None:
         span_context = span_context_saved or None
 
-    with tracer.start_active_span(operation_name=span_name,
-                                  child_of=span_context) as scope:
+    with tracer.start_active_span(
+        operation_name=span_name, child_of=span_context
+    ) as scope:
         token = span_context_var.set(scope.span.context)
         yield scope
         span_context_var.reset(token)
--- a/bentoml/server/opentrace.py
+++ b/bentoml/server/opentrace.py
@@ -28,15 +28,15 @@ def initialize_tracer(service_name, log_level=logging.DEBUG):
 
 @contextmanager
 def trace(
-        server_url=None,  # @UnusedVariable
-        request_headers=None,
-        async_transport=False,  # @UnusedVariable
-        sample_rate=1.0,  # @UnusedVariable
-        standalone=False,  # @UnusedVariable
-        is_root=False,  # @UnusedVariable
-        service_name="some service",
-        span_name="service procedure",
-        port=0,  # @UnusedVariable
+    server_url=None,  # @UnusedVariable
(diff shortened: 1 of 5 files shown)


bentoml/server/opentrace.py
```diff
@@ -0,0 +1,67 @@
...
+    )
+
+    return config.initialize_tracer()
+
+
+@contextmanager
+def trace(
+    server_url=None,
+    request_headers=None,
+    async_transport=False,
```
bojiang, line 33: How does `async_transport` work here?

bojiang:
Thanks for your contribution!
It's great to support the opentracing API.

Please note that the marshal server(the micro-batching server) is fully non-blocking. So if the tracing lib sends messages in a blocking way, the performance of the batching server would be significantly affected.

bojiang (dismissed):
Please make sure the transport is non-blocking when `async_transport=True`

vgeshiktor:
> Please make sure the transport is non-blocking when `async_transport=True`

Please look into [jaeger client architecture](https://www.jaegertracing.io/docs/1.17/architecture/#jaeger-client-libraries) to address this concern. I hope it clears everything. Currently, jaeger client does not provide non blocking communication, but according to jaeger documentation, the impact should be minimal. Sampling can also be configured to minimize impact.

bojiang:
Context:
* `jaeger-client-python` uses `tornado` as the transporter(reporter) implementation. https://github.com/jaegertracing/jaeger-client-python/blob/d83b7b1d43c003607682b927b1b251eec0bb61df/jaeger_client/reporter.py#L121
* `jaeger-client-python` reports spans: https://github.com/jaegertracing/jaeger-client-python/blob/master/jaeger_client/reporter.py#L133

bojiang:
Have you tested the API server with both trace URL provided and `--enable-microbatch`?
@vgeshiktor

bojiang:
My previous concerns have almost been resolved. But I still have two questions.
* Have you tested the API server with both trace URL provided and `--enable-microbatch`?
* Do we need to pass the event loop to tracer initialization as mentioned in the [readme](https://github.com/jaegertracing/jaeger-client-python#initialization--configuration) of `jaeger-client-python`?
@vgeshiktor

vgeshiktor:
> My previous concerns have almost been resolved. But I still have two questions.
> 
> 1. Have you tested the API server with both trace URL provided and `--enable-microbatch`? Is the server working properly?
> 2. Do we need to pass the event loop to tracer initialization as mentioned in the [readme](https://github.com/jaegertracing/jaeger-client-python#initialization--configuration) of `jaeger-client-python`?
> 
> For Q1, I suggest simply modifying the current integration tests of the API server, starting it with a dummy tracing address.
> @vgeshiktor

1. I ran test_microbatch.py and had the same results with and without jaeger(4 passed, 4 ignored).
2. No, we do not need to pass event loop to the tracer.
3. I ran test_microbatch.py with/without jaeger server running and had same results (4 passed, 4 ignored)
4. same happens with/without server address in configuration. I implemented the same logic as zipkin implementation to test for url and respond accordingly.

Please let me know if you having other concerns.

parano:
Thanks for the PR @vgeshiktor - it all looks good to me with two caveats:

1. We are introducing a new config system here: https://github.com/bentoml/BentoML/pull/1422, we plan to merge #1422 soon and it would be great if you could update the config part to adopt the new pattern
2. Could you make `opentracing` and `jaeger_client` optional dependency? Meaning that we should not add it the `setup.py` and require the user to install it in order to use BentoML while we can add it to our default docker image [here](https://github.com/bentoml/BentoML/blob/v0.11.0/docker/model-server/Dockerfile#L19). This will require BentoML code to do a conditional import, to only import those modules if the config `opentracing_server_address` was set by the user. We plan to do the same for the `py_zipkin` module.

vgeshiktor:
Hi @parano,

1. updated configuration with new style configuration.
2. imports are already conditional, including import for py_zipkin

parano:
@vgeshiktor  could you merge with master and resolve conflicts?  It looks like this PR is still using the old config.

If those are already conditional import, could we remove them from the `setup.py` file and add them to the base docker image instead?

vgeshiktor:
Hi, 

 I am closing this pull request and instead of it will create a new pull request that will incorporate latest requirements from @parano.

Issue #1415, opened 2021-01-20
Allow BentoML User to set custom access_token headers


jack1902:
**Is your feature request related to a problem? Please describe.**

I'm frustrated by the static `access_token` header set when i set the `access_token` in my config file. This severly limits me using a custom access-token header with JWT and an OIDC provider

**Describe the solution you'd like**

I'd like to be able to define a new configuration option in the `config` file similar to the below:

https://github.com/bentoml/BentoML/blob/506d5026f89916b684493198db5bb2707a11537f/bentoml/yatai/yatai_service.py#L26

Which lets me define `access_token_header` which allows me to set a new config option like:

```python
access_token_header = access_token_header or config('yatai_service').get('access_token_header')

header_adder_interceptor = header_client_interceptor.header_adder_interceptor(
    access_token_header, access_token
)
```

**Describe alternatives you've considered**

- Have the end-user authenticate somewhere else (Copying their `access_token` into their local config) and then ensure Istio checks the `access_token` header. This does mean that web browser requests to the service won't be redirected if they don't contain the valid headers though making for a less than enjoyable experience 😞) 

**Additional context**

I am running BentoML which is fronted by Istio. Using Istio `AuthorizationPolicy` i can dictate that a valid JWT is present before letting traffic proceed to `yatai`. In front of this, i would be running an ALB which has a target group for `gRPC` traffic but the first action on the ALB is to valid the JWT on the incoming request and re-route the user to the OIDC Provider if it is not valid. I took inspiration for this from this blog post: [Blog Post](https://zdenek-papez.medium.com/istio-authorization-using-okta-user-groups-in-jwt-claims-behind-aws-application-load-balancer-191f26779cad)

Issue #1374, opened 2021-01-04
BentoML Configuration Redesign


ssheng:
### Background
BentoML introduced a configuration system in the early days to simplify user customizations and scale developer productivity. The design starts with an out-of-box configuration that just works. For advanced users who need additional customizations, configuration can be overridden with user supplied values. A number of pain points summarized below have been encountered by our developers and users with the current design. We want to take this opportunity to redesign the configuration system and scale BentoML for the next five years.
#### Too Many Ways to Override
There are too many ways to override the same configuration key making it difficult for users and developers to establish a clear mental model.
- CLI explicit parameters
- CLI config settings ($ bentoml config set key=value)
- Environment variables
- Configuration file override
#### Convoluted Override Precedence
Keys defined in the bentoml.cfg already have a well established override precedence. However, the override precedence between CLI parameters and BentoML configurations still rely on application logic to define, particularly for configuration keys prefixed with “default”.
#### All-or-Nothing Override
Users must provide a full bentoml.cfg configuration file even if they wish to override only a handful of keys. Due to the same limitation, BentoML config CLI copies the entire default configuration to the BentoML home directory when the user sets a single configuration key. Once a copy of bentoml.cfg is placed, evolving the default configuration becomes impossible unless the user manually deletes the bentoml.cfg file.
#### Lack of Configuration Scope
Overridden configurations apply to all subsequent BentoML CLI commands, users may not wish to apply the same configuration to two different commands.
#### Configuration Coupling with Implementation
Implementation classes all have to import and be aware of the configuration module, making unit testing and code modularization challenging.
#### Lack of Hierarchy and Data Types
The current “.ini” format of the configuration file does not easily allow configuration key hierarchy and collection data types, such as list and dict.
### Goals
1. Default configurations should continue to work for most of the use cases
2. Deterministic precedence order of configuration override
3. Allow partial overrides through any means
4. Limit configuration override to only the current session (no persistence)
5. Decouple configuration with implementation classes
6. Introduce hierarchy and collection data types
7. Backward compatible with the current configuration system
### Design
Started a design draft. Comments are welcome. https://docs.google.com/document/d/1kriInVhjCEoXblh_hBBRwFfi6VwgzwFogQ2tyDoRa5U/edit#heading=h.l0i6f5osh4wr

### Pull Requests
- [x] Introduce Dependency Injection and Configuration Revamp (https://github.com/bentoml/BentoML/pull/1422)
[...]

PR #1416, opened 2021-01-21
Lets user set a custom access_token header

jack1902:
## Description

Will allow the user to set a custom `access_token_header` within their config which will then ensure the set `access_token` is used under that Header defined when making requests with `BentoML` to `Yatai`

I am happy to contribute some docs around this if this works as expected so others are able to use authentication in-front of yatai in the future 👍 

## Motivation and Context

relates to: https://github.com/bentoml/BentoML/issues/1415

Essentially want to pass a JWT via a custom header to whatever sits in front of `Yatai`. Currently the hard-coded `access_token` header limits my options on the service that can sit in front of `yatai` to valid these tokens

## How Has This Been Tested?

Wasn't able to get the pip packages installed on my MacBook, will reattempt then but hoping for tests to run upon this PR being opened

## Types of changes

- [ ] Documentation
- [ ] Test, CI, or build

## Component(s) if applicable
- [x] BentoML CLI

## Checklist:

### DIFF
--- a/bentoml/configuration/default_bentoml.cfg
+++ b/bentoml/configuration/default_bentoml.cfg
@@ -64,6 +64,7 @@ tls_root_ca_cert =
 tls_client_key =
 tls_client_cert =
 access_token =
+access_token_header =
 
 [apiserver]
 default_port = 5000
--- a/bentoml/configuration/default_bentoml.cfg
+++ b/bentoml/configuration/default_bentoml.cfg
@@ -64,7 +64,7 @@ tls_root_ca_cert =
 tls_client_key =
 tls_client_cert =
 access_token =
-access_token_header =
+access_token_header = access_token
 
 [apiserver]
 default_port = 5000
--- a/bentoml/yatai/yatai_service.py
+++ b/bentoml/yatai/yatai_service.py
@@ -17,6 +17,7 @@ from bentoml.yatai.utils import ensure_node_available_or_raise, parse_grpc_url
 def get_yatai_service(
     channel_address=None,
     access_token=None,
+    access_token_header=None,
     db_url=None,
     repo_base_url=None,
     s3_endpoint_url=None,
@@ -24,6 +25,9 @@ def get_yatai_service(
 ):
     channel_address = channel_address or config('yatai_service').get('url')
     access_token = access_token or config('yatai_service').get('access_token')
+    access_token_header = access_token_header or config('yatai_service').get(
+        'access_token_header'
+    )
     channel_address = channel_address.strip()
     if channel_address:
         # Lazily import grpcio for YataiSerivce gRPC related actions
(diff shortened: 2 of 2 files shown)


bentoml/configuration/default_bentoml.cfg
```diff
@@ -64,6 +64,7 @@ tls_root_ca_cert =
 tls_client_key =
 tls_client_cert =
 access_token =
+access_token_header =
```
parano, line 67: Use `access_token_header = access_token` here as the default value to make sure we don't introduce a breaking change?

jack1902: Would be better to do it in the codes kwargs if we want to ensure its backwards compatible? As a user's config will only be reset to this files of they run bentonl config reset

parano: This is a new config key so it doesn't have that issue.

I think you must have noticed that the current config module in BentoML has a flaw that it prevents us from changing the default config value easily because it copies the entire default config file to the local bentoml home. But we are fixing that issue in our upcoming new config system https://github.com/bentoml/BentoML/issues/1374

parano:
@jack1902 could you tell me why you need a custom token header? Isn't that something very easily configurable on the frontend authentication side?

jack1902:
> @jack1902 could you tell me why you need a custom token header? Isn't that something very easily configurable on the frontend authentication side?

As mentioned in the linked issue, i wanted to ensure i could set a specific Header which is looked for by an AWS ALB, as the current statically defined header `access_token` is too restrictive in terms of being able to use this tool for a true auth flow

parano:
@jack1902 what is the default auth header in AWS ALB?

jack1902:
> @jack1902 what is the default auth header in AWS ALB? I'm just wondering if that's a better default value. Could you share link to related AWS documentation?

The header key would be `x-amzn-oidc-accesstoken` See below for the link to the supporting documentation and a screenshot of the relevant bit @parano 

https://docs.aws.amazon.com/elasticloadbalancing/latest/application/listener-authenticate-users.html

Screenshot 2021-01-29 at 15 52 33

parano:
Thanks, @jack1902 that's super helpful, I think this PR makes sense and would love to merge it.  Could you make the change I requested above, to add a default value in the default config? `access_token_header = access_token`. Some tests are failing due to this issue.

jack1902:
> Thanks, @jack1902 that's super helpful, I think this PR makes sense and would love to merge it. Could you make the change I requested above, to add a default value in the default config? `access_token_header = access_token`. Some tests are failing due to this issue.

I've pushed the requested change, however the config will still blow up as this is basically a breaking change for anyone who already has a `config` file

parano:
@jack1902 thank you for updating the PR! merged!

PR #1422, opened 2021-01-24
[Enhancement] Introduce Dependency Injection and Configuration Revamp (#1374)

ssheng:
## Description, Motivation, and Context
Introduced dependency injection mechanism to BentoML through [dependency-injector](https://pypi.org/project/dependency-injector/). New configuration system is introduced with backward compatibility support. Migrated server and marshal packages as proof-of-concept. For more details, please see the overall [design document](https://docs.google.com/document/d/1kriInVhjCEoXblh_hBBRwFfi6VwgzwFogQ2tyDoRa5U/edit#heading=h.gqmla76xy4u9).

## How Has This Been Tested?
- ./ci/unit_test.sh
- Manual testing

## Types of changes

- [ ] Documentation
- [ ] Test, CI, or build

## Component(s) if applicable
- [x] BentoML CLI

## Checklist:


bentoml/marshal/marshal.py
```diff
@@ -197,14 +208,12 @@ def setup_routes_from_pb(self, bento_service_metadata_pb):
...
-                    or self.DEFAULT_MAX_LATENCY
+                    api_pb.mb_max_latency
+                    or self.mb_max_latency
                 )
                 max_batch_size = (
-                    self.mb_max_batch_size
-                    or api_pb.mb_max_batch_size
-                    or self.DEFAULT_MAX_BATCH_SIZE
+                    api_pb.mb_max_batch_size
+                    or self.mb_max_batch_size
```
ssheng, line 216: This logic is the only backward incompatible part. CLI parameter `self.mb_max_latency` and legacy configuration key `self.DEFAULT_MAX_LATENCY` were already resolved outside of this class. Therefore we cannot apply the override priority as specified in the old logic. The new logic will prioritize the `max_latency` and `max_batch_size` specified in the Bento metadata over CLI parameter.

We should discuss our long term plan fo the metadata parameters. How they will co-exist (or not) with the configuration in the bento bundle.

bentoml/configuration/containers.py
```diff
@@ -0,0 +1,170 @@
...
+    "yatai": {
+        "url": Or(str, None),
+    },
+    "tracing": {
+        "zipkin_api_url": Or(str, None),
+    },
+    "instrument": {
+        "namespace": str,
+    },
+})
```
ssheng, line 56: Compared the cerberus and schema libraries. I prefer the schema library because it is more pythonic and allow python logic in the definition of schema. However, cerberus has better documentation. Happy to discuss more.

yubozhao: Both of them are good libraries, hard to go wrong.  I think `schema`'s more pythonic approach is better in our case.

For allowing python logic, we are using Cerberus this way: https://github.com/bentoml/BentoML/blob/93cd4b1b6ef8488473736399a979e5800cee2075/bentoml/yatai/validator/deployment_pb_validator.py#L154

Seems like `Schema` is cleaner to deal with.

bentoml/configuration/__init__.py
```diff
@@ -61,6 +61,7 @@ def parameterized_config(template):
 
 DEFAULT_BENTOML_HOME = expand_env_var(os.environ.get("BENTOML_HOME", "~/bentoml"))
 BENTOML_HOME = DEFAULT_BENTOML_HOME
+BENTOML_CONFIG = os.path.join(BENTOML_HOME, "bentoml.yml")
```
parano, line 64: This is interesting, it means users will be able to put a default config file under the BENTOML_HOME directory which is expected to affect all BentoML usage on that machine?

For configs that affect the BentoML usage during the model packaging stage, does a regular `import bentoml` pick up this default config file? And how do users set a config file in that scenario where no CLI command is involved?

We talked about adding a python API for setting a config before, I think it could be something like this:

```python
import bentoml
bentoml.set_config("~/foo/bar/my_bentoml_config.yml")

@env(...)
@artifacts(...)
class MyClassifier(bentoml.BentoService):

     @api(...)
     def classify(self, ...):
             ...
```
And in the `bentoml/__init__.py` file, we can have it load the default config in BENTOML_HOME, with `bentoml.set_config(benotml.configuration.BENTOML_CONFIG)`.

bentoml/configuration/default_bentoml.yml
```diff
@@ -0,0 +1,42 @@
...
+  enable_feedback: True
+  max_request_size: 20971520
+
+  #
+  # Number of worker threads to be configured for the API server. If Null is
+  # configured, then BentoML will determine an optimal number of worker threads
+  # automatically, see `bentoml.configuration.containers.BentoMLContainer.
+  # api_server_workers` for details.
+  #
+  workers: Null
```
parano, line 21: we discussed changing this default worker number to 1, do you want to make that change in this PR? I can help update related documentation before the next release and make sure we communicate the breaking change with users in the release note

bentoml/configuration/__init__.py
ssheng: > This is interesting, it means users will be able to put a default config file under the BENTOML_HOME directory which is expected to affect all BentoML usage on that machine?

The goal behind reading config from BentoML home is that, when user containerize with a config option, the config will be placed under BentoML home and be automatically picked up when the docker image is run.
```zsh
$ bentoml containerize bento --config ~/foo/bar/bentoml.yml
```

> For configs that affect the BentoML usage during the model packaging stage, does a regular import bentoml pick up this default config file? And how do users set a config file in that scenario where no CLI command is involved?

We have the option to include model packaging stage in the scope of the config. Do you see a need to have config there as the packaging logic is pretty standard and non-customizable.

> We talked about adding a python API for setting a config before, I think it could be something like this:

To package a config into a bundle, I was thinking something like the following.

```py
import bentoml
from iris_classifier import IrisClassifier  # example service from the quick start guide

model = ... # training a model

svc = IrisClassifier()
svc.pack(name='model', model=model, config="~/foo/bar/my_bentoml_config.yml")
svc.save()
```

bentoml/configuration/default_bentoml.yml
ssheng: Sure. I can include that change in this PR. What is the reason for changing default workers to 1? The current logic based on CPU cores seems more correct.

parano: @ssheng it is because the current calculation based on CPU doesn't really make sense for many use cases and caused lots of confusion for the user. I think worker=1 as a default makes it more obvious that it's a deployment option that users should decide. Plus that 1 is a better default than a variable number associate with CPU: with our adaptive micro-batching layer, `worker=1` still allows the framework itself to take advantage of multiple CPUs when applicable.

ssheng: @parano I would like to learn more about the use cases that don't work with CPU-based worker count. IMHO, users are not best equipped to determine the number of works. The configuration would be different from environment to environment. In an async system like gunicorn, we are in a good position to determine worker count on user's behalf and a constant * the number of cores is typically a very good proxy.

tests/utils/test_usage_stats.py
```diff
@@ -120,6 +120,7 @@ def test_track_cli_with_keyboard_interrupt(bento_bundle_path):
         cli = create_bentoml_cli()
         runner.invoke(cli.commands['serve'], [bento_bundle_path])
         _, properties = mock.call_args_list[0][0]
+        print(properties)
```
johnhuangclever, line 123: remove this pls

bentoml/configuration/containers.py
```diff
@@ -0,0 +1,172 @@
...
+                        "Configuration after user override does not conform to"
+                        " the required schema."
+                    ) from e
+
+    def override(self, keys: list, value):
+        if keys is None:
+            raise BentoMLConfigException("Configuration override key is None.")
+        if len(keys) == 0:
+            raise BentoMLConfigException("Configuration override key is empty.")
+        if value is None:
```
parano, line 141: Is it possible that the user is actually trying to set the value to None?

ssheng: This pattern will prevent the user from overriding a value to None. I don't know if we have any use cases where it would be valid for overriding the value to None. Maybe after we change the workers default to 1, the user might want BentoML to automatically set workers. In those use cases, we can perhaps use a special value, e.g. -1 or add another configuration flag `--automatic-workers` to allow this configuration.

parano: that makes sense, I was just curious if the user wants to reset a config value to the default, which can often be None. But I guess that's not supposed to happen because, with your new design, we encourage users to only have one source of config value. And I think an `auto worker` flag is a great idea, we should explore that.

bentoml/marshal/marshal.py
```diff
@@ -141,6 +140,13 @@ def __init__(
         outbound_workers=1,
         mb_max_batch_size: int = None,
         mb_max_latency: int = None,
+        request_header_flag: str = Provide[
+            BentoMLContainer.config.marshal_server.request_header_flag
```
parano, line 144: use `from bentoml.configuration.containers import BentoMLContainer as BTC` to make it shorter? 🤔🤔🤔

parano:
@ssheng is there any pending work on your end that you'd like to add to this PR? I feel pretty good with it and would suggest we merge this one and add other related work in follow up PR, which also allow others to start adopting this new config system. What do you think?

bentoml/marshal/marshal.py
ssheng: Elon Must would agree. :)

ssheng:
> @ssheng is there any pending work on your end that you'd like to add to this PR? I feel pretty good with it and would suggest we merge this one and add other related work in follow up PR, which also allow others to start adopting this new config system. What do you think?

I'm working on addressing the incompatibility of the `bentoml serve` command on Windows, by separating gunicorn modules to a different package. While doing that, I encountered a problem with Python multiprocess start method. It was addressed in https://github.com/bentoml/BentoML/pull/1440. We can merge after I get that addressed.

PR #1532, opened 2021-03-25
Fix zipkin module not found issue

parano:
## Types of changes

- [ ] Documentation
- [ ] Test, CI, or build

## Component(s) if applicable
- [ ] BentoML CLI

## Checklist:


bentoml/tracing/_zipkin.py
```diff
@@ -12,6 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# This module name starts with underscore because it contains optional Python package
```
parano, line 15: @ssheng looks like walk_packages will skip modules starting with underscore, so I used it for these two modules that has optional dependency and we want to avoid loading if it's not used

ssheng: Correct and this is probably fine for now. The long term solution is still limit `wire` to BentoML modules.

bentoml/configuration/containers.py
```diff
@@ -47,9 +49,10 @@
         },
         "yatai": {"url": Or(str, None)},
         "tracing": {
-            "zipkin_api_url": Or(str, None),
+            Optional("zipkin_server_address"): Or(str, None),
```
bojiang, line 52: Why `zipkin_server_address` and port rather than `zipkin_api_url`?
zipkin service is a REST HTTP service, `zipkin_server_address` and `zipkin_server_port` can not represent the API.
For example, `http://127.0.0.1:8080/zipkin/v1/trace`

bentoml/tracing/zipkin.py
```diff
@@ -109,66 +89,64 @@ async def _async_post(url, data, headers):
                 return resp
 
     def send(self, payload):
+        path, content_type = self._get_path_content_type(payload)
```
bojiang, line 92: Where is the path from?

bentoml/tracing/opentrace.py
```diff
@@ -1,74 +1,99 @@
...
-        yield scope
-        span_context_var.reset(token)
+        if standalone:
+            yield None
+            return
+        else:
+            token = span_context_var.set(scope.span.context)
+            if request_headers:
+                tracer.inject(
+                    scope.span.context, opentracing.Format.HTTP_HEADERS, request_headers
```
bojiang, line 92: 👍

bentoml/tracing/zipkin.py
parano: It's from the `SimpleHTTPTransport`

bentoml/configuration/containers.py
parano: @bojiang what is the convention used in zipkin's documentation & community?  I tried to follow zipkin quickstart guide and setup it with BentoML, it seems to me that specifying an address and host is a bit easier for the user since the user don't need to worry about whether BentoML is using API v1 or v2

parano: I looked at a few other Zipkin libraries and it seems like a URL is more often used, I guess we just need to better document this and ask users to provide a full URL like `http://zipkinhost:9411/api/v2/spans`

bentoml/configuration/containers.py
```diff
@@ -47,9 +47,10 @@
         },
         "yatai": {"url": Or(str, None)},
         "tracing": {
-            "zipkin_api_url": Or(str, None),
+            "zipkin_server_address": Or(str, None),
```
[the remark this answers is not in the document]
parano: Looks like wiring only the tracing module fixed the issue for us. I will try another level of config here, that looks much better.

I think creating a provider for trace and async_trace is a great idea, although they are currently being used at places where there's no `@inject` call. I probably will leave that out of this PR for now.

bentoml/configuration/containers.py
bojiang: > I looked at a few other Zipkin libraries and it seems like a URL is more often used

It makes sense because zipkin is using HTTP as the transport protocol. (host, port) pair is for L4 protocols.

bojiang: > it seems to me that specifying an address and host is a bit easier for the user since the user don't need to worry about whether BentoML is using API v1 or v2

But the `path` won't be guaranteed by zipkin implements, for example, jaeger. There are no specs that claim a zinkin server has to listen on a specific `path`.

parano: @bojiang depends if we want to target the use of default zipkin server https://github.com/openzipkin/zipkin/tree/master/zipkin-server or custom zipkin collector that implements this endpoint

I think the default zipkin server is more commonly used although the other approach is a bit more flexible

bentoml/server/gunicorn_server.py
```diff
@@ -17,11 +17,11 @@
 from typing import Optional
 
 import psutil
-from dependency_injector.wiring import Provide as P
+from dependency_injector.wiring import Provide
```
ssheng, line 20: nit: move to same line as wire.

ssheng (dismissed):
I have no other concerns beside abstracting different trace implementation behind a common `Tracer` class and update the config to reflect it.

bentoml/server/gunicorn_server.py
parano: End up changing this back to avoid conflicts with other ongoing PRs

PR #1543, opened 2021-03-29
Update dependency injection to wire only once

ssheng:
## Description

Update dependency injection logic to run wire only once during the top level initialization.

## Motivation and Context

Pros:

- Unify all container and wire logic in one place at the top level initialization.
- Simplify the logic of dependency injection and explicit argument parsing, such that Python API and CLI now share the same library function. 
- Direct use Python APIs will also have dependencies and configuration injected.

Cons:

- Top level dependency injection adds an average of 0.75s latency when either the Python API or CLI is first invoked, because the dependency injection logic tries to recursively import all modules. I started a separate [discussion](https://github.com/ets-labs/python-dependency-injector/discussions/437) with dependency-injector community on avoiding the cascading imports. Meanwhile, we can have modules that import large libraries do lazy import inside the function.

## Types of changes

- [ ] Documentation
- [ ] Test, CI, or build

## Component(s) if applicable
- [ ] BentoML CLI

## Checklist:


bentoml/__init__.py
```diff
@@ -34,6 +37,10 @@
     save,
 )
 
+from bentoml.cli import create_bentoml_cli
+
+commandline_interface = create_bentoml_cli()
```
parano, line 42: What's the reason for moving it up here?

bentoml/__init__.py
```diff
@@ -17,12 +17,15 @@
...
 
-from bentoml.configuration import config
+from bentoml.configuration import config, inject_dependencies
 from bentoml.utils.log import configure_logging
 
 # Configuring logging properly before loading other modules
 configure_logging()
 
+# Inject dependencies and configurations
+inject_dependencies()
```
parano, line 27: Is this intentionally behind the `configure_logging` call? I'm curious what would happen if we want to inject configs in the logging module in the future

ssheng: It is intentional for now because `inject_dependenies` has some logging statements.

bentoml/__init__.py
ssheng: This is because we need `inject_dependencies` to happen before the creation of CLI.

PR #1567, opened 2021-04-05
Lock Module Implementation

### LINKED ISSUE #1540: Bundle/Deployment Locking Module
## Background

Why do we need to lock? As of now, there's no way to guarantee mutual exclusion of resources like

- Bento Bundles
- Deployment Operators

As certain operations (like deployment) take a long time to execute, there's no way to guarantee mutual exclusion on the resources that these operations touch. For example, if one user is trying to deploy a Bento and another is trying to delete it through the same YataiService, what happens when the deletion finishes before the deployment finishes?

As BentoML scales to larger companies, often just a single YataiService isn't enough to handle all of the deployment needs. However, in order to support multiple concurrent YataiServices, we need to be able to deploy multiple things concurrently and for that we need a locking module to make access concurrently safe.

## User-facing API Proposal

Developers just need to wrap code in a `with` block. Lock will automatically release after everything inside it is done. If the user wishes to retry the operation with some back-off strategy, they can do that too.

```python
from bentoml.locking import (Lock, LockType)
from bentoml.exceptions import FailedLockAcquisition
import backoff

class YataiServiceImpl(base):

	def ApplyDeployment(self, request, context=None):
		with Lock(request.name, request.namespace, type=LockType.READ, timeout_seconds=3): # aquire lock here, will throw BentoMLException if failed
          try:
              # apply default namespace if not set
              request.deployment.namespace = (
                  request.deployment.namespace or self.default_namespace
              )

	def DeleteDeployment(self, request, context=None):
		with Lock(request.name, request.namespace, type=LockType.WRITE): # acquire write lock here
          try:
              request.namespace = request.namespace or self.default_namespace
              deployment_pb = self.deployment_store.get(
                  request.deployment_name, request.namespace
              )
```

### Abstract Store

The same underlying operations of `bentoml/yatai/repository/metadata_store.py`, `bentoml/yatai/deployment/store.py`, and `bentoml/yatai/label_store.py` are all the same. Could pull all common operations to `bentoml/yatai/db.py` (which all 3 are subclasses of). That way, creating a new store to keep track of lock state will be easy to implement.

### Lock State Table

```python
class RESOURCE_TYPE(enum.Enum):
    deployment = 1
    bento = 2

class LOCK_TYPE(enum.Enum):
    mutex = 1
    rw_mutex = 2

class LOCK_STATUS(enum.Enum):
    unlocked = 1
    read_lock = 2
	write_lock = 3

class Lock(Base):
    __tablename__ = 'locks'
    __table_args__ = tuple(
        UniqueConstraint(
            'resource_type',
            'resource_id',
            name='_resource_type_resource_id_uc',
        )
    )
    id = Column(Integer, primary_key=True)
    resource_type = Column(Enum(RESOURCE_TYPE))
    resource_id = Column(Integer, nullable=False)
	lock_type = Column(Enum(LOCK_TYPE))
	lock_status = Column(Enum(LOCK_STATUS))
```

### User-facing API

Implement with a [context manager](https://book.pythontips.com/en/latest/context_managers.html). 

```python
class LockType:
	READ = 2
	WRITE = 3

class Lock():
	def __init__(self, lock_identifier, type=LockType.READ, timeout_seconds=None, retry_interval_seconds=None, max_retry_count=5):
		self.type = type
    # create proper database session
		# maybe use the other context manager create_session(session_maker) here?

	def __enter__(self):
		# attempt to acquire lock based off of retry strategy
        # when timeout is None and failed to acquire lock, raise an exception immediately
        # when timeout_seconds is an integer number:
        # retry acquiring the lock every `retry_interval_seconds` seconds 
        # until `timeout_seconds` seconds or `max_retry_count` has reached, raise an exception
		if (self.type == LockType.READ):
			# wait for col lock_status to be either unlocked or read_lock
		else:
			# wait for col lock_status to be unlocked
		# set lock_status to be self.type on row resource_id = name_namespace

	def __exit__(self, type, value, traceback):
		# release lock, set lock_status to 1

```

[https://book.pythontips.com/en/latest/context_managers.html](https://book.pythontips.com/en/latest/context_managers.html)

## Plan of Action

- [x]  YataiService Store Abstraction → abstract existing stores to reduce duplicate code
    - [x]  label store
    - [x]  metadata store
    - [x]  deployment store
- [x]  Create new SQL table for locks
    - [x]  Write schema
- [x]  Implement general locking module

## Revisions
* Should also have a TTL (time-to-live)
* Should check lock holder identity before releasing

jackyzha0:
## Description
Closes #1540

## Motivation and Context
See #1540 

## How Has This Been Tested?
* local unit and integration tests

## Types of changes

- [ ] Documentation
- [ ] Test, CI, or build

## Component(s) if applicable
- [x] YataiService gRPC server (model registry, cloud deployment automation)
- [ ] BentoML CLI

## Checklist:


bentoml/yatai/db/__init__.py
```diff
@@ -99,7 +102,8 @@ def create_all_or_upgrade_db(self):
         inspector = inspect(self.engine)
         tables = inspector.get_table_names()
 
-        if 'deployments' not in tables or 'bentos' not in tables:
+        table_names = ['deployments', 'bentos', 'labels', 'locks']
+        if not all([table_name in tables for table_name in table_names]):
```
parano, line 106: This should be "only create table if none of the table exist", right?
```
if all([table_name in tables for table_name not in table_names]):
```

bentoml/yatai/yatai_service_impl.py
```diff
@@ -251,7 +254,8 @@ def DeleteDeployment(self, request, context=None):
                     return DeleteDeploymentResponse(status=Status.INTERNAL(str(e)))
 
         def GetDeployment(self, request, context=None):
-            with self.db.create_session() as sess:
+            deployment_id = f"{request.deployment.name}_{request.deployment.namespace}"
```
bojiang, line 257: Mark

bentoml/yatai/db/stores/lock.py
```diff
@@ -0,0 +1,54 @@
...
+    @staticmethod
+    def acquire(sess, lock_type, resource_id):
+        lock = LockStore._find_lock(sess, resource_id)
+
+        # no lock found; free to acquire
+        if not lock:
+            lock = Lock()
+            lock.lock_status = lock_type
+            lock.resource_id = resource_id
+            sess.add(lock)
```
bojiang, line 36: https://docs.sqlalchemy.org/en/13/orm/session_api.html#sqlalchemy.orm.session.Session.add

> Its state will be persisted to the database on the next flush operation.

So the lock won't work between instances of Yatai?

bojiang, line 36: https://docs.sqlalchemy.org/en/13/orm/session_api.html#sqlalchemy.orm.session.Session.ad
> Its state will be persisted to the database on the next flush operation.

So the lock doesn't work between multiple Yatai workers?

bentoml/yatai/db/stores/lock.py
```diff
@@ -0,0 +1,54 @@
...
+            # read lock acquisition success, no change required
+            # as current state is already read_lock
+            return True
+        else:
+            # can't acquire read/write lock when write lock is already held
+            raise LockUnavailable("Failed to acquire write lock, lock held")
+
+    @staticmethod
+    def release(sess, resource_id):
+        lock = LockStore._find_lock(sess, resource_id)
```
bojiang, line 53: It doesn't verify the owner of the lock. A worker may release the lock held by other processes.

bentoml/yatai/db/stores/lock.py
```diff
@@ -0,0 +1,54 @@
...
+class LOCK_STATUS(enum.Enum):
+    read_lock = 1
+    write_lock = 2
+
+
+class Lock(Base):
+    __tablename__ = 'locks'
+    __table_args__ = tuple(UniqueConstraint('resource_id', name='_resource_id_uc',))
+    id = Column(Integer, primary_key=True)
+    resource_id = Column(Integer, nullable=False)
```
bojiang, line 17: `unique=True`?

bentoml/yatai/db/stores/lock.py
```diff
@@ -0,0 +1,54 @@
...
+    lock_status = Column(Enum(LOCK_STATUS))
+
+
+class LockStore(object):
+    @staticmethod
+    def _find_lock(sess, resource_id):
+        lock_obj = sess.query(Lock).filter_by(resource_id=resource_id).first()
```
bojiang, line 24: The value may be changed between `_find_lock` and `add`

bentoml/yatai/db/stores/lock.py
jackyzha0: adding an extra `sess.commit()` after each of these so it gets persisted to database

bentoml/yatai/db/stores/lock.py
jackyzha0: hmm what's the best way to fix this?

bentoml/yatai/db/stores/lock.py
jackyzha0: I was thinking that we don't expose this acquire/release API to users and only lock using the context manager in `yatai/locking/lock.py`. That way, we don't need to store lock owner in the table.

bentoml/yatai/locking/lock.py
```diff
@@ -0,0 +1,95 @@
...
+                lock_objs = []
+                # acquire all locks in lock list
+                for (lock_identifier, lock_type) in locks:
+                    lock_obj = LockStore.acquire(
+                        sess, lock_type, lock_identifier, ttl_min
+                    )
+                    lock_objs.append(lock_obj)
+
+                # try to commit all locks to db
+                sess.commit()
```
jackyzha0, line 69: session is created on line 59 and changes are finally commited on line 69. this *may* cause lock contention if another yataiservice tries to acquire a lock in the time that the for loop runs for

tests/integration/test_locking.py
```diff
@@ -0,0 +1,80 @@
...
+    ):
+        def function():
+            self.result = target(*args, **kwargs)
+
+        super().__init__(group=group, target=function, name=name, daemon=daemon)
+
+
+def run_delayed_thread(t1, t2):
+    t1.start()
+    time.sleep(1)
```
jackyzha0, line 25: anything less than 0.5s here causes some instability in the guarantee that t1 acquires lock before t2 here (related to concern above)

jackyzha0 (commented):
*could* be potential problems

PR #1586, opened 2021-04-20
[Enhancement]Better local FS support for remote yatai

yubozhao:
## Description

Add support for retrieve/load functions for remote yatai with 

Add support for upload/download functionality for yatai service with local FS as repository storage.
- Add `download_bento` and `upload_bento` grpc calls (via streaming)
- Upload flow: Client tarball bento bundle and streaming to server, server unzip and then place bundle in to the repository 
- Download flow: server tar bento bundle and then streaming to client, client unzip into local directory
   

## Types of changes

- [ ] Documentation
- [ ] Test, CI, or build

## Component(s) if applicable
- [ ] BentoML CLI

## Checklist:


bentoml/yatai/client/bento_repository_api.py
```diff
@@ -52,13 +57,67 @@
...
 )
+from bentoml.yatai.proto.yatai_service_pb2_grpc import YataiStub
 from bentoml.yatai.status import Status
 
 
 logger = logging.getLogger(__name__)
 yatai_proto = LazyLoader('yatai_proto', globals(), 'bentoml.yatai.proto')
 
 
+class BentoUploadStreamingRequests:
```
ssheng, line 68: Maybe `BentoUploadStreamRequests`? Why does upload behavior need to be modeled as a class?

bentoml/yatai/client/bento_repository_api.py
```diff
@@ -534,6 +629,74 @@ def load(self, bento):
...
+                    saved_bento_bundle_path, tarfile_dir, bento_version
+                )
+                result = self.yatai_service.UploadBento(
+                    iter(
+                        BentoUploadStreamingRequests(
+                            bento_name, bento_version, tarfile_path
+                        ),
+                    )
+                )
+                if result.status.status_code != status_pb2.Status.OK:
```
ssheng, line 654: Does GRPC come with a default timeout behavior?

bentoml/yatai/client/bento_repository_api.py
```diff
@@ -273,12 +346,34 @@ def _update_bento_upload_progress(
...
+        Returns:
+            None
+        """
+        if bento_pb.uri.type == BentoUri.S3 or bento_pb.uri.type == BentoUri.GCS:
+            bento_service_bundle_path = (
+                bento_pb.uri.s3_presigned_url
+                if bento_pb.uri.type == BentoUri.S3
+                else bento_pb.uri.gcs_presigned_url
+            )
+        elif bento_pb.uri.type == BentoUri.LOCAL:
```
ssheng, line 365: Can we rename `BentoUri.LOCAL` to `BentoUri.FILE_SYSTEM` in this PR?

ssheng:
Thanks for addressing the FS support for remote Yatai. While this use case is addressed, I think it has further demonstrated the confusing nature of the current Yatai concepts. I think we have an opportunity to simplify the Yatai concepts by re-organizing functionality into a few high level building blocks.

**Yatai Local**

Yatai Local is conceptually the staging environment for testing and validating bentos. Every BentoML installation comes with a Yatai Local. Yatai Local can **only** use the file system repository (not S3 or GCS). Python API talks to the Yatai Local through the Yatai Client. Bentos can be pulled from and pushed to Yatai Remote **only** through gRPC calls. Yatai Local still supports all deployments types.

**Yatai Remote**

Yatai Remote is conceptually the collaboration environment that manages bentos created by a team. Yatai Remote is a standalone gRPC server. Yatai Remote hides the repository implementation away from the caller. The caller can only upload and download bentos through the gRPC interface. Direct uploading to and downloading from the cloud repositories is not allowed. BentoML installation communicates with a Yatai Remote using the Yatai Client. Multiple Yatai Remote can be configured by a BentoML installation.

BentoML Yatai Discussion

bentoml/yatai/client/bento_repository_api.py
yubozhao: streaming grpc calls requires an iterator.  The `BentoUploadStreamRequests` class will be iterating the file and generate requests for the call.

yubozhao:
> Yatai Local

 > Yatai Local is conceptually the staging environment for testing and validating bentos. Every BentoML installation comes with a Yatai Local. Yatai Local can only use the file system repository (not S3 or GCS). Python API talks to the Yatai Local through the Yatai Client. Bentos can be pulled from and pushed to Yatai Remote only through gRPC calls. Yatai Local still supports all deployments types.

I think this is a good idea.  We need to be mindful of the breaking changes to existing users. 

> Yatai Remote

> Yatai Remote is conceptually the collaboration environment that manages bentos created by a team. Yatai Remote is a standalone gRPC server. Yatai Remote hides the repository implementation away from the caller. The caller can only upload and download bentos through the gRPC interface. Direct uploading to and downloading from the cloud repositories is not allowed. BentoML installation communicates with a Yatai Remote using the Yatai Client. Multiple Yatai Remote can be configured by a BentoML installation.

One advantage of direct upload/download via resigned URLs is taking away the middleman, another potentials point of failure.
Also, it could release some pressure from yatai service.

bentoml/yatai/client/bento_repository_api.py
yubozhao: grpc provides a timeout parameter as part of the request method. That's something we should add.  Good catch

bentoml/yatai/client/bento_repository_api.py
yubozhao: This change might impact a lot of files. Let's do it in a separate pr.

bentoml/yatai/yatai_service_impl.py
```diff
@@ -552,6 +562,99 @@ def ContainerizeBento(self, request, context=None):
...
+                        return UploadBentoResponse(status=result_status)
+                except BentoMLException as e:
+                    logger.error("RPC ERROR GetBento: %s", e)
+                    return UploadBentoResponse(status=e.status_proto)
+                except Exception as e:  # pylint: disable=broad-except
+                    logger.error("RPC ERROR GetBento: %s", e)
+                    return UploadBentoResponse(status=Status.INTERNAL())
+
+        def DownloadBento(self, request, context=None):
+            with self.db.create_session() as sess:
```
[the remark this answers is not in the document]
yubozhao: Since we are doing the entire operation(steam out responses) within the session context, we are fine.  It will use the same session.

bentoml/yatai/client/bento_repository_api.py
ssheng: Got it. Not a big deal but perhaps we can do this inline with `yield`.

```python
while sent_chunk_count <  bundle_chunk_count:
    chunk = bundle.read(CHUNK_SIZE)
    yield UploadBentoRequest(chuck)
```

bentoml/yatai/client/bento_repository_api.py
ssheng: Will we timeout here if server does not respond?

bentoml/yatai/client/bento_repository_api.py
```diff
@@ -535,6 +630,74 @@ def load(self, bento):
...
+                    if response.status.status_code != status_pb2.Status.OK:
+                        raise BentoMLException(response.status.error_message)
+                    file.write(response.bento_bundle)
+                file.seek(0)
+                temp_bundle_path = os.path.join(
+                    temp_dir, f'{bento_name}_{bento_version}'
+                )
+                with tarfile.open(fileobj=file, mode='r:gz') as tar:
+                    tar.extractall(path=temp_bundle_path)
+                file.close()
```
ssheng, line 690: Could we open with `with` statement? If anything throws, will it close file?

ssheng (dismissed):

ssheng:
> One advantage of direct upload/download via resigned URLs is taking away the middleman, another potentials point of failure.
Also, it could release some pressure from yatai service.

Maybe the additional implementation complexity is worth it to simplify the interface and concepts.

bentoml/utils/__init__.py
```diff
@@ -261,3 +265,19 @@ def resolve_bento_bundle_uri(bento_pb):
         return bento_pb.uri.gcs_presigned_url
     else:
         return bento_pb.uri.uri
+
+
+def archive_directory_to_tar(source_dir, tarfile_dir, tarfile_name):
+    file_name = f'{tarfile_name}.tar'
+    tarfile_path = os.path.join(tarfile_dir, file_name)
+    with tarfile.open(tarfile_path, mode="w:gz") as tar:
+        # Use arcname to prevent storing the full path to the bundle,
```
parano, line 274: 👍

bentoml/yatai/client/bento_repository_api.py
```diff
@@ -51,13 +57,74 @@
...
 
 logger = logging.getLogger(__name__)
 yatai_proto = LazyLoader('yatai_proto', globals(), 'bentoml.yatai.proto')
 
 
+class BentoUploadStreamRequests:
+    """
+    A class for iterating over a file to generate upload bento requests
+
+    Args:
```
parano, line 72: I think conventionally this part of the docstring goes under the `__init__` method?

bentoml/yatai/client/bento_repository_api.py
```diff
@@ -200,21 +267,38 @@ def upload_from_dir(self, saved_bento_path, labels=None):
...
 
         if response.uri.type == BentoUri.LOCAL:
-            if os.path.exists(response.uri.uri):
-                # due to copytree dst must not already exist
-                shutil.rmtree(response.uri.uri)
-            shutil.copytree(saved_bento_path, response.uri.uri)
-
-            self._update_bento_upload_progress(bento_service_metadata)
+            # For remote yatai service, call upload method.
+            if isinstance(self.yatai_service, YataiStub):
```
parano, line 271: 👍

bentoml/yatai/client/bento_repository_api.py
```diff
@@ -200,21 +267,38 @@ def upload_from_dir(self, saved_bento_path, labels=None):
...
+                        bento_service_metadata.name,
+                        bento_service_metadata.version,
+                        saved_bento_path,
+                    )
+                    upload_status = UploadStatus.DONE
+                except Exception as e:  # pylint: disable=broad-except
+                    logger.error(e)
+                    upload_status = UploadStatus.ERROR
+            else:
+                if os.path.exists(response.uri.uri):
```
parano, line 283: not related to this PR - this should probably raise an exception here instead, let's revisit this

bentoml/yatai/client/bento_repository_api.py
```diff
@@ -200,21 +267,38 @@ def upload_from_dir(self, saved_bento_path, labels=None):
...
         if response.uri.type == BentoUri.LOCAL:
-            if os.path.exists(response.uri.uri):
-                # due to copytree dst must not already exist
-                shutil.rmtree(response.uri.uri)
-            shutil.copytree(saved_bento_path, response.uri.uri)
-
-            self._update_bento_upload_progress(bento_service_metadata)
+            # For remote yatai service, call upload method.
+            if isinstance(self.yatai_service, YataiStub):
+                try:
```
parano, line 272: Let's just have `self.upload_bento` return the UploadStatus here?  It seems weird to me to catch a generic Exception here for this

bentoml/yatai/client/bento_repository_api.py
```diff
@@ -200,21 +267,38 @@ def upload_from_dir(self, saved_bento_path, labels=None):
...
+                    shutil.rmtree(response.uri.uri)
+                shutil.copytree(saved_bento_path, response.uri.uri)
+                upload_status = UploadStatus.DONE
 
-            logger.info(
-                "BentoService bundle '%s:%s' saved to: %s",
-                bento_service_metadata.name,
-                bento_service_metadata.version,
-                response.uri.uri,
+            self._update_bento_upload_progress(
```
parano, line 289: do we still need this - in the case of calling UploadBento?
UploadBento gRPC should be able to set the upload progress itself once the upload is finished right?

bentoml/yatai/client/bento_repository_api.py
```diff
@@ -226,14 +310,12 @@ def upload_from_dir(self, saved_bento_path, labels=None):
...
 
-            if response.uri.type == BentoUri.S3:
-                http_response = requests.put(
-                    response.uri.s3_presigned_url, data=fileobj
-                )
-            elif response.uri.type == BentoUri.GCS:
-                http_response = requests.put(
-                    response.uri.gcs_presigned_url, data=fileobj
-                )
+            upload_path = (
```
parano, line 313: why this change?  it seems better previously - except that it should also handle the "else" case and report error

bentoml/yatai/client/bento_repository_api.py
yubozhao: I am planning on revisit how bento is added to yatai in a different PR.  One of the change in the upcoming PR will be on progression

bentoml/yatai/client/bento_repository_api.py
```diff
@@ -274,12 +356,34 @@ def _update_bento_upload_progress(
...
+
+        Args:
+            bento_pb: bento bundle protobuf dict
+            target_dir: path
+
+        Returns:
+            None
+        """
+        if bento_pb.uri.type == BentoUri.S3 or bento_pb.uri.type == BentoUri.GCS:
+            bento_service_bundle_path = (
```
parano, line 370: same as above, prefer specific multi-line if-else - which makes it a lot easier when we need to add support or a new blob storage service.

bentoml/yatai/client/bento_repository_api.py
yubozhao: It feels redundant.  I can revert the change

bentoml/yatai/yatai_service_impl.py
```diff
@@ -578,6 +588,101 @@ def ContainerizeBento(self, request, context=None):
...
+                                        f"{request.bento_version}"
+                                    )
+                        bento_id = f"{bento_name}_{bento_version}"
+                        with lock(self.db, [(bento_id, LockType.READ)]) as (sess, _):
+                            bento_pb = self.db.metadata_store.get(
+                                sess, bento_name, bento_version
+                            )
+                            file.seek(0)
+                            if bento_pb:
+                                # If there is a directory exist, it will be removed
```
parano, line 623: let's raise an exception here instead :D

bentoml/yatai/yatai_service_impl.py
```diff
@@ -578,6 +588,101 @@ def ContainerizeBento(self, request, context=None):
...
+                return UploadBentoResponse(status=Status.INTERNAL())
+
+        def DownloadBento(self, request, context=None):
+            bento_id = f"{request.bento_name}_{request.bento_version}"
+            with lock(self.db, [(bento_id, LockType.READ)]) as (sess, _):
+                try:
+                    bento_pb = self.db.metadata_store.get(
+                        sess, request.bento_name, request.bento_version
+                    )
+                    with TempDirectory() as temp_dir:
```
parano, line 651: can we refactor this part out to something that can be tested more easily?

bentoml/utils/__init__.py
```diff
@@ -261,3 +265,19 @@ def resolve_bento_bundle_uri(bento_pb):
...
+        # from source_dir/path/to/file to path/to/file
+        tar.add(source_dir, arcname='')
```
parano, line 276: can we use `arcname="bento_name"` so it's consistent with how we currently create zip and upload bundle to s3? https://github.com/bentoml/BentoML/blob/master/bentoml/saved_bundle/bundler.py#L60

bentoml/yatai/client/bento_repository_api.py
```diff
@@ -51,13 +57,74 @@
...
+            )
+            self.sent_chunk_count += 1
+            return request
+        else:
+            self.bundle.close()
+            raise StopIteration
+
+
+def process_grpc_error(grpc_error):
+    if grpc_error.code() == grpc.StatusCode.DEADLINE_EXCEEDED:
```
parano, line 122: why special case this error type?  what's the output of `grpc_error.details()` when `grpc_error.code() == grpc.StatusCode.DEADLINE_EXCEEDED`?

bentoml/yatai/client/bento_repository_api.py
```diff
@@ -51,13 +57,74 @@
...
+        file_path: path
+        chunk_size optional: int
+    """
+
+    def __init__(
+        self, bento_name, bento_version, file_path,
+    ):
+        self.bento_name = bento_name
+        self.bento_version = bento_version
+        self.bundle = open(file_path, 'rb')
```
parano, line 84: rename to `self.bundle_file_handler`

bentoml/yatai/client/bento_repository_api.py
```diff
@@ -535,6 +639,88 @@ def load(self, bento):
...
+                    ),
+                    timeout=6,  # 6 seconds timeout
+                )
+                if result.status.status_code != status_pb2.Status.OK:
+                    raise BentoMLException(result.status.error_message)
+            except BentoMLException as e:
+                logger.error(f'RPC ERROR upload bento: {e}')
+                raise BentoMLException(
+                    f'Failed to upload {bento_name}:{bento_version} to remote yatai '
+                    f'server {e}'
```
parano, line 671: add ":" here? - `f"server: {e}"`

bentoml/yatai/client/bento_repository_api.py
```diff
@@ -535,6 +639,88 @@ def load(self, bento):
...
+                tarfile_path, _ = archive_directory_to_tar(
+                    saved_bento_bundle_path, tarfile_dir, bento_version
+                )
+                result = self.yatai_service.UploadBento(
+                    iter(
+                        BentoUploadStreamRequests(
+                            bento_name, bento_version, tarfile_path
+                        ),
+                    ),
+                    timeout=6,  # 6 seconds timeout
```
parano, line 663: move out to a constant

bentoml/yatai/client/bento_repository_api.py
```diff
@@ -535,6 +639,88 @@ def load(self, bento):
...
+                    f'server {error_message}'
+                )
+            except Exception as e:  # pylint: disable=broad-except
+                logger.error(f'ERROR upload bento: {e}')
+                raise BentoMLException(
+                    f'Failed to upload {bento_name}:{bento_version} to remote yatai '
+                    f'server {e}'
+                )
+
+    def download_bento(self, bento_name, bento_version):
```
parano, line 686: as discussed, renaming to `_download_bento` and `_upload_bento`

bentoml/yatai/yatai_service_impl.py
```diff
@@ -578,6 +588,101 @@ def ContainerizeBento(self, request, context=None):
...
+                                else:
+                                    raise BentoMLException(
+                                        f"Incoming stream request doesn't match "
+                                        f"with initial request info "
+                                        f"{bento_name}:{bento_version} - "
+                                        f"{request.bento_name}:"
+                                        f"{request.bento_version}"
+                                    )
+                        bento_id = f"{bento_name}_{bento_version}"
+                        with lock(self.db, [(bento_id, LockType.READ)]) as (sess, _):
```
parano, line 617: use WRITE lock here?

bentoml/yatai/grpc_stream_utils.py
```diff
@@ -0,0 +1,134 @@
...
+            # Add the directory path to tar
+            for dir_path in dirs:
+                tar.add(
+                    name=os.path.join(path, dir_path),
+                    arcname=os.path.join(arc_path, dir_path),
+                    recursive=False,
+                )
+            # Stream out directories info in the tar
+            yield self.create_request_or_response(out_stream.read_value())
+        tar.close()
```
ssheng, line 132: We probably need clean up logic here. May be we will end up with memory leak In an unhappy path when bytes are not all read.

bentoml/yatai/client/bento_repository_api.py
```diff
@@ -524,6 +563,82 @@ def load(self, bento):
...
+            result = self.yatai_service.UploadBento(
+                iter(
+                    BentoBundleStreamRequestsOrResponses(
+                        bento_name=bento_name,
+                        bento_version=bento_version,
+                        directory_path=saved_bento_bundle_path,
+                        is_request=True,
+                    ),
+                ),
+                timeout=DEFAULT_REQUEST_TIMEOUT,
```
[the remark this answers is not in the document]
yubozhao: 1. Yep. The timeout is applied to individual requests.
2. Added clean up resources function.  However it shouldn't part of `BentoBundlesStreamRequestsOrResponses` to handling error, but it should provide ways to clean it up. (it's added in the latest commit)

bentoml/yatai/yatai_service_impl.py
```diff
@@ -578,6 +584,79 @@ def ContainerizeBento(self, request, context=None):
...
+                        bento_version=request.bento_version,
+                        directory_path=bento_pb.uri.uri,
+                        is_request=False,
+                    )
+                    for response in responses_generator:
+                        yield response
+                except BentoMLException as e:
+                    responses_generator.clear()
+                    logger.error("RPC ERROR DownloadBento: %s", e)
+                    return DownloadBentoResponse(status=e.status_proto)
```
ssheng, line 654: We may be able to use a simpler pattern by throwing the exception back into the generator with `response_generator.throw(e)`. Then inside the generator, we can handle the exception and close the files in a `finally` block. Let's quickly chat over Zoom if my explanation isn't clear.

ssheng (commented):
Thanks for adding timeout and clean up logic. I had a slightly different pattern in mind of clean up. Let's chat over Zoom and wrap this PR.

bentoml/yatai/yatai_service_impl.py
yubozhao: I understand your explanation.  Let's still have a quick zoom call with where the responsibility ends for the generator.

ssheng: Discussed offline, will invoke `close()` in `finally`.

bentoml/yatai/grpc_stream_utils.py
```diff
@@ -0,0 +1,138 @@
...
+            self.buffer = io.BytesIO()
+
+
+class BentoBundleStreamRequestsOrResponses:
+    def __init__(
+        self,
+        bento_name,
+        bento_version,
+        directory_path,
+        file_chunk_size=DEFAULT_FILE_CHUNK_SIZE,
```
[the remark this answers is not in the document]
ssheng: Discussed offline. Will address this when we restructure Yatai architecture.

bentoml/yatai/client/bento_repository_api.py
parano: @yubozhao could you make this change in this PR and add a test?

bentoml/yatai/client/bento_repository_api.py
parano: can we just do that in this PR? it's pretty simple - 1) remove this line here, 2) update the upload status on server side where it handles upload

parano: Let's do it in this PR because the logic is wrong here

For local yatai, let's copy the files and call update upload progress to set status to "DONE"

For remote yatai, server-side should update the status upon upload finish

bentoml/yatai/client/bento_repository_api.py
```diff
@@ -524,6 +563,82 @@ def load(self, bento):
             saved_bundle_path = bento
         else:
             bento_pb = self.get(bento)
-            saved_bundle_path = resolve_bento_bundle_uri(bento_pb)
+            if bento_pb.uri.type == BentoUri.LOCAL and isinstance(
```
parano, line 566: let's make  a helper method for it?

```
@property
def is_remote_yatai(self)
      return isinstance(self.yatai_service, YataiStub)
```

bentoml/yatai/yatai_service_impl.py
```diff
@@ -578,6 +584,79 @@ def ContainerizeBento(self, request, context=None):
...
+                                else:
+                                    raise BentoMLException(
+                                        f"Incoming stream request doesn't match "
+                                        f"with initial request info "
+                                        f"{bento_name}:{bento_version} - "
+                                        f"{request.bento_name}:"
+                                        f"{request.bento_version}"
+                                    )
+                        bento_id = f"{bento_name}_{bento_version}"
+                        with lock(self.db, [(bento_id, LockType.WRITE)]) as (sess, _):
```
parano, line 613: can we place the lock at the beginning of the upload method?

tests/yatai/client/test_bento_repository_api.py
```diff
@@ -1,150 +1,109 @@
...
+    pytest.skip(
+        "Requires docker, skipping test for Mac OS on Github Actions",
+        allow_module_level=True,
+    )
 
 
-@pytest.fixture(scope='module')
-def yatai_server_container():
+@pytest.fixture(scope='session')
+def local_yatai_service_url():
```
parano, line 29: just `yatai_service_url `?  "local" is a bit misleading here since it's not what we call "local yatai" in the code

parano: can we just reuse the same `local_yatai_service_container` function from `tests/yatai/local_yatai_service.py`?

parano:
Could you also update the tests in `test_local_fs.py` to cover the changes in this PR? 

https://github.com/bentoml/BentoML/blob/master/tests/integration/yatai_server/test_local_fs.py

bentoml/yatai/grpc_stream_utils.py
```diff
@@ -0,0 +1,138 @@
...
+    def __iter__(self):
+        if self.is_request:
+            yield UploadBentoRequest(
+                bento_name=self.bento_name, bento_version=self.bento_version
+            )
+        # Include length of separator.
+        prefix_len = len(self.directory_path) + len(os.path.sep)
+
+        # Add the directory path to tar
+        self.tar.add(name=self.directory_path, arcname='', recursive=False)
```
parano, line 103: can we use `arcname =bento_service.name` here?

bentoml/yatai/yatai_service_impl.py
```diff
@@ -578,6 +584,79 @@ def ContainerizeBento(self, request, context=None):
...
+                            bento_pb = self.db.metadata_store.get(
```
parano, line 614: we should move this up as well

bentoml/yatai/yatai_service_impl.py
```diff
@@ -578,6 +584,79 @@ def ContainerizeBento(self, request, context=None):
...
+        def UploadBento(self, request_iterator, context=None):
+            try:
+                with TempDirectory() as temp_dir:
+                    temp_tar_path = os.path.join(
+                        temp_dir, f'{uuid.uuid4().hex[:12]}.tar'
+                    )
+                    with open(temp_tar_path, 'wb+') as file:
+                        for request in request_iterator:
+                            if not request.bento_bundle:
+                                bento_name = request.bento_name
```
parano, line 596: Is the bento_name, bento_version request always the first item in the iterator?

I think a few things we can do after it received the bento name and version:

* check if the bento with name, version is already in the DB, return error if not (you have it down in this function, let's move it up)
* check the upload status(this is missing in the PR right now), if it's already in progress or done, return an error
* acquire the write lock, make sure it's not already being uploaded in another client
* extract the bento files to the right URL
* Set the status to "uploading" with the metadata store API
* Return the response

bentoml/yatai/yatai_service_impl.py
```diff
@@ -578,6 +584,79 @@ def ContainerizeBento(self, request, context=None):
...
+                            bento_pb = self.db.metadata_store.get(
+                                sess, bento_name, bento_version
+                            )
+                            file.seek(0)
+                            if bento_pb:
+                                with tarfile.open(fileobj=file, mode='r') as tar:
+                                    tar.extractall(path=bento_pb.uri.uri)
+                                result_status = Status.OK()
+                            else:
+                                result_status = Status.NOT_FOUND(
```
parano, line 623: just raise `bentoml.execeptions.NotFound("BentoService ....")` here?

bentoml/yatai/yatai_service_impl.py
```diff
@@ -578,6 +584,79 @@ def ContainerizeBento(self, request, context=None):
...
+                                bento_name = request.bento_name
+                                bento_version = request.bento_version
+                            else:
+                                if (
+                                    bento_name == request.bento_name
+                                    and bento_version == request.bento_version
+                                ):
+                                    file.write(request.bento_bundle)
+                                else:
+                                    raise BentoMLException(
```
parano, line 605: let's use `bentoml.exceptions.BadInput` here

bentoml/yatai/client/bento_repository_api.py
```diff
@@ -41,21 +43,26 @@
...
 )
+from bentoml.yatai.proto.yatai_service_pb2_grpc import YataiStub
 from bentoml.yatai.status import Status
-
+from bentoml.yatai.utils import process_grpc_error
 
 logger = logging.getLogger(__name__)
 yatai_proto = LazyLoader('yatai_proto', globals(), 'bentoml.yatai.proto')
 
+DEFAULT_REQUEST_TIMEOUT = 6
```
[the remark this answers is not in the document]
parano: Can you add some inline comments on this TIMEOUT value? at least explain that it's not the time out for entire upload request, just for each chunk

bentoml/yatai/client/bento_repository_api.py
```diff
@@ -524,6 +563,82 @@ def load(self, bento):
...
+            except grpc.RpcError as e:
+                error_message = process_grpc_error(e)
+                raise BentoMLException(
+                    f'Failed to upload {bento_name}:{bento_version} to remote yatai '
+                    f'server: {error_message}'
+                )
+            except Exception as e:  # pylint: disable=broad-except
+                logger.error(f'ERROR download bento: {e}')
+                raise BentoMLException(
+                    f'Failed to download {bento_name}:{bento_version} to remote yatai '
```
parano, line 642: s/to/from/

parano: how about `from yatai server at {self.yatai_service.url}`?

bentoml/yatai/grpc_stream_utils.py
```diff
@@ -0,0 +1,138 @@
...
+            self.buffer.close()
+            self.buffer = io.BytesIO()
+
+
+class BentoBundleStreamRequestsOrResponses:
+    def __init__(
+        self,
+        bento_name,
+        bento_version,
+        directory_path,
```
parano, line 38: rename to bento_bundle_path for consistency?

bentoml/yatai/grpc_stream_utils.py
```diff
@@ -0,0 +1,138 @@
...
+                    tar_info = self.tar.gettarinfo(
+                        name=file_path, arcname=os.path.join(arc_path, f), fileobj=file,
+                    )
+                    self.tar.addfile(tar_info)
+                    # stream the file content into the tar
+                    for _ in self._stream_file_into_tar(
+                        tar_info, self.tar, file, self.file_chunk_size
+                    ):
+                        # Stream out as tarfile
+                        yield self.create_request_or_response()
```
parano, line 124: the code here is a bit hard to follow, can we make `create_request_or_response` a static helper method:

```
def create_request_or_response(bento_bundle_chunk_value):
  ...
```

And here we can have `yield create_request_or_response(self.tar.fileobj.read_value(), self.is_request)` ?

bentoml/yatai/grpc_stream_utils.py
```diff
@@ -0,0 +1,138 @@
...
+            bento_version:
+            directory_path:
+            file_chunk_size:
+            is_request:
+        """
+        self.bento_name = bento_name
+        self.bento_version = bento_version
+        self.directory_path = directory_path
+        self.file_chunk_size = file_chunk_size
+        self.is_request = is_request
```
parano, line 57: might be better to just separate them to two different classes and have them use some shared utility functions?

bentoml/yatai/grpc_stream_utils.py
```diff
@@ -0,0 +1,138 @@
...
+                bento_bundle=self.out_stream.read_value(),
+            )
+        else:
+            return DownloadBentoResponse(bento_bundle=self.out_stream.read_value())
+
+    def close(self):
+        self.tar.close()
+        self.out_stream.close()
+
+    def __iter__(self):
```
parano, line 94: could we add some standalone tests to verify this behavior?  e.g. have a test case that generates random files under nested directories and verifies the received file has identical structure and content?

bentoml/yatai/grpc_stream_utils.py
```diff
@@ -0,0 +1,138 @@
...
+        return self.offset
+
+    def close(self):
+        self.buffer.close()
+
+    def read_value(self):
+        try:
+            return self.buffer.getvalue()
+        finally:
+            self.buffer.close()
```
parano, line 29: is there a `flush` method for BytesIO?

bentoml/yatai/grpc_stream_utils.py
```diff
@@ -0,0 +1,138 @@
...
+                self.tar.add(
+                    name=os.path.join(path, dir_path),
+                    arcname=os.path.join(arc_path, dir_path),
+                    recursive=False,
+                )
+            # Stream out directories info in the tar
+            yield self.create_request_or_response()
+        self.tar.close()
+        # Stream out any value that are left over.
+        yield self.create_request_or_response()
```
parano, line 137: this last request could be size=0 right? could you explain why there might be any value left over?

bentoml/yatai/grpc_stream_utils.py
```diff
@@ -0,0 +1,138 @@
...
+            yield UploadBentoRequest(
+                bento_name=self.bento_name, bento_version=self.bento_version
+            )
+        # Include length of separator.
+        prefix_len = len(self.directory_path) + len(os.path.sep)
+
+        # Add the directory path to tar
+        self.tar.add(name=self.directory_path, arcname='', recursive=False)
+
+        # Manually walk the directory and add to tarfile
```
parano, line 105: nit: not really "Manually" :P , just `# walk the directory and ...`

bentoml/yatai/grpc_stream_utils.py
```diff
@@ -0,0 +1,138 @@
...
+            # Add files to tar. Add the tar info and then stream the file into tar
+            for f in files:
+                file_path = os.path.join(path, f)
+                with open(file_path, 'rb') as file:
+                    # Reading the file info and generate tarinfo from that.
+                    # No file content is added to the tar at this point
+                    tar_info = self.tar.gettarinfo(
+                        name=file_path, arcname=os.path.join(arc_path, f), fileobj=file,
+                    )
+                    self.tar.addfile(tar_info)
```
parano, line 118: Looking at related docs, `addfile` will 1) add tar info 2) actually write the file to the `tar.fileobj`. whereas `add` will only add the tar info.  In this case we want to stream file content to tar.fileobj in the next step, so I thought it should be `self.tar.add(tar_info)` instead of `self.tar.addfile` here. Could you help me understand this?

bentoml/yatai/grpc_stream_utils.py
```diff
@@ -0,0 +1,138 @@
...
+            blocks += 1
+        tar.offset += blocks * tarfile.BLOCKSIZE
+        yield
+
+    def create_request_or_response(self):
+        if self.is_request:
+            return UploadBentoRequest(
+                bento_name=self.bento_name,
+                bento_version=self.bento_version,
+                bento_bundle=self.out_stream.read_value(),
```
parano, line 85: does our gRPC server enforce a size limit for each request here?

bentoml/yatai/yatai_service_impl.py
```diff
@@ -578,6 +584,79 @@ def ContainerizeBento(self, request, context=None):
...
+                                )
+                    return UploadBentoResponse(status=result_status)
+            except BentoMLException as e:
+                logger.error("RPC ERROR UploadBento: %s", e)
+                return UploadBentoResponse(status=e.status_proto)
+            except Exception as e:  # pylint: disable=broad-except
+                logger.error("RPC ERROR UploadBento: %s", e)
+                return UploadBentoResponse(status=Status.INTERNAL())
+
+        def DownloadBento(self, request, context=None):
```
parano, line 636: Should we add a check here to verify that `self.repository` is a FileSystemRepository? Looks like this RPC method only supports FileSystemRepository right now.

parano: Same for the UploadBento rpc above?

bentoml/yatai/grpc_stream_utils.py
yubozhao: `tar.add` is typically used for directory, fifo, symbolic link and etc...It generated created a tarinfo and then call `tar.addfile` internally.

We wants to control that process, that's why we are using `tar.addfile`

bentoml/yatai/yatai_service_impl.py
yubozhao: We checked bento bundle's URI type to be file system before making calling the internal methods for uploading and downloading.

bentoml/yatai/grpc_stream_utils.py
yubozhao: In tarfile write mode, two finishing zero blocks are appended to the archive.
https://docs.python.org/3/library/tarfile.html?highlight=tarfile#tarfile.TarFile.close

bentoml/yatai/yatai_service_impl.py
parano: That's on the client side, would be better yo verify it on server side as well

bentoml/yatai/grpc_stream_utils.py
yubozhao: `bytesio.close` does the flushing and once closed, any operations will raise exception. It is recommend.

bentoml/yatai/grpc_stream_utils.py
yubozhao: Adding the comments as inline notes for code

bentoml/yatai/yatai_service_impl.py
```diff
@@ -578,6 +592,114 @@ def ContainerizeBento(self, request, context=None):
                     logger.error(f"RPC ERROR ContainerizeBento: {e}")
                     return ContainerizeBentoResponse(status=Status.INTERNAL(e))
 
-            # pylint: enable=unused-argument
+        def UploadBento(self, request_iterator, context=None):
+            if not is_file_system_repo(self.repo):
+                logger.error(
+                    "Yatai only supports file system repository for uploading "
```
parano, line 598: nit: "UploadBento RPC only works with File System based repository, for other types of repositories(s3, gcs, minio), use pre-signed URL for download."

bentoml/yatai/yatai_service_impl.py
```diff
@@ -578,6 +592,114 @@ def ContainerizeBento(self, request, context=None):
...
+                    bento_pb = None
+                    with TempDirectory() as temp_dir:
+                        temp_tar_path = os.path.join(
+                            temp_dir, f'{uuid.uuid4().hex[:12]}.tar'
+                        )
+                        file = open(temp_tar_path, 'wb+')
+                        for request in request_iterator:
+                            # Initial request is without bundle
+                            if not request.bento_bundle:
+                                print('in init request')
```
parano, line 614: remove

bentoml/yatai/yatai_service_impl.py
```diff
@@ -578,6 +592,114 @@ def ContainerizeBento(self, request, context=None):
...
+                                bento_name = request.bento_name
+                                bento_version = request.bento_version
+                                if lock_obj is None:
+                                    print('we are getting lock obj')
```
parano, line 618: remove..

bentoml/yatai/yatai_service_impl.py
```diff
@@ -578,6 +592,114 @@ def ContainerizeBento(self, request, context=None):
...
+                                    lock_obj = LockStore.acquire(
+                                        sess=sess,
+                                        lock_type=LockType.WRITE,
+                                        resource_id=f'{bento_name}_{bento_version}',
+                                        ttl_min=DEFAULT_TTL_MIN,
+                                    )
+                                bento_pb = self.db.metadata_store.get(
```
parano, line 625: let's acquire the lock after checking the bento exist?

parano: also check the upload status - see my comments from last review

bentoml/yatai/yatai_service_impl.py
```diff
@@ -578,6 +592,114 @@ def ContainerizeBento(self, request, context=None):
...
+                                    file.close()
+                                    return UploadBentoResponse(status=result_status)
+                            else:
+                                if (
+                                    bento_name == request.bento_name
+                                    and bento_version == request.bento_version
+                                ):
+                                    file.write(request.bento_bundle)
+                                else:
+                                    lock_obj.release(sess)
```
parano, line 644: move this to `finally`?

bentoml/yatai/yatai_service_impl.py
```diff
@@ -578,6 +592,114 @@ def ContainerizeBento(self, request, context=None):
...
+                                    )
+                        file.seek(0)
+                        with tarfile.open(fileobj=file, mode='r') as tar:
+                            tar.extractall(path=bento_pb.uri.uri)
+                        upload_status = UploadStatus(status=UploadStatus.DONE)
+                        upload_status.updated_at.GetCurrentTime()
+                        self.db.metadata_store.update_upload_status(
+                            sess, bento_name, bento_version, upload_status
+                        )
+                        lock_obj.release(sess)
```
parano, line 660: remove this and the line below - you already have it in `finally`

bentoml/yatai/yatai_service_impl.py
```diff
@@ -578,6 +592,114 @@ def ContainerizeBento(self, request, context=None):
...
+                        return UploadBentoResponse(status=Status.OK())
+            except BentoMLException as e:
+                logger.error("RPC ERROR UploadBento: %s", e)
+                return UploadBentoResponse(status=e.status_proto)
+            except Exception as e:  # pylint: disable=broad-except
+                logger.error("RPC ERROR UploadBento: %s", e)
+                return UploadBentoResponse(status=Status.INTERNAL())
+            finally:
+                file.close()
+                lock_obj.release(sess)
```
parano, line 671: add a `is not None` check for both `file` and `lock_obj`

bentoml/yatai/yatai_service_impl.py
```diff
@@ -578,6 +592,114 @@ def ContainerizeBento(self, request, context=None):
...
+
+        def DownloadBento(self, request, context=None):
+            if not is_file_system_repo(self.repo):
+                logger.error(
+                    "Yatai only supports file system repository for download at the "
```
parano, line 676: same as above

tests/yatai/client/test_bento_repository_api.py
```diff
@@ -1,150 +1,65 @@
...
+        "Requires docker, skipping test for Mac OS on Github Actions",
+        allow_module_level=True,
     )
 
-    yield f'localhost:{port}'
 
-    logger.info(f"Shutting down docker container: {container_name}")
-    os.kill(docker_proc.pid, signal.SIGINT)
+@pytest.fixture(scope='session')
+def local_yatai_container_url():
```
parano, line 26: can we rename these fixtures and functions as suggested in last review?

local_yatai_service_container => yatai_service_container
local_yatai_container_url => yatai_service_url

tests/yatai/client/test_bento_repository_api.py
```diff
@@ -227,65 +130,49 @@ def test_pull_with_labels(yatai_server_container, example_bento_service_class):
...
     assert len(bentos) == 7
 
 
-@pytest.mark.skipif(
-    sys.platform == "darwin",
-    reason="Requires docker, skipping test for Mac OS for Github Action",
-)
-@pytest.mark.skipif('not psutil.POSIX')
-def test_load(yatai_server_container, example_bento_service_class):
+def test_load(local_yatai_container_url, example_bento_service_class):
```
parano, line 162: looks like you didn't change anything in this test, why does it work before prior to this PR?

tests/yatai/client/test_bento_repository_api.py
```diff
@@ -193,28 +98,26 @@ def test_push_with_labels(yatai_server_container, example_bento_service_class):
     pushed_path = yc.repository.push(f'{svc.name}:{svc.version}')
     assert pushed_path != saved_path
     remote_bento_pb = yc.repository.get(f'{svc.name}:{svc.version}')
+    print("GETTTTTING REMOTE")
```
parano, line 101: remove..

bentoml/yatai/client/bento_repository_api.py
```diff
@@ -195,12 +205,29 @@ def upload_from_dir(self, saved_bento_path, labels=None):
             )
 
         if response.uri.type == BentoUri.LOCAL:
-            if os.path.exists(response.uri.uri):
-                # due to copytree dst must not already exist
-                shutil.rmtree(response.uri.uri)
-            shutil.copytree(saved_bento_path, response.uri.uri)
+            # For remote yatai service, call upload method.
```
parano, line 208: nit:

# When using Yatai backed by File System repository, if Yatai is a local instance, copy the files directly.
# Otherwise, use UploadBento RPC to stream files to remote Yatai server

bentoml/yatai/client/bento_repository_api.py
```diff
@@ -195,12 +205,29 @@ def upload_from_dir(self, saved_bento_path, labels=None):
...
+                    bento_service_metadata.name,
+                    bento_service_metadata.version,
+                    saved_bento_path,
+                )
+                update_bento_service = UpdateBentoRequest(
+                    bento_name=bento_service_metadata.name,
+                    bento_version=bento_service_metadata.version,
+                    service_metadata=bento_service_metadata,
+                )
+                self.yatai_service.UpdateBento(update_bento_service)
```
parano, line 220: why do we need it here?

bentoml/yatai/yatai_service_impl.py
parano: also set the status to " UploadStatus.UPLOADING" here

bentoml/yatai/client/bento_repository_api.py
yubozhao: We combined update upload status and bento metadata in the `_update_bento_upload_progress` call.  For remote yatai with FS repo, we will need to make sure the metadata is updated.

bentoml/yatai/client/bento_repository_api.py
parano: could you address this as well?

bentoml/yatai/client/bento_repository_api.py
```diff
@@ -258,6 +285,7 @@ def upload_from_dir(self, saved_bento_path, labels=None):
     def _update_bento_upload_progress(
         self, bento_service_metadata, status=UploadStatus.DONE, percentage=None
     ):
+        # TODO separate update upload progress and bento metadata.
```
parano, line 288: I think it is ok leaving them in the same gRPC endpoint, we can also use it for adding and removing labels going forward

parano: remove the TODO?

tests/yatai/client/test_bento_repository_api.py
yubozhao: It was 'working' previously on accident. Previously, it will attempt to save 'remotely' and the bundle ends up in the local FS path, `/tmp`. The load method was only looks at the uri path on the bento_pb, and is able to find it in `/tmp`

bentoml/yatai/grpc_stream_utils.py
```diff
@@ -0,0 +1,190 @@
...
+    @staticmethod
+    def _create_request_or_response(bento_name, bento_version, value, is_request):
+        if is_request:
+            return UploadBentoRequest(
+                bento_name=bento_name, bento_version=bento_version, bento_bundle=value,
+            )
+        else:
+            return DownloadBentoResponse(bento_bundle=value)
+
+    def close(self):
```
ssheng, line 90: Generator has a built-in `close()` function. What I meant was for `__iter()__` function to close the resources with either a `with` block or a `finally` block.

ssheng: https://realpython.com/introduction-to-python-generators/#how-to-use-close

bentoml/yatai/db/stores/metadata.py
```diff
@@ -105,13 +105,20 @@ def _bento_orm_obj_to_pb(bento_obj, labels=None):
     bento_uri = BentoUri(
         uri=bento_obj.uri, type=BentoUri.StorageType.Value(bento_obj.uri_type)
     )
+    if not bento_obj.upload_status:
+        upload_status = DEFAULT_UPLOAD_STATUS
+    else:
+        upload_status = UploadStatus(
+            status=UploadStatus.Status.Value(bento_obj.upload_status['status'])
```
parano, line 112: could you also include the other fields here? there are also `updated_at`, `percentage`, and `error_message`

bentoml/yatai/grpc_stream_utils.py
parano: It might be hard to do that because the StopIteration was raised in `iter(lambda: file_handler.read(buf_size), b'')`, not from its own `close`

bentoml/yatai/client/bento_repository_api.py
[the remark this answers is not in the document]
parano: Let's rename it to `DEFAULT_STREAMING_PER_CHUNK_TIMEOUT `

bentoml/yatai/yatai_service_impl.py
```diff
@@ -578,6 +593,126 @@ def ContainerizeBento(self, request, context=None):
...
+                                    )
+                        file.seek(0)
+                        with tarfile.open(fileobj=file, mode='r') as tar:
+                            tar.extractall(path=bento_pb.uri.uri)
+                        upload_status = UploadStatus(status=UploadStatus.DONE)
+                        upload_status.updated_at.GetCurrentTime()
+                        self.db.metadata_store.update_upload_status(
+                            sess, bento_name, bento_version, upload_status
+                        )
+                        lock_obj.release(sess)
```
parano, line 674: why not put `lock_obj.release(sess)` in the finally block?

parano:
Merged the PR for QA the next release, @yubozhao could you revisit the last few comments in a follow up PR?

ShaohongBai:
> Thanks for addressing the FS support for remote Yatai. While this use case is addressed, I think it has further demonstrated the confusing nature of the current Yatai concepts. I think we have an opportunity to simplify the Yatai concepts by re-organizing functionality into a few high level building blocks.
> 
> **Yatai Local**
> 
> Yatai Local is conceptually the staging environment for testing and validating bentos. Every BentoML installation comes with a Yatai Local. Yatai Local can **only** use the file system repository (not S3 or GCS). Python API talks to the Yatai Local through the Yatai Client. Bentos can be pulled from and pushed to Yatai Remote **only** through gRPC calls. Yatai Local still supports all deployments types.
> 
> **Yatai Remote**
> 
> Yatai Remote is conceptually the collaboration environment that manages bentos created by a team. Yatai Remote is a standalone gRPC server. Yatai Remote hides the repository implementation away from the caller. The caller can only upload and download bentos through the gRPC interface. Direct uploading to and downloading from the cloud repositories is not allowed. BentoML installation communicates with a Yatai Remote using the Yatai Client. Multiple Yatai Remote can be configured by a BentoML installation.
> 
> BentoML Yatai Discussion

First, thank you all for this amazing product. As discussed with @sencenan, I would like give some feedback and users cases when it comes to next Yatai Design. Indeed, the current implementation of local/remote yatai is slightly confusing. Personally, I would expect something like this https://www.mlflow.org/docs/latest/tracking.html, in which it is logically aligned with what have been discussed on local yatai, i.e. only supports local file system in localhost situation. However, I do not think yatai local should be able to handle any cloud deployment requests, those should be passed back into remote yatai for better configuration and security. Particularly, we understand current implementation flow is for a deployment, one needs to pull from remote yatai and triggering deployment from local - this requires we need pre-config aws cli and aws sam cli for any machine that likes to do deployment, that could be individual team member or production pipeline. This is not ideal right?  Ideally, we would like to only config remote yatai to be able to do the jobs and that will be one-off job for a team setting-up. Regarding the yatai remote, I am not sure why we still need file system repository, in most cases I found it is empty when I have S3. Might need to clarify the expected behavior. Additionally, I would also suggest we have in remote yatai some functionality that one can refuse to approve a deployment - this will give you guys some room in the future one asks about larget team alert and etc. Let me know these make sense or not.

sencenan:
Hi @ShaohongBai, you probably meant to tag @ssheng ^ ?

yubozhao:
@ShaohongBai Thank you for the thoughtful write up!

> I do not think yatai local should be able to handle any cloud deployment requests, those should be passed back into remote yatai for better configuration and security. 

Yes, I totally agree. We are moving toward to that direction where remote yatai will be managing and operating deployments. 

Local yatai will not handle any deployment. We will also update the name to better reflect its responsibilities. Currently we are thinking local registry... any suggestions?

>  I would also suggest we have in remote yatai some functionality that one can refuse to approve a deployment (UI button) - this will give you guys some room in the future one asks about larget team alert and etc

Yep, it makes sense. We are going to put more focus on Yatai and the deployment flow in the coming months. Would love to get your feedback on things.

ShaohongBai:
> Hi @ShaohongBai, you probably meant to tag @ssheng ^ ?

hah, sorry, wrong tag.

PR #1636, opened 2021-05-23
Image Output Adapter

canturan10:
## Description
**ImageOutput** has been added to make the output format as an image (bytes).
In this way, efficiency in web integrations was achieved and at the same time, preview was provided from the swagger page.

```python
    @bentoml.api(input=ImageInput(), output=ImageOutput(extension_format='.png'))
    def test(self, imgs):
        return imgs
```
image

## Motivation and Context
My service was supposed to edit the image input and reply back as image again. When I did this without converting to byte, I had both integration and slowness problem. Due to this, I obtained a solution by creating a new output adapter.

## How Has This Been Tested?
I tested the new feature in python 3.6, 3.7 and 3.8. I verified that the requests to the service are returned as byte image.

## Types of changes

- [ ] Documentation
- [ ] Test, CI, or build

## Component(s) if applicable
- [ ] BentoML CLI

## Checklist:


yubozhao:
I updated the spelling_check CI on the master branch. You can discard its failing, if caused by:
```
Exception occurred:
286    File "<string>", line 1, in <module>
287    AttributeError: module 'subprocess' has no attribute 'Process'
```

bentoml/adapters/image_output.py
```diff
@@ -0,0 +1,99 @@
...
+        Pack the return value of user defined API function into InferenceResults
+        """
+        results = []
+        for bytes_, _ in regroup_return_value(return_result, tasks):
+            try:
+                results.append(
+                    InferenceResult(
+                        data=bytes_,
+                        http_status=200,
+                        http_headers={"Content-Type": "image/png"},
```
bojiang, line 52: The `Content-Type` value should be configurable.
```python
ImageOutput(cors=True, content_type="image/png")
```

or
```python
ImageOutput(cors=True, extension="png")
```

yubozhao:
@canturan10 Thank you for the PR. This is really great addition to the community!

One minor change, can you add an example as part of the docstring? Other users can see it in the API reference.
You can find an example of that at:
https://github.com/bentoml/BentoML/blob/73a84a136fcd06231ff727cc07c9e06358f16c9b/bentoml/adapters/image_input.py#L60-L84

canturan10:
> @canturan10 Thank you for the PR. This is really great addition to the community!
> 
> One minor change, can you add an example as part of the docstring? Other users can see it in the API reference.
> You can find an example of that at:
> https://github.com/bentoml/BentoML/blob/73a84a136fcd06231ff727cc07c9e06358f16c9b/bentoml/adapters/image_input.py#L60-L84

It was my pleasure. I will add an example as soon as possible.

bojiang:
Hi @canturan10 
I just realize our `ImageInput` convert service input to numpy ndarrays, and you have
```python
out_img = Image.fromarray(...)
buf = io.BytesIO()
out_img.save(buf, format='png')
return buf.getvalue()
```
in your example.

We could make this converting inside of the output adapter, just like `ImageInput` does. Thus the `ImageOutput(format, <other convert options>)` takes numpy ndarrays as the user function return value.

Plus if we wanna handle bytes return values, just use `FileOutput` #1320 in the future.

parano:
echo to @bojiang's comment, I think it makes sense to accept a ndarray as return value here and does the conversion for user - 

Instead of doing the conversion to bytes in user code:

```python
@bentoml.api(input=ImageInput(), output=ImageOutput())
    def test(self, imgs):
        ...
        out_img = Image.fromarray(img_ndarray)
        buf = io.BytesIO()
        out_img.save(buf, format='png')
        return buf.getvalue()
```

the user can simply return the ndarray and expect the ImageOutput to convert it to bytes:
```python
@bentoml.api(input=ImageInput(), output=ImageOutput())
    def test(self, imgs):
        ...
        return img_ndarray
```

Currently, ImageInput supports only the numpy.ndarray type, we do plan to add support for other image input format such as bytes and PIL.image - I think we can do something similar for ImageOutput in the future, adding support for bytes and PIL.image as an option. e.g. `ImageOutput(type="ndarray|bytes|PIL")`. But for now, it would be great to make it consistent with ImageInput and use ndarray as the default.

And thanks for contributing the PR! It would be great if you could also add some unit tests for the ImageOutput class.

canturan10:
Thank you for your comments. but we can already return as ndarray.

```python
@bentoml.api(input=ImageInput(), output=DefaultOutput())
    def test(self, imgs):
        ...
        return img_ndarray
```
image

My problem is that ndarray is slow in web integration. Because of this, I wanted to send it as bytes.

bojiang:
@canturan10

I mean, now the ImageInput is converting image bytes (in HTTP requests) to ndarrays.
Correspondingly, it'd be better to have the ImageOutput converting ndarrays to image bytes(in HTTP responses).

It will be also more convenient for future users.

parano:
> @parano @bojiang
> Thank you for your comments. but we can already return as ndarray.
> 
> ```python
> @bentoml.api(input=ImageInput(), output=DefaultOutput())
>     def test(self, imgs):
>         ...
>         return img_ndarray
> ```
> 
> My problem is that ndarray is slow in web integration. Because of this, I wanted to send it as bytes.

What we meant was that the user can have `return img_ndarray` in their code, but the API server will convert that to an image and send the image back to the client - if it is using ImageOutput.  let me know what you think, happy to chat over a call if that's easier.

aarnphm:
@canturan10 We are currently looking into this problem for the 1.0 releases. I think it would be best to freeze current development for this specific PR

parano:
This feature will be covered in the upcoming 1.0 release, in which we redesigned the IO adapter and moved out all batching logic from it. Since there will be significant changes from the master branch, we will close this PR now. Thank you again for your contribution @canturan10!

Issue #1647, opened 2021-06-01
Enable micro-batch by default for "batch-True" APIs


parano:
**Is your feature request related to a problem? Please describe.**

* remove the current "--enable-microbatch" flag for `bentoml serve` and `bentoml serve-gunicorn` (show a deprecation warning when user use this flag)
* (next step) add a tool for sending test requests to an API server to test and verify batching behavior

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #1652, opened 2021-06-03
[Enhancement] Deprecate --enable-microbatch flag for bentoml serve/serve-gunicorn

larme:
## Description

Enable micro batch by default.

Implement first part of #1647

## Motivation and Context

Implementing the first part of #1647

## Types of changes

- [ ] Documentation
- [ ] Test, CI, or build

## Component(s) if applicable
- [x] BentoML CLI

## Checklist:

### DIFF
--- a/bentoml/configuration/default_bentoml.yml
+++ b/bentoml/configuration/default_bentoml.yml
@@ -10,7 +10,6 @@ bento_server:
   timeout: 60
   max_request_size: 20971520
   microbatch:
-    enabled: True
     workers: 1
     max_batch_size: Null
     max_latency: Null
--- a/bentoml/configuration/default_bentoml.yml
+++ b/bentoml/configuration/default_bentoml.yml
@@ -10,7 +10,7 @@ bento_server:
   timeout: 60
   max_request_size: 20971520
   microbatch:
-    enabled: False
+    enabled: True
     workers: 1
     max_batch_size: Null
     max_latency: Null
--- a/docs/source/concepts.rst
+++ b/docs/source/concepts.rst
@@ -891,7 +891,7 @@ Adaptive Micro-Batching
 
 .. note::
   The micro-batching option has become the default behavior starting release 0.12.0.
-  Use the --disable-microbatch option to turn off batching behavior.
+  Set --mb-max-batch-size=1 to turn off batching behavior.
 
 Micro batching is a technique where incoming prediction requests are grouped into small
 batches to achieve the performance advantage of batch processing in model inference
--- a/bentoml/cli/bento_service.py
+++ b/bentoml/cli/bento_service.py
@@ -66,9 +66,6 @@ def create_bento_service_cli(
     pip_installed_bundle_path: str = None,
     default_yatai_url: str = None,
     default_port: int = Provide[BentoMLContainer.config.bento_server.port],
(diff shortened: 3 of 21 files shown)


bentoml/configuration/default_bentoml.yml
```diff
@@ -10,7 +10,7 @@ bento_server:
   timeout: 60
   max_request_size: 20971520
   microbatch:
-    enabled: False
+    enabled: True
```
bojiang, line 13: I suggest to just removing this option, and change this field to optional in the schema.

bojiang:
In the integration tests, both `enable_microbatch=True/False` were covered by provide `enable_microbatch` fixture. We may removing it or stick it to `True` to make the CI faster.

docs/source/concepts.rst
```diff
@@ -891,7 +891,7 @@ Adaptive Micro-Batching
 
 .. note::
   The micro-batching option has become the default behavior starting release 0.12.0.
-  Use the --disable-microbatch option to turn off batching behavior.
+  Set --mb-max-batch-size=1 to turn off batching behavior.
```
parano, line 894: cc @bojiang  do we have any existing integration tests covering the behavior of `--mb-max-batch-size=1`?

bentoml/configuration/default_bentoml.yml
larme: Seems that we cannot simply change this field to optional (like `Optional("enabled", default=True)`) in the schema because inside [`BentoMLConfiguration.__init__()`](https://github.com/bentoml/BentoML/blob/master/bentoml/configuration/containers.py#L127), when `validate_schema=True`, [current code only validates the config dictionary](https://github.com/bentoml/BentoML/blob/master/bentoml/configuration/containers.py#L144), but does not reassign `self.config` to the validated result. That means `microbatch.enabled` is missing and this will crash the entire framework.

Solution is simply change every `SCHEMA.validate(self.config)` to `self.config = SCHEMA.validate(self.config)`. However this means when `validate_schema=False` the `microbatch.enabled` is still missing. I cannot think of a good way to solve this. Maybe we can just remove `validate_schema` argument and make the validation part mandatory. What is `validate_schema=False`'s use case?

docs/source/concepts.rst
bojiang: now we have https://github.com/bentoml/BentoML/pull/1661

bentoml/configuration/default_bentoml.yml
bojiang: Since we would have "mb_max_batch_size=1" to disable the auto-batching behaviour, we may fully remove the `config.microbatch.enabled`, but keep it in the validation for backwards compatibility.

larme: done in https://github.com/bentoml/BentoML/pull/1652/commits/d1afffdb9183afae505714a55df1302406c57b2d

larme:
> In the integration tests, both `enable_microbatch=True/False` were covered with the `enable_microbatch` fixture. We may removing it or stick it to `True` to make the CI faster.

In https://github.com/bentoml/BentoML/pull/1652/commits/d1afffdb9183afae505714a55df1302406c57b2d I just remove the `enable_microbatch` fixture.

bojiang (approved):

Issue #1654, opened 2021-06-03
Docker management


aarnphm:
## Background

- Currently BentoML manages docker images via `./dev/release-docker-images.sh` every docker directory contains a small `release.sh` to generate and push the docker images to the registry.  This approach is pretty manual and a lot of repetitions are invoked in every `release.sh`. 

- Additionally there aren't any ways for developer to develop `Dockerfile`  and test the docker images locally. Developers has to go into each ``release.sh`` to comment out ``push`` commands.

  

### Project structure under `docker/`

```shell
.
├── generated #DO NOT EDIT 
│   └── {package}
│      └── {distro}{distro_version}
│               ├── Dockerfile   # base dockerfile
│               ├── runtime       # runtime dockerfile
│               ├── cudnn         # cuda enabled dockerfile
│               └── devel          # devel image with bentoml installed from git
├── templates
│   ├── alpine
│   ├── debian
│   ├── docs
│   └── rhel
├── tools
│   ├── bashrc
│   └── {package}
├── Dockerfile
├── manager.py
├── manifest.yml
├── pyproject.toml
└── README.md

```

- when install different version of conda &rarr; lack of testing &rarr; CI-related using project integration test. &rarr; [Docker-in-docker](https://jpetazzo.github.io/2015/09/03/do-not-use-docker-in-docker-for-ci/) or [Sysbox](https://github.com/nestybox/sysbox) 
- supports multiple OS: amazonlinux - ubuntu - debian - alpine - centos
- deprecate `bentoml/azure-functions`

## Implementation

Dependency will be separated from BentoML's main package &rarr; introduces the package as a  _side-car_

We will render `Dockerfile` from jinja templates via `manager.py`.

Introduces ``manifest.yaml`` to control our release and supports for wider range of OS and architecture. 

Inside our ``manifest.yaml``:

```yaml
---
# allows users to define custom registries
# this can also include GCR, NVCR, Heroku, etc.
repository:
  docker.io:
    only_if: HUB_PASSWORD
    user: HUB_USERNAME
    pwd: HUB_PASSWORD
    registry:
      model-server: docker.io/bentoml/model-server
      yatai-service: docker.io/bentoml/yatai-service

packages:
  model-server:
    devel: &default_specs
      - debian
      - centos
    runtime:
      - *default_specs
      - amazonlinux
      - alpine
    cudnn:
      - *default_specs
  yatai-service:
    runtime:
      - *default_specs

cuda_dependencies:
[...]

PR #1683, opened 2021-06-16
feat: revamp docker management

aarnphm:
## Description

#1654 

## How Has This Been Tested?

Will be addressed in a separate PR

## Types of changes

- [ ] Documentation
- [ ] Test, CI, or build

## Checklist:


docker/.gitignore
```diff
@@ -0,0 +1,4 @@
+**/*.tmp.*
```
aarnphm, line 1: separate from main .gitignore

docker/generated/model-server/3.6/alpine3.14/Dockerfile
```diff
@@ -0,0 +1,60 @@
+# syntax = docker/dockerfile:experimental
```
aarnphm, line 1: enable developer to quickly build image with `ENABLE_BUILDKIT=1`. Othewise we will take advantage of `docker.api.build` from docker-py

docker/README.md
```diff
@@ -0,0 +1,46 @@
+notes:
```
aarnphm, line 1: We can ignore the generated file when reviewing, refers to all `*.j2` file

docker/manager.py
```diff
@@ -0,0 +1,821 @@
...
+from typing import Dict, Optional
+
+from absl import flags, app
+from cerberus import Validator
+from glom import glom, Path, PathAccessError, Assign, PathAssignError
+from jinja2 import Environment
+from ruamel import yaml
+from prettytable import PrettyTable
+
+logging.basicConfig(level=logging.NOTSET)
```
aarnphm, line 22: There is currently a bug somewhere that prevent `logging` to setup basic config. Could be `cerberus` related.

docker/manager.py
```diff
@@ -0,0 +1,821 @@
...
+
+# defined global vars.
+FLAGS = flags.FLAGS
+
+DOCKERFILE_PREFIX = ".Dockerfile.j2"
```
aarnphm, line 27: Feedback needed for managing constant.

docker/manager.py
```diff
@@ -0,0 +1,821 @@
...
+    type: dict
+    keysrules:
+      regex: '([\d\.]+)'
+    valuesrules:
+      type: dict
+    
+"""
+
+
+def validate_template_name(names):
```
aarnphm, line 202: This seems to be a little bit excessive. Can remove later

docker/manager.py
```diff
@@ -0,0 +1,821 @@
...
+    """
+    try:
+        _ = glom(obj, Assign(Path(*path), value))
+    except PathAssignError:
+        logging.error(
+            f"Exception occurred in update_data, unable to update {Path(*path)} with {value}"
+        )
+
+
+def _build_release_args(bentoml_version):
```
aarnphm, line 257: We will define our `ENV` with this function. Hardcoded `PYTHON_VERSION` and `BENTOML_VERSION`. Then later update values with build context.

docker/manager.py
```diff
@@ -0,0 +1,821 @@
...
+            )
+            pass
+    elif isinstance(names, list):
+        for _name in names:
+            validate_template_name(_name)
+    else:
+        raise TypeError(f"{names} has unknown type {type(names)}")
+
+
+def get_data(obj, *path, skip=False):
```
aarnphm, line 227: [pretty neat library](https://glom.readthedocs.io/en/latest/api.html?highlight=glom.Path#glom.Path)

docker/manager.py
```diff
@@ -0,0 +1,821 @@
...
+def flatten(arr):
+    for it in arr:
+        if isinstance(it, Iterable) and not isinstance(it, str):
+            for x in flatten(it):
+                yield x
+        else:
+            yield it
+
+
+def load_manifest_yaml(file):
```
aarnphm, line 340: load our `manifest.yml` and validate with a custom Cerberus validator

docker/manager.py
```diff
@@ -0,0 +1,821 @@
...
+
+def _update_build_ctx(ctx_dict, spec_dict: Dict, *ignore_keys):
+    # flatten release_spec to template context.
+    for k, v in spec_dict.items():
+        if k in list(flatten([*ignore_keys])):
+            continue
+        ctx_dict[k] = v
+
+
+def _add_base_release_type(build_ctx, release_type: list):
```
aarnphm, line 366: quick hack to setup `base` image, which have conda and python that can be used later to build runtime, devel, etc. Definitely improvements here since I wrote this at 3am 😄

docker/manager.py
```diff
@@ -0,0 +1,821 @@
...
+        if obj is None:
+            return self
+        value = obj.__dict__.get(self.__name__, self._missing)
+        if value is self._missing:
+            value = self.func(obj)
+            obj.__dict__[self.__name__] = value
+        return value
+
+
+class TemplatesMixin(object):
```
aarnphm, line 493: `TemplatesMixin` will handle our templates generation and also provide a `build_path` dictionary where:
  - `keys`: image tag
  - `values`: path to generated Dockerfile

We can later use this `build_path` to build our image with `docker-py`

docker/manager.py
```diff
@@ -0,0 +1,821 @@
...
+    """Mixin for generating Dockerfile from templates."""
+
+    _build_ctx = {}
```
aarnphm, line 496: This is our placeholder for all releases metadata, refers below.

docker/manager.py
```diff
@@ -0,0 +1,821 @@
...
+    def _generate_distro_release_mapping(self, package):
+        """Return a dict containing keys as distro and values as release type (devel, cudnn, runtime)"""
+        releases = defaultdict(list)
+        for k, v in self._packages[package].items():
+            _v = list(flatten(v))
+            for distro in _v:
+                releases[distro].append(k)
+        return releases
+
+    def prepare_template_context(
```
aarnphm, line 568: use ``--dump_metadata`` to see each image specs that will be parsed into our jinja templates.

docker/manager.py
```diff
@@ -0,0 +1,821 @@
...
+            python_version: target list of supported python version. Overwrite default supports list if
+                            FLAGS.python_version is given.
+
+        Returns:
+            Let the magic happens.
+        """
+        _tag_metadata = defaultdict(list)
+
+        logging.info(f"Generating dockerfile to {target_dir}...\n")
+        for package in self._packages.keys():
```
aarnphm, line 699: Feedback needed. As I mentioned above, I wrote this around 3am 😄 so there are like 5 nested for loops

docker/templates/alpine/base.Dockerfile.j2
```diff
@@ -0,0 +1,52 @@
+{{ metadata.header }}
```
aarnphm, line 1: Currently I was thinking to structured our generated image directory as below:
``` bash
generated
  - model-server
    - python_version
      - {distro}{distro_version}
         - Dockerfile (this is the base image)
         - cudnn
         - devel
         - runtime 
```
This is just a personal preference, where I would like to define all build_args into `ENV`and have unique Dockerfile for each releases.

Also it would be better for debugging.

docker/templates/rhel/nvidia-ml.repo.j2
```diff
@@ -0,0 +1,6 @@
+[nvidia-ml]
```
aarnphm, line 1: `centos` nvidia related

docker/manager.py
```diff
@@ -0,0 +1,821 @@
...
+            # update our packages ctx.
+            logging.info(f"> Working on {package}")
+            update_data(self._build_ctx, package, "package")
+            for _py_ver in python_version:
+                logging.info(f">> Using python{_py_ver}")
+                for distro, _release_type in self._generate_distro_release_mapping(
```
aarnphm, line 705:
```python
dict: {"ubuntu":["cudnn","runtime","devel"],"alpine":["runtime"]}
```
We control generation base image via `"manifest.yml".needs_deps_image`

docker/manager.py
```diff
@@ -0,0 +1,821 @@
...
+            _build_ctx['base_tags'] = 'base'
+        _build_ctx.pop('needs_deps_image')
+
+        # setup python version
+        update_data(
+            _build_ctx, python_version, "envars", "PYTHON_VERSION",
+        )
+
+        # setup cuda deps
+        if _build_ctx['cuda_prefix_url']:
```
aarnphm, line 613: build context logics can be found here

docker/manager.py
```diff
@@ -0,0 +1,821 @@
...
+
+def _add_base_release_type(build_ctx, release_type: list):
+    try:
+        return release_type + [build_ctx['base_tags']]
+    except KeyError as e:
+        # when base_tags is either not set or available
+        return release_type
+
+
+class MetadataSpecValidator(Validator):
```
aarnphm, line 374: `MetadataSpecValidator` with `SPEC_SCHEMA` will achieve a few things:
- it makes sure that our defined `registry` under `repository` keys are also defined under `packages`
- ensure envars has correct format `ARG=foobar`. (could also have support for `ARG foobar`)
- subkeys are dependent on main keys: e.g:
  - `pwd`, `user` for `repository` keys depends on `only_if` field to be defined
  - `cuda_*` under `release_spec` keys depends `cuda` version, etc.
- supported distros are also cross-validation between `releases` and one defined under `packages.{model-server,yatai-service}`

aarnphm (commented):
still has a lot of things to do:
- [x] Finishing building steps with `docker-py`
- [x] Generate README with files directory
- [ ] Add `Dockerfile` to bundle `manager.py` for easier development.
- [ ] Documentation
- [ ] add usage for `start_dev.sh` and `build_dev.sh` for `yatai-service` image.

docker/templates/alpine/base.Dockerfile.j2
aarnphm: probably parsed python_version at build-time to reduce boilerplate

aarnphm: There are a lot of similarity between centos and ubuntu releases (difference are pretty much os version). This is due to `docker.api.build` requires a file object for Dockerfile to be found in given directory to build the image.

I was thinking to use 
```docker
ARG OS_IMAGE

FROM $OS_IMAGE
``` 
and generate from one Dockerfile (eg: for both `ubuntu20.04` and `ubuntu18.04`). However, the current implementation helps easier debugging and exploring image

docker/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+LABEL maintainer="BentoML Contributor <contact@bentoml.ai>"
+
+ENV DOCKER_BUILDKIT=1
+
+RUN apt-get update && \
+    apt-get install -y python3 python3-pip bash curl build-essential
+
+SHELL ["/bin/bash", "-exo", "pipefail", "-c"]
+
+# https://jpetazzo.github.io/2015/09/03/do-not-use-docker-in-docker-for-ci/
```
aarnphm, line 14: I think for the given need of having an environment up, docker in docker should not be too much of an issue

aarnphm:
@parano

aarnphm:
One possible tags for nightly release could be `<bentoml.__version__>-<python_version>-nightly`, so for example : `0.13.0+60.geb2d816-python3.7-nightly`

.flake8
```diff
@@ -2,3 +2,6 @@
 ignore = E203, E266, W503, E402
 max-line-length = 88
 exclude = tests/yatai/proto, proto, _version.py
+per-file-ignores =
+    docker/manager.py:F405, F403, E501
```
bojiang, line 6: Any reason to ignore these flags?

dev/format.sh
```diff
@@ -1,9 +1,13 @@
...
-GIT_ROOT=$(git rev-parse --show-toplevel)
-# format yatai webui
-WEB_UI_DIR=$GIT_ROOT/bentoml/yatai/web
+GIT_ROOT="$(git rev-parse --show-toplevel)"
 
 black -S "$GIT_ROOT"
 
+# format docker directory format.
+DOCKER_DIR="$GIT_ROOT"/docker
+isort "$DOCKER_DIR"
```
bojiang, line 9: https://github.com/bentoml/BentoML/pull/1734

dev/lint.sh
```diff
@@ -15,4 +15,4 @@ pylint --rcfile="./pylintrc" bentoml
 echo "Running pylint on test module.."
 pylint --rcfile="./pylintrc" tests e2e_tests
 
-echo "Done"
+echo "Done"
```
bojiang, line 18:
```suggestion
echo "Done"

```

.flake8
```diff
@@ -2,3 +2,6 @@
...
+    docker/utils.py:W605, E501
```
bojiang, line 7:
```suggestion
    docker/utils.py:W605, E501

```

.flake8
aarnphm: I will keep E501 for now (line-too-long). Before i use star imports

bojiang:
Discussed with aarnphm on Slack.
We are introducing poetry for these benefits:

1. dependency isolation
1. resolve&lock dependencies, generate a dependency spec
1. restore the dependencies

For 1, I think in most cases, the deployment is in a container, which is already isolated
For 3, `pyproject.toml` (PEP 518) is already supported by pip https://github.com/pypa/pip/pull/4144

So we just need a dependency resolver.

`poetry` is mainly for development, we don't need to ship it in each deployment. It also has overlaps with `conda`.
The lightweight `pip-tools` should be a better choice.

aarnphm:
> For the resolver, `pip-tools` should be a lightweight alternative.
For the most part our main dependency for `model_server` hasn't shown any conflict, mostly `yatai_service`. However this would be great to introduce for future proof.

bojiang:
> For the most part our main dependency for model_server hasn't shown any conflict, mostly yatai_service. However this would be great to introduce for future proof.

Conflict is not the major concern. I just say it's not a good idea to includes a development purpose lib in our deployments, especially when we do not really need it.

.flake8
```diff
@@ -2,3 +2,6 @@
 ignore = E203, E266, W503, E402
 max-line-length = 88
 exclude = tests/yatai/proto, proto, _version.py
+per-file-ignores =
+    docker/manager.py:E501
+    docker/utils.py:W605, E501
```
bojiang, line 7: W605: Are you encountering escape issue in long strings?
You may need r"\" or r''' \ '''

aarnphm: This is limited to how `cerberus` parse regex. It only accept regex that encapsulate between single quotes, eg. 

```yaml

field:
  regex: '(\w+)\,{1}(\d+)'
```

docker/manager.py
```diff
@@ -0,0 +1,742 @@
...
+    log.debug("Creating an instance of DockerClient")
+    docker_client = DockerClient(
+        base_url="unix://var/run/docker.sock", timeout=3600, tls=True
+    )
+
+
+class LogsMixin(object):
+    """Logs Mixin for docker build process"""
+
+    def build_logs(self, resp: Iterator, image_tag: str):
```
alxmke, line 75: Your documentation overall is very good. I would appreciate some more explicit, extended comments associated with your function headers, generally, similar to what you do with some of your classes, like so:
```suggestion
    """
    does what build_logs() does, using what it uses, etc.
    Args:
        resp: is what a resp is, does what a resp does
        image_tag: is what an image_tag is, does what an image_tag does
    """
    def build_logs(self, resp: Iterator, image_tag: str):
```

Issue #1588, opened 2021-04-20
Env var BENTOML_PORT should set API server port


parano:
**Describe the bug**

The environment variable BENTOML_PORT is expected to be used as the port number for API server, and this is currently implemented using `click`'s `env_var` option: https://github.com/bentoml/BentoML/blob/v0.12.1/bentoml/cli/bento_service.py#L254

This works when launching API server from CLI, e.g.:
```bash
BENTOML_PORT=3030 bentoml serve-gunicorn IrisClassifier:latest
```

But failed when using `docker run` to pass the env var, see steps to reproduce below:

**To Reproduce**

```bash
cd $(bentoml get IrisClassifier:latest --print-location --quiet)
docker build . -t iris_classifier:latest
docker run -p 5001:5001 -e BENTOML_PORT='5001' iris_classifier
```

See user report: https://github.com/bentoml/BentoML/discussions/1585#discussioncomment-632910

**Expected behavior**

The server is expected to run on port 5001, but instead it still runs on default 5000 port.

**Screenshots/Logs**

**Environment:**
 - OS: [e.g. MacOS 10.14.3] MacOS
 - Python Version [e.g. Python 3.7.1] Python 3.7
 - BentoML Version [e.g. BentoML-0.8.6] 0.12.1

**Additional context**

ssheng:
Why should `bentoml run` start a server?

ssheng:
The issue is only limited to `BENTOML_PORT` environment variable. Other click parameters work just fine. The problem is in how we generate Dockerfile. In the Dockerfile template, we set a environment variable `PORT` to 5000, seemingly for Heroku.

https://github.com/bentoml/BentoML/blob/238bc9757033e5b3ffad6810c5dba95ac596c70e/bentoml/saved_bundle/templates.py#L113

In the ENTRYPOINT script, we override `BENTOML_PORT` to with the value of `PORT`, which explained why `--port` parameter works but not `BENTOML_PORT`.

https://github.com/bentoml/BentoML/blob/238bc9757033e5b3ffad6810c5dba95ac596c70e/bentoml/saved_bundle/docker-entrypoint.sh#L19

Anyone knows the contexts of Heroku port 5000? The solution maybe to set `PORT` with the value of `BENTOML_PORT` instead of vice versa.

jjmachan:
[...]

PR #1706, opened 2021-06-21
Use $BENTOML_PORT as Dockerfile default port

jjmachan:
## Description

This fixes the bug mentioned in #1588. Also checks if $PORT is present
and overides $BENTOML_PORT to $PORT if found.

## Motivation and Context

Solves #1588 
## How Has This Been Tested?

I've tested it locally and on Heroku service. 

## Types of changes

- [x] Documentation
- [ ] Test, CI, or build

## Component(s) if applicable
- [ ] BentoML CLI

## Checklist:

### DIFF
--- a/bentoml/saved_bundle/templates.py
+++ b/bentoml/saved_bundle/templates.py
@@ -109,9 +109,9 @@ RUN rm ./bundled_pip_dependencies/docker-entrypoint.sh && ./bentoml-init.sh inst
 # copy over model files
 COPY --chown=bentoml:bentoml . ./
 
-# the env var $PORT is required by heroku container runtime
-ENV PORT 5000
-EXPOSE $PORT
+# Default port for service
+ENV BENTOML_PORT 5000
+EXPOSE $BENTOML_PORT
 
 USER bentoml
 RUN chmod +x ./docker-entrypoint.sh
--- a/bentoml/saved_bundle/templates.py
+++ b/bentoml/saved_bundle/templates.py
@@ -109,9 +109,8 @@ RUN rm ./bundled_pip_dependencies/docker-entrypoint.sh && ./bentoml-init.sh inst
 # copy over model files
 COPY --chown=bentoml:bentoml . ./
 
-# Default port for service
-ENV BENTOML_PORT 5000
-EXPOSE $BENTOML_PORT
+# Default port for BentoML Service
+EXPOSE 5000
 
 USER bentoml
 RUN chmod +x ./docker-entrypoint.sh
--- a/docs/source/deployment/heroku.rst
+++ b/docs/source/deployment/heroku.rst
@@ -105,6 +105,10 @@ BentoService and available for sending test request:
 Build and deploy to Heroku
 ==========================
 
+.. note::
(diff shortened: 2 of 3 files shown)


jjmachan:
Also, I was wondering if I should mention this in the cli reference as well. Right now in effect, we have to environment variables that can be set to specify the port to which our docker service will listen to.

bentoml/saved_bundle/templates.py
```diff
@@ -109,9 +109,9 @@
 # copy over model files
 COPY --chown=bentoml:bentoml . ./
 
-# the env var $PORT is required by heroku container runtime
-ENV PORT 5000
-EXPOSE $PORT
+# Default port for service
+ENV BENTOML_PORT 5000
+EXPOSE $BENTOML_PORT
```
parano, line 114: I think we also need `EXPOSE $PORT` here in order for it to work with Heroku right?  are you sure this change works when you also set the `port` to a different value on heroku?

docs/source/deployment/heroku.rst
```diff
@@ -105,6 +105,10 @@ BentoService and available for sending test request:
 Build and deploy to Heroku
 ==========================
 
+.. note::
+  Apps deployed on Heroku must listen to a specific port number which is
+  specified by the $PORT environment variable in the dyno. Bentoml will
```
parano, line 110: is it a typo?  I guess you mean `...in the demo`?

bentoml/saved_bundle/templates.py
parano: never mind, I think I misunderstood the `EXPOSE` command. Looking at related documentation here https://docs.docker.com/engine/reference/builder/#expose, I think it's ok if we just remove `ENV BENTOML_PORT 5000` here and change this line to `EXPOSE 5000`, that way we only define the default port once in the BentoML config, what do you think?

docs/source/deployment/heroku.rst
jjmachan: no dyno is what heroku calls their containers, so I thought I'd use the same term here.

bentoml/saved_bundle/templates.py
jjmachan: I'm not sure. So one doubt is, suppose we use this with something like a docker-compose and override the BENTOML_PORT with  the environment variable to another port, should we also use EXPOSE on that port number or is EXPOSE suppose to be always a fixed port number?

parano: @jjmachan per the linked documentation, EXPOSE doesn't actually do anything: `The EXPOSE instruction does not actually publish the port. It functions as a type of documentation between the person who builds the image and the person who runs the container, about which ports are intended to be published. To actually publish the port when running the container, use the -p flag on docker run to publish and map one or more ports`

If the person who runs the container overrides it via  BENTOML_PORT env var, this person should also be aware to use `-p` to publish that port.  In other cases, `EXPOSE 5000` tells people that this is the default port that will be used.

jjmachan: Thanks for explaining it @parano! I've made the changes

parano:
thank you @jjmachan!

Issue #1591, opened 2021-04-23
[Tracing] Jaeger tracing need dict type header


hsiehjackson:
**Describe the bug**

When using tracing with Jaeger in bentoml, there is an error from Jaeger client because of the type of header.
[Bug](https://github.com/bentoml/BentoML/blob/master/bentoml/tracing/jaeger.py#L104) - header type is 'werkzeug.datastructures.EnvironHeaders'
[Reason](https://github.com/jaegertracing/jaeger-client-python/blob/master/jaeger_client/codecs.py#L57) - header type must be 'dict'

**To Reproduce**

**Expected behavior**

**Screenshots/Logs**

**Environment:**
 - OS: [e.g. MacOS 10.14.3]
 - Python Version [e.g. Python 3.7.1]
 - BentoML Version [e.g. BentoML-0.12.1]

**Additional context**

parano:
Fixed via https://github.com/bentoml/BentoML/pull/1720

PR #1720, opened 2021-06-29
Fix opentracing.propagation.InvalidCarrierException: carrier not a collection

wei840222:
issue: https://github.com/bentoml/BentoML/issues/1591
When open the jaeger tracing like guide: https://docs.bentoml.org/en/stable/guides/tracing.html.
It will raise an exception on runtime and all api request become s to 500.
log:
```bash
opentracing.propagation.InvalidCarrierException: carrier not a collection
[2021-06-29 11:48:20,120] ERROR - Traceback (most recent call last):
  File "/opt/conda/lib/python3.6/site-packages/bentoml/marshal/marshal.py", line 109, in request_dispatcher
    resp = await func(request)
  File "/opt/conda/lib/python3.6/site-packages/bentoml/marshal/marshal.py", line 329, in request_dispatcher
    resp = await self.relay_handler(request)
  File "/opt/conda/lib/python3.6/site-packages/bentoml/marshal/marshal.py", line 342, in relay_handler
    request_headers=request.headers,
  File "/opt/conda/lib/python3.6/contextlib.py", line 81, in __enter__
    return next(self.gen)
  File "/opt/conda/lib/python3.6/site-packages/bentoml/tracing/jaeger.py", line 112, in async_span
    with self.span(*args, async_transport=True, **kwargs) as ctx:
  File "/opt/conda/lib/python3.6/contextlib.py", line 81, in __enter__
    return next(self.gen)
  File "/opt/conda/lib/python3.6/site-packages/bentoml/tracing/jaeger.py", line 104, in span
    scope.span.context, Format.HTTP_HEADERS, request_headers,
  File "/opt/conda/lib/python3.6/site-packages/jaeger_client/tracer.py", line 266, in inject
    codec.inject(span_context=span_context, carrier=carrier)
  File "/opt/conda/lib/python3.6/site-packages/jaeger_client/codecs.py", line 59, in inject
    raise InvalidCarrierException('carrier not a collection')
opentracing.propagation.InvalidCarrierException: carrier not a collection
```

and found inject function in https://github.com/jaegertracing/jaeger-client-python/blob/master/jaeger_client/codecs.py#L57 only accept dict type, and seems request header not need to pass to the span.
so, I follow the opentracing example on github https://github.com/opentracing/opentracing-python#outbound-request only pass a empty dict to it.


wei840222:
I think maybe ```dict(request_headers)``` is beter?

wei840222:
The failure on CI does not seem to be related to this modification?

parano:
> The failure on CI does not seem to be related to this modification?

That's right, I think that particular test is flaky on github CI environment, let me trigger a re-run

bentoml/tracing/jaeger.py
```diff
@@ -101,7 +101,7 @@ def span(
                 token = span_context_var.set(scope.span.context)
                 if request_headers and TRACE_ID_HEADER not in request_headers:
                     tracer.inject(
-                        scope.span.context, Format.HTTP_HEADERS, request_headers,
+                        scope.span.context, Format.HTTP_HEADERS, dict(request_headers),
```
parano, line 104: I think this will break how the headers are injected between marshal and API server here - since the injected headers dict here will not being used by the aiohttp client that sends the request to downstream service: https://github.com/bentoml/BentoML/blob/master/bentoml/marshal/marshal.py#L358

Looks like this will need a bit of refactoring to actually fix the problem, @bojiang any suggestions?

bojiang: I don't like the `inject` approach in fact. Is this necessary?

wei840222: as https://github.com/opentracing/opentracing-python#outbound-request saiad.
if no any outbound http request. just put a empty dict() also acceptable.
or even remove inject function call.

wei840222: which one is preferd?

bentoml/tracing/jaeger.py
```diff
@@ -101,7 +101,7 @@ def span(
                 token = span_context_var.set(scope.span.context)
                 if request_headers and TRACE_ID_HEADER not in request_headers:
                     tracer.inject(
```
bojiang, line 103:
```suggestion
                   headers_dict = dict(request_headers)
                    tracer.inject(
```

bentoml/tracing/jaeger.py
bojiang:
```suggestion
                        scope.span.context, Format.HTTP_HEADERS, headers_dict,
```

bentoml/tracing/jaeger.py
```diff
@@ -101,7 +101,7 @@ def span(
...
-                        scope.span.context, Format.HTTP_HEADERS, request_headers,
+                        scope.span.context, Format.HTTP_HEADERS, dict(request_headers),
                     )
```
bojiang, line 105:
```suggestion
                    )
                    request_headers.update(headers_dict)
```

bentoml/tracing/jaeger.py
wei840222: done

bentoml/tracing/jaeger.py
wei840222: done

bentoml/tracing/jaeger.py
wei840222: done

PR #1729, opened 2021-07-04
fix: adapt changes introduced by pylint 2.9

larme:
## Description

adapt changes introduced by pylint 2.9, by which`arguments-differ` is split into `arguments-differ` and `arguments-renamed`

## Checklist:

### DIFF
--- a/bentoml/frameworks/evalml.py
+++ b/bentoml/frameworks/evalml.py
@@ -63,7 +63,7 @@ class EvalMLModelArtifact(BentoServiceArtifact):
     def _model_file_path(self, base_path):
         return os.path.join(base_path, self.name + EVALML_MODEL_PICKLE_EXTENTION)
 
-    def pack(self, evalml_model, metadata=None):  # pylint:disable=arguments-differ
+    def pack(self, evalml_model, metadata=None):  # pylint:disable=arguments-renamed
         self._validate_package()
         self._model = evalml_model
         return self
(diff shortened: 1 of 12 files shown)


bentoml/frameworks/evalml.py
```diff
@@ -63,7 +63,7 @@ def _validate_package(self):
     def _model_file_path(self, base_path):
         return os.path.join(base_path, self.name + EVALML_MODEL_PICKLE_EXTENTION)
 
-    def pack(self, evalml_model, metadata=None):  # pylint:disable=arguments-differ
+    def pack(self, evalml_model, metadata=None):  # pylint:disable=arguments-renamed
```
larme, line 66: The root cause of these recent pylint warnings is that in pylint 2.9, `arguments-differ` is split into `arguments-differ` (which identifies any changes at the parameter number of overridden functions) and `arguments-renamed` (which identifies changes at the parameter
names of overridden functions). We can change every `pylint: disable=arguments-differ` into `pylint: disable=arguments-differ, arguments-renamed` to keep backward compatibility with the older pylint, but I think just change `pylint: disable=arguments-differ` to `pylint:
disable=arguments-renamed` and ask contributors to bump their pylint version is more suitable.

parano: sounds great, maybe we can enforce the pylint version in the `make lint` script in the future. I think fixing CI is a higher priority for now.

Issue #1698, opened 2021-06-20
Fix PyTorchModelArtifact store_as_json_and_weights=True option


parano:
**Describe the bug**

The fix in https://github.com/bentoml/BentoML/pull/1696 exposed an error with PyTorchModelArtifact when setting the  store_as_json_and_weights option to `True`.

**To Reproduce**

**Expected behavior**

**Screenshots/Logs**

**Environment:**
 - OS: [e.g. MacOS 10.14.3]
 - Python Version [e.g. Python 3.7.1]
 - BentoML Version [e.g. BentoML-0.8.6]

**Additional context**

PR #1736, opened 2021-07-07
[WIP] fix(frameworks): keras multi model loading

larme:
This commit is not meant to be the final solution for the current keras loading problem (#1698), but try to show what may go wrong in the original approach. The main point is that we want to have separated graph for every keras model loaded

## Checklist:

### DIFF
--- a/bentoml/frameworks/keras.py
+++ b/bentoml/frameworks/keras.py
@@ -204,7 +204,8 @@ class KerasModelArtifact(BentoServiceArtifact):
         self.create_session()
         with self.graph.as_default(), self.sess.as_default():
             if self._store_as_json_and_weights:
-                # load keras model via json and weights if store_as_json_and_weights=True
+                # load keras model via json and weights if
+                # store_as_json_and_weights=True
                 with open(self._model_json_path(path), 'r') as json_file:
                     model_json = json_file.read()
                 model = keras_module.models.model_from_json(
--- a/bentoml/frameworks/keras.py
+++ b/bentoml/frameworks/keras.py
@@ -130,9 +130,9 @@ class KerasModelArtifact(BentoServiceArtifact):
 
         self.graph = tf.compat.v1.get_default_graph()
         self.sess = tf.compat.v1.Session(graph=self.graph)
-        tf.compat.v1.keras.backend.set_session(self.sess)
+        # tf.compat.v1.keras.backend.set_session(self.sess)
 
-    def pack(self, data, metadata=None):  # pylint:disable=arguments-renamed
+    def _pack_without_bind_session(self, data, metadata=None):
         try:
             import tensorflow as tf
         except ImportError:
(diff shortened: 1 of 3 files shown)


bentoml/frameworks/keras.py
```diff
@@ -168,9 +168,13 @@ def pack(self, data, metadata=None):  # pylint:disable=arguments-renamed
...
                 raise InvalidArgument(error_msg)
 
-        self.bind_keras_backend_session()
         self._model = model
         self._custom_objects = custom_objects
+
+    def pack(self, data, metadata=None):  # pylint:disable=arguments-renamed
+
+        self.bind_keras_backend_session()
+        self._pack_without_bind_session(data, metadata)
```
larme, line 177: In old codes we use `pack` in two contexts, when packing a in-memory model to the artifacts and when loading the model from disk. In latter case we may not want to bind current session because we already create the session/graph when loading the model from disk.

tests/bento_service_examples/keras_classifier.py
```diff
@@ -4,20 +4,28 @@
...
+        KerasModelArtifact('model2', store_as_json_and_weights=True),
     ]
 )
 class KerasClassifier(bentoml.BentoService):
     @bentoml.api(input=JsonInput(), batch=True)
     def predict(self, jsons):
-        return self.artifacts.model.predict(np.array(jsons))
+        raw_artifact = self._artifacts['model']
+        with raw_artifact.graph.as_default(), raw_artifact.sess.as_default():
+            return self.artifacts.model2.predict(np.array(jsons))
```
larme, line 25: This is the ugly part. For every model we may want to use its own graph when doing inference, which means we need to access the raw `KerasModelArtifact` instead of `self.artifacts.model`. Maybe we can try to make a customized context manger to make this look better?

tests/bento_service_examples/keras_classifier.py
```diff
@@ -4,20 +4,28 @@
...
+        KerasModelArtifact('model2', store_as_json_and_weights=True),
     ]
 )
 class KerasClassifier(bentoml.BentoService):
     @bentoml.api(input=JsonInput(), batch=True)
     def predict(self, jsons):
-        return self.artifacts.model.predict(np.array(jsons))
+        raw_artifact = self._artifacts['model']
+        with raw_artifact.graph.as_default(), raw_artifact.sess.as_default():
+            return self.artifacts.model.predict(np.array(jsons))
```
larme, line 25: This is ugly. We want to access the session/graph associated with the keras model, so we need to access the raw `KerasModelArtifact` instead of `self.artifacts.model`.

Maybe we can create a customized context manager to make this look better?

bentoml/frameworks/keras.py
```diff
@@ -197,24 +201,24 @@ def load(self, path):
...
-                    model.load_weights(self._model_weights_path(path))
-        else:
-            # otherwise, load keras model via standard load_model
-            model = keras_module.models.load_model(
-                self._model_file_path(path),
-                custom_objects=self._default_custom_objects,
-            )
-        return self.pack(model)
+        self.create_session()
+        with self.graph.as_default(), self.sess.as_default():
```
aarnphm, line 205: We only need to set context to the current session only, since we will have the session bind to the default graph. This is fixed in the 1.0 branch on my end

aarnphm:
Fixed in 1.0

PR #1760, opened 2021-07-17
fix: Exclude '.ipynb_checkpoints' when seeking pip packages by infer_pip_packages(True)

withsmilo:
## Description

When using BentoML in a Jupyter notebook environment, I found that the `.ipynb_checkpoints` directory was scanned when the `infer_pip_packages` option was turned on. I don't think this makes any sense, it just slows down BentoML. So it would be better to exclude it.

## Checklist:


parano:
@withsmilo thank you for the PR and making it to the 1.0 branch!

PR #1785, opened 2021-07-28
fix(type): make some decorator utility functions type-safe and add some type annotations

yetone:
## Description

Some internal decorator utility functions are not type-safe now, for example: 

before:

image

after:

image

image

I fixed the error above in this PR, but some decorators like `cached_contextmanager` and `catch_exceptions` are still not type-safe, this problem that relies on this issue: https://github.com/python/mypy/issues/8645

## Checklist:

### DIFF
--- a/bentoml/_internal/utils/__init__.py
+++ b/bentoml/_internal/utils/__init__.py
@@ -5,10 +5,16 @@ import os
 import socket
 import tarfile
 from io import StringIO
-from typing import Callable, Optional
+from typing import (
+    Optional, TypeVar, Type, Union, overload, Dict, Iterator, Any, Tuple,
+    TYPE_CHECKING, Generic, Callable)
 from urllib.parse import urlparse, uses_netloc, uses_params, uses_relative
 
 from google.protobuf.message import Message
+from mypy.typeshed.stdlib.contextlib import _GeneratorContextManager
+
+if TYPE_CHECKING:
+    from bentoml._internal.yatai_client import YataiClient
 
 from ..utils.gcs import is_gcs_url
 from ..utils.lazy_loader import LazyLoader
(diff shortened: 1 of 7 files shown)


bentoml/_internal/utils/__init__.py
```diff
@@ -5,10 +5,16 @@
...
 import tarfile
 from io import StringIO
-from typing import Callable, Optional
+from typing import (
+    Optional, TypeVar, Type, Union, overload, Dict, Iterator, Any, Tuple,
+    TYPE_CHECKING, Generic, Callable)
 from urllib.parse import urlparse, uses_netloc, uses_params, uses_relative
 
 from google.protobuf.message import Message
+from mypy.typeshed.stdlib.contextlib import _GeneratorContextManager
```
aarnphm, line 14: This means adding mypy into our library dependency. I would rather to reserve all `typeshed` import under `TYPE_CHECKING`

aarnphm (approved):
LGTM

aarnphm:
I will fix the CI related in a separate PR

bentoml/_internal/utils/__init__.py
yetone: Sorry, I forgot to add the relevant dependencies.

aarnphm: no worries since its only a type call i just added to the files. I think we shouldn't add mypy to library dependency anw

yetone: Because `mypy` uses `typeshed` types to annotate the standard libraries types during static analysis, so that `mypy` can only recognize the types in `typeshed`.

image

aarnphm: I see. I don't know if its a good practice to include mypy into library dependency. I don't see a lot of other library doing so.

yetone: I'm not sure how to do it, I'll research how other libraries are doing it.

aarnphm:
We should revisit sometimes once we decide whether to go to 3.10 or 3.9 as default

Issue #1796, opened 2021-08-03
Deploy Yatai By remote db and local path report  can't locate config file 'bentoml.yml' 


wuxutao:
**Describe the bug**
 I deploy Yatai Service by a remote PostgreSQL database and local repo base,  
When I use command:  bentoml containerize IrisClassifier:latest -t iris-classifier  --yatai-url=127.0.0.1:50051 
It report error:  Error: bentoml-cli containerize failed: BentoML can't locate config file 'bentoml.yml' in saved bundle in path: /bentoml/repository/IrisClassifier/20210803122208_2E7460

But the file "bentoml.yml" is in the docker volume

**To Reproduce**

1. Start Yaitai service: "docker run -p 3000:3000 -p 50051:50051     -v ~/bentoml:/bentoml   bentoml/yatai-service:latest    --db-url postgresql://postgres:postgres@192.168.0.47:15433/wt     --repo-base-url /bentoml/repository"
2. # Save Bento bundle to Yatai server
bento_svc.save(yatai_url="127.0.0.1:50051")
3. bentoml containerize IrisClassifier:latest --yatai-url=127.0.0.1:50051

It will report error

![image](https://user-images.githubusercontent.com/2825910/127980581-d8451bdc-a885-47ff-a8b3-0ae42eca1879.png)

and what  is "api.amplitude.com"?

mqk:
I also see this same error. In my setup I deployed the yatai service on kubernetes (minikube).
```
$ bentoml containerize --yatai-url http://127.0.0.1:62838 IrisClassifier:latest
Found Bento: /root/bentoml/repository/IrisClassifier/20210823162836_FDA855
Error: bentoml-cli containerize failed: BentoML can't locate config file 'bentoml.yml' in saved bundle in path: /root/bentoml/repository/IrisClassifier/20210823162836_FDA855
```

parano:
Hi @wuxutao @mqk - sorry about the delay in getting back to you. The reason for this issue is that in BentoML 0.13, Yatai requires a storage backend for storing the Models and it supports both local file system and blob storage(AWS S3, MinIO, Azure blob storage). However, when Yatai is deployed in production mode and connected remotely, only blob storage is supported. Local storage is only meant for local Yatai only. This is definitely something we should document better.

[...]

PR #1840, opened 2021-09-09
fix: issue#1796

mqk:
## Description

This PR fixes https://github.com/bentoml/BentoML/issues/1796: calling `bentoml containerize` fails when you try to use a remote yatai service that has been configured to locally (postgres backed) store bentos.

The problem occurs because we attempt to fetch the BentoServiceMetadata using `load_bento_service_metadata(saved_bundle_path)`. That calls `load_saved_bundle_config(bundle_path)`, which has been wrapped with `@resolve_remote_bundle`. That wrapper deduces that this is a local bundle (`_is_remote_path(bundle_path)` returns `False`), and so we attempt to load from a local path on the filesystem that is making the `bentoml` call, whereas it should really refer to the filesystem in the Docker container of the yatai grpc service.

An easy fix is to just call `yatai_client.repository.get(bento)` to obtain BentoServiceMetadata, because this API is aware of how to deal with the grpc service.

## Motivation and Context

https://github.com/bentoml/BentoML/issues/1796

## How Has This Been Tested?

`./ci/unit_tests.sh`

## Checklist:

### DIFF
--- a/bentoml/cli/bento_service.py
+++ b/bentoml/cli/bento_service.py
@@ -430,12 +430,9 @@ def create_bento_service_cli(
 
         _echo(f"Found Bento: {saved_bundle_path}")
 
-        # fmt: off
-        bento_metadata: "BentoServiceMetadata" = load_bento_service_metadata(saved_bundle_path)  # noqa: E501
-        # fmt: on
-
-        bento_tag = f'{bento_metadata.name}:{bento_metadata.version}'
         yatai_client: "YataiClient" = get_yatai_client(yatai_url)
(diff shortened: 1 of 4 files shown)


aarnphm:
Hi @mqk ty for the PR. We are currently working in the transition to 1.0 release, and thus most of the code are will be completely different than what we have in the master branch. We will still open the issues if the problem still comes up in the new release, but in the meantime let's just freeze the work for now. I'm hoping you will contribute again once the 1.0 is released.

mqk:
OK, no worries. Looking forward to 1.0.

PR #1842, opened 2021-09-10
feat: modelstore + new API design(wip)

aarnphm:
## Description

modelstore + examples for xgboost
Migrate metadata into `model_details.yaml`

image

## Notes

API redesign for frameworks will be defined in a separate PR.

## Checklist:


bentoml/xgboost.py
```diff
@@ -32,22 +31,32 @@
     )
 
 
-def _save_model(
+def save(
```
aarnphm, line 34: @parano we will enable context options to pass per framework implementation, but all of them can have the context of function call via `sys._getframe().f_back.f_code.co_filename`

bentoml/_internal/models/stores.py
```diff
@@ -85,14 +109,14 @@ def add(self, name: str, module: str, save_options: t.Dict[str, t.Any]):
...
-        ctx = StoreCtx(path=path, version=version)
-
-        yield ctx
+        # save bentoml_model.yml
+        model_yaml = Path(path, BENTOML_MODEL_YAML)
+        with model_yaml.open("w", encoding="utf-8") as f:
+            yaml.safe_dump(_gen_model_yaml(**yaml_kwargs), f)
 
-        # TODO save yaml file after model_yaml is ready
+        yield StoreCtx(path=path, version=version)
```
larme, line 117: I think we should either save yaml after yield or do a clean up (deleting the model directory) if the yield failed?

bentoml/xgboost.py
```diff
@@ -32,22 +31,32 @@
...
+        _version = ctx.version
+        _instance.save(ctx.path)
+    return f"{name}:{_version}"
+
+
+def load(name: str, **load_kwargs) -> "xgb.core.Booster":
+    model_info = _stores.get(name)
+    return _XgBoostModel.load(
+        model_info.path, **{**model_info.save_options, **load_kwargs}
+    )
```
larme, line 56: Should we check model_info's module is the same as current module?

bentoml/_internal/models/stores.py
```diff
@@ -67,12 +91,12 @@ def list_model(self, name: t.Optional[str] = None) -> t.List[str]:
         return [_f.name for _f in path.iterdir()]
 
     @contextmanager
-    def add(self, name: str, module: str, save_options: t.Dict[str, t.Any]):
+    def add_model(self, name: str, **yaml_kwargs) -> "StoreCtx":
```
parano, line 94:
```
def register_model(
   self,
   name: str,
   options: Dict[str, Any],
   metadata: Dict[str, Any],
   labels: Dict[str, str],
)
```

bentoml/_internal/models/store.py
```diff
@@ -119,35 +120,33 @@ def register_model(
...
+            path=model_path,
             version=model_tag.version,
             labels=labels,
             metadata=metadata,
             options=options,
         )
         try:
-            dump_model_yaml(model_yaml, ctx, framework_context=framework_context)
             yield ctx
+        except StopIteration:
```
parano, line 136: what's the reason for `StopIteration` here? maybe just `except Exception`?

bentoml/xgboost.py
```diff
@@ -30,34 +35,60 @@
         "pd", globals(), "pandas", exc_msg="Install pandas with `pip install pandas`"
     )
 
+LOAD_RETURNS_DOCS = """\
```
parano, line 38: why not just put this in the docstring?

bentoml/_internal/models/store.py
```diff
@@ -187,6 +186,24 @@ def export_model(self, name: str):
     def import_model(self, name: str):
         ...
 
+    def __getattribute__(self, item):
+        inherited = object.__getattribute__(self, item)
+        if "model" in item:
+
+            def wrapped_check_identifier(*args, **kwargs):
```
parano, line 193: let's just add a `validate_model_name` func call in get_model and register_model

bentoml/xgboost.py
```diff
@@ -30,34 +35,60 @@
...
-def load(name: str, **load_kwargs) -> "xgb.core.Booster":
-    model_info = _stores.get(name)
-    return _XgBoostModel.load(
-        model_info.path, **{**model_info.save_options, **load_kwargs}
-    )
-
-
-def load_runner(*args, **kw) -> "_XgBoostRunner":
-    ...
+def load_runner(*args, **kwargs) -> "_XgBoostRunner":
```
parano, line 92: this implementation needs to be updated as well right?

bentoml/xgboost.py
```diff
@@ -30,34 +35,60 @@
...
     ) as ctx:
-        _version = ctx.version
-        _instance.save(ctx.path)
-    return f"{name}:{_version}"
+        model.save_model(os.path.join(ctx.path, f"{SAVE_NAMESPACE}{JSON_EXT}"))
+    return f"{name}:{ctx.version}"
 
 
-def load(name: str, **load_kwargs) -> "xgb.core.Booster":
-    model_info = _stores.get(name)
```
parano: this actually looks right to me. And since `load` and `load_runner` are user-facing, we should put the detailed arguments list and docstring here

bentoml/_internal/models/store.py
```diff
@@ -187,6 +186,24 @@ def export_model(self, name: str):
     def import_model(self, name: str):
```
parano, line 186: on a side note, similar to the bento_store, the `delete_model` above will need to search for the `latest` based on file timestamp and recreate the symlink iff the model being deleted is the current "latest"

bentoml/xgboost.py
```diff
@@ -1,14 +1,19 @@
...
+    LOAD_INIT_DOCS,
+    SAVE_INIT_DOCS,
+    SAVE_NAMESPACE,
+    SAVE_RETURNS_DOCS,
+)
 from ._internal.service.runner import Runner
-from ._internal.types import GenericDictType, PathType
-from ._internal.utils import LazyLoader
+from ._internal.types import GenericDictType
+from ._internal.utils import LazyLoader, init_docstrings, returns_docstrings
```
parano, line 16: let's put init_docstrings, returns_docstrings in the `models` namespace?

bentoml/_internal/models/store.py
aarnphm: This is mainly for catching the ctx when saving is failed.

bentoml/xgboost.py
aarnphm: We can reuse it in many frameworks

bentoml/_internal/models/store.py
aarnphm: I was thinking to use it in every function since it will invoked the `name` args

bentoml/xgboost.py
aarnphm: yep

bentoml/xgboost.py
aarnphm: sounds good

aarnphm:
@larme modelstore is finished

PR #1847, opened 2021-09-16
chore: add runner

bojiang:
## Checklist:


bentoml/_internal/service/runner.py
```diff
@@ -1,24 +1,22 @@
+import abc
+
+from simple_di import Provide, inject
+from bentoml._internal.configuration.containers import BentoMLContainer
 import os
-import typing as t
```
aarnphm, line 2: We will use this since it is faster than explicitly import the types 😄

bentoml/_internal/service/runner.py
```diff
@@ -1,24 +1,22 @@
...
+from .utils import _cpu_converter, _mem_converter, _query_cgroup_cpu_count
 
 
 @attr.s
-class RunnerResourceLimits:
-    cpu = attr.ib(converter=_cpu_converter, type=float)
-    mem = attr.ib(converter=_mem_converter, type=int)
+class ResourceQuota:
+    cpu = attr.ib(converter=_cpu_converter, type=Optional[float])
+    ram = attr.ib(converter=_mem_converter, type=Optional[int])
```
aarnphm, line 17: I think `memory` make more sense for users

bentoml/_internal/service/runner.py
```diff
@@ -1,24 +1,22 @@
...
     gpu = attr.ib(type=float)
+    gram = attr.ib(converter=_mem_converter, type=Optional[int], default=None)
```
aarnphm, line 19: the convention is `VRAM` i believe (which is video ram :))

bentoml/_internal/service/runner.py
parano: let's make it `memory` to keep it consistent with kubernetes' resource quota API

bentoml/_internal/service/runner.py
parano: Let's just remove GPU memory here since we won't be able to use it anytime soon

parano:
I will address comments in follow up PR

PR #1886, opened 2021-10-13
refactor(framework): pycaret impl

dtemir:
### Implements PyCaret
Finished:
* save and load functions
* runner class

Next:
* add integration tests

## Motivation and Context

The framework is slow, but is a good tool for Data Scientists to train multiple models at once.

## How Has This Been Tested?

It has been tested with the examples provided in Docstrings. Integration tests will be added soon.

## Possible Issues

In `pycaret.py`, instead of using various `from pycaret.classification import *` and `from pycaret.regression import *`, I use their parent class `pycaret.internal.tabular` that defines all the default functions for those categories. 

However, it introduces problems that the user needs to provide **parameters to input when setting up the environment**
```python
# create parmeters
# note: change my_usercase according to the type of model
params = {"ml_usecase": "classification", "available_plots": {},
            "data": data, "target": "default", "session_id": 123}

# initialize runner
runner = bentoml.pycaret.load_runner(tag="my_model:latest", **params)

# set up the runner and pycaret environment
runner._setup()
```
 and some functions in classification, for example, might need different values from the ones listed by default in tabular.  

## Checklist:

### DIFF
--- a/tests/integration/frameworks/test_pycaret_impl.py
+++ b/tests/integration/frameworks/test_pycaret_impl.py
@@ -46,7 +46,7 @@ def pycaret_model(get_pycaret_data) -> t.Any:
     return final_dt
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture()
 def save_proc(
     pycaret_model,
     modelstore: "ModelStore",


aarnphm:
If you can run `make lint` that would be great

aarnphm:
NOTE(future reference): https://github.com/pycaret/pycaret#pycaret-on-gpu

PR #1916, opened 2021-10-22
feat(internal): add common store implementation

sauyon:
This very much needs tests, but I thought I'd put this up so I can get feedback anyway.


parano:
Merging now 🎉  let's address the remaining comments and model store changes in follow-up PRs.

PR #1930, opened 2021-10-27
internal(io): the rest of the IO implementation (except multipart)

aarnphm:
## Description

After discussion, we will remove Bytes IO supports for now, since it is pretty similar to File implementation

We will tackle multipart data on the following PR

## How Has This Been Tested?

Native testing, will add testcases for PR

## Checklist:


aarnphm:
@aarnphm multipart and tests will be addressed in consequent PR

bentoml/_internal/io_descriptors/json.py
```diff
@@ -63,10 +75,59 @@ def default(self, o: _SerializableObj) -> t.Any:
...
+            JSON encoder.
+    Returns:
+        IO Descriptor that in JSON format.
+    """
+
     def __init__(
         self,
         pydantic_model: t.Optional["pydantic.BaseModel"] = None,
         validate_json: bool = True,
+        orient: t.Literal[
```
parano, line 128: I'd suggest removing this parameter here,  it might be confusing to the user - if user is looking to control the format of a dataframe output, might be better to just use `PandasDataFrame` or convert df to json before returning the function

bentoml/_internal/io_descriptors/pandas.py
```diff
@@ -182,15 +179,144 @@ async def to_http_response(self, obj: "pd.DataFrame") -> Response:
...
+            dtype=None,  # TODO: not breaking atm
+        )
+
+
+_repl = {"PandasDataFrame": "PandasSeries", "DataFrame": "Series"}
+_rgx = re.compile("|".join(dict((re.escape(k), v) for k, v in _repl.items()).keys()))
+
+
+class PandasSeries(PandasDataFrame):
+    __doc__ = _rgx.sub(lambda m: _repl[re.escape(m.group(0))], PandasDataFrame.__doc__)
```
parano, line 254: i'd suggest copying over the docs here, especially we want to add some sample client HTTP requests and they will be different between these two

bentoml/_internal/io_descriptors/text.py
```diff
@@ -1,18 +1,70 @@
...
+        import bentoml.transformers
+
+        # If you don't have a gpt2 model previously saved under BentoML modelstore
+        # tag = bentoml.transformers.import_from_huggingface_hub('gpt2')
+        runner = bentoml.transformers.load_runner('gpt2',tasks='text-generation')
+
+        svc = bentoml.Service("gpt2-generation", runners=[runner])
+
+        @svc.api(input=Text(), output=Text())
+        def predict(input_arr):
```
parano, line 29: it would be great to add type annotation on the sample code here

bentoml/_internal/utils/dataframe.py
```diff
@@ -30,37 +30,37 @@ def check_dataframe_column_contains(
 def guess_orient(
     table: t.Union[t.List[t.Mapping[str, t.Any]], t.Dict[str, t.Any]],
     strict: bool = False,
-) -> t.Optional[t.Union[str, t.Set[str]]]:
+) -> t.Optional[t.Set[str]]:
```
parano, line 33: are the changes here used?

bentoml/transformers.py
```diff
@@ -394,6 +394,21 @@ def _internal_save(
     _check_flax_supported()  # pragma: no cover
```
parano, line 394: what's the reason for the change from `_save` to `_internal_save`?

bentoml/_internal/utils/dataframe.py
aarnphm: You can't assign type Union to a variable, so in some consequent function using `guess_orient` variable should just have one type instead of two. Mypy can't seem to detect it automatically

bentoml/transformers.py
aarnphm: just for consistency with paddlepaddle

bentoml/_internal/io_descriptors/json.py
aarnphm: noted

bentoml/_internal/io_descriptors/pandas.py
aarnphm: noted

PR #1950, opened 2021-11-02
fix: create model path if not exists when build tmp bento

larme:
## Checklist:

### DIFF
--- a/bentoml/_internal/bento/bento.py
+++ b/bentoml/_internal/bento/bento.py
@@ -71,6 +71,7 @@ class Bento:
 
             model_name, model_version = model_tag.split(":")
             target_path = os.path.join("models", model_name, model_version)
+            bento_fs.makedirs(target_path, recreate=True)
             copy_dir(fs.open_fs(model_info.path), "/", bento_fs, target_path)
 
         # Copy all files base on include and exclude, into `{svc.name}` directory


bentoml/_internal/bento/bento.py
```diff
@@ -71,6 +71,7 @@ def create(
 
             model_name, model_version = model_tag.split(":")
             target_path = os.path.join("models", model_name, model_version)
+            bento_fs.makedirs(target_path, recreate=True)
```
aarnphm, line 74: can we create a tmp repo path?

parano: i think bento_fs is already a tmp path `bento_fs = fs.open_fs(f"temp://bentoml_bento_{svc.name}")`

bentoml/_internal/bento/bento.py
```diff
@@ -71,6 +71,7 @@ def create(
...
             copy_dir(fs.open_fs(model_info.path), "/", bento_fs, target_path)
```
parano, line 75: for `copy_dir` here, do we want to `preserve_time=Ture` here? since we will be using that for finding the `latest` model?

parano: cc @sauyon

PR #1956, opened 2021-11-03
fix(internal): fixed bento store implementation

sauyon:


bentoml/bentos.py
```diff
@@ -85,56 +85,51 @@
...
     tag: t.Union[Tag, str],
-    _bento_store: "Store" = Provide[BentoMLContainer.bento_store],
+    _bento_store: "BentoStore" = Provide[BentoMLContainer.bento_store],
 ):
     _bento_store.delete(tag)
 
 
 def import_bento(path: str) -> Bento:
-    return Bento.import_from_fs(fs.open_fs(path))
+    return Bento.from_fs(fs.open_fs(path))
```
parano, line 125: probably for future PR, here we may need to choose the fs type based on the path or extenstion

parano (approved):

PR #1663, opened 2021-06-07
[docs] added project structures

aarnphm:
## Description

Add a minimal project structure for newcomers and others to follow as a convention for BentoML projects.

## Motivation and Context

- allow new users to see other usecases other than jupyter notebook.
- avoid error when new users try to put everything into one file, that by running scripts model loading and saving code are invoked again, when loading the BentoService class

## How Has This Been Tested?

N/A

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature and improvements (non-breaking change which adds/improves functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Code Refactoring (internal change which is not user facing)
- [x] Documentation
- [ ] Test, CI, or build

## Component(s) if applicable
- [ ] BentoService (service definition, dependency management, API input/output adapters)
- [ ] Model Artifact (model serialization, multi-framework support)
- [ ] Model Server (mico-batching, dockerisation, logging, OpenAPI, instruments)
- [ ] YataiService gRPC server (model registry, cloud deployment automation)
- [ ] YataiService web server (nodejs HTTP server and web UI)
- [ ] Internal (BentoML's own configuration, logging, utility, exception handling)
- [ ] BentoML CLI

## Checklist:

- [ ] My code follows the bentoml code style, both `./dev/format.sh` and
  `./dev/lint.sh` script have passed
  ([instructions](https://github.com/bentoml/BentoML/blob/master/DEVELOPMENT.md#style-check-and-auto-formatting-your-code)).
- [ ] My change reduces project test coverage and requires unit tests to be added
- [ ] I have added unit tests covering my code change
- [ ] My change requires a change to the documentation
- [x] I have updated the documentation accordingly

PR #1961, opened 2021-11-04
[docs] Fix custom metrics for prometheus (#1663)

ahallermed:
## Motivation and Context

Help the next reader to directly use custom metrics without research.

## How Has This Been Tested?

Tested on local machine, but it is a straight forward change.

## Checklist:


parano:
Hi @ahallermed - it is recommended to use what was suggested in the document, using the `prometheus_client` directly only works when there is one single worker, it does not work when there are multiple workers.

PR #1962, opened 2021-11-05
feat(BentoServer): implement --reload option

parano:
## Checklist:

### DIFF
--- a/bentoml/_internal/cli/bento_server.py
+++ b/bentoml/_internal/cli/bento_server.py
@@ -46,7 +46,7 @@ def add_serve_command(cli) -> None:
     @click.option(
         "--reload-delay",
         type=click.FLOAT,
-        help="Delay between each check if the Service needs to be reloaded. Default is 0.25s.",
+        help="Delay in seconds between each check if the Service needs to be reloaded. Default is 0.25.",
         show_default=True,
         default=0.25,
     )
--- a/bentoml/_internal/cli/bento_server.py
+++ b/bentoml/_internal/cli/bento_server.py
@@ -36,18 +36,34 @@ def add_serve_command(cli) -> None:
         show_default=True,
     )
     @click.option(
+        "--reload",
+        type=click.BOOL,
+        is_flag=True,
+        help="Reload Service when code changes detected, this is only available in development mode.",
+        default=False,
+        show_default=True,
+    )
+    @click.option(
+        "--reload-delay",
+        type=click.FLOAT,
+        help="Delay between each check if the Service needs to be reloaded. Default is 0.25s.",
+        show_default=True,
+        default=0.25,
+    )
+    @click.option(
         "--run-with-ngrok",  # legacy option
         "--ngrok",
         is_flag=True,
         default=False,
-        help="Use ngrok to relay traffic on a public endpoint to the local BentoServer",
-        envvar="BENTOML_ENABLE_NGROK",
+        help="Use ngrok to relay traffic on a public endpoint to the local BentoServer, this is only available in development mode.",
         show_default=True,
     )
     def serve(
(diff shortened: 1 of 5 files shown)


bentoml/_internal/server/__init__.py
```diff
@@ -71,20 +74,13 @@ def _start_ngrok_server() -> None:
 
 def serve_production(
     bento_path_or_tag: str,
-    working_dir: Optional[str] = None,
-    port: Optional[int] = None,
+    working_dir: t.Optional[str] = None,
+    app_workers: t.Optional[int] = None,
```
parano, line 78: not related to this PR, I was thinking about how should we expose the `workers` option to end-user and thought it  would be helpful to name them a bit more specific, `api_worders` and `runner_workers`

bentoml/_internal/server/__init__.py
```diff
@@ -71,20 +74,13 @@ def _start_ngrok_server() -> None:
...
+    runner_workers: t.Optional[int] = None,
+    port: t.Optional[int] = None,
 ) -> None:
-    if working_dir is not None:
-        working_dir = os.path.abspath(working_dir)
-
     svc = load(bento_path_or_tag, working_dir=working_dir)
 
-    import psutil
```
parano, line 82: @bojiang we no longer need to limit `serve --production` to POSIX users right?  I think uvicorn have good support for windows

bentoml/_internal/server/__init__.py
parano: on a related note, `workers` should be limited to 1 for development mode

bentoml/_internal/server/__init__.py
```diff
@@ -32,7 +35,7 @@ def serve_development(
     for _, runner in svc._runners.items():
         runner._setup()
 
-    if with_ngroxy:
```
parano, line 35: also changed from `ngroxy` to `ngrok` cc @bojiang  I think the latter is the right name?

bojiang: whops where is `ngroxy ` from lol

bentoml/_internal/server/__init__.py
bojiang: Theoretically, yes. I plan to reorganize windows serving when other parts' ready

bentoml/_internal/cli/bento_server.py
```diff
@@ -35,19 +35,35 @@ def add_serve_command(cli) -> None:
...
+        type=click.BOOL,
+        is_flag=True,
+        help="Reload Service when code changes detected, this is only available in development mode.",
+        default=False,
+        show_default=True,
+    )
+    @click.option(
+        "--reload-delay",
+        type=click.FLOAT,
+        help="Delay between each check if the Service needs to be reloaded. Default is 0.25s.",
```
bojiang, line 49:
```suggestion
        help="Delay in seconds between each check if the Service needs to be reloaded. Default is 0.25.",
```

bentoml/_internal/server/__init__.py
bojiang: IMO we should only expose this in the configuration file, allows users to config workers(or quota) for api server and each runner.

bentoml/_internal/service/service.py
```diff
@@ -15,7 +15,7 @@
     from starlette.middleware import Middleware
     from starlette.types import ASGIApp
 
-WSGI_APP = t.NewType("WSGI_APP", object)
+WSGI_APP = t.Callable[[t.Callable, t.Mapping[str, t.Any]], t.Iterable[bytes]]
```
bojiang, line 18: 👍 a typing_extension.Protocol is a little better here i guess.

PR #1966, opened 2021-11-05
fix: fix flake8 errors

sauyon:


tests/integration/frameworks/paddle/test_paddle_runner.py
```diff
@@ -5,10 +5,11 @@
 
 import bentoml.paddle
 
-from .test_paddle_impl import test_df, train_paddle_model
+# for review: ???
+from .test_paddle_impl import test_df, train_paddle_model  # noqa: F401
```
sauyon, line 9: Why do these model imports exist?

tests/integration/frameworks/test_pycaret_impl.py
```diff
@@ -38,7 +38,8 @@ def get_pycaret_data():
 def pycaret_model(get_pycaret_data) -> t.Any:
     # note: silent must be set to True to avoid the confirmation input of data types
     train_data, _ = get_pycaret_data
-    env = pycaret_setup(data=train_data, target="default", session_id=123, silent=True)
+    # for review
+    pycaret_setup(data=train_data, target="default", session_id=123, silent=True)
```
sauyon, line 42: Is this change ok?

tests/utils/frameworks/tensorflow_utils.py
```diff
@@ -12,7 +12,8 @@ def __init__(self, units=32, **kwargs):
         super(CustomLayer, self).__init__(**kwargs)
         self.units = tf.Variable(units, name="units")
 
-    def call(self, inputs, training=False, **kwargs):
+    # for review
+    def call(self, inputs, training=False):
```
sauyon, line 16: Are the `kwargs` for call compatibility? Should I add a pylint ignore here instead?

sauyon:
I actually did a bit of rewording and grammar fixes for documentation but gave up halfway through since I realized some of it may not end up actually used.

Also, documentation reflowing may have caused some format issues in the output; I'm assuming for now that's not a big deal.

docker/manager.py
```diff
@@ -712,9 +712,10 @@ class attributes:
 
         - _push_context (:obj:`Dict`):
 
-            contains mapping for docker.Images object from docker-py with correct releases type
+            contains mapping for docker.Images object from docker-py with correct
+            releases type
 
-    """
+    """  # noqa: LN001
```
sauyon, line 718: Is this file currently used? It seems like it doesn't even evaluate without errors.

aarnphm: we are using everything related under docker for creating releases for our docker image.

tests/integration/frameworks/paddle/test_paddle_runner.py
aarnphm: i don't like having multiple conftest under smaller directory. It is just a way to write test quickly once I created a model.

sauyon: Oh, these are test fixtures, aren't they.

aarnphm: yes

sauyon: Hm, seems like the recommended way to make these lint is to add a `conftest.py`. Is there any particular reason you don't like having more than one conftest?

The alternative seems to be that or what I've done here which seems a bit ugly too.

docker/manager.py
sauyon: Hm, do I have the wrong docker version or something? Flake8 [complains](https://github.com/bentoml/BentoML/runs/4115558301?check_suite_focus=true) that `docker.DockerClient` and `docker.errors` don't exist

parano: @sauyon it seems to work with the latest version of docker python library 5.0.3, but for version 4.x, `docker.DockerClient` and `docker.errors` don't exist

tests/utils/frameworks/tensorflow_utils.py
parano: it should be ok to remove `**kwargs` here

parano:
@sauyon merging now to avoid more conflicts in other dev branches, let's discuss those questions and address them in follow-up if needed

docker/manager.py
aarnphm: Could be because pylint detect the docker at our top level modules instead of the actual docker library.

parano: @aarnphm yes I got rid of that error by removing the `__init__.py` file under the `docker` directory

PR #1969, opened 2021-11-05
feat(internal): change the format of generated versions

sauyon:
This is somewhat a proposal, but I've changed the version format to `YYYYMMDD.<stamp>`, where `<stamp>` is a custom base64 encode of the unix timestamp in seconds.

This way, versions are lexically sortable if we ever decide that they should be. This is currently only important because the store implementations are sorting by tag when listing, which may not be the "correct" order for the user.

There's definitely an argument to be made that the store implementation should also sort by item timestamp too.

### DIFF
--- a/bentoml/_internal/bento/bento.py
+++ b/bentoml/_internal/bento/bento.py
@@ -53,14 +53,16 @@ class Bento(StoreItem):
         build_ctx: PathType,
         models: t.List[str],
         version: t.Optional[str],
-        description: t.Optional[str],
+        description: str,
         include: t.List[str],
         exclude: t.List[str],
-        env: t.Optional[t.Dict[str, t.Any]],
-        labels: t.Optional[t.Dict[str, str]],
+        env: t.Dict[str, t.Any],
+        labels: t.Dict[str, str],
         model_store: "ModelStore" = Provide[BentoMLContainer.model_store],
     ) -> "Bento":
         tag = Tag(svc.name, version)
+        if version is None:
+            tag = tag.make_new_version()
 
         logger.debug(f"Building BentoML service {tag} from build context {build_ctx}")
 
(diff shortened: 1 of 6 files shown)


parano:
Great idea, my main concern is version format like '20201106.--0WWnml', '20201106.__0WWnn-', '20201106.__0WWnnm'  might not be very recognizable for users looking at a few models/bentos they created during development time, also for organizations that have multiple pipelines creating many models, it's more likely to run into collision. Perhaps we can remove the `.` add an underscore and 4 digit hash to the end? e.g.  '20201106__0WWnml_BC3A'.

Another concern is - this still does not solve the problem of timezone. Users may have pipelines running in a different timezone pushing model/bento to Yatai and when they pull the model to the local machine, the sorting may not work as expected in some cases.

sauyon:
> Great idea, my main concern is version format like '20201106.--0WWnml', '20201106.__0WWnn-', '20201106.__0WWnnm' might not be very recognizable for users looking at a few models/bentos they created during development time, also for organizations that have multiple pipelines creating many models, it's more likely to run into collision. Perhaps we can remove the `.` add an underscore and 4 digit hash to the end? e.g. '20201106__0WWnml_BC3A'.

Something I was thinking about doing was just implementing a custom base-62 instead of 64, which would solve the issue of having `-` and `_` as additional characters to make the round 64. It would make the conversion a little harder, but as long as we don't care about this format being really readable to anybody it would enable just `0-0a-zA-Z`, which would much nicer.

> Another concern is - this still does not solve the problem of timezone. Users may have pipelines running in a different timezone pushing model/bento to Yatai and when they pull the model to the local machine, the sorting may not work as expected in some cases.

Yeah. I was thinking about prefixing with the timestamp but that didn't seem very pretty.

sauyon:
I ended up using the entire UUID and therefore generating 25 character versions; I can probably trim parts of it if you want them to be shorter.

parano:
@sauyon all look great! there is one linter error on unused import but otherwise good to go

PR #1970, opened 2021-11-06
feat: finish multipart io

bojiang:
## Checklist:


bentoml/_internal/io_descriptors/multipart.py
```diff
@@ -27,7 +29,7 @@ def predict(image: "PIL.Image.Image", annotations: t.Dict[str, t.Any]) -> t.Dict
     curl -i -F image=@test.jpg -F annotations=@test.json localhost:5000/predict
     """
 
-    def __init__(self, **inputs: t.Dict[str, IODescriptor]):
+    def __init__(self, **inputs: IODescriptor):
```
parano, line 32: do you mean inputs: MultiPartIO here?

aarnphm: I think typehint `**` can't take a `t.Dict`

bentoml/_internal/io_descriptors/multipart.py
```diff
@@ -55,44 +59,26 @@ def openapi_responses_schema(self):
...
-        :param obj:
-        :type obj: MultipartIO
-        :rtype: Response
-        """
-        pass
+        for k, i in self._inputs.items():
+            req = reqs[k]
+            v = await i.from_http_request(req)
+            res[k] = v
+        print(res)
```
parano, line 75: remove?

aarnphm: yep, I'm fixing the return type

bentoml/_internal/utils/_formparser.py
```diff
@@ -0,0 +1,195 @@
...
+        ori_headers = t.cast(t.Dict[bytes, bytes], ori_headers)
+        ori_headers.update(dict(headers))
+        scope["headers"] = list(ori_headers.items())
+        req = Request(scope)
+        req._body = data
+        reqs[field_name] = req
+    return reqs
+
+
+class _MultiResponse(Response):
```
aarnphm, line 133: @bojiang

bentoml/_internal/io_descriptors/image.py
```diff
@@ -21,11 +19,10 @@
 
     # NOTE: pillow-simd only benefits users who want to do preprocessing
     # TODO: add options for users to choose between simd and native mode
-    _exc = const.IMPORT_ERROR_MSG.format(
```
parano, line 24: 👍 we should probably get rid of the IMPORT_ERROR_MSG constant, an inline string like this makes it more readable

bentoml/_internal/io_descriptors/multipart.py
parano: looks like `**` just assumes all the args are the same type, so `IODescriptor` here is correct

bentoml/_internal/runner/runner.py
```diff
@@ -281,6 +280,7 @@ def run_batch(self, *args, **kwargs):
         if self._state is RunnerState.INIT:
             self._setup()
         if isinstance(self._runner, Runner):
+            print(args, kwargs)
```
parano, line 283: remove?

bentoml/_internal/server/service_app.py
```diff
@@ -289,7 +297,7 @@ async def api_func(
                     )
                 else:
                     response = JSONResponse("", status_code=e.error_code)
-            except Exception:  # pylint: disable=broad-except
+            except _StopIteration:
```
parano, line 300: What's the reason for this change?

bentoml/_internal/tracing/jaeger.py
```diff
@@ -2,7 +2,7 @@
 from contextlib import contextmanager
 from contextvars import ContextVar
 
-from bentoml._internal.tracing import Tracer
+from ..tracing import Tracer
```
parano, line 5: this can be just `from . import Tracer`?

bentoml/_internal/server/service_app.py
aarnphm: remove the pylint error

bentoml/_internal/tracing/jaeger.py
aarnphm: will fix it

bentoml/_internal/io_descriptors/image.py
aarnphm: we already removed the import_error_msg

bentoml/_internal/utils/_formparser.py
```diff
@@ -0,0 +1,180 @@
...
+
+from ...exceptions import BentoMLException
+
+_ItemsBody = t.TypeVar(
+    "_ItemsBody",
+    bound=t.List[t.Tuple[str, t.List[t.Tuple[bytes, bytes]], bytes]],
+)
+
+
+class MultiPartParser(_StarletteMultiPartParser):
```
bojiang, line 30: Relying on external codes is not recommended since they may change without warning. copy/paste is much better.

bentoml/_internal/utils/_formparser.py
```diff
@@ -0,0 +1,180 @@
...
+                    #         content_type = content_type.decode('latin-1')
+                    #     )
+                    # else:
+                    #     file = None
+                elif message_type == MultiPartMessage.PART_DATA:
+                    data += message_bytes
+                    # if not file:
+                    #     data += message_bytes
+                    # else:
+                    #     await file.write(message_bytes)
```
bojiang: why does we keep these lines?

bentoml/_internal/utils/_formparser.py
```diff
@@ -0,0 +1,180 @@
...
+        req = Request(scope)
+        req._body = data
+        reqs[field_name] = req
+    return reqs
+
+
+async def concat_multipart_responses(
+    responses: t.List[t.Tuple[str, t.Union[Response, StreamingResponse]]]
+) -> Response:
+    boundary = binascii.hexlify(os.urandom(16)).decode("ascii")
```
bojiang: i think we can use multipart utils to do this rather than concat manually.

bentoml/_internal/utils/_formparser.py
```diff
@@ -0,0 +1,180 @@
+import binascii
```
bojiang, line 1: why the renaming?

bentoml/_internal/server/service_app.py
bojiang: we want to catch all the exceptions caused by the user function. It's necessary.

parano: @aarnphm let's change this back to `Exception`

bentoml/_internal/utils/_formparser.py
parano: there's no need to use leading `_` for files like this right? it's already under `_internal`

bentoml/_internal/server/service_app.py
aarnphm: ok

bentoml/_internal/utils/_formparser.py
aarnphm: you are correct

bentoml/_internal/io_descriptors/image.py
```diff
@@ -1,11 +1,11 @@
 import io
+import mimetypes
 import typing as t
 
+import imageio
```
parano, line 5: `imageio` is no longer a dependency, is it possible to remove it? ideally we want to use only `pillow`, which is imported with lazyloader below. cc @aarnphm @bojiang

parano: this is breaking the documentation checker rn

aarnphm: I will update it to Pillow

bojiang: oh never expected that. yeah these commits is for quick testing, not well formated for dependencies

PR #1977, opened 2021-11-09
chore(internal): fix linting

sauyon:
@aarnphm I've deleted the ONNX tests because it doesn't seem like it works.


sauyon:
Mypy was actually complaining that the `type: ignore` lines were unused in the `pandas` file, so I just deleted them. We should revisit once we get mypy actually passing.

aarnphm:
@sauyon I'm fixing onnx atm

PR #1979, opened 2021-11-09
feat(bento): bento build part i

parano:
## Checklist:


bentoml/_internal/service/loader.py
```diff
@@ -156,19 +156,16 @@ def load_bento(
         load_bento("FraudDetector:latest")
         load_bento("FraudDetector:20210709_DE14C9")
     """
-    bento_info = bento_store.get(bento_tag)
-    bento_config = yaml.load(os.path.join(bento_info.path, "bento.yml"))
-    # TODO: verify bento_config.bentoml_version, bentoml_version
+    bento = bento_store.get(bento_tag)
```
parano, line 159: cc @sauyon

sauyon (commented):
Probably a bit early for this thorough of a review, but oh well...

bentoml/_internal/bento/docker.py
```diff
@@ -68,78 +70,78 @@ class ImageProvider(object):
...
 
     def __init__(
         self,
-        distro: str,
+        distro: t.Optional[
+            Literal["slim", "ami2", "alpine", "centos7", "centos8"]
+        ] = None,
         python_version: t.Optional[str] = None,
-        gpu: bool = False,
+        gpu: t.Optional[bool] = None,
```
aarnphm, line 92: this can just be 
```python
gpu: bool = False
```
If users want to use GPU they can then specify `gpu=True`

bentoml/_internal/bento/docker.py
```diff
@@ -68,78 +70,78 @@ class ImageProvider(object):
...
+                gpu=True
+            )
         )
     )
     """
 
     def __init__(
         self,
-        distro: str,
+        distro: t.Optional[
```
aarnphm, line 88: same here
```python
distro: Literal[...]='slim'
```

bentoml/_internal/bento/docker.py
```diff
@@ -68,78 +70,78 @@ class ImageProvider(object):
...
+            Literal["slim", "ami2", "alpine", "centos7", "centos8"]
```
aarnphm, line 89:
```suggestion
```

bentoml/_internal/bento/docker.py
```diff
@@ -3,6 +3,8 @@
 import sys
 import typing as t
 
+from typing_extensions import Literal
+
 from ...exceptions import BentoMLException
 from ..configuration import BENTOML_VERSION, is_pip_installed_bentoml
```
aarnphm, line 10: also I changed `ami2` to `amazonlinux2`

bentoml/_internal/bento/docker.py
parano: why removing it here?

bentoml/_internal/bento/docker.py
parano: sounds good

bentoml/_internal/bento/bento.py
```diff
@@ -50,24 +58,28 @@ def create(
         bento_fs = fs.open_fs(f"temp://bentoml_bento_{svc.name}")
         ctx_fs = fs.open_fs(build_ctx)
 
-        # Copy required models from local modelstore, into
-        # `models/{model_name}/{model_version}` directory
+        # Add Runner required models if models is not specified
```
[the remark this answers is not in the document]
parano: I think it should still run even if models is not Nont, I will update the comment here

bentoml/_internal/bento/bento.py
```diff
@@ -159,9 +183,63 @@ def push(self):
         pass
 
     def validate(self):
-        return self.fs.isfile("bento.yaml")
+        return self.fs.isfile(BENTO_YAML_FILENAME)
```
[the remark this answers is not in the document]
parano: Probably just put out a warning if `bentoml.__version__ != self.metadata.bentoml_version`, and only raise an error if we break backward compatibility for certain versions

bentoml/_internal/types.py
```diff
@@ -33,15 +33,16 @@ class Tag:
...
+    def __init__(self, name: str, version: t.Optional[str] = None):
         lname = name.lower()
         if name != lname:
             logger.warning(f"converting {name} to lowercase: {lname}")
 
         validate_tag_str(lname)
 
-        if version is not None:
-            validate_tag_str(version)
+        # TODO: fix version validation - allow version str to start with numbers?
```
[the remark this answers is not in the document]
parano: yea I ran into an issue, it raises an error with the old generated version that starts with `2021...`

bentoml/_internal/bento/docker.py
parano: Actually, this is because we want to allow upstream to call it with `gpu=None`.

Let's say the user has the following code:
```
svc.set_build_options( 
  env = dict(
    docker = dict(
        base_image = ...
    )
  )
)
```

We want to distinguish if the user has provided other options like `distro="slim"` at the same time, and provide the user a proper warning. But if we have `gpu=False, distro='slim'` as the default, we won't be able to tell if it's user-specified or if it's just the default.

aarnphm: noted

bentoml/_internal/bento/env.py
```diff
@@ -0,0 +1,139 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+@contextmanager
+def cwd(path):
+    oldpwd = os.getcwd()
+    os.chdir(path)
+    try:
+        yield
```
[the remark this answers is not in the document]
parano: I guess if `os.chdir` call failed, there is no need to change it back in finally?

bentoml/_internal/bento/bento.py
```diff
@@ -50,24 +58,28 @@ def create(
...
         ctx_fs = fs.open_fs(build_ctx)
 
-        # Copy required models from local modelstore, into
-        # `models/{model_name}/{model_version}` directory
+        # Add Runner required models if models is not specified
         for runner in svc._runners.values():
             models += runner.required_models
 
-        # TODO: remove duplicates in the models list
+        models_list = []
```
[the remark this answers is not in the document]
parano: Probably after the modelstore change?  right now we have `ModelInfo` instance here, and they will be changed to `StoreItem` soon

bentoml/_internal/bento/bento.py
```diff
@@ -111,33 +123,45 @@ def create(
...
         # Create `readme.md` file
-        with bento_fs.open("readme.md", "w") as f:
+        with bento_fs.open("README.md", "w") as f:
             f.write(description)
 
         # Create 'apis/openapi.yaml' file
         bento_fs.makedir("apis")
         with bento_fs.open(fs.path.combine("apis", "openapi.yaml"), "w") as f:
+            # Add representer to allow dumping OrderedDict as a regular map
+            yaml.add_representer(
```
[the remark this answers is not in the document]
parano: moved up to top of this module

bentoml/_internal/bento/bento.py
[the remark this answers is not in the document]
parano: We should definitely do that in the `Bento#from_fs` call, will add a TODO note for now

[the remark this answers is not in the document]
parano: in the older version of BentoML, we check base on the semver of BentoML itself, and it will only raise an error if the major version digit is different. I think we can probably continue that strategy

bentoml/_internal/types.py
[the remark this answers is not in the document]
parano: actually, I think it's because the old generated version has uppercase letters

bentoml/_internal/io_descriptors/image.py
```diff
@@ -85,6 +85,7 @@ def predict(input_img):
     """
```
parano, line 85: cc @bojiang @aarnphm

parano:
@bojiang could you help take a look here? this test still times out waiting for the API server docker to come up: https://github.com/bentoml/BentoML/runs/4152318798?check_suite_focus=true

.github/workflows/frameworks-on-pr.yml
```diff
@@ -564,24 +568,25 @@ jobs:
...
-        uses: actions/setup-python@v2
-        with:
-          python-version: 3.8
-      - name: Install test dependencies
-        run: ./ci/install_dependencies.sh.cmd
-      - name: Run tests
-        run: ./ci/integration/frameworks/run_spacy_tests.sh
-      - name: Upload test coverage to Codecov
-        uses: codecov/codecov-action@v1.0.12
+#  TODO: fix me
```
aarnphm, line 571: @parano why do we disable spacy tests? I think it passed right?

parano: @aarnphm I did that by mistake, I will uncomment this in a follow up PR

aarnphm: 👍🏼

PR #1987, opened 2021-11-10
refactor: pytorch multiargs better support and partial_kwargs impl

larme:
This is more like a proposal for adding partial_kwargs to pytorch.

The main motivation is that pytorch's `forward` is quite flexible and may accept arbitrary arguments. For a runner, sometimes all input share the same value for a particular argument (most likely a configuration parameter) and in this case it's better to specify it at load_runner stage.

Will add docs and tests later.

## Checklist:


bentoml/pytorch.py
```diff
@@ -230,28 +232,33 @@ def _setup(self) -> None:  # type: ignore[override]
     @torch.no_grad()
     def _run_batch(  # type: ignore[override]
```
aarnphm, line 233: I don't think we need this `# type: ignore` anymore
```suggestion
    def _run_batch(
```

bentoml/pytorch.py
```diff
@@ -230,28 +232,33 @@ def _setup(self) -> None:  # type: ignore[override]
...
+                if isinstance(item, torch.Tensor):
+                    item = item.cuda()
 
-        res: "torch.Tensor"
+            return item
+
+        params = params.map(_mapping)
+
+        # merge kwargs with partial_kwargs
+        merged_kwargs = {**self._partial_kwargs, **kwargs}
```
aarnphm, line 254: any updates from our previous chat? @larme

larme: @aarnphm done, just wanted to finish the test then do this so the refactoring is tested~

bentoml/pytorch.py
larme: let me correct this part for all frameworks in another commit

bentoml/pytorch.py
aarnphm: i saw the push, well done 👍🏼

PR #1989, opened 2021-11-11
tests: add pyright ci check

sauyon:
Using pyright because `mypy` is unbearably slow and doesn't seem to limit itself to the files passed on the CLI.

### DIFF
--- a/ci/format_and_lint.sh
+++ b/ci/format_and_lint.sh
@@ -7,17 +7,24 @@ cd "$GIT_ROOT" || exit
 has_errors=0
 
 # Code auto formatting check with black & isort
-./dev/format.sh
-GIT_STATUS="$(git status --porcelain)"
-if [ "$GIT_STATUS" ];
-then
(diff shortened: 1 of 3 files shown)


sauyon:
Fixed the fact this doesn't actually work...

sauyon:
Seems like this works now.

ci/format_and_lint.sh
```diff
@@ -7,17 +7,24 @@ cd "$GIT_ROOT" || exit
...
+if ! (black --check --config "$GIT_ROOT/pyproject.toml" tests docker); then
+  has_errors=1
+fi
+
+echo "Running isort on bentoml module..."
+if ! (isort --check bentoml); then
+  has_errors=1
+fi
+
+echo "Running isort on tests and docker modules..."
```
parano, line 25: cc @yubozhao @jjmachan could you guys add this for the bcdt repo?

PR #1990, opened 2021-11-11
fix(internal): remove unnecessary usage of OrderedDict

sauyon:


sauyon:
I rebased this over #1989 to make sure that works.

PR #1992, opened 2021-11-11
fix(internal): store fixes and basic tests

sauyon:


bentoml/_internal/types.py
```diff
@@ -59,11 +61,20 @@ def __str__(self):
         else:
             return f"{self.name}:{self.version}"
```
parano, line 62: @sauyon  should we do `f"{self.name}:{self.version[:6]}"` here?

sauyon: I think the python api should surface the whole version, and the cli should do the work of truncating. It should also probably check to make sure 6 characters is enough for uniqueness when displaying just in case too.

PR #1994, opened 2021-11-12
feat(ci): `bentoml/checker` docker images

aarnphm:

We should have a docker images that contains all of the tools we want to use for code quality, and use it on both CI and locally to address the need to change both `./dev/lint.sh` and `./ci/format_and_lint.sh`.

Also I refactor CI and Makefile to make it consistent with `bentoml/yatai` when we open source it :)

I also refactor the frameworks into one file now to remove duplication.

## Checklist:


Makefile
```diff
@@ -1,35 +1,60 @@
 .DEFAULT_GOAL := help
 
+CHECKER_IMG := bentoml/checker:1.0
+BASE_ARGS := -u root -v $(PWD):/bentoml
```
parano, line 4: also add `--rm` here?

scripts/ci/style/pyright_check.sh
```diff
@@ -0,0 +1,26 @@
...
+    exit 1
+  fi
+else
+  echo "Running pyright..."
+  pyright . | tee /tmp/"$name"_full
+  if [ -s /tmp/"$name"_full ]; then
+    FAIL "pyright failed."
+    cat /tmp/"$name"_full
+    exit 1
+fi
```
sauyon, line 23: pyright seems to output a bunch of stuff even when the lint passes, so this doesn't seem quite right to me; I'd think using exit code in general is cleaner than parsing output, too.

Also, since you're `tee`ing and not redirecting above isn't the `cat` unnecessary?

aarnphm: ye true

PR #1998, opened 2021-11-12
chore(internal): fix pyright type checking

sauyon:
I've now disabled pyright strict checking for `reportUnknownMemberType` because third-party stubs seem to be quite bad about fully defining all types, and I've spent far too much time trying to fix type stubs already.


sauyon:
The type stubs aren't complete, and honestly I've probably filled in a few too many `t.Any`s, but oh well...

bentoml/_internal/service/service.py
```diff
@@ -144,15 +155,12 @@ def openapi_doc(self):
...
+    def set_build_options(self, **build_options):  # type: ignore
         ...
 
     def __str__(self):
         if self.bento:
-            return (
-                f'bentoml.Service(tag="{self.tag}", '
-                f'path="{self.bento.fs.getsyspath("/")}")'
-            )
+            return f'bentoml.Service(tag="{self.tag}", ' f'path="{self.bento.path}")'
```
parano, line 163: 👍

aarnphm:
we should call it `typing_extensions` instead of `typings`? I think we can do something similar to what python is doing.

Another quick question: can this stub be used with mypy as well?

sauyon:
> we should call it `typing_extensions` instead of `typings`? I think we can do something similar to what python is doing.

It's just the default `pyright` stub path. I don't think any of the other tools have defaults

bentoml/_internal/bento/bento.py
```diff
@@ -131,7 +137,10 @@ def create(
         # Create `readme.md` file
         with bento_fs.open("README.md", "w") as f:
             description = description if description else svc.__doc__
-            f.write(description)
+            if description is None:
+                logger.warning(f"no description found while creating Bento {tag}")
```
parano, line 141: there's no need to warn user about this, we can just do a `if description is not None` check here

parano:
Somehow this PR breaks the CI's install test dependencies step?

poetry.lock
```diff
@@ -0,0 +1,3755 @@
+[[package]]
```
parano, line 1: is this meant to be committed here?

bentoml/_internal/bento/bento.py
```diff
@@ -170,45 +183,45 @@ def save(
...
+    def dump(self, stream: t.IO[str]):
         return yaml.dump(self.data, stream, sort_keys=False)
 
     @classmethod
     def from_build_args(
         cls, tag: Tag, svc: "Service", labels: t.Dict[str, t.Any], models: t.List[str]
     ):
         bento_metadata = cls()
-        bento_metadata["service"] = svc._import_str
+        bento_metadata["service"] = svc._import_str  # type: ignore[reportPrivateUsage]
```
parano, line 217: is it possible to do this ignore for everything in  the `Service` class?

bentoml/_internal/models/store.py
```diff
@@ -61,7 +61,7 @@ def _process_model_tag(tag: str) -> t.Tuple[str, str]:
 class StoreCtx(object):
     name = attr.ib(type=str)
     version = attr.ib(type=str)
-    path = attr.ib(type=PathType)
+    path = attr.ib(type=str)
```
parano, line 64: it seems to me there's no real need for the PathType, should we remove that?

bentoml/_internal/bento/bento.py
aarnphm: I think the better way to do this is to have a property to access the protected attributes.
```python
@property
def import_str(self):
	return self._import_str
```

sauyon: > is it possible to do this ignore for everything in  the `Service` class? 

Not that I know of; I'll take a look.

bentoml/_internal/models/store.py
sauyon: I think it's still valuable to allow `PathType` in user-facing APIs, but we can normalize to string internally.

poetry.lock
sauyon: No, I'll delete it.

bentoml/_internal/bento/bento.py
sauyon: > I think the better way to do this is to have a property to access the protected attributes.
> 
> ```python
> @property
> def import_str(self):
> 	return self._import_str
> ```

Hm I thought I replied to this, but the reason I didn't do this was because that would expose these to the user, and I don't know that that's what we want.

parano: that's right, this is not supposed to be exposed to the user and only meant to be used within BentoML, same for `_apis` and `_runners`.

aarnphm: would a function like
```python
def privateattr(cls_, attr: str):
	return getattr(cls, f"_{attr}")
```
fix our issue?
I'm fine with writing the `type: ignore` but it will make our codebase very tiresome to look into :(

aarnphm:
seems like https://github.com/bentoml/BentoML/runs/4207474709?check_suite_focus=true

bentoml/_internal/bento/bento.py
sauyon: We should probably just have a service subclass that we can cast into when we use them internally that surfaces the attributes.

sauyon:
> seems like https://github.com/bentoml/BentoML/runs/4207474709?check_suite_focus=true

Ugh. It seems like `PathLike` was changed in 3.9 to be a generic, before which it wasn't. I'll wrap it in a version check.

bentoml/_internal/bento/bento.py
sauyon: I'd say let's draft this in another PR, this is large enough as it is.

aarnphm: noted

aarnphm:
hmm the carryforward flags doesn't seem to work. I will fix it in a different PR

sauyon:
It was another instance of the same issue; hopefully there isn't another lol

aarnphm:
If possible I think we should merge this in order not to block others workflow, sincce I think you touched a lot of files in this PR

bentoml/_internal/bento/bento.py
```diff
@@ -32,10 +32,14 @@
 BENTO_YAML_FILENAME = "bento.yaml"
 
 
-@attr.define(repr=False)
+@dataclass
```
bojiang, line 35: @sauyon hi sauyon. just curious why we changed to dataclass here?

sauyon: Sorry, I missed this---I'll change it back, it's because version 20 of attrs doesn't support type checking; I'm going to just force attrs 21+ for dev-dependencies so that everything type-checks nicely.

PR #2000, opened 2021-11-13
feat(openapi): text, multipart, file, image, numpy

aarnphm:
## Description

Per usual I got a very bad habit of making large PR that fix alot of code :)

Text, File, Image, Numpy OpenAPI schema is defined. Multipart is partially done. 

Will add tests and Pandas schema after.

Fixed some CI related tasks, added typestubs and create a separate repository for wip stubs

**TODO**:
- [ ] `Multipart` to accept components' IODescriptor `openapi_schema`
- [ ] added tests for e2e and unit
- [ ] add docs for `Multipart`
- [ ] fix type compatibility for `mypy` checker


bentoml/_internal/io_descriptors/json.py
```diff
@@ -135,16 +113,6 @@ def __init__(
         json_encoder: t.Type[json.JSONEncoder] = DefaultJsonEncoder,
     ):
         if pydantic_model is not None:
-            if pydantic is None:  # pragma: no cover
```
parano, line 138: can we add this check back?  this warns the user about missing dependency immediately when the API server boot up, not until it receives a request

aarnphm: done

pyproject.toml
```diff
@@ -38,9 +46,8 @@ starlette = "*"
 python-multipart = "*"
 rich = "*"
 pyyaml = "*"
-attrs = ">=20"
 cattrs = "*"
-importlib_metadata = "*"
+attrs = ">=21"
```
sauyon, line 50: This should be `>=21.1` for typing to work, and I think leaving things as they were is fine?

pyproject.toml
```diff
@@ -53,18 +60,35 @@ opentracing = {version = "*", optional = true}
...
+types-chardet = {version = "*", optional = true}
+types-setuptools = {version = "*", optional = true}
+types-PyYAML = {version = "*", optional = true}
+pyspark-stubs = {version = "*", optional = true}
+types-filelock = {version = "*", optional = true}
+
+[tool.poetry.extras]
+model-server = ["prometheus-client", "jaeger-client", "py-zipkin", "opentracing"]
+docs = ["recommonmark", "sphinx", "sphinx-click", "sphinx_rtd_theme", "sphinxcontrib-fulltoc", "sphinxcontrib-spelling", "sphinx_copybutton", "pyenchant"]
+types = ["types-click", "types-chardet", "types-setuptools", "pyspark-stubs", "types-filelock", "types-PyYAML"]
```
sauyon, line 81: I think types are pretty essential to the dev env and should be included by default.

sauyon (commented):
Few comments, and in general should we try to keep history sane? Especially if we're going to squash-and-merge I think there should be separate PRs for a lot of this stuff.

pyproject.toml
aarnphm: ok lets open a follow up PR about this

pyproject.toml
aarnphm: But then the package will become a lot heavier. I think we should keep the behaviour from setup.py

sauyon: Do packages contain dev dependencies by default?

aarnphm: yes

PR #2001, opened 2021-11-13
chore: minor fixes for io descripters & e2e test

bojiang:
## Checklist:


bentoml/_internal/io_descriptors/image.py
```diff
@@ -101,47 +102,56 @@ def __init__(self, pilmode: str = DEFAULT_PIL_MODE, mime_type: str = "image/jpeg
...
-
-        if (
-            content_type.decode("utf-8").startswith("image/")
-            or content_type == self._mime_type
-        ):
-            return PIL.Image.open(io.BytesIO(await request.body()))
-
-        raise BentoMLException(
-            f"{self.__class__.__name__} should have `Content-Type: multipart/form-data`, got {content_type} instead"
+        mime_type = content_type.decode()
```
aarnphm, line 118: ugh shouldn't we also accept processing content-type as `multipart/form-data` as well?

parano: We definitely should, @bojiang what's the reason for removing that?

bentoml/_internal/io_descriptors/image.py
```diff
@@ -101,47 +102,56 @@ def __init__(self, pilmode: str = DEFAULT_PIL_MODE, mime_type: str = "image/jpeg
 
         self._mime_type = mime_type.lower()
         self._pilmode = pilmode
-
-        ext = mimetypes.guess_extension(self._mime_type)
-        self._format = PIL.Image.EXTENSION[ext]
+        self._mime_type = mime_type
+        self._ext = mimetypes.guess_extension(mime_type)
+        if not self._ext:
+            raise ValueError(f"{mime_type} is not supported by the Image IO")
```
parano, line 108: can we move this check to line 92 above and share the same exception message?

bentoml/_internal/io_descriptors/numpy.py
```diff
@@ -1,16 +1,17 @@
 import json
 import logging
 import typing as t
+from typing import TYPE_CHECKING  # make linter happy
```
parano, line 4: what's the issue with linter here?  I think we use `t.TYPE_CHECKING` everywhere

bentoml/_internal/service/service.py
```diff
@@ -90,6 +90,16 @@ def api(
         """Decorator for adding InferenceAPI to this service"""
 
         def decorator(func: t.Callable) -> t.Callable:
+            from ..io_descriptors import Multipart
+
+            if isinstance(output, Multipart):
+                logger.warning(
```
parano, line 96: this is probably not the best place to add it, move this to `InferenceAPI.__init__`?

parano: actually, let's just remove this? I think it is ok to just warn user about it in the documentation - otherwise this message will be all over the place in the logs if the user actually wants to use it

bentoml/_internal/io_descriptors/image.py
bojiang: this code actually doesn't work in my testing, because PIL needs the extension to read the image bytes.

bojiang: user may define `Multipart(img=Image())` to support the `multipart-form` mime?

parano:
Could you address the comments in a follow up? @bojiang

PR #2004, opened 2021-11-15
refactor(framework): onnx multi args support

larme:
## Checklist:

### DIFF
--- a/bentoml/onnx.py
+++ b/bentoml/onnx.py
@@ -1,3 +1,4 @@
+import functools
 import logging
 import os
 import shutil
@@ -177,6 +178,7 @@ class _ONNXRunner(Runner):
         disable_copy_in_default_stream: bool,
         providers: t.Optional[_ProviderType],
         session_options: t.Optional["ort.SessionOptions"],
+        partial_kwargs: t.Optional[t.Dict[str, t.Any]],
         resource_quota: t.Optional[t.Dict[str, t.Any]],
         batch_options: t.Optional[t.Dict[str, t.Any]],
         model_store: "ModelStore",
@@ -190,6 +192,7 @@ class _ONNXRunner(Runner):
         self._model_info, self._model_file = _get_model_info(tag, model_store)
         self._model_store = model_store
         self._backend = backend
+        self._partial_kwargs = partial_kwargs or dict()
 
         if backend not in SUPPORTED_ONNX_BACKEND:
             raise BentoMLException(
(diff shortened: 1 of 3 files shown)


bentoml/onnx.py
```diff
@@ -274,28 +277,34 @@ def _setup(self) -> None:  # type: ignore[override]
...
+        def _mapping(item) -> t.Any:
+            if isinstance(item, np.ndarray):
+                item = item.astype(np.float32)
+            elif isinstance(item, pd.DataFrame):
+                item = item.to_numpy()
+            else:
+                raise TypeError(
+                    f"`_run_batch` of {self.__class__.__name__} only takes "
+                    "`numpy.ndarray` or `pd.DataFrame` as input parameters"
+                )
```
larme, line 298: should we support torch.Tensor? @aarnphm

aarnphm: sure, but I think we might need to do lazy loading for torch and tensorflow

bentoml/onnx.py
```diff
@@ -274,28 +277,34 @@ def _setup(self) -> None:  # type: ignore[override]
...
-        else:
-            raise TypeError(
-                f"`_run_batch` of {self.__class__.__name__} only takes "
-                "`numpy.ndarray` or `pd.DataFrame` as input parameters"
-            )
+        params = Params[t.Union[np.ndarray, "pd.DataFrame"]](*args)
+
+        def _mapping(item) -> t.Any:
+            if isinstance(item, np.ndarray):
+                item = item.astype(np.float32)
```
aarnphm, line 291: I was thinking when writing this line is to allow users to define given type for numpy here. But I'm not entirely sure if that is necessary since I'm unsure about some usecases.

larme: If we add Tensors, do we still want to do conversion here? Or should we just check if the input data is types we support?

aarnphm: would onnx accept tensor directly?

bentoml/onnx.py
larme: done

bentoml/onnx.py
larme: no tensor so I think current codes is ok

larme:
@aarnphm I have read doc about onnxruntime and my understanding is that onnx.InferenceSession.run usally only take `output_names` and `input_names`, and maybe with a RunOptions. If we want user has the flexibility, we can put RunOptions in `load_runner` part. So do we really need `kwargs` in `_run_batch`?

OTOH because `input_names` is a dictionary, I think it may make sense to have a `partial_inputs` for runner so the one runner can have some parameter fixed for all inputs it received. I tried to make a prototype and write a test case for this usage. However I'm not sure if this make sense. Any thoughts?

aarnphm:
> @aarnphm I have read doc about onnxruntime and my understanding is that onnx.InferenceSession.run usally only take `output_names` and `input_names`, and maybe with a RunOptions. If we want user has the flexibility, we can put RunOptions in `load_runner` part. So do we really need `kwargs` in `_run_batch`?

You are right, I don't think we need kwargs for this case.

 
> OTOH because `input_names` is a dictionary, I think it may make sense to have a `partial_inputs` for runner so the one runner can have some parameter fixed for all inputs it received. I tried to make a prototype and write a test case for this usage. However I'm not sure if this make sense. Any thoughts?

With ONNX the cases for `input_names` and `output_names` is dependant on the model itself. Some model has multiple inputs, multiple outputs. Current implementation assume that when users put in inputs for `run_batch` they will keep the order of the inputs to how they do it with the torch model, and internally we mapped it over to `input_names` of the converted models. I don't really see the advantage of having a `partial_inputs` in this case.

bentoml/onnx.py
```diff
@@ -36,6 +37,8 @@
     )
 
 pd = LazyLoader("pd", globals(), "pandas")  # noqa: F811
+torch = LazyLoader("torch", globals(), "torch")  # noqa: F811
+tf = LazyLoader("tf", globals(), "tensorflow")  # noqa: F811
 
 _ProviderType = t.TypeVar(
```
aarnphm, line 43:
```suggestion
_ProviderType = t.List[t.Union[str, t.Tuple[str, t.Dict[str, t.Any]]]]
```

aarnphm (approved):
LGTM

PR #2014, opened 2021-11-16
fix(internal): use importlib for version

sauyon:
I think this is somewhat nicer.

At the moment this will still give the versioneer version inherited from the `setup.py`. If we move away from versioneer / `setup.py` it will give whatever we specify in the `pyproject.toml`, but we can use a poetry plugin to give us VCS versions if we want that.

### DIFF
--- a/bentoml/__init__.py
+++ b/bentoml/__init__.py
@@ -1,7 +1,9 @@
-from ._version import get_versions
-
-__version__ = get_versions()["version"]
-del get_versions
+try:
+    import importlib.metadata as importlib_metadata
+except ModuleNotFoundError:
+    import importlib_metadata
+__version__: str = importlib_metadata.version(__name__)
+del importlib_metadata
 
 from ._internal.configuration import load_global_config  # noqa: E402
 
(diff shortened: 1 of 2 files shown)


parano:
The unit tests CI has not been fixed yet although it looks like codestyle_check did break

bentoml/_internal/configuration/__init__.py
```diff
@@ -34,14 +38,14 @@ def expand_env_var(env_var):
 # the BentoML module to be used when loading and using a saved BentoService bundle.
 # This is useful when using customized BentoML fork/branch or when working with
 # development branches of BentoML
-BENTOML_VERSION = __version__
+BENTOML_VERSION: str = importlib_metadata.version(__name__)
```
parano, line 41: I assume this reads the PKG-INFO file from the package sdist?  what does it return in a dev environment?   I'm confused how this would interact with versioneer

sauyon: I wrote this up somewhere and then promptly forgot, probably---in a dev environment it evaluates the setup.py or pyproject.toml versions, it seems like.

So it works as "expected" giving 0.13.x+g<whatever>, at least for me right now with the poetry venv.

PR #2015, opened 2021-11-16
Bento object fixes

sauyon:
Thought I'd factor this out of the model store refactor because those changes were getting a bit unwieldy.


tests/unit/_internal/bento/test_bentos.py
```diff
@@ -15,11 +15,11 @@ def test_create_simplebento(tmpdir):
         version="1.0",
         build_ctx="./simplebento",
         additional_models=[],
-        include=["*.py", "config.json", "somefile", "*dir*"],
+        include=["*.py", "config.json", "somefile", "*dir*", ".bentomlignore"],
         exclude=[
             "*.storage",
             "/somefile",
-        ],  # + anything specified in .bentoml_ignore file
+        ],  # + anything specified in .bentomlignore file
```
parano, line 22: not related to this PR - one thing I noticed when building bento with 1.0 was lots of `.pyc` files and `__pycache__` directory was included in the final Bento. Do you think we should add those to the default exclude list?

tests/unit/_internal/bento/simplebento/.bentomlignore
```diff
@@ -0,0 +1 @@
+__pycache__
```
parano, line 1: not related to this PR - one thing I noticed when building bento with 1.0 was lots of .pyc files and __pycache__ directory was included in the final Bento. Do you think we should add those to the default exclude list?

bentoml/_internal/bento/bento.py
```diff
@@ -200,45 +209,68 @@ def __str__(self):
...
+    models: t.List[str]
+
+    def dump(self, stream: t.IO[t.Any]):
+        return yaml.dump(self, stream, sort_keys=False)
 
     @classmethod
     def from_build_args(
         cls, tag: Tag, svc: "Service", labels: t.Dict[str, t.Any], models: t.List[str]
     ):
-        bento_metadata = cls()
```
parano, line 222: really nice refactoring here 👍

PR #2016, opened 2021-11-16
fix(internal): more miscellaneous bento fixes

sauyon:


bentoml/_internal/bento/bento.py
```diff
@@ -279,8 +265,11 @@ def from_yaml_file(cls, stream: t.IO[t.Any]):
             logger.error(exc)
             raise
 
+        yaml_content["tag"] = Tag(yaml_content["name"], yaml_content["version"])
+        del yaml_content["name"]
+        del yaml_content["version"]
+        yaml_content["creation_time"] = datetime.fromisoformat(yaml_content["creation_time"])
         bento_info = cls(**yaml_content)
```
parano, line 272: should we catch the `TypeError` for unexpected keyword argument here and give user a message about unexpected content in `bento.yaml` ?

bentoml/_internal/bento/bento.py
```diff
@@ -144,13 +143,14 @@ def create(
...
+            BentoInfo(tag, svc, labels, models).dump(bento_yaml)  # type: ignore
 
-        return OSBento(tag, bento_fs)
+        return SysPathBento(tag, bento_fs)
 
     @classmethod
-    def from_fs(cls, tag: Tag, fs: FS) -> "Bento":
-        res = cls(Tag("", ""), fs)
+    def from_fs(cls, item_fs: FS) -> "Bento":
+        res = cls(None, item_fs)  # type: ignore
```
aarnphm, line 153:
```suggestion
        res = cls(None, item_fs)  # type: Bento
```

bentoml/_internal/bento/bento.py
```diff
@@ -277,8 +264,23 @@ def from_yaml_file(cls, stream: t.IO[t.Any]):
             logger.error(exc)
             raise
 
-        bento_info = cls(**yaml_content)
-        bento_info.validate()
+        yaml_content["tag"] = Tag(yaml_content["name"], yaml_content["version"])
+        del yaml_content["name"]
```
aarnphm, line 268: Why do we have to do this?

bentoml/_internal/bento/bento.py
```diff
@@ -234,41 +236,26 @@ def save(
...
-            labels=labels,  # TODO: validate user provided labels
-            apis=svc._apis,  # type: ignore[reportPrivateUsage]
-            models=models,  # TODO: populate with model & framework info
-        )
-
-        bento_info.validate()
-        return bento_info
-
     @classmethod
     def from_yaml_file(cls, stream: t.IO[t.Any]):
```
aarnphm, line 260: maybe it is type `BinaryIO` instead?

```suggestion
    def from_yaml_file(cls, stream: t.BinaryIO):
```

bentoml/_internal/bento/bento.py
sauyon: The issue is the `None` tag argument, not the result type of the expression.

bentoml/_internal/bento/bento.py
sauyon: I'm passing the yaml directly as kwargs to init.

bentoml/_internal/bento/bento.py
sauyon: Why? Does the method not support `TextIO`?

aarnphm: Does passing a file in directly considered a TextIO? Passing files would be considered as a BinaryIO no?

sauyon: If you open a file with an encoding it's `TextIO`, I believe.

sauyon:
Oh the test failures are because of the `DummyItem` class...

sauyon:
@aarnphm @parano  seems like the e2e tests are failing but this should be good to go in---codestyle checks are running on everything as I said, so it's not very useful.

sauyon:
...nevermind I introduced an `isort` issue while cleaning up `pylint` issues, of course.

bentoml/_internal/bento/bento.py
bojiang: why we cannot make this class `tag: t.Optiional[Tag]` here as usual?

parano:
@sauyon I think the issue `No module named pytest` in test CI has been fixed, you can try re-run tests on this PR now

PR #2017, opened 2021-11-16
test: fix pytests

sauyon:


bentoml/_internal/bento/bento.py
```diff
@@ -17,6 +17,7 @@
 from ...exceptions import BentoMLException, InvalidArgument
 from ..configuration import BENTOML_VERSION
 from ..configuration.containers import BentoMLContainer
+from ..service.inference_api import InferenceAPI
```
aarnphm, line 20: Since we only need this for type_checking, you just need to quote it instead of import it at top level

PR #2018, opened 2021-11-16
test: make flake8 slightly more strict 

sauyon:


setup.cfg
```diff
@@ -1,10 +1,10 @@
 [flake8]
```
parano, line 1: cc @aarnphm I think changes in this file should also trigger the style_check CI to run?

aarnphm: ok we can add that into ci

PR #2019, opened 2021-11-17
fix(ci): missing dependencies

aarnphm:
Addressing some CI problems wrt inconsistency between runs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,7 @@ jobs:
               - .github/workflows/ci.yml
               - setup.cfg
               - pyproject.toml
+              - scripts/ci/config.yml
             bentoml:
               - *related
               - "bentoml/**"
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -98,8 +99,8 @@ jobs:
         run: timeout 10s poetry run pip --version || rm -rf .venv
 
       - name: Install dependencies
-        if: steps.cache-pip.outputs.cache-hit != 'true'
-        run: poetry install
+        if: steps.cache-pip.outputs.cache-hit != 'true' || needs.diff.outputs.related == 'true'
+        run: poetry install -E "docs"
```
sauyon, line 103: Why don't we just cache all the dependencies we need for CI at once?

aarnphm: yep we should do that instead.

PR #2020, opened 2021-11-17
chore: revert attrs dependency changes

sauyon:
We only need attrs 21.1 for type checking.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,17 +68,10 @@ sphinxcontrib-fulltoc = {version = "*", optional = true}
 sphinxcontrib-spelling = {version = "*", optional = true}
 sphinx-copybutton = {version = "*", optional = true}
 pyenchant = {version = "*", optional = true}
-types-click = {version = "*", optional = true}
-types-chardet = {version = "*", optional = true}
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -81,6 +81,7 @@ docs = ["recommonmark", "sphinx", "sphinx-click", "sphinx_rtd_theme", "sphinxcon
 types = ["types-click", "types-chardet", "types-setuptools", "pyspark-stubs", "types-filelock", "types-PyYAML"]
 
 [tool.poetry.dev-dependencies]
+attrs = ">=21.1"
```
aarnphm, line 84: but then I think by default this will get install nonetheless, so it will override with line 50

sauyon: Yes, but it won't be an issue if user packages require old attrs for some reason.

sauyon:
I've moved the type packages into poetry's dev dependencies; I'm pretty sure those aren't included in pypi packages published by poetry.

bojiang (approved):

pyproject.toml
bojiang: (really detailed)

PR #2024, opened 2021-11-18
ci: fix CI

sauyon:


bentoml/_internal/utils/formparser.py
```diff
@@ -96,7 +96,6 @@ async def parse(self) -> _ItemsBody:
             parser.write(chunk)
             messages = list(self.messages)
             self.messages.clear()
```
bojiang, line 98: Then what happens here? There is already a stub now?

bojiang (approved):

bojiang:
sauyon: codestyle check will still fail because of an error fixed in #2016

PR #2025, opened 2021-11-18
feat: `--use-poetry` for CI scripts

aarnphm:

CI scripts now enable optional using poetry via `--use-poetry`. By default we won't use poetry

```bash
# This will use poetry
./scripts/ci/run_tests.sh --use-poetry ...
# default behaviour
./scripts/ci/run_tests.sh general_features
```

Btw all e2e tests should have `type` under `config.yml` to be `e2e` and it will cd into the directory @bojiang


bojiang:
rebased to the newest bentoml-1.0

bojiang:
let's leave the macos e2e test issue to another PR

PR #2026, opened 2021-11-19
fix: removing optional deps from dev deps

bojiang:
## Motivation and Context

To avoid possible false negatives

## Checklist:


sauyon:
The reason I added pydantic as a dev dependency was because it's required for a unit test---we should probably have a core suite of tests that are runnable without optional dependencies.

bojiang:
@sauyon Is `pydantic` only required by unittests of Json IO?

sauyon:
Yeah. Maybe that test should be moved somewhere else?

bojiang:
@sauyon then we can just put `pydantic` into the `ci/config.yaml`

sauyon:
Sure, but I think that there should be a separation between tests you can run locally without having all the optdepends and stuff you can't, is all.

Ideally a contributor who just ran `poetry install` would be able to run some test script that runs just the stuff that they can run.

bojiang:
Surprisingly the unit tests didn't fail.

bojiang:
>Sure, but I think that there should be a separation between tests you can run locally without having all the optdepends and stuff you can't, is all.
Ideally a contributor who just ran poetry install would be able to run some test script that runs just the stuff that they can run.

I'm not sure if I understand well. Contributors can always run `make test_*` locally right? @aarnphm

aarnphm:
> Contributors can always run `make test_*` locally right? @aarnphm
`make tests-<name>` where `<name>` is defined under `./scripts/ci/config.yml`. But the `run_tests.sh` assumes that you install all the dependencies including tests.

parano:
looks like this is no longer relevant after the latest CI changes, closing now

PR #2027, opened 2021-11-19
feat: launch prod server & runner app

bojiang:
## Checklist:


bentoml/_internal/server/service_app.py
```diff
@@ -12,7 +13,7 @@
 from ..server.base_app import BaseAppFactory
 from ..service.service import Service
 
-if t.TYPE_CHECKING:
+if TYPE_CHECKING:  # try not use t.TYPE_CHECKING to avoid pylint unused-import
```
aarnphm, line 16:
```suggestion
if TYPE_CHECKING:   # pragma: no cover # pylint: disable=unused-import 
```

bojiang: Reject. this will also omit the the imports that really not used

aarnphm: ugh usually for types they will be detected as unused-import?

bojiang: No. pylint actually know if you are using `if TYPE_CHECKING`. But not that smart at lease now.

aarnphm: but doing individual imports from typing is a lot slower than just `import typing as t` 🤷🏼

bentoml/_internal/server/__init__.py
```diff
@@ -40,18 +40,26 @@ def serve_development(
         watchers.append(
             Watcher(
                 name="ngrok",
-                cmd=f'{sys.executable} -c "import bentoml; bentoml.server._start_ngrok_server()"',
+                cmd=f'{sys.executable} -c "import bentoml; bentoml.server.start_ngrok_server()"',
```
parano, line 43: this should be under `bentoml._internal` right?

bentoml/_internal/server/service_app.py
bojiang: You mean from `8.11us` to `10us` ?

aarnphm: in my case it is some us to some ms

bentoml/_internal/server/service_app.py
```diff
@@ -244,24 +251,30 @@ def middlewares(self) -> t.List["Middleware"]:
...
-            on_shutdown=[self.bento_service._on_asgi_app_shutdown],
-            on_startup=[self.bento_service._on_asgi_app_startup, self.mark_as_ready],
-        )
+        app = super().__call__()
 
         for mount_app, path, name in self.bento_service._mount_apps:
-            app.mount(app=mount_app, path=path, name=name)
+            app.mount(
+                app=mount_app, path=path, name=name
+            )  # TODO(jiang): mount wsgi app ?
```
parano, line 272: wsgi app is already supported, there's no change here needed, https://github.com/bentoml/BentoML/blob/bentoml-1.0/bentoml/_internal/service/service.py#L190

scripts/ci/config.yml
```diff
@@ -18,6 +18,11 @@ general_features:
   is_dir: true
   type_tests: "e2e"
   dependencies:
+    - "pytest"
```
parano, line 21: we don't need these here, everything here besides `Pillow` is already defined in the base test-requirements.txt

bentoml/_internal/server/runner_app.py
```diff
@@ -9,51 +10,53 @@
...
+        """
+        Setup routes for bento model server, including:
+
+        /               Index Page, shows readme docs, metadata, and Swagger UI
+        /docs.json      Returns Swagger/OpenAPI definition file in json format
+        /healthz        liveness probe endpoint
+        /readyz         Readiness probe endpoint
+        /metrics        Prometheus metrics endpoint
+
+        And user defined InferenceAPI list into routes, e.g.:
```
parano, line 47: update this docstring for runner app?

bentoml/_internal/server/base_app.py
```diff
@@ -9,44 +9,70 @@
...
             return PlainTextResponse("\n", status_code=200)
         raise HTTPException(500)
 
-    @abc.abstractmethod
     def __call__(self) -> "Starlette":
-        ...
+        from starlette.applications import Starlette
+
+        return Starlette(
+            debug=True,  # TDOO: inject this from `debug=True` or `--production` flag
```
parano, line 58:
```
from bentoml._internal.configuration import get_debug_mode

...

debug=get_debug_mode()
```

bentoml/_internal/server/service_app.py
bojiang: This is more about typing

scripts/ci/config.yml
bojiang: And that's what we should not do I guess

bentoml/_internal/server/__init__.py
```diff
@@ -171,24 +203,43 @@ def _start_dev_api_server(
         uvicorn.run(svc.asgi_app, **uvicorn_options)  # type: ignore
 
 
-"""
-def _start_prod_api_server(bento_path_or_tag: str, instance_id: int, runners_map: str):
+def start_prod_api_server(
```
aarnphm, line 206: need tests for this 😢

bojiang: just covered by e2e tests

bentoml/_internal/runner/container.py
```diff
@@ -82,13 +86,15 @@ def batch_to_payloads(cls, batch: BatchType, batch_axis=None) -> t.List[Payload]
 
 class NdarrayContainer(DataContainer["tnp.ndarray", "tnp.ndarray"]):
     @classmethod
-    def singles_to_batch(cls, singles, batch_axis=0):
+    def singles_to_batch(
+        cls, singles: t.Sequence["tnp.ndarray"], batch_axis: t.Optional[int] = 0
```
aarnphm, line 90: this kinda also depends on the stubs PR to get merged as well, but in numpy 1.20+:
```
... singles: t.Sequence["tnp.ndarray[t.Any, np.dtype[t.Any]"], ...
```

bentoml/_internal/server/runner_app.py
```diff
@@ -1,59 +1,96 @@
...
     @property
-    def asgi_app(self) -> "Starlette":
-        from starlette.applications import Starlette
+    def name(self) -> str:
+        return self.runner.name
+
+    @property
+    def routes(self) -> t.List["BaseRoute"]:
+        """
+        Setup routes for bento model server, including:
```
parano, line 48:
```suggestion
        Setup routes for Runner server, including:
```
```suggestion
        Setup routes for bento model server, including:
```

bentoml/_internal/server/service_app.py
parano:
```suggestion
            )
```

bentoml/_internal/service/service.py
```diff
@@ -37,7 +37,6 @@ class Service:
     """
 
     _apis: t.Dict[str, InferenceAPI] = {}
```
parano, line 39: Should we just make `_apis` also public, same as runners?

Also setting the default to `{}` was a huge mistake by me 😂 we should change that in a separate PR

bentoml/sklearn.py
```diff
@@ -125,7 +125,7 @@ def __init__(
         batch_options: t.Optional[t.Dict[str, t.Any]],
         model_store: "ModelStore" = Provide[BentoMLContainer.model_store],
     ):
-        super().__init__(str(tag), resource_quota, batch_options)
+        super().__init__(f"{tag}-{function_name}", resource_quota, batch_options)
```
parano, line 128: is this change necessary? I think it's ok to just use the model name so everything is consistent. Doing this will be hard for some other frameworks, xgboost's load_runner for example takes a whole dictionary, and you can't easily get a readable and unique str to represent the runner.

A slightly nicer way to do it is:

```suggestion
        super().__init__(f"{tag}-{id(self)}", resource_quota, batch_options)
```

scripts/ci/config.yml
```diff
@@ -18,6 +18,11 @@ general_features:
...
+    - "pytest-asyncio"
+    - "pandas"
+    - "pydantic"
+    - "scikit-learn"
```
parano, line 25:
```suggestion
```

typings/starlette/formparsers.pyi
```diff
@@ -30,7 +30,6 @@ class MultiPartMessage(enum.Enum):
     HEADERS_FINISHED: Literal["7"] = ...
     END: Literal["8"] = ...
 
-def _user_safe_decode(src: bytes, codec: str) -> str: ...
```
parano, line 33: did we add this manually before? curious why a private function got added to stub

bentoml/_internal/runner/runner.py
```diff
@@ -78,19 +77,30 @@ class BatchOptions:
...
     ):
+        # probe an unique name
+        i = 0
+        while True:
+            name = display_name if i == 0 else f"{display_name}-{i}"
+            if name not in self.EXIST_NAMES:
+                self.EXIST_NAMES.add(name)
+                break
+            else:
+                i += 1
```
parano, line 97: Maybe we can apply the object id here instead of in the framework runner class?
```suggestion
        name = f"{display_name}-{id(self)}"
```

parano: This is still not ideal tho, in an idea case we would have identical runner name even if we restart the entire API server or run it from a different machine, if the runner is loaded from the same model with the same config

bentoml/_internal/runner/runner.py
```diff
@@ -78,19 +77,30 @@ class BatchOptions:
...
         self.name = name
-        self.resource_quota = ResourceQuota(**resource_quota if resource_quota else {})
```
parano, line 92: this should also work, right?

typings/starlette/formparsers.pyi
aarnphm: Because we uses this function

bentoml/_internal/runner/container.py
parano:
```suggestion
        cls, singles: t.Sequence["tnp.ndarray[t.Any, np.dtype[t.Any]"], batch_axis: t.Optional[int] = 0
```

bentoml/_internal/runner/container.py
```diff
@@ -110,7 +116,7 @@ def single_to_payload(
             )
 
         return cls.create_payload(
-            cloudpickle.dumps(single_data),
+            pickle.dumps(single_data),
```
parano, line 119: 👍

bentoml/_internal/runner/container.py
```diff
@@ -156,7 +164,7 @@ def singles_to_batch(
 
     @classmethod
     def batch_to_singles(
-        cls, batch: "pd.DataFrame", batch_axis=0
+        cls, batch: "pd.DataFrame", batch_axis: t.Optional[int] = 0
```
parano, line 167: can batch_axis be None at all?

bentoml/_internal/server/runner_app.py
```diff
@@ -1,59 +1,96 @@
...
     @inject
     def __init__(
         self,
         runner: "Runner",
-        tracer=Provide[BentoMLContainer.tracer],
-    ):
+        instance_id: t.Optional[int] = None,
+        tracer: "Tracer" = Provide[BentoMLContainer.tracer],
+    ) -> None:
         self.runner = runner
```
parano, line 37: should we do `self.runner._setup()` here?

bentoml/_internal/server/__init__.py
```diff
@@ -25,35 +26,46 @@ def serve_development(
...
+                cmd=f"{sys.executable} -c '{cmd_ngrok}'",
                 env=env,
                 numprocesses=1,
                 stop_children=True,
                 working_dir=working_dir,
             )
         )
 
-    dev_server_cmd = f'import bentoml._internal.server; bentoml._internal.server._start_dev_api_server("{bento_identifier}", {port}, working_dir="{working_dir}", reload={reload}, reload_delay={reload_delay}, instance_id=$(CIRCUS.WID))'
+    cmd_api_server = f"""
```
parano, line 55:
```suggestion
    cmd_api_server = f"""\
```

bentoml/_internal/server/__init__.py
```diff
@@ -25,35 +26,46 @@ def serve_development(
...
 
     env = dict(os.environ)
 
-    watchers = []
-    for _, runner in svc._runners.items():
-        runner._setup()
-
+    watchers: t.List[Watcher] = []
     if with_ngrok:
+        cmd_ngrok = """
```
parano, line 41:
```suggestion
        cmd_ngrok = """\
```

bentoml/sklearn.py
parano:
```suggestion
        super().__init__(str(tag), resource_quota, batch_options)
```

bojiang:
ops the `tnp.ndarray` patch breaks everything

bentoml/_internal/frameworks/keras.py
```diff
@@ -188,43 +191,38 @@ def __init__(
...
-    @property
-    def session(self) -> "BaseSession":
-        return self._session
+        self._session = _sess
+        self._tag = in_store_tag
 
-    # pylint: disable=attribute-defined-outside-init
-    def _setup(self) -> None:
+    # pylint: disable=arguments-differ,attribute-defined-outside-init
+    def _setup(self) -> None:  # type: ignore[override] # noqa
```
aarnphm, line 208: why do we need to do this? `_setup` is not user facing and not supposed to accept any extra arguments

bentoml/_internal/bento/bento.py
```diff
@@ -133,7 +133,7 @@ def create(
 
         model_tags = build_config.additional_models
         # Add Runner required models to models list
-        for runner in svc._runners.values():  # type: ignore[reportPrivateUsage]
+        for runner in svc.runners.values():  # type: ignore[reportPrivateUsage]
```
aarnphm, line 136:
```suggestion
        for runner in svc.runners.values():
```

PR #2029, opened 2021-11-19
refactor(internal): move model store to common store implementation

sauyon:
I was trying to fix pyright errors everywhere, but that seemed like a losing battle.

This depends on #2016 for now.

TODO: added numpy stubs for numpy < 1.20


bentoml/_internal/bento/bento.py
```diff
@@ -11,33 +11,46 @@
...
     from ..service import Service
-    from ..service.inference_api import InferenceAPI
 
 logger = logging.getLogger(__name__)
 
 BENTO_YAML_FILENAME = "bento.yaml"
 BENTO_PROJECT_DIR_NAME = "src"
 
 
+def _Bento_set_fs(bento: "Bento", _: t.Any, new_fs: "FS") -> "FS":
```
parano, line 38: what's the reason for the second parameter here?

sauyon: That's the `attrs` converter spec; it takes the old object, the field being assigned, and the new value. The field being assigned in this case is always `.fs`, so I just ignored it.

bentoml/_internal/models/model.py
```diff
@@ -0,0 +1,257 @@
...
+    def __attrs_post_init__(self):
+        self.validate()
+
+    def dump(self, stream: t.IO[t.Any]):
+        return yaml.dump(self, stream, sort_keys=False)
+
+    @classmethod
+    def from_yaml_file(cls, stream: t.IO[t.Any]):
+        try:
+            yaml_content = yaml.safe_load(stream)
```
sauyon, line 220: Note: `safe_load` isn't able to load a decent amount of yaml objects that we're dumping, but I'm not what the security model should be here. We might want to support custom save/load routines for metadata or just require that all the relevant data is `str`.

aarnphm:
> @aarnphm should I be trying to fix all the tests or are some of them currently broken?

None of them are broken

bentoml/_internal/bento/bento.py
```diff
@@ -33,10 +35,24 @@
...
+    bento._model_store = ModelStore(new_fs.opendir("models"))  # type: ignore[reportPrivateUsage]
+    return new_fs
+
+
 @attr.define(repr=False)
 class Bento(StoreItem):
     _tag: Tag
-    _fs: FS
+    _fs: "FS" = attr.field(on_setattr=_Bento_set_fs)
+    _model_store: ModelStore
```
aarnphm, line 48:
```suggestion
    _model_store: Provider[BentoMLContainer.model_store]
```

bentoml/__init__.py
```diff
@@ -8,17 +9,15 @@
 
 configure_logging()
 
-from ._internal.models.store import ModelStore  # noqa: E402
-from ._internal.service import Service  # noqa: E402
-from ._internal.service.loader import load  # noqa: E402
-from ._internal.types import Tag  # noqa: E402
+import bentoml.models as models
```
aarnphm, line 12:
```suggestion
from . import models
```

bentoml/_internal/models/model.py
```diff
@@ -0,0 +1,257 @@
...
+from ..types import PathType, Tag
+
+if t.TYPE_CHECKING:
+    from ..runner import Runner
+
+logger = logging.getLogger(__name__)
+
+MODEL_YAML_FILENAME = "model.yaml"
+
+PathRes = t.TypeVar("PathRes", bound=t.Union[t.Optional[str], str])
```
aarnphm, line 29: dead code?
if not then maybe
```suggestion
PathRes = t.Union[t.Optional[str], str]
```
as a typealias rather than a typevar?

bentoml/_internal/models/model.py
```diff
@@ -0,0 +1,257 @@
...
+        metadata: t.Optional[t.Dict[str, t.Any]] = None,
+        framework_context: t.Optional[t.Dict[str, t.Any]] = None,
+    ) -> "SysPathModel":
+        tag = Tag(name).make_new_version()
+        labels = {} if labels is None else labels
+        options = {} if options is None else options
+        metadata = {} if metadata is None else metadata
+        framework_context = {} if framework_context is None else framework_context
+
+        model_fs = fs.open_fs("temp://bentoml_model_{name}")
```
aarnphm, line 65: This lines broke a lot of tests, seems that f string is wrongly formatted
```suggestion
        model_fs = fs.open_fs(f"temp://bentoml_model_{name}")
```
also the path tests look like this
```bash
/tmp/pytest-of-aarnphm/pytest-25/bentoml0/test_mlflow_save_load/nn5yl7cnffy6zkeafb746k2pu/tmp/tmpznwfm0fgbentoml_model_test_mlflow_save_load/sklearn_clf
```

bentoml/_internal/models/model.py
```diff
@@ -0,0 +1,257 @@
...
+        pass
+
+    def load_runner(self) -> "Runner":
+        raise NotImplementedError
+
+    def validate(self):
+        return self._fs.isfile(MODEL_YAML_FILENAME)
+
+    def __str__(self):
+        return f'Bento(tag="{self.tag}", path="{self.path}")'
```
aarnphm, line 148:
```suggestion
        return f'Model(tag="{self.tag}", path="{self.path}")'
```

bentoml/_internal/bento/bento.py
sauyon: This is the model store inside the bento.

bentoml/_internal/models/model.py
sauyon: It's old code I forgot to delete; it used to be the generic parameter to `Bento` before I just made the syspathbento because generics in python are absolutely awful lol.

aarnphm:
@sauyon @parano everything should be fixed now

bentoml/_internal/bento/bento.py
bojiang: btw this is not a correct usage.

aarnphm: yep thats my bad

sauyon:
Fixed lint.

aarnphm (dismissed):
Looks and feels good to me on first pass

aarnphm (approved):

bentoml/_internal/models/model.py
```diff
@@ -0,0 +1,255 @@
...
+            with self._fs.open(MODEL_YAML_FILENAME, "w") as model_yaml:
+                self._info.dump(model_yaml)
+
+        if not self.validate():
+            logger.warning(f"Failed to create Model for {self.tag}, not saving.")
+            raise BentoMLException("Failed to save Model because it was invalid")
+
+        with model_store.register(self.tag) as model_path:
+            os.rmdir(model_path)
+            shutil.move(self._fs.getsyspath("/"), model_path)
```
parano, line 179: This should probably be `shutil.copy`?  I ran into an issue: when building a new Bento, the `bento.create` method calls this to copy over models from the local model store to a new Bento's fs, and it actually removes the files in the local model store unexpectly

aarnphm: @sauyon

PR #2035, opened 2021-11-20
refactor(ci): migrate back to `setuptools`, PEP517 compatible

aarnphm:

- Unified config system to `setup.cfg` and `pyproject.toml`
- separate `requirements/docs-requirements.txt`, `requirements/tests-requirements.txt`, `requirements/dev-requirements.txt`
- uses `setuptools_scm` to generate version, in favor of PEP518 and PEP660. If we go this route then `bentoml/_version.py` will exists within releases package so there is no need to version control it. Which leads to deprecating `versioneer.py` [NEED MORE DISCUSSION]
- two pip commands that is exposed to end users : `pip install -e .` and `pip install -e ".[bento-server]"`
- refactor some docs and documents changes on how to use `./scripts/ci/run_tests.sh`
- remove poetry completely from both CI and scripts
- Fix some caching issues in CI

Idw whats wrong with ONNXMLir tests but I will fix it in the consequent PR


aarnphm:
I will merge it so that it won't block others workflow

PR #2043, opened 2021-11-22
test: automatically pragma no cover type checking

sauyon:
Should we also do this for `except ImportError`?


sauyon:
Hm, can somebody give me access to the codecov?

sauyon:
This doesn't actually work, sadly. I'll look into this more later.

PR #2044, opened 2021-11-22
fix: revert&fix typing of di containers

bojiang:
## Checklist:


scripts/tools/type_checker.sh
```diff
@@ -1,3 +1,6 @@
 #!/usr/bin/env bash
 
-pyright --stats
```
parano, line 3: the CI is using `scripts/ci/style/pyright_check.sh`, running in the checker docker image, not this file

bojiang: yeah modifying this for local test

requirements/dev-requirements.txt
```diff
@@ -11,3 +11,4 @@ types-setuptools
 types-PyYAML
 types-filelock
 attrs>=21.1
+pyright
```
aarnphm, line 14: rejected this

aarnphm (changes requested):
remove uncorrect requirements

requirements/dev-requirements.txt
aarnphm:
```suggestion
```

bentoml/_internal/configuration/containers.py
```diff
@@ -174,7 +175,7 @@ def as_dict(self) -> t.Dict[str, t.Any]:
         return self.config
 
 
-@container
+@dataclass
```
aarnphm, line 178: I thought we stopped using dataclass?

aarnphm:
@bojiang fixed to uses attrs instead of dataclass

bentoml/_internal/configuration/containers.py
```diff
@@ -170,11 +171,11 @@ def override(self, keys: t.List[str], value: t.Any):
...
             ) from e
 
-    def as_dict(self) -> t.Dict[str, t.Any]:
-        return self.config
+    def as_dict(self) -> providers.ConfigDictType:
+        return t.cast(providers.ConfigDictType, self.config)
 
 
-@container
+@attr.s
```
sauyon, line 178: Use `attr.define` or pyright will complain.

bentoml/_internal/configuration/containers.py
```diff
@@ -223,24 +224,21 @@ def tracer(
 
             return NoopTracer()
 
-    logging_file_directory = t.cast(
-        Provider[t.Callable[[], str]],
+    logging_file_directory = providers.Factory[str](
+        lambda default, customized: customized or default,
```
sauyon, line 228: we should decide whether we're ok with using `or` like this.

bentoml/_internal/configuration/containers.py
```diff
@@ -223,24 +224,21 @@ def tracer(
...
         ),
+        config.logging.file.directory,
     )
 
 
 BentoMLContainer = BentoMLContainerClass()
 
 
-@container
+@attr.s
```
sauyon, line 241: Same as above

bentoml/_internal/configuration/containers.py
```diff
@@ -315,8 +300,8 @@ def metrics_client(
         )
 
     # Mapping from runner name to RunnerApp file descriptor
-    remote_runner_mapping = t.cast(Provider[t.Dict[str, int]], providers.Static(dict()))
-    plasma_db: "PlasmaClient" = providers.Static(None)  # type: ignore[reportUnknownVariableType]
+    remote_runner_mapping = providers.Static[t.Dict[str, int]](dict())
+    plasma_db = providers.Placeholder["PlasmaClient"]()  # type: ignore
```
sauyon, line 304: Any reason this isn't just `t.Optional` and `None`?

scripts/ci/run_tests.sh
```diff
@@ -41,9 +37,9 @@ getval(){
 run_python(){
   if [ "$USE_POETRY" -eq 1 ]; then
     need_cmd poetry;
-    poetry run python -m "$@";
+    poetry run python -m "$@"
```
sauyon, line 40: We should remove this.

bentoml/_internal/configuration/containers.py
aarnphm: attr.define broke the whole thing

bentoml/_internal/configuration/containers.py
aarnphm: also doesn't seem like pyright is complaining

bentoml/_internal/configuration/containers.py
parano: @aarnphm it is ok to use dataclass for this use case if it works ok - it does not add extra dependency, right?

bentoml/_internal/configuration/containers.py
parano: I prefer specifically checking `is None`

bentoml/_internal/configuration/containers.py
parano: I think for some of the configs, `None` actually means something, whereas `Placeholder` means user's config yaml did not touch this field

parano: but in this case, I think `t.Optional` and `None` is better. @bojiang any reason we are not doing that?

bentoml/_internal/configuration/containers.py
aarnphm: nvm

bentoml/_internal/configuration/containers.py
aarnphm: yep, just for consistency sake

bentoml/_internal/configuration/containers.py
aarnphm: @bojiang so you suggest not to do this?

typings/schema.pyi
```diff
@@ -2,18 +2,39 @@
 This type stub file was generated by pyright.
 """
 from typing import Any, Callable, Dict, Generic, List
```
aarnphm, line 4: ignore this file, I was meant to open a separate PR to address this

aarnphm:
@bojiang good to merge?

aarnphm (approved):

bentoml/_internal/configuration/containers.py
bojiang: just dataclass here works for both pyright and runtime. actually it should be `simple_di.container` (which is just an alias of `dataclass` in that lib but sadly pyright didn't understand it)

bentoml/_internal/configuration/containers.py
bojiang: yeah Optional makes sense here. will modify it.

bentoml/_internal/configuration/containers.py
bojiang: Actually I don't know why the default value get so twisted. But it's not about the current PR.

PR #2048, opened 2021-11-24
fix: schema typing

aarnphm:
Some errors wrt schema type annotations


typings/schema.pyi
```diff
@@ -1,17 +1,23 @@
...
-from typing import Protocol, Tuple, TypeVar, Union, overload
+from types import BuiltinFunctionType, FunctionType
+from typing import Any, Callable, Dict, List
+from typing import Literal as LiteralType
+from typing import Optional as OptionalType
+from typing import Tuple, Union, overload
+
+GenericType = Union["Schema", "And", "Or", "Use", "Optional", "Regex", "Literal", "Const"]
+AcceptedDictType = Dict[Union[str, GenericType], Any]
+_CallableLike = Union[FunctionType, BuiltinFunctionType, Callable[..., Any]]
```
bojiang, line 12: `Callable` should already include the above. Any exceptions that requires this

bojiang: @aarnphm

aarnphm: Hmm I don't think so

bojiang: Any exception?

PR #2050, opened 2021-11-24
feat: multipart openapi

aarnphm:
added openapi supports for IO

also I added pandas and stubs :(

This could mainly due to setup issues. However, for consistency and doesn't let these problems fail CI for future contributor, I think it would be best if we keep a copy of stubs for the library. (applies to library that doesn't have py.typed supports, pandas).

Another solution is to have the correct setup, in which I don't really know how to make it work with python management system like pyenv or pipenv

image


aarnphm:
codestyle check will definitely fail because the stubs are incomplete. I think we can work the stubs along the way

bojiang:
https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github

let's just collapse those generated files automatically

bentoml/_internal/io_descriptors/base.py
```diff
@@ -7,6 +8,16 @@
 IOPyObj = t.TypeVar("IOPyObj")
 
 
+def _mk_repr(obj: t.Any) -> str:
```
bojiang, line 11: `__repr__` goal is to be unambiguous. Mostly you can eval a repr to create the same object.
`__str__` goal is to be readable.

bentoml/_internal/io_descriptors/json.py
```diff
@@ -123,30 +124,31 @@ def __init__(
...
     async def from_http_request(self, request: Request) -> JSONType:
         json_obj = await request.json()
         if hasattr(self, "_pydantic_model") and self._validate_json:
+            if pydantic is None:
+                raise MissingDependencyException(
+                    "`pydantic` must be installed to use `pydantic_model`"
+                )
             try:
-                return self._pydantic_model.parse_obj(json_obj)
+                self._pydantic_model.parse_obj(json_obj)
```
bojiang, line 148: So just verifying but not regulating the json_obj right?

aarnphm: `parse_obj` actually return a `pydantic.BaseModel` instead of a JSON object. I think we should only use pydantic for validation

bentoml/_internal/io_descriptors/multipart.py
```diff
@@ -44,28 +59,30 @@ def predict(img, annotations):
     curl -i -F img=@test.jpg -F annotations=@test.json localhost:5000/predict
     """
 
-    def __init__(self, **inputs: IODescriptor[_DescriptorType]):
+    def __init__(self, **inputs: _DescriptorType):
         for descriptor in inputs.values():
-            if isinstance(descriptor, Multipart):
+            if isinstance(descriptor, Multipart):  # pragma: no cover
```
sauyon, line 64: Why is this no cover?

bentoml/_internal/io_descriptors/multipart.py
```diff
@@ -44,28 +59,30 @@ def predict(img, annotations):
...
                 raise InvalidArgument(
                     "Multipart IO can not contain nested Multipart item"
                 )
 
-        self._inputs = inputs
+        self._inputs = inputs  # type: t.Dict[str, _DescriptorType]
```
sauyon, line 69: can you use `self._inputs: TYPE = inputs` here?

bentoml/_internal/io_descriptors/multipart.py
```diff
@@ -83,7 +100,41 @@ async def from_http_request(self, request: Request) -> MultipartIO:
             res[k] = v
         return res
 
-    async def to_http_response(self, obj: MultipartIO) -> Response:
+    @t.overload
+    async def to_http_response(self, obj: t.Dict[str, str]) -> Response:  # noqa: F811
```
sauyon, line 104: Seems like pyflakes is supposed to detect overload now---maybe it has to not be qualified?

And what's the point of this, anyway? Why not just have a union type? If there isn't separate documentation I don't think it's very valuable.

bentoml/_internal/io_descriptors/numpy.py
```diff
@@ -22,7 +22,7 @@
 def _is_matched_shape(
     left: t.Optional[t.Tuple[int, ...]],
     right: t.Optional[t.Tuple[int, ...]],
-) -> bool:
+) -> bool:  # pragma: no cover
```
sauyon, line 25: Why no cover?

bentoml/_internal/io_descriptors/pandas.py
```diff
@@ -16,10 +16,46 @@
 else:
     pd = LazyLoader("pd", globals(), "pandas")
 
+try:
+    from typing import Literal
+except ImportError:  # pragma: no cover
```
sauyon, line 21: Does this not get covered by our 3.7 test?

bentoml/_internal/io_descriptors/pandas.py
```diff
@@ -16,10 +16,46 @@
...
+    from typing_extensions import Literal
+
 logger = logging.getLogger(__name__)
 
 
-class PandasDataFrame(IODescriptor):
+def _infer_type(item: str) -> str:  # pragma: no cover
```
sauyon, line 27: seems like there are multiple infer type methods?

bentoml/_internal/io_descriptors/pandas.py
```diff
@@ -16,10 +16,46 @@
...
+        return "string"
+    elif item.startswith("bool"):
+        return "boolean"
+    else:
+        return "object"
+
+
+def _schema_type(
+    mime_type: str, dtype: t.Optional[t.Union[bool, t.Dict[str, t.Any]]]
+) -> t.Dict[str, t.Any]:  # pragma: no cover
```
sauyon, line 42: why no cover?

bentoml/_internal/types.py
```diff
@@ -36,7 +36,7 @@
 
 
 @total_ordering
-class Tag:
+class Tag(object):
```
sauyon, line 39: Should I have done this everywhere?

bentoml/_internal/service/inference_api.py
```diff
@@ -45,9 +45,11 @@ def __init__(
...
 
-    def __repr__(self):
-        return "<{}({} -> {})>".format(
-            self.__class__.__name__, repr(self.input), repr(self.output)
+    def __str__(self):
+        return u"%s(%s\u2192 %s)" % (
+            self.__class__.__name__,
+            str(self.input),
+            str(self.output),
         )
```
sauyon, line 53: I feel like f strings are more readable than formats like this.

Also, why not just have the actual unicode character?

sauyon (commented):
A mostly stylistic review, since I still don't really know much about what this is supposed to do.

typings/PIL/GimpPaletteFile.pyi
```diff
@@ -1,6 +1,3 @@
-"""
-This type stub file was generated by pyright.
-"""
```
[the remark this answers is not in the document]
aarnphm: cleanup

bentoml/_internal/io_descriptors/base.py
```diff
@@ -7,6 +8,20 @@
 IOPyObj = t.TypeVar("IOPyObj")
 
 
+def _mk_str(obj: t.Any) -> str:
```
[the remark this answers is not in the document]
aarnphm: I don't think so, this is mainly for IO descriptor only. I don't see other use for it

bentoml/_internal/io_descriptors/base.py
```diff
@@ -15,10 +30,18 @@ class IODescriptor(ABC, t.Generic[IOPyObj]):
...
+        return f"%s(%s)" % (
+            self.__class__.__name__,
+            ", ".join(
+                [f'{k.strip("_")}={_mk_str(v)}' for k, v in self.__dict__.items()]
+            ),
+        )
 
     # fmt: off
+    @abstractmethod
+    def openapi_schema_type(self) -> t.Dict[str, str]: ...  # noqa: E704
```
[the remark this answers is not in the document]
aarnphm: we should enable fmt: on and remove E704 ignore. I forgot why I did this :)

bentoml/_internal/io_descriptors/file.py
```diff
@@ -54,20 +54,19 @@ def predict(input_pdf):
         IO Descriptor that represents file-like objects.
     """  # noqa: LN001
 
-    def __init__(self, media_type: t.Optional[str] = None):
-        self._media_type = media_type or "application/octet-stream"
+    def __init__(self, mime_type: t.Optional[str] = None):
+        self._mime_type = mime_type or "application/octet-stream"
```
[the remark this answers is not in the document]
aarnphm: maybe we should just do 
```python
self._mime_type = 'application/octet-stream' if mime_type is None else mime_type
```

bentoml/_internal/io_descriptors/file.py
```diff
@@ -54,20 +54,19 @@ def predict(input_pdf):
...
 
-    def openapi_schema(self) -> t.Dict[str, t.Dict[str, t.Any]]:
-        # TODO: supports multipart OpenAPI
-        return {self._media_type: dict(schema=dict(type="string", format="binary"))}
+    def openapi_schema_type(self) -> t.Dict[str, str]:
+        return dict(type="string", format="binary")
```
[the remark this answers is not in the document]
aarnphm: sure, I don't really have a preference

bentoml/_internal/io_descriptors/multipart.py
aarnphm: sure I will add it tmr morning

bentoml/_internal/io_descriptors/multipart.py
aarnphm: Union type doesn't seem to work with pyright

bentoml/_internal/io_descriptors/numpy.py
aarnphm: I don't think we need to write unit test for this

bentoml/_internal/io_descriptors/pandas.py
aarnphm: only for pandas and numpy

bentoml/_internal/io_descriptors/pandas.py
aarnphm: yep

aarnphm: well usually I don't think we need to write test to cover import

bentoml/_internal/service/inference_api.py
aarnphm: sure I will change it to f-string

bentoml/_internal/io_descriptors/multipart.py
sauyon: I haven't seen an issue with that? What was wrong with it?

bentoml/_internal/io_descriptors/numpy.py
sauyon: Are we even using it? What's it for?

bentoml/_internal/io_descriptors/pandas.py
sauyon: no, but in this particular case shouldn't it be covered?

bentoml/_internal/io_descriptors/pandas.py
sauyon: should those be unified or do they need to be distinct?

aarnphm: distinct

bentoml/_internal/types.py
aarnphm: pyright was angry b4 that

bentoml/_internal/io_descriptors/pandas.py
aarnphm: technically this should be covered in the tests, but it didn't for some reason

bentoml/_internal/io_descriptors/multipart.py
```diff
@@ -83,7 +144,35 @@ async def from_http_request(self, request: Request) -> MultipartIO:
...
+    async def to_http_response(
+        self, obj: t.Dict[str, "np.ndarray[t.Any, np.dtype[t.Any]]"]
+    ) -> Response:
+        ...
+
+    @t.overload
+    async def to_http_response(self, obj: t.Dict[str, "FileLike"]) -> Response:
+        ...
+
+    @t.overload
```
bojiang, line 169: we should have just Union here rather than overloads?

bentoml/_internal/io_descriptors/base.py
```diff
@@ -1,12 +1,52 @@
...
+from ..types import FileLike
+
+if TYPE_CHECKING:
+    import numpy as np
+    import pandas as pd
+    import PIL.Image
+    from pydantic import BaseModel
+
+
+JSONType = t.Union[str, t.Dict[str, t.Any], "BaseModel"]
```
bojiang, line 18: BaseModel?

bentoml/_internal/io_descriptors/multipart.py
```diff
@@ -10,62 +10,123 @@
...
+    Args:
+        inputs (`Dict[str, IODescriptor]`):
+            Dictionary consisting keys as inputs definition for a Multipart request/response, values
+             as IODescriptor supported by BentoML. Currently we support Image, NumpyNdarray, PandasDataFrame,
+             PandasSeries, Text, File.
+
+    .. notes::
+        Make sure to match your input params in your API function to the keys defined under `Multipart`::
+
+            ┌───────────────────────────────────────────────────────┐
```
bojiang, line 89: 👍

bentoml/_internal/io_descriptors/base.py
aarnphm: oops my bad forgot to remove it

bentoml/_internal/io_descriptors/multipart.py
aarnphm: for some reason pyright make a false negative with Union type

bentoml/_internal/io_descriptors/multipart.py
aarnphm: actually when writing code if users do this then they already show it to them

aarnphm: no need to cover since type checker should detect this when users write their code. This acts as safety net

bentoml/_internal/io_descriptors/base.py
```diff
@@ -1,44 +1,116 @@
...
+            if len(args) != 0:
+                default_params.update(zip_longest(args, defaults))
+            if spec.varkw is not None:
+                default_params[spec.varkw] = spec.kwonlydefaults
+            setattr(self, "__io_default_params__", default_params)
 
-    def __repr__(self) -> str:
-        return f"{self.__class__.__name__}"
+    def __str__(self) -> str:
+        default_params = getattr(self, "__io_default_params__")
```
bojiang, line 73: I guess u want `getattr(self, "__io_default_params__", None)`?
The current will raise the same exception if `__io_default_params__ ` doesn't exist

aarnphm: fixed

bentoml/_internal/io_descriptors/base.py
```diff
@@ -1,44 +1,116 @@
...
+            if spec.varkw is not None:
+                default_params[spec.varkw] = spec.kwonlydefaults
+            setattr(self, "__io_default_params__", default_params)
 
-    def __repr__(self) -> str:
-        return f"{self.__class__.__name__}"
+    def __str__(self) -> str:
+        default_params = getattr(self, "__io_default_params__", None)
+        filtered: t.List[str] = []
+        for k, v in self.__dict__.items():
```
bojiang, line 75: it might be simpler if implemented in __new__

bentoml/_internal/io_descriptors/json.py
```diff
@@ -42,7 +41,7 @@ def default(self, o: _SerializableObj) -> t.Any:
         if isinstance(o, np.ndarray):
             return o.tolist()
 
-        if any(isinstance(o, i) for i in [pd.DataFrame, pd.Series]):
+        if isinstance(o, (pd.DataFrame, pd.Series)):
             raise BentoMLException(
```
parano, line 45: Just noticed this exception -why did we make this change? why not just try converting it to JSON here? it is quite common to return a dictionary that contains dataframe and it would be nice to just support that - PandasDataFrame doesn't really handle that case neither

bentoml/_internal/io_descriptors/json.py
parano: maybe we should consider returning the pydantic obj instead of json_obj here? or make that an option?  let's have a follow up discussion

bentoml/_internal/io_descriptors/multipart.py
```diff
@@ -10,62 +10,140 @@
...
+        input_spec = Multipart(arr=NumpyNdarray(), annotations=JSON())
+        output_spec = Multipart(output=NumpyNdarray(), result=JSON())
+
+        @svc.api(input=input_spec, output=output_spec)
+        def predict(arr, annotations):
+            res = runner.run(arr)
+            return {"output":res, "result":annotations}
+
+    Users then can then serve this service with `bentoml serve`::
+        % bentoml serve ./sklearn_svc.py:svc --auto-reload
```
parano, line 51: `s/--auto-reload/--reload/`

bentoml/_internal/io_descriptors/json.py
bojiang: how about
```
@api(output=Multipart(df=DataFrame()))
def api():
    return {"df": df}

```

bojiang: shall we encourage this usage since we already supported

```
@api(output=Multipart(df=DataFrame()))
def api():
    return {"df": df}

```
?

bentoml/_internal/io_descriptors/multipart.py
bojiang:
```suggestion
        % bentoml serve ./sklearn_svc.py:svc --reload
```

bojiang (approved):

bentoml/_internal/io_descriptors/json.py
parano: it's quite different, right? one is a multi-part response vs. a simple JSON response

bentoml/_internal/io_descriptors/base.py
```diff
@@ -1,44 +1,86 @@
...
 
     HTTP_METHODS = ["POST"]
+    _init_str: str = ""
+
+    def __new__(cls: t.Type[_T], *args: t.Any, **kwargs: t.Any) -> _T:
+        self = super().__new__(cls)
+        arg_strs = tuple(repr(i) for i in args) + tuple(
+            f'{k}={repr(v)}' for k, v in kwargs.items()
+        )
+        setattr(self, "_init_str", f"{cls.__name__}({', '.join(arg_strs)})")
```
parano, line 51: should we do anything for IO descriptors that are created via `from_sample`?

aarnphm: maybe add a attribute `_from_sample` = True?

aarnphm: we can save it as an example, the present it on openapi as a $ref to that example

tests/integration/conftest.py
```diff
@@ -21,10 +40,61 @@ def pytest_collection_modifyitems(config, items):
...
+def fixture_modelstore(tmp_path_factory: "TempPathFactory") -> "ModelStore":
     # we need to get consistent cache folder, thus tmpdir is not usable here
     # NOTE: after using modelstore, also use `delete_cache_model` to remove model after
     #  load tests.
     path = tmp_path_factory.mktemp("bentoml")
     return ModelStore(path)
+
+
+@pytest.fixture(scope="session", name="pipeline")
+def fixture_pipeline(modelstore: "ModelStore") -> Pipeline:
```
parano, line 53: this seems way too much complexity for something that can be done in a few lines of code, can we just use a utility function instead?

tests/integration/frameworks/test_catboost_impl.py
```diff
@@ -1,23 +1,23 @@
...
-if t.TYPE_CHECKING:
-    from bentoml._internal.models import ModelStore
+if TYPE_CHECKING:
+    from bentoml._internal.models import Model, ModelStore
 
-TEST_MODEL_NAME = __name__.split(".")[-1]
 
-
-def catboost_model() -> cbt.core.CatBoostClassifier:
+def create_catboost_model() -> cbt.core.CatBoostClassifier:
```
parano, line 20: let's modify this to `create_catboost_model(save_params, metadata)` and replace the `pipleline` fixture?

tests/integration/conftest.py
parano: I think in general fixtures are better fitted for initializing service/object and states. For a reusable "process", simple function calls are easier to manage

bentoml/detectron.py
```diff
@@ -59,18 +67,18 @@ def load(
             "my_detectron_model:20201012_DE43A2")
     """  # noqa
 
-    model = model_store.get(tag)
```
parano, line 62: what's the reason for this chagne?

bentoml/detectron.py
```diff
@@ -208,33 +219,44 @@ def num_replica(self) -> int:
...
             device = "cpu"
         self._model = load(self._tag, device, self._model_store)
         self._predict_fn = getattr(self._model, self._predict_fn_name)
 
-    # pylint: disable=arguments-differ
-    def _run_batch(  # type: ignore[override]
-        self, input_data: t.Union[np.ndarray, torch.Tensor]
-    ) -> np.ndarray:
-        images = np.split(input_data, input_data.shape[0], 0)
+    def _run_batch(
```
parano, line 231: should we move this to a separate PR?

bentoml/sklearn.py
```diff
@@ -137,31 +147,32 @@ def __init__(
...
     def num_concurrency_per_replica(self) -> int:
-        # NOTE: sklearn doesn't use GPU, so return max. no. of CPU's.
-        return int(round(self.resource_quota.cpu))
+        return 1
 
     @property
     def num_replica(self) -> int:
-        # NOTE: sklearn doesn't use GPU, so just return 1.
-        return 1
+        return int(round(self.resource_quota.cpu))
```
parano, line 154: with `parallel_backend`, this should just return 1

bentoml/sklearn.py
```diff
@@ -137,31 +147,32 @@ def __init__(
 
     @property
     def num_concurrency_per_replica(self) -> int:
-        # NOTE: sklearn doesn't use GPU, so return max. no. of CPU's.
-        return int(round(self.resource_quota.cpu))
+        return 1
```
parano, line 150: also need to change this back

bentoml/sklearn.py
parano: for this kind of changes it is better to put in separate PR for visibility

bentoml/sklearn.py
```diff
@@ -137,31 +147,32 @@ def __init__(
...
     @property
     def required_models(self) -> t.List[Tag]:
         return [self._model_info.tag]
 
-    # pylint: disable=arguments-differ,attribute-defined-outside-init
-    def _setup(self) -> None:  # type: ignore[override]
+    # pylint: disable=attribute-defined-outside-init
+    def _setup(self) -> None:
         self._model = joblib.load(filename=self._model_file)
+        self._infer_func = getattr(self._model, self._function_name)
```
parano, line 163: move this to a separate class method? e.g.:

```python
def _infer(self, *args, **kwargs):
    with self._parallel_ctx:
          return getattr(self._model, self._function_name)(*args, **kwargs)
```

requirements/dev-requirements.txt
```diff
@@ -5,6 +5,7 @@ setuptools
 twine
 wheel
 pydantic
+attrs>=21.1
```
parano, line 8: duplicated below in line 13

scripts/Dockerfile-checker
```diff
@@ -36,6 +36,10 @@ RUN npm i -g npm@^7 yarn pyright
 
 RUN python -m pip install -U pip setuptools
 
+ENV VIRTUAL_ENV=/opt/venv
```
parano, line 39: I thought we are getting rid of the checker image? is it still being used?

scripts/ci/config.yml
```diff
@@ -24,6 +24,7 @@ catboost:
   <<: *tmpl
   dependencies:
     - "catboost"
+    - "psutil"
```
parano, line 27: already a bentoml dependency

parano:
```suggestion
```

requirements/dev-requirements.txt
parano:
```suggestion
```

bentoml/sklearn.py
```diff
@@ -137,31 +147,32 @@ def __init__(
```
parano, line 147: we should also change the parallel_backend mode above, from 'threading' to 'loky': https://scikit-learn.org/stable/modules/generated/sklearn.utils.parallel_backend.html

bentoml/_internal/io_descriptors/base.py
parano: that would be really nice

bentoml/_internal/io_descriptors/json.py
```diff
@@ -42,7 +40,7 @@ def default(self, o: _SerializableObj) -> t.Any:
         if isinstance(o, np.ndarray):
             return o.tolist()
 
-        if any(isinstance(o, i) for i in [pd.DataFrame, pd.Series]):
+        if isinstance(o, (pd.DataFrame, pd.Series)):
```
parano, line 43:
```
if pd is not None:
           if isinstance(o, (pd.DataFrame, pd.Series)):
                       return o.to_json(orient=default_orient)
```

bentoml/_internal/io_descriptors/json.py
```diff
@@ -42,7 +40,7 @@ def default(self, o: _SerializableObj) -> t.Any:
...
             raise BentoMLException(
```
parano, line 44: should we also add support for returning a pydantic model here?

bentoml/_internal/io_descriptors/json.py
parano: just discussed with @bojiang, we all agree that it should just return the pydantic object in this case, cc @aarnphm

bentoml/_internal/io_descriptors/json.py
aarnphm: I don't think that is the job of a serializer. we should just convert a pydantic model to json

bentoml/_internal/io_descriptors/json.py
aarnphm: where do we defin this default_orient?

bentoml/_internal/io_descriptors/json.py
parano: yes that's that I meant, if it's a pydantic instance here, convert it to dict

aarnphm: it should do it now

bentoml/_internal/io_descriptors/pandas.py
```diff
@@ -124,6 +160,9 @@ def __init__(
         enforce_dtype: bool = False,
         shape: t.Optional[t.Tuple[int, ...]] = None,
         enforce_shape: bool = False,
+        mime_type: t.Union[
+            Literal["text/csv"], Literal["application/json"]
```
parano, line 164: "text/csv" doesn't work with `read_json` call below right?

bentoml/_internal/io_descriptors/pandas.py
```diff
@@ -151,6 +196,18 @@ async def from_http_request(self, request: Request) -> "pd.DataFrame":
...
+        if mime_type not in ["application/json", "text/csv"]:
+            raise BentoMLException(
+                f"mimetype {mime_type} is not supported, BentoML will only"
+                " supports `application/json` or `text/csv`"
+            )
+        if mime_type != self._mime_type:
+            logger.warning(
+                f"Current mimetype is {self._mime_type}, while receiving mimetype"
+                f" to be {mime_type}. This will lead to mismatch OpenAPI specification."
+            )
```
parano, line 210: if we were to support `text/csv`, we also need to change l219 below

bentoml/_internal/io_descriptors/pandas.py
```diff
@@ -368,18 +425,29 @@ def __init__(
...
+    def openapi_request_schema(self) -> t.Dict[str, t.Any]:
+        """Returns OpenAPI schema for incoming requests"""
+        return {self._mime_type: {"schema": self.openapi_schema_type()}}
+
+    def openapi_responses_schema(self) -> t.Dict[str, t.Any]:
+        """Returns OpenAPI schema for outcoming responses"""
+        return {self._mime_type: {"schema": self.openapi_schema_type()}}
 
-    async def from_http_request(self, request: Request) -> "pd.Series":
+    async def from_http_request(self, request: Request) -> "pd.Series[t.Any]":
```
parano, line 450: same here in L470

bentoml/_internal/io_descriptors/text.py
```diff
@@ -56,17 +55,23 @@ def predict(input_arr):
         IO Descriptor that represents strings type.
     """
 
-    @staticmethod
-    def openapi_schema() -> t.Dict[str, t.Dict[str, t.Any]]:
-        return {MIME_TYPE: {"schema": {"type": "string"}}}
+    def __init__(self, *args: t.Any, **kwargs: t.Any) -> None:
+        if len(args) > 0 or len(kwargs) > 0:
+            raise EnvironmentError(
```
parano, line 60: this should be a TypeError?

parano: can we just remove the `__init__` here?

bentoml/_internal/io_descriptors/pandas.py
parano: we can do that in a follow up PR

bentoml/_internal/io_descriptors/text.py
```diff
@@ -56,17 +55,23 @@ def predict(input_arr):
...
+                f"{self.__class__.__name__} is not designed to take any arguments."
+                " You should only call `Text()` instead."
+            )
+
```
parano, line 64:
```suggestion
```

PR #2051, opened 2021-11-24
feat: build refactor and build with bentofile.yaml

parano:
## Checklist:


bentoml/_internal/bento/README.md
```diff
@@ -1,11 +1,155 @@
- # Bento
+# Bento
```
parano, line 1: cc @ssheng we can probably move this to user-facing docs somewhere

bentoml/_internal/bento/README.md
```diff
@@ -1,11 +1,157 @@
...
+```
+
+
+When running the CLI command `bentoml build` without any parameter, it will look for
+a `bentofile.yaml` file in current directory and use it as build target. User may also
+specify which build file to use with the `-f` option, e.g.:
+
+```bash
+bentoml build -f ./bentofile-proj2.yaml
+```
```
ssheng, line 131: Is there a use case for having multiple build files?

parano: yes, when there are multiple bento service defined in the same python project

PR #2053, opened 2021-11-24
fix: bento metadata

dtemir:
## Description

When running
```python
>>> import bentoml
>>> bentoml.get("iris_classifier:latest")
```
I would get two errors: 
```bash
....
  File "/home/damir/Git/BentoML/bentoml/_internal/bento/bento.py", line 170, in info
    return BentoInfo.from_yaml_file(bento_yaml)
  File "/home/damir/Git/BentoML/bentoml/_internal/bento/bento.py", line 273, in from_yaml_file
    yaml_content["creation_time"]
TypeError: fromisoformat: argument must be str
```
and then having fixed that, it would throw
```bash
...
  File "/home/damir/Git/BentoML/bentoml/_internal/bento/bento.py", line 281, in from_yaml_file
    bento_info = cls(**yaml_content)
  File "<attrs generated init bentoml._internal.bento.bento.BentoInfo>", line 4, in __init__
  File "/home/damir/Git/BentoML/bentoml/_internal/bento/bento.py", line 248, in <lambda>
    service: str = attr.field(converter=lambda svc: svc._import_str)  # type: ignore[reportPrivateUsage]
AttributeError: 'str' object has no attribute '_import_str'
```
So I made some small changes to make it work, but let me know if you think it should be dealt with differently.

## Checklist:

### DIFF
--- a/bentoml/_internal/bento/bento.py
+++ b/bentoml/_internal/bento/bento.py
@@ -245,7 +245,7 @@ class BentoStore(Store[SysPathBento]):
 @attr.define(repr=False, frozen=True)
 class BentoInfo:
     tag: Tag
-    service: str = attr.field(converter=lambda svc: svc._import_str)  # type: ignore[reportPrivateUsage]
+    service: str  # type: ignore[reportPrivateUsage]
     labels: t.Dict[str, t.Any]  # TODO: validate user-provide labels
     models: t.List[str]  # TODO: populate with model & framework info
     bentoml_version: str = BENTOML_VERSION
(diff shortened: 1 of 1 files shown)


bentoml/_internal/bento/bento.py
```diff
@@ -245,7 +245,7 @@ def __init__(self, base_path: PathType):
 @attr.define(repr=False, frozen=True)
 class BentoInfo:
     tag: Tag
-    service: str = attr.field(converter=lambda svc: svc._import_str)  # type: ignore[reportPrivateUsage]
+    service: str  # type: ignore[reportPrivateUsage]
```
aarnphm, line 248: rejected this. This should be fixed in #2049

aarnphm (commented):
rejected

bentoml/_internal/bento/bento.py
dtemir: oh didn't see that one, i'll close the PR

PR #2054, opened 2021-11-24
feat(typing): stubs and ML framework

aarnphm:
Ugh big stubs PR

depends on #2050 

We will ignore stubbing library that doesn't have type supports: PyTorch, Tensorflow, paddlpaddle, etc.

If library have partial stubs, we will try to improve it: Xgboost, transformers with our own copy of stubs, until the library have better type support

If library have full type support then hurray 😄 

Frameworks that hasn't fixed yet:
- H2O
- PySpark
- pycaret (dc library :))
- Gluon
- Fasttext
- FastAI
- EvalML
- EasyOCR
- Detectron2
- CatBoost

- [ ] added stubs from numpy main repo. The main reason for this is for backward compatibility with numpy <1.20. Major framework like tensorflow will still use numpy <1.20, where the majority of type were introduced. This should be in sync with the main numpy repo until the majority of frameworks supports numpy 1.20+


bentoml/_internal/runner/runner.py
```diff
@@ -105,7 +105,7 @@ def required_models(self) -> t.List[Tag]:
         return []
 
     @abstractmethod
-    def _setup(self, **kwargs: t.Any) -> None:
+    def _setup(self) -> None:
```
aarnphm, line 108: @bojiang can you confirm this won't break anything? Test does pass. The reason behind this is that `_setup()` is not user facing, so anything a custom runner needs to pass into setup can be passed via `__init__`

parano: I think this is correct, setup should not take additional arguments

bentoml/xgboost.py
```diff
@@ -225,21 +237,22 @@ def _setup_booster_params(
...
-    # pylint: disable=arguments-differ
-    def _run_batch(  # type: ignore[override]
-        self, input_data: t.Union[np.ndarray, "pd.DataFrame", xgb.DMatrix]
-    ) -> "np.ndarray":
-        if not isinstance(input_data, xgb.DMatrix):
-            input_data = xgb.DMatrix(input_data)
-        res = self._predict_fn(input_data)
-        return np.asarray(res)
+    def _run_batch(
+        self, *args: t.Union[AnyNdarray, "DataFrame", xgb.DMatrix], **kwargs: str
```
parano, line 248: what's the kwargs here for?

bentoml/xgboost.py
```diff
@@ -225,21 +237,22 @@ def _setup_booster_params(
...
+    ) -> AnyNdarray:
+        params = Params[t.Union[AnyNdarray, "DataFrame", xgb.DMatrix]](*args, **kwargs)
+        params = params.map(
+            lambda x: xgb.DMatrix(x) if not isinstance(x, xgb.DMatrix) else x
+        )
+        res = self._predict_fn(*params.args, **params.kwargs)
+        return t.cast(AnyNdarray, np.asarray(res))  # type: ignore[reportUnknownMemberType]
```
parano, line 255: why do we need t.cast here?

scripts/tools/formatter.sh
```diff
@@ -1,5 +1,14 @@
 #!/usr/bin/env bash
 
-black --config ./pyproject.toml bentoml/ tests/ docker/ typings/
+GIT_ROOT=$(git rev-parse --show-toplevel)
 
-isort bentoml/ tests/ docker/ typings/
+cd "$GIT_ROOT" || exit
+
+source "$GIT_ROOT/scripts/ci/helpers.sh"
```
parano, line 7: is this used here?

.gitignore
```diff
@@ -109,3 +109,4 @@ tests/integration/data
 
 # setuptools_scm generated version file
 bentoml/_version.py
+pyminifier
```
parano, line 112: add Pyminifier to dev dependencies?

bentoml/_internal/server/service_app.py
```diff
@@ -149,10 +148,9 @@ async def index_view_func(
             media_type="text/html",
         )
 
+    # pylint: disable=unused-argument
```
parano, line 151: I think this will disable unused-argument check for everything below the lin in this file/

.gitignore
aarnphm: ugh no, its a python2 dependencies. I have a scripts to do that under `./scripts/tools/stubs_cleanup.sh`

bentoml/xgboost.py
aarnphm: asarray types is impartial, actually the cast is not necessary. I will just remove it

scripts/tools/formatter.sh
aarnphm: updated to the mian branch

bentoml/_internal/server/service_app.py
aarnphm: should only be for that function?

bojiang: if unused, just use `_` as the arg name.

bentoml/mlflow.py
```diff
@@ -211,15 +218,14 @@ def num_concurrency_per_replica(self) -> int:
     def num_replica(self) -> int:
         return 1
 
-    # pylint: disable=arguments-differ,attribute-defined-outside-init
-    def _setup(self) -> None:  # type: ignore[override]
+    # pylint: disable=attribute-defined-outside-init,used-before-assignment
```
bojiang, line 221: used-before-assignment !!?

bojiang (dismissed):

bojiang:
- attribute-defined-outside-init 
- used-before-assignment

They should not be ignored anytime.

aarnphm:
> * attribute-defined-outside-init
> 
>     * used-before-assignment
> 
> 
> They should not be ignored anytime.

only ignored in same of the frameworks since for `_setup()` we are setting up models and predict function there

.gitignore
parano: looking through pyminifier's GitHub issues, I'm not so confident about using it, could remove it from this PR?

aarnphm:
refers to #2085 and #2087

PR #2057, opened 2021-11-28
Fix: additional model tags error and attribution error on bento._info

parano:
## Checklist:

### DIFF
--- a/bentoml/_internal/bento/bento.py
+++ b/bentoml/_internal/bento/bento.py
@@ -79,12 +79,12 @@ def _Bento_set_fs(bento: "Bento", _: t.Any, new_fs: "FS") -> "FS":
     return new_fs
 
 
-@attr.define(repr=False)
+@attr.define(repr=False, auto_attribs=False)
 class Bento(StoreItem):
-    _tag: Tag
+    _tag: Tag = attr.field()
     _fs: "FS" = attr.field(on_setattr=_Bento_set_fs)
-    _model_store: ModelStore
 
+    _model_store: ModelStore
     _info: t.Optional["BentoInfo"] = None
     _doc: t.Optional[str] = None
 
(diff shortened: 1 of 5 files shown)


bentoml/_internal/types.py
```diff
@@ -89,8 +89,6 @@ def from_taglike(cls, taglike: t.Union["Tag", str]) -> "Tag":
             return taglike
         return cls.from_str(taglike)
 
-        raise TypeError(f"can't make a Tag from {type(taglike)}")
```
parano, line 92: @sauyon what was the intention of this line? ok to remove?

bentoml/_internal/bento/bento.py
```diff
@@ -79,12 +79,12 @@ def _Bento_set_fs(bento: "Bento", _: t.Any, new_fs: "FS") -> "FS":
     return new_fs
 
 
-@attr.define(repr=False)
+@attr.define(repr=False, auto_attribs=False)
```
parano, line 82: @sauyon `bentoml.get` currently raise an AttributionError and failed to access `_info`, the same happens to `_doc`. setting `auto_attribs=False` solved the issue but I'm not entirely sure why that was happening. seem like a bug in attrs to me.

bentoml/sklearn.py
```diff
@@ -36,14 +36,10 @@ def _get_model_info(
...
     if model.info.module != __name__:
-        if model.info.module == "bentoml.mlflow":
-            pass
-        else:
-            raise BentoMLException(  # pragma: no cover
-                f"Model {tag} was saved with module"
-                f" {model.info.module}, failed loading"
-                f" with {__name__}."
-            )
+        raise BentoMLException(  # pragma: no cover
```
parano, line 39: cc @aarnphm - this check should be removed right?

aarnphm: we can just have it so that users can't load untrusted pickle file, since for sklearn we don't even use sklearn at all

parano: We still have the module check, I mean the removed part that allows loading bentoml.mlflow saved model with bentoml.sklearn module

aarnphm: oh it should be removed in some of the other PR i had

PR #2062, opened 2021-11-30
fix(internal): don't move models

sauyon:
At some point we should think about how exactly this should be done, but for now this should be fine.

### DIFF
--- a/bentoml/_internal/models/model.py
+++ b/bentoml/_internal/models/model.py
@@ -2,6 +2,7 @@ import logging
 import typing as t
 from collections import UserDict
 from datetime import datetime, timezone
+from typing import TYPE_CHECKING
 
 import attr
 import fs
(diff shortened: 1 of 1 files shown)


bentoml/_internal/models/model.py
```diff
@@ -163,25 +158,6 @@ def path(self) -> str:
     def path_of(self, item: str) -> str:
         return self._fs.getsyspath(item)
 
-    def save(
```
aarnphm, line 166: Isn't this the same with line 78?

sauyon: No, it's OSFS-specific and runs `shutil.move`.

parano: I saw some similar code in `bento.py`, should we remove that as well?

parano: and could you fix the CI/codestyle_check?

sauyon: I thought I'd only do this one, since technically it's nice that the files don't get copied and we aren't doing save for bentos anywhere else, but I can.

PR #2064, opened 2021-11-30
fix: support load from bento dir

parano:
## Checklist:

### DIFF
--- a/bentoml/_internal/server/__init__.py
+++ b/bentoml/_internal/server/__init__.py
@@ -16,15 +16,15 @@ logger = logging.getLogger(__name__)
 
 @inject
 def serve_development(
-    bento_path_or_tag: str,
+    bento_identifier: str,
     working_dir: str,
     port: int = Provide[BentoServerContainer.config.port],
     with_ngrok: bool = False,
     reload: bool = False,
     reload_delay: float = 0.25,
 ) -> None:
-    working_dir = os.path.realpath(working_dir)
-    svc = load(bento_path_or_tag, working_dir=working_dir)
+    working_dir = os.path.realpath(os.path.expanduser(working_dir))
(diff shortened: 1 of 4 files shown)


bentoml/_internal/server/__init__.py
```diff
@@ -44,17 +44,19 @@ def serve_development(
...
-    dev_server_cmd = f'import bentoml._internal.server; bentoml._internal.server._start_dev_api_server("{bento_path_or_tag}", {port}, working_dir="{working_dir}", reload={reload}, reload_delay={reload_delay}, instance_id=$(CIRCUS.WID))'
+    dev_server_cmd = f'import bentoml._internal.server; bentoml._internal.server._start_dev_api_server("{bento_identifier}", {port}, working_dir="{working_dir}", reload={reload}, reload_delay={reload_delay}, instance_id=$(CIRCUS.WID))'
     watchers.append(
         Watcher(
             name="ngrok",
             cmd=f"{sys.executable} -c '{dev_server_cmd}'",
             env=env,
             numprocesses=1,
             stop_children=True,
+            working_dir=working_dir,
```
parano, line 59: cc @bojiang - we will need this line for `serve_production` below as well, will leave that to the other PR that you are working on

bojiang: do we still need `working_dir` in dev_server_cmd if specified here

bojiang: record of pm: We have both `bento_identifier` and `working_dir`, plus `bento_identifier` could also be a path to a bento, which the parent process may change to that directory during load. Because of that, we may align the starting work dir of parent and child processes to avoid redundant path navigation, and always let load command change the working directory.

PR #2065, opened 2021-11-30
fix(types): keras+tf refactor LazyLoader with type-safe

aarnphm:

## Description

A few things tackle in this PR
- depends on #2054 for stubs
- Move keras and tensorflow into `_internal.frameworks`, then lazy import it in `__init__.py`
- adds stubs for type hint since static type checker won't be able to know the type
image
image

## Checklist:


bentoml/keras.py
```diff
@@ -187,39 +176,59 @@ def __init__(
...
-        with self._device:
+        raw_predict_fn = getattr(self._model, self._predict_fn_name)
+        self._predict_fn = functools.partial(raw_predict_fn, **self._partial_kwargs)
+
+    def _run_batch(
+        self,
+        *args: t.Union[
+            t.List[t.Union[int, float]], "np.ndarray[t.Any, np.dtype[t.Any]]", tf.Tensor
+        ],
+        **kwargs: t.Any,
```
larme, line 198: @aarnphm  does keras really need this? The `predict` should only accept one input (multiple inputs are inside a list). So we may just need partial kwargs. And here partial kwargs is not flexible like in pytorch's case, just some configuration parameter for keras's predict function. So in my pr I renamed it into predict_kwargs (#2069)

aarnphm: noted. Was trying to make pyright happy :)

tests/integration/frameworks/test_keras_impl.py
```diff
@@ -54,7 +54,7 @@ def test_keras_save_load(model, kwargs, modelstore):
     if not TF2:
         # Initialize variables in the graph/model
         bentoml.keras._sess.run(tf.global_variables_initializer())
-        with bentoml.keras._default_sess():
+        with bentoml.keras._sess.as_default():
```
parano, line 57: is this line required for our users as well?

parano: it would be great to add some basic sample usage to the docstring

aarnphm: for tf1 case then yes. i will add some docstring for that one.

aarnphm: this can be accessed via `get_session()`

bojiang:
_Had a discussion with Aaron on Slack._

* keep using LazyLoader combine with `if TYPE_CHECKING`. it's the best solution so far, before a typed LazyLoader implementation with pyright plugin
* `overload` is not a better `Union`/`Optional`. Use it only when the possible combination of params are less than cartesian product

bentoml/__init__.py
```diff
@@ -13,11 +15,18 @@
 from ._internal.service import Service
 from ._internal.service.loader import load
 from ._internal.types import Tag
+from ._internal.utils import LazyLoader as _LazyLoader
 
 # bento APIs are top-level
 from .bentos import list  # pylint: disable=W0622
 from .bentos import build, delete, export_bento, get, import_bento
 
+if TYPE_CHECKING:
```
aarnphm, line 24: @parano

bentoml/_internal/utils/tensorflow.py
```diff
@@ -1,14 +1,17 @@
-import logging
+import importlib.util
```
aarnphm, line 1: Also I should fix this, we shouldn't rely on this https://github.com/asottile/scratch/wiki/ImportSideEffects

aarnphm: However, autoflake `--remove-unused-imports` seems to rely on this behaviour

aarnphm: we don't use autoflake but there might be scenario where developers use it. Put it here for future reference

parano (approved):

bentoml/_internal/utils/tensorflow.py
parano: does it help if we move this to the get_tf_version function below?

PR #2066, opened 2021-11-30
chore(pyright-safe): add Params for pytorch

aarnphm:

## Checklist:


bentoml/pytorch.py
```diff
@@ -35,7 +30,16 @@
...
+try:
+    import importlib.metadata as importlib_metadata
+except ImportError:
+    import importlib_metadata
+
+_ModelType = t.Union["torch.nn.Module", "torch.jit.ScriptModule"]  # type: ignore[reportPrivateUsage]
+
+_torch_version = importlib_metadata.version("torch")
+
+infer_mode_compat = _torch_version.startswith("1.9")
```
larme, line 43: why we want to decide pytorch version this way?

bentoml/pytorch_lightning.py
```diff
@@ -18,18 +19,27 @@
 """
 
 if TYPE_CHECKING:  # pragma: no cover
+    import pytorch_lightning as pl
+
```
larme, line 23: considering this module is bentoml.pytorch_lightning, I think it's ok to import pytorch_lightning directly?

bentoml/pytorch.py
aarnphm: pyright fix

sauyon: We should just have a utility function that gets a module version.

bentoml/pytorch.py
```diff
@@ -35,7 +30,16 @@
...
```
parano, line 42: suggest renaming: infer_mode_compat -> `is_pytorch_1_9`

bentoml/pytorch.py
```diff
@@ -228,33 +227,32 @@ def _setup(self) -> None:  # type: ignore[override]
...
-                if isinstance(item, torch.Tensor):
-                    item = item.cuda()
-
+                item = item.cuda()
             return item
 
         params = params.map(_mapping)
 
-        res: torch.Tensor
         if infer_mode_compat:
```
parano, line 256:
```suggestion
        # inference mode is required for PyTorch version 1.9.*
        if _torch_version.startswith("1.9"):
```

bentoml/pytorch.py
parano, line 42:
```suggestion
```

bentoml/pytorch.py
```diff
@@ -67,20 +71,15 @@ def load(
         booster = bentoml.pytorch.load(
             'lit_classifier:20201012_DE43A2', device_id="cuda:0")
     """  # noqa
-    model = model_store.get(tag)
-    weight_file = model.path_of(f"{SAVE_NAMESPACE}{PT_EXT}")
+    model_info = model_store.get(tag)
+    weight_file = model_info.path_of(f"{SAVE_NAMESPACE}{PT_EXT}")
     # TorchScript Models are saved as zip files
     if zipfile.is_zipfile(weight_file):
```
parano, line 77: perhaps in a follow-up PR, could we store the model type somewhere in the model metadata YAML instead of relying on file format? that way is a bit more future proof - say if there's a new model format for PyTorch, and if it also uses zip file format. If it's well structured in the metadata file, we don't need to worry about backward compatibility.

parano: cc @larme

bentoml/pytorch_lightning.py
aarnphm: funny thing is we don't even use pytorch_lightning for this. Mostly just served for typing

bentoml/pytorch.py
```diff
@@ -228,34 +226,34 @@ def _setup(self) -> None:  # type: ignore[override]
...
         )
 
-    # pylint: disable=arguments-differ
     @torch.no_grad()
-    def _run_batch(  # type: ignore[override]
+    def _run_batch(
         self,
-        *args: t.Union[np.ndarray, torch.Tensor],
+        *args: t.Union["np.ndarray[t.Any, np.dtype[t.Any]]", torch.Tensor],
         **kwargs: str,
```
parano, line 237: @larme @aarnphm what's the reason for `**kwargs: str` here?

parano: it should be impossible to have an str type here, data container does not know how to aggregate a batch for it

bentoml/pytorch_lightning.py
```diff
@@ -54,13 +64,11 @@ def load(
...
-    weight_file = model.path_of(f"{SAVE_NAMESPACE}{PT_EXT}")
-    _load: t.Callable[[str], "pl.LightningModule"] = functools.partial(
-        torch.jit.load, map_location=device_id
-    )
-    return _load(weight_file)
+    """  # noqa: LN001
+    bentoml_model = model_store.get(tag)
+    weight_file = bentoml_model.path_of(f"{SAVE_NAMESPACE}{PT_EXT}")
+    model: "pl.LightningModule" = torch.jit.load(weight_file, map_location=device_id)  # type: ignore[reportPrivateImportUsage] # noqa: LN001
+    return model
```
larme, line 71: @parano here I make the name different because we have both BentoML Model and pytorch model

larme (approved):

PR #2080, opened 2021-12-02
fix(internal): fix Model and add tests

sauyon:


sauyon:
actually drafting this one because I need to rework yaml stuff.

sauyon:
Will do that later.

bentoml/models.py
```diff
@@ -82,8 +101,13 @@ def create(
...
 
-def load_runner(tag: t.Union[Tag, str]) -> "Runner":
-    model = get(tag)
+@inject
+def load_runner(
+    tag: t.Union[Tag, str],
+    *,
+    _model_store: "ModelStore" = Provide[BentoMLContainer.model_store],
+) -> "Runner":
+    model = get(tag, _model_store=_model_store)
```
parano, line 110: cc @larme

tests/unit/_internal/models/test_model.py
```diff
@@ -0,0 +1,164 @@
...
+from bentoml import Tag
+from bentoml.exceptions import BentoMLException
+from bentoml._internal.models.model import Model
+from bentoml._internal.models.model import ModelInfo
+from bentoml._internal.configuration import BENTOML_VERSION
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+expected_yaml = (
```
parano, line 19: would it be easier to write this with a multi-line string?

sauyon: probably, I wrote this originally as a bunch of f-strings before I realized that I needed to use format

tests/unit/_internal/models/test_model.py
```diff
@@ -0,0 +1,164 @@
...
+    with open(
+        sys_written_path, mode="w", encoding="utf-8", newline=""
+    ) as sys_written_file:
+        sys_written_file.write(sys_written_content)
+
+    with open(model_b.path_of("sys_written/file"), encoding="utf-8") as f:
+        assert f.read() == sys_written_content
+
+    b_tar_path = f"tar://{fs.path.join(str(tmpdir), 'modelb.tar')}"
+    model_b.export(b_tar_path)
```
parano, line 115: we we expect users to add `tar://` prefix when using import and export?

tests/unit/_internal/models/test_model.py
```diff
@@ -0,0 +1,164 @@
...
+    "  label: stringvalue\n"
+    "options:\n"
+    "  option:\n"
+    "    dictkey: dictvalue\n"
+    "metadata:\n"
+    "  metadatakey:\n"
+    "  - list\n"
+    "  - of\n"
+    "  - str\n"
+)
```
parano, line 38:
```suggestion
expected_yaml = """
name: test
version: version
bentoml_version: {bentoml_version}
creation_time: {creation_time}
api_version: v1
module: testmodule
context:
  ctx: 1
labels:
  label: stringvalue
options:
  option:
    dictkey: dictvalue
metadata:
  metadatakey:
  - list
  - of
  - str
"""
```

something like this?

parano (approved):

PR #2081, opened 2021-12-02
fix(internal): fix Store and add tests

sauyon:


sauyon:
drafting this because I think I need to rework the yaml representation code.

sauyon:
Er, wrong PR...

PR #2085, opened 2021-12-03
chore: stubs instruction

aarnphm:

## Checklist:


scripts/tools/install_stubs.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+
+LIBRARIES=(cloudpickle deepmerge fs huggingface_hub importlib_metadata joblib multipart pandas pathspec PIL starlette transformers xgboost yaml)
```
parano, line 3: should we have a requirements.txt file just for typing, that locks to a version of these libraries?

.gitignore
```diff
@@ -109,3 +109,5 @@ tests/integration/data
 
 # setuptools_scm generated version file
 bentoml/_version.py
+typings/pandas
+typings/numpy
```
parano, line 113: we will need to add all the other libs in the list below?

maybe let's ignore everything under `typings` here?  and when we actually need to commit some stub files to the repo, we can manually do that.

aarnphm: sure

scripts/tools/install_stubs.sh
aarnphm: sure we can do that

aarnphm: done

.gitignore
aarnphm: done

sauyon:
What was the point of generating stubs again?

aarnphm:
> What was the point of generating stubs again?

For local use with pyright.

sauyon:
Does pyright not pick up the types from the libraries? I was under the impression type stubs were only required if the typing was incomplete or non-existent.

aarnphm:
> Does pyright not pick up the types from the libraries? I was under the impression type stubs were only required if the typing was incomplete or non-existent.

not for most ml frameworks

DEVELOPMENT.md
```diff
@@ -302,6 +279,17 @@ BentoML also provides git hooks that developers can install with:
 ```bash
 make hooks
 ```
+## Stubs
+Refers to [Installation](https://github.com/microsoft/pyright#installation) to install pyright correctly.
+
+In order to make pyright function correctly one also need to run the following scripts alongside with the stubs provided
+ in the main repository.
+```bash
+./script/tools/install_stubs.sh
```
bojiang, line 288: where is the install_stubs.sh ?

scripts/ci/helpers.sh
```diff
@@ -9,6 +9,10 @@ PASS() {
     echo -e "$GREEN""[PASS]""$NC" "$*"
 }
 
+INFO() {
+    echo -e "$YELLOW""[INFO]""$NC" "$*"
```
sauyon, line 13: Personally find this to look nicer.

```suggestion
    echo -e "${YELLOW}[INFO]${NC}" "$*"
```

Or alternatively just

```suggestion
    echo -e "$YELLOW[INFO]$NC" "$*"
```

sauyon:
lol github didn't show the review comments I had on the nonexistent file

PR #2086, opened 2021-12-03
feat: add yatai cli

yetone:


bentoml/_internal/bento/bento.py
```diff
@@ -288,8 +315,77 @@ def export(self, path: str):
...
+                    def filter_(tar_info: tarfile.TarInfo) -> t.Optional[tarfile.TarInfo]:
+                        if tar_info.path == './models' or tar_info.path.startswith('./models/'):
+                            return None
+                        return tar_info
+                    tar.add(bento_dir_path, arcname='./', filter=filter_)
+            tar_io.seek(0, 0)
+            with yaspin(text=f'Starting upload bento {self.tag}'):
+                yatai_cli.start_upload_bento_version(bento_name=bento_obj.name, version=version)
+            file_size = tar_io.getbuffer().nbytes
+            with tqdm(desc=f'Uploading bento {self.tag}', total=file_size, unit='B', unit_scale=True, unit_divisor=1024) as progress_bar:
```
parano, line 366: should we just use the progress bar from Rich? https://github.com/willmcgugan/rich#rich-library to avoid adding another dependency?

bentoml/_internal/cli/bento_management.py
```diff
@@ -64,7 +77,19 @@ def list_bentos(bento_name, output, no_trunc):
...
+            "tag": str(bento.tag),
+            "path": bento.path,
+            "creation_time": bento.info.creation_time,
+        } for bento in sorted(bentos, key=lambda x: x.info.creation_time, reverse=True)]
+        if output == "json":
+            info = json.dumps(res, indent=2)
+        elif output == "yaml":
+            info = yaml.safe_dump(res, indent=2)
+        else:
+            info = tabulate([item.values() for item in res], headers=["Tag", "Path", "Creation Time"])
```
parano, line 91: similarly, let's use the table component from Rich instead of tabulate

bentoml/_internal/bento/bento.py
```diff
@@ -312,23 +408,65 @@ def from_Bento(bento: Bento) -> "SysPathBento":
...
+class BentoStore(RemoteStore[SysPathBento]):
+    @inject
+    def pull(self, tag: t.Union[str, Tag], *, force: bool = False,
+             model_store: "ModelStore" = Provide[BentoMLContainer.model_store]) -> RemoteItem:
+        if not force:
+            try:
+                bento = self.get(tag)
+                logger.info(f"Bento {tag} already exists locally, skipping pull")
+                return bento
+            except NotFound:
```
parano, line 437: Could we do a delete here when force is True? instead of adding it to all the store APIs?

The reason is in almost all other cases, `force` can be a dangerous behavior and we would never recommend users doing that. Although having it in the API parameter is tempting. I slightly prefer asking the user do a explicitly delete and then add, if they actually need the `force` behavior

bentoml/_internal/models/model.py
```diff
@@ -134,6 +150,70 @@ def export(self, path: str):
         fs.mirror.mirror(self._fs, out_fs, copy_if_newer=False)
         out_fs.close()
 
+    def push(self, *, force: bool = False):
+        yatai_cli = get_current_yatai_cli()
+        name = self.tag.name
+        version = self.tag.version
+        info = self.info
+        with yaspin(text=f'Fetching model {name}'):
```
parano, line 158: yaspin can be replaced with `Status` in Rich

setup.cfg
```diff
@@ -60,6 +60,13 @@ install_requires =
     typing-extensions>=3.7.4
     uvicorn
     importlib-metadata;python_version<"3.8"
+    simple-term-menu
```
parano, line 63: are we using simple-term-menu?

bentoml/_internal/yatai_client/schemas.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+datetime_field_kwargs = dict(metadata={'dataclasses_json': {
+    'encoder': datetime_encoder,
+    'decoder': datetime_decoder,
+    'mm_field': fields.DateTime(format='iso')
+}})
+
+
+@dataclass
```
parano, line 33: can we use `attrs` for this?  the rest of the framework has decided on attrs, and cattrs can replace `dataclasses_json`

bentoml/_internal/yatai_client/yatai.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def get_current_organization(self) -> Optional[OrganizationSchema]:
+        url = urljoin(self.endpoint, '/api/v1/current_org')
+        resp = self.session.get(url)
+        if self._is_not_found(resp):
+            return None
+        self._check_resp(resp)
+        return OrganizationSchema.from_json(resp.text)
+
+    def list_organizations(self, start: int, count: int) -> OrganizationListSchema:
```
parano, line 48: we are removing this API right?

bentoml/_internal/yatai_client/yatai.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def get_current_user(self) -> Optional[UserSchema]:
+        url = urljoin(self.endpoint, '/api/v1/auth/current')
+        resp = self.session.get(url)
+        if self._is_not_found(resp):
+            return None
+        self._check_resp(resp)
+        return UserSchema.from_json(resp.text)
+
+    def get_current_organization(self) -> Optional[OrganizationSchema]:
```
parano, line 40: also remove?

bentoml/_internal/yatai_client/schemas.py
```diff
@@ -0,0 +1,240 @@
...
+    items: List[OrganizationSchema]
+
+
+@dataclass
+class ClusterSchema(ResourceSchema):
+    description: str = None
+
+
+@dataclass
+class BentoSchema(ResourceSchema):
```
parano, line 98: We have plan to "flatten" the Bento & versions page right?  Let's also reflect that in the API and backend storage. There's no need for both BentoSchema and a BentoVersionSchema, it should be just one BentoSchema

bentoml/_internal/yatai_client/config.py
```diff
@@ -0,0 +1,82 @@
...
+    return Path.home() / '.yatai.yaml'
+
+
+@attr.define
+class YataiClientContext:
+    name: str
+    endpoint: str
+    api_token: str
+
+    def get_yatai_cli(self) -> YataiClient:
```
parano, line 25: rename to `get_yatai_client`? it might be a bit confusing since  `cli` usually means `command-line` by convention especially with python and click

bentoml/_internal/bento/bento.py
```diff
@@ -288,8 +315,77 @@ def export(self, path: str):
         out_fs = fs.open_fs(path, create=True, writeable=True)
         fs.mirror.mirror(self._fs, out_fs, copy_if_newer=False)
 
-    def push(self):
-        pass
+    @inject
+    def push(self, *, force: bool = False, model_store: "ModelStore" = Provide[BentoMLContainer.model_store]):
```
parano, line 319: Could we put bento push/pull and models push/pull in yatai client class?  

* yatai_rest_api_client: low-level API, methods directly mapping to REST APIs
* yatai_client: high-level user facing APIs: push_bento, pull_bento, push_model, pull_bento, login, swith_context,

bentoml/_internal/bento/bento.py
```diff
@@ -93,13 +107,46 @@ def __init__(self, tag: Tag, bento_fs: "FS"):
...
+                model_info = model_store.get(model_tag)
+                if model_info.tag not in seen_model_tags:
+                    seen_model_tags.add(model_info.tag)
+                    models.append(model_info)
+            except FileNotFoundError:
+                raise BentoMLException(
+                    f"Model {model_tag} not found in local model store"
+                )
+
+        logger.info(
```
parano, line 136: this logging message is specifically for the bento build process, it will be confusing showing it in `pull`

bentoml/_internal/bento/bento.py
```diff
@@ -80,7 +94,7 @@ def _Bento_set_fs(bento: "Bento", _: t.Any, new_fs: "FS") -> "FS":
 
 
 @attr.define(repr=False, auto_attribs=False)
-class Bento(StoreItem):
+class Bento(RemoteStoreItem):
```
parano, line 97: We should probably not do this - `bentoml.get` returns a `Bento` which means a local Bento, and naming it RemoteStoreItem doesn't seem right

bentoml/_internal/bento/bento.py
```diff
@@ -93,13 +107,46 @@ def __init__(self, tag: Tag, bento_fs: "FS"):
         self._fs = _Bento_set_fs(self, None, bento_fs)
         self.validate()
 
+    @classmethod
+    def get_download_url(cls, tag: Tag) -> str:
```
parano, line 111: let's just move this to YataiClient as a utility function

bentoml/_internal/yatai_client/schemas.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+@dataclass
+class BentoVersionSchema(ResourceSchema):
+    description: str = ''
+    version: str = ''
+    image_build_status: BentoVersionImageBuildStatus = ''
+    upload_status: BentoVersionUploadStatus = ''
+    upload_finished_reason: str = ''
+    presigned_s3_url: str = ''
```
parano, line 146: could we change the API to just `presigned_download_url`? that way we don't need to change this API when we start adding support for other cloud blob storage services (azure blob storage, google cloud storage etc)

bentoml/_internal/yatai_client/yatai.py
```diff
@@ -0,0 +1,161 @@
...
+        self._check_resp(resp)
+        return BentoVersionSchema.from_json(resp.text)
+
+    def presign_bento_version_s3_upload_url(self, bento_name: str, version: str) -> BentoVersionSchema:
+        url = urljoin(self.endpoint, f'/api/v1/bentos/{bento_name}/versions/{version}/presign_s3_upload_url')
+        resp = self.session.patch(url)
+        self._check_resp(resp)
+        return BentoVersionSchema.from_json(resp.text)
+
+    def presign_bento_version_s3_download_url(self, bento_name: str, version: str) -> BentoVersionSchema:
```
parano, line 91: rename: presign_bento_download_url

bentoml/_internal/yatai_client/yatai.py
```diff
@@ -0,0 +1,161 @@
...
+        self._check_resp(resp)
+        return BentoVersionSchema.from_json(resp.text)
+
+    def create_bento_version(self, bento_name: str, req: CreateBentoVersionSchema) -> BentoVersionSchema:
+        url = urljoin(self.endpoint, f'/api/v1/bentos/{bento_name}/versions')
+        resp = self.session.post(url, data=req.to_json())
+        self._check_resp(resp)
+        return BentoVersionSchema.from_json(resp.text)
+
+    def presign_bento_version_s3_upload_url(self, bento_name: str, version: str) -> BentoVersionSchema:
```
parano, line 85: rename: presign_bento_upload_url

bentoml/_internal/yatai_client/yatai.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def get_bento_version(self, bento_name: str, version: str) -> Optional[BentoVersionSchema]:
+        url = urljoin(self.endpoint, f'/api/v1/bentos/{bento_name}/versions/{version}')
+        resp = self.session.get(url)
+        if self._is_not_found(resp):
+            return None
+        self._check_resp(resp)
+        return BentoVersionSchema.from_json(resp.text)
+
+    def create_bento_version(self, bento_name: str, req: CreateBentoVersionSchema) -> BentoVersionSchema:
```
parano, line 79: rename: create_bento

bentoml/_internal/yatai_client/yatai.py
```diff
@@ -0,0 +1,161 @@
...
+        url = urljoin(self.endpoint, f'/api/v1/bentos/{bento_name}/versions')
```
parano, line 80: this endpoint should be just `/api/v1/bentos`

bentoml/_internal/models/model.py
```diff
@@ -46,6 +56,12 @@ def __eq__(self, other: "Model") -> bool:
     def __hash__(self) -> int:
         return hash(self._tag)
 
+    @classmethod
+    def get_download_url(cls, tag: Tag) -> str:
```
parano, line 60: same here, let's move this to yatai client `get_model_download_url(tag)`

bentoml/_internal/types.py
```diff
@@ -94,7 +94,7 @@ def from_str(cls, tag_str: str) -> "Tag":
         if ":" not in tag_str:
             return cls(tag_str, None)
         try:
-            name, version = tag_str.split(":")
+            name, _, version = tag_str.partition(":")
```
parano, line 97: 👍

bentoml/_internal/yatai_client/config.py
```diff
@@ -0,0 +1,82 @@
...
+import yaml
+
+from bentoml._internal.yatai_client.yatai import YataiClient
+from bentoml._internal.yatai_client.errors import YataiCliError
+
+default_context_name = 'default'
+
+
+def get_config_path() -> Path:
+    return Path.home() / '.yatai.yaml'
```
parano, line 16: could we put this file under BENTOML_HOME? https://github.com/bentoml/BentoML/blob/cc111ce/bentoml/_internal/configuration/containers.py#L31

bentoml/_internal/yatai_client/config.py
```diff
@@ -0,0 +1,82 @@
...
+    with open(get_config_path(), 'r') as f:
+        dct = yaml.safe_load(f)
+        return cattr.structure(dct, YataiClientConfig)
+
+
+def add_context(context: YataiClientContext) -> None:
+    config = get_config()
+    for idx, ctx in enumerate(config.contexts):
+        if ctx.name == context.name:
+            config.contexts[idx] = context
```
parano, line 68: add warning here? `logger.warning(f"Overriding existing Yatai context config: {ctx.name}")

bentoml/_internal/yatai_client/errors.py
```diff
@@ -0,0 +1,2 @@
+class YataiCliError(Exception):
```
parano, line 1: rename YataiClientError?  and should we just put this in `bentoml.exceptions`? which makes it "public" to the user, when using our Python API

bentoml/_internal/yatai_client/schemas.py
yetone: @parano I think leveling is just something to do on the front-end presentation, and blending Bento and BentoVersion together in the back-end abstraction is a very irrational approach

bentoml/_internal/bento/bento.py
yetone: Yes, I'm trying to use rich's progressbar, but the API of rich's progressbar is really stupid compared to tqdm

bentoml/_internal/yatai_client/schemas.py
yetone: Well, that's why I said bentoml would be confusing for new developers, because now bentoml has both dataclass and attrs

bentoml/_internal/yatai_client/yatai.py
yetone: We need this function to get the current organization from the api_token, which mainly serves to inform the user

bentoml/_internal/bento/bento.py
```diff
@@ -131,32 +159,12 @@ def create(
         bento_fs = fs.open_fs(f"temp://bentoml_bento_{svc.name}")
         ctx_fs = fs.open_fs(build_ctx)
 
-        model_tags = build_config.additional_models
+        model_tags = build_config.additional_models or []
```
parano, line 162: there's no need for `or []` here, see BuildConfig's `with_defaults`

bentoml/_internal/store.py
```diff
@@ -129,11 +132,9 @@ def register(self, tag: t.Union[str, Tag]):
         _tag = Tag.from_taglike(tag)
 
         item_path = _tag.path()
-        if self._fs.exists(item_path):
-            raise BentoMLException(
-                f"Item '{_tag}' already exists in the store {self._fs}"
-            )
-        self._fs.makedirs(item_path)
+        is_exists = self._fs.exists(item_path)
```
parano, line 135: let's change this back - it is expected behavior to raise that exception, it is not allowed to re-register with the same tag

bentoml/_internal/store.py
```diff
@@ -82,11 +82,14 @@ def list(self, tag: t.Optional[t.Union[Tag, str]] = None) -> t.List[Item]:
         else:
             return [self._get_item(_tag)] if self._fs.isdir(_tag.path()) else []
 
+    def get_item_from_fs(self, fs: FS) -> Item:
```
parano, line 85: no need for this change, simply use `Bento.from_fs` or `Model.from_fs`

bentoml/_internal/yatai_client/__init__.py
```diff
@@ -0,0 +1,263 @@
...
+            with fs.open_fs('temp://') as temp_fs:
+                for member in tar.getmembers():
+                    f = tar.extractfile(member)
+                    if f is None:
+                        continue
+                    p = Path(member.name)
+                    if p.parent != Path('.'):
+                        temp_fs.makedirs(str(p.parent), recreate=True)
+                    temp_fs.writebytes(member.name, f.read())
+                Bento.save_models(temp_fs, remote_bento.manifest.models, model_store=model_store)
```
parano, line 157: I'm not so sure about this change here, the logic in `save_models` is pretty specific for Bento creation. I'd suggest undo the refactoring around `save_models` method, and change this line to:

```
bento = Bento.from_fs(temp_fs)
for model_tag in remote_bento.manifest.models:
    copy_model_if_not_found(model_tag, src=bento.model_store, target=model_store)
```
And for the utility function:
```
def copy_model_if_not_found(model_tag, src, target):
    """copy a model from src model store to taget modelstore, and do nothing if the model tag
    already exist in target model store
    """
    try:
        target_model_store.get(model_tag):  # if model tag already found in target
        return
    except NotFound:
         pass
 
    model = src_model_store.get(model_tag)
    model.save(model_store=target_model_store)

```

bentoml/_internal/yatai_client/__init__.py
```diff
@@ -0,0 +1,263 @@
...
+                return bento_store.get_item_from_fs(temp_fs).save()
```
parano, line 158: we should pass in the bento_store parameter to save: `save(bento_store)`

bentoml/_internal/bento/bento.py
yetone: image
the type is `Optional` so we must check it

parano: yes, although the expectation is this will never be `None` at this point, let's add `assert build_config.additional_models is not None` before this line?

bentoml/_internal/yatai_rest_api_client/schemas.py
```diff
@@ -0,0 +1,247 @@
...
+    metadata: Dict[str, Any] = attr.field(factory=dict)
+    context: Dict[str, Any] = attr.field(factory=dict)
+    options: Dict[str, Any] = attr.field(factory=dict)
+
+
+@attr.define
+class ModelSchema(ResourceSchema):
+    description: str
+    version: str
+    image_build_status: ModelImageBuildStatus
```
parano, line 217: should model image build status be exposed here? My understanding is that it is a platform internal detail and not suppose to be user facing

PR #2090, opened 2021-12-03
feat: Pytorch Tensor data container

larme:
## Checklist:

### DIFF
--- a/bentoml/pytorch.py
+++ b/bentoml/pytorch.py
@@ -1,3 +1,4 @@
+import pickle
 import typing as t
 import zipfile
 import functools
@@ -15,6 +16,8 @@ from ._internal.models import PT_EXT
 from ._internal.models import SAVE_NAMESPACE
 from ._internal.runner import Runner
 from ._internal.runner.utils import Params
+from ._internal.runner.container import Payload
+from ._internal.runner.container import DataContainer
 from ._internal.configuration.containers import BentoMLContainer
 
 _RV = t.TypeVar("_RV")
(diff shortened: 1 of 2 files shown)


bentoml/pytorch.py
```diff
@@ -318,3 +321,38 @@ def load_runner(
         batch_options=batch_options,
         model_store=model_store,
     )
+
+
+class PytorchTensorContainer(DataContainer[torch.Tensor, torch.Tensor]):
```
parano, line 326: we still need to add a `DataContainerRegistry.register_container` call here in this file right?

tests/integration/frameworks/test_pytorch_impl.py
```diff
@@ -120,3 +122,20 @@ def test_pytorch_runner_with_partial_kwargs(modelstore, bias_pair):
     # tensor to float may introduce larger errors, so we bump rel_tol
     # from 1e-9 to 1e-6 just in case
     assert math.isclose(res1 - res2, bias1 - bias2, rel_tol=1e-6)
+
+
+@pytest.mark.parametrize("batch_axis", [0, 1])
+def test_pytorch_container(batch_axis):
```
parano, line 128: could you add a test that loads a custom runner with PyTorch tensor input, and test both its `run` and `run_batch` method?

aarnphm:
there is some weird formatting with all the CI files on the main branch atm. Maybe because of the recent PR merge.

bentoml/pytorch.py
larme: my bad

tests/integration/frameworks/test_pytorch_impl.py
larme: Can we test with current pytorch runner? I push a draft test in the new commit

parano: that is great, it would be great to also add a minimal test case to test data container separately, e.g. create a dummy runner class with PyTorch tensor input type and test its behavior.  I will go ahead and merge this PR, feel free to add that in follow up PR.

```python
class DummpyPyTorchRunner(Runner):
   def _run_batch(self, input_tensor):
          ...
```

PR #2091, opened 2021-12-03
feat: Model add custom objects support

larme:
first add custom objects support to the base Model class. After this pr is merged, we can add custom objects to each runner's `save`

## Checklist:


bentoml/_internal/models/model.py
```diff
@@ -102,11 +108,16 @@ def from_fs(cls, item_fs: FS) -> "Model":
         try:
             with item_fs.open(MODEL_YAML_FILENAME, "r") as model_yaml:
                 info = ModelInfo.from_yaml_file(model_yaml)
+            if item_fs.isfile(CUSTOM_OBJECTS_FILENAME):
+                with item_fs.open(CUSTOM_OBJECTS_FILENAME, "rb") as cofile:
+                    custom_objects = cloudpickle.load(cofile)
```
parano, line 113: we can add a `custom_objects` property method in Model class, where it reads the custom objects from `fs` lazily upon access?

larme: I pushed a new commit for this change

bentoml/_internal/models/model.py
```diff
@@ -124,13 +149,20 @@ def flush_info(self):
         with self._fs.open(MODEL_YAML_FILENAME, "w") as model_yaml:
             self.info.dump(model_yaml)
 
+    def flush_custom_objects(self):
+        if self.custom_objects is None:
+            return
+        with self._fs.open(CUSTOM_OBJECTS_FILENAME, "wb") as cofile:
+            cloudpickle.dump(self.custom_objects, cofile)
```
larme, line 156: this part I want to make sure if user pass an empty dictionary, after loading the model from fs he/she will get back an empty dictionary

bentoml/_internal/models/model.py
```diff
@@ -75,6 +97,7 @@ def create(
                 metadata=metadata,
                 context=framework_context,
             ),
+            custom_objects,
```
parano, line 100: should this be part of ModelInfo?

sauyon:
I don't currently understand the difference in nuance between all the dictionaries passed into the model constructor now; is there a document explaining them?

bentoml/_internal/models/model.py
```diff
@@ -33,11 +35,30 @@ class Model(StoreItem):
...
+        if self._custom_objects is None:
+            if self._fs.isfile(CUSTOM_OBJECTS_FILENAME):
+                with self._fs.open(CUSTOM_OBJECTS_FILENAME, "rb") as cofile:
+                    self._custom_objects = cloudpickle.load(cofile)
+            else:
+                # we need to distinguish the case of property not
+                # accessed and model has no custom objects
+                self._custom_objects = False
+
+        if self._custom_objects is False:
```
sauyon, line 57: I think if you use `isinstance(..., bool)` you can avoid the type ignore below.

bentoml/_internal/models/model.py
sauyon: I feel like it's best if we leave `ModelInfo` to correspond exactly to the information in the `model.yaml`.

bentoml/_internal/models/model.py
```diff
@@ -104,7 +128,8 @@ def from_fs(cls, item_fs: FS) -> "Model":
             logger.warning(f"Failed to import Model from {item_fs}.")
             raise BentoMLException("Failed to create Model because it was invalid")
 
-        res = cls(info.tag, item_fs, info)
+        # set _custom_objects to None for lazy loading
+        res = cls(info.tag, item_fs, info, None)
```
sauyon, line 132: Does the `None` not happen automatically with the default provided in the field definition?

tests/unit/_internal/models/test_model.py
```diff
@@ -38,6 +38,14 @@
 """
 
 
+class AdditionClass:
+    def __init__(self, x):
+        self.x = x
+
+    def __call__(self, y):
+        return self.x + y
```
sauyon, line 46:
```suggestion
class AdditionClass:
    def __init__(self, x: int):
        self.x = x

    def __call__(self, y: int) -> int:
        return self.x + y
```

Should make the type ignore below unnecessary.

parano:
I will follow up with a PR and run formatting for everything,  @aarnphm

parano:
> I don't currently understand the difference in nuance between all the dictionaries passed into the model constructor now; is there a document explaining them?

Just added some doctoring to Model#create to explain the differences

bentoml/_internal/models/model.py
larme: yep, this way we can keep the model info lightweight. Custom objects may be heavy

PR #2106, opened 2021-12-07
fix: regular runner names

bojiang:
## Checklist:

- [ ] My code follows the bentoml code style, both `./dev/format.sh` and
  `./dev/lint.sh` script have passed
  ([instructions](https://github.com/bentoml/BentoML/blob/master/DEVELOPMENT.md#style-check-and-auto-formatting-your-code)).
- [ ] My change reduces project test coverage and requires unit tests to be added
- [ ] I have added unit tests covering my code change
- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly

PR #2107, opened 2021-12-07
fix: error when serve --production using bento path

parano:
## Checklist:

### DIFF
--- a/bentoml/_internal/server/__init__.py
+++ b/bentoml/_internal/server/__init__.py
@@ -98,6 +98,8 @@ def serve_production(
     port: int = Provide[BentoServerContainer.config.port],
     app_workers: t.Optional[int] = None,
 ) -> None:
+    working_dir = os.path.realpath(os.path.expanduser(working_dir))
+
     svc = load(bento_identifier, working_dir=working_dir)
     env = dict(os.environ)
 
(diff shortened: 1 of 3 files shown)


bentoml/_internal/server/__init__.py
```diff
@@ -116,19 +118,19 @@ def serve_production(
...
         )
+        sockets_map[runner_name] = runner_socket
         cmd_runner = f"""import bentoml._internal.server
 bentoml._internal.server.start_prod_runner_server(
     "{bento_identifier}",
     "{runner_name}",
     working_dir="{working_dir}",
     instance_id=$(CIRCUS.WID),
-    fd=$(circus.sockets.{runner_name}),
+    fd={runner_socket.fileno()},
```
parano, line 133: @bojiang I'm not sure why circus is not replacing it here, but changing it to `.fileno` call seems to be working well

bentoml/_internal/server/__init__.py
```diff
@@ -159,6 +163,7 @@ def serve_production(
             numprocesses=app_workers or 1,
             stop_children=True,
             use_sockets=True,
+            working_dir=working_dir,
```
parano, line 166: @bojiang we need this change per https://github.com/bentoml/BentoML/pull/2064#discussion_r759091488

tests/integration/conftest.py
```diff
@@ -6,10 +6,10 @@
 from bentoml._internal.models import ModelStore
 
 if TYPE_CHECKING:
-    from _pytest.config import Config
-    from _pytest.config.argparsing import Parser
     from _pytest.nodes import Item
+    from _pytest.config import Config
```
parano, line 10: auto formatting change caused by another PR

bentoml/_internal/server/__init__.py
bojiang: because we didn't ensure the runner_name to be a valid identifier.
Fixed in #2106

PR #2118, opened 2021-12-10
fix(internal): fix bento implementation

sauyon:
The new tests are currently somewhat broken, and so are the old ones.

This will fail style checks because of some type checking and pylint errors I'm not sure what the best way to deal with are; suggestions appreciated.


bentoml/_internal/bento/build_config.py
```diff
@@ -58,13 +58,15 @@ def _convert_python_version(py_version: t.Optional[str]) -> t.Optional[str]:
 @attr.frozen
 class DockerOptions:
     # Options for choosing a BentoML built-in docker images
-    distro: t.Optional[str] = attr.ib(
-        validator=attr.validators.optional(
-            attr.validators.in_(DOCKER_SUPPORTED_DISTROS)
-        ),
-        default=None,
+    distro: "t.Literal['slim', 'amazonlinux2', 'alpine', 'centos7', 'centos8']" = (
```
parano, line 61: this should be `t.Optional[t.Literal[...]]` ?

sauyon: Yep.

parano:
```

W0613:unused-argument [193:29] CondaOptions.write_to_bento: Unused argument 'bento_fs'
15
W0613:unused-argument [193:43] CondaOptions.write_to_bento: Unused argument 'build_ctx'
16
```
CondaOptions is not implemented, it's okay to ignore for now

```
E1133:not-an-iterable [233:28] PythonOptions.write_to_bento: Non-iterable value self.wheels is used in an iterating context
17
E1133:not-an-iterable [265:24] PythonOptions.write_to_bento: Non-iterable value self.trusted_host is used in an iterating context
18
E1133:not-an-iterable [268:24] PythonOptions.write_to_bento: Non-iterable value self.find_links is used in an iterating context
19
E1133:not-an-iterable [271:24] PythonOptions.write_to_bento: Non-iterable value self.extra_index_url is used in an iterating context
20
```
I thought having checks like `if self.wheels is not None:` in front should be enough?  


```
W0223:abstract-method [433:00] FilledBentoBuildConfig: Method 'to_yaml' is abstract in class 'BentoBuildConfig' but is not overridden
```
Should be ok to ignore this for now? Since `to_yaml` is just a place holder for now

bentoml/_internal/bento/bento.py
```diff
@@ -138,14 +140,12 @@ def create(
...
 
-        assert build_config.additional_models is not None
-
         model_tags = build_config.additional_models
         # Add Runner required models to models list
         for runner in svc.runners.values():
             model_tags += runner.required_models
 
-        models: t.List[Model] = []
+        models: "t.List[Model]" = []
```
aarnphm, line 148: Why do we have to do this instead of 
```
t.List["Model"]
```?

sauyon: I prefer fully quoting the type stylistically. I suspect that unquoted type hints have a startup performance impact and are in general more annoying to use, so at some point I think it would be nice to start quoting all annotations.

That, or we could use future annotations but sadly that's 3.7+.

sauyon:
> I thought having checks like `if self.wheels is not None:` in front should be enough?  

Sadly pylint doesn't seem to pick this up, though pyright does. It seems like adding `.all()` fixes it for pylint, but we may want to just disable it for those lines.


> ```
> W0223:abstract-method [433:00] FilledBentoBuildConfig: Method 'to_yaml' is abstract in class 'BentoBuildConfig' but is not overridden
> ```
> Should be ok to ignore this for now? Since `to_yaml` is just a place holder for now
> 

Yeah, I wasn't entirely sure what the best way to deal with this was so I left it.

PR #2143, opened 2021-12-25
feat(codebase): add LazyType.isinstance with TypeGuard(PEP647) support

bojiang:
## Description

    TypeRef provides solutions for several conflicts when applying lazy dependencies,
         type annotations and runtime class checking.
     It works both for runtime and type checking phases.
     * conflicts 1
     isinstance(obj, class) requires importing the class first, which breaks
     lazy dependencies
     solution:
     >>> TypeRef("numpy.ndarray").isinstance(obj)
     * conflicts 2
     `isinstance(obj, str)` will narrow obj types down to str. But it only works for the
     case that the class is the type at the same time. For numpy.ndarray which the type
     is actually numpy.typing.NDArray, we had to hack the type checking.
     solution:
     >>> if TYPE_CHECKING:
     >>>     from numpy.typing import NDArray
     >>> TypeRef["NDArray"]("numpy.ndarray").isinstance(obj)`
     >>> #  this will narrow the obj to NDArray with PEP-647
     * conflicts 3
     compare/refer/map classes before importing them.
     >>> HANDLER_MAP = {
     >>>     TypeRef("numpy.ndarray"): ndarray_handler,
     >>>     TypeRef("pandas.DataFrame"): pdframe_handler,
     >>> }
     >>>
     >>> HANDLER_MAP[TypeRef(numpy.ndarray)]](array)
     >>> TypeRef("numpy.ndarray") == numpy.ndarray

## Checklist:

### DIFF
--- a/bentoml/_internal/io_descriptors/image.py
+++ b/bentoml/_internal/io_descriptors/image.py
@@ -7,7 +7,7 @@ from starlette.requests import Request
 from multipart.multipart import parse_options_header
 from starlette.responses import Response
 
-from bentoml._internal.types import TypeRef
+from bentoml._internal.types import LazyType
 
 from .base import ImageType
 from .base import IODescriptor
@@ -146,9 +146,9 @@ class Image(IODescriptor[ImageType]):
         return PIL.Image.open(io.BytesIO(bytes_))
 
     async def to_http_response(self, obj: ImageType) -> Response:
-        if TypeRef["NDArray"]("numpy.ndarray").isinstance(obj):
+        if LazyType["NDArray"]("numpy.ndarray").isinstance(obj):
             image = PIL.Image.fromarray(obj, mode=self._pilmode)
-        elif TypeRef["PIL.Image.Image"]("PIL.Image.Image").isinstance(obj):
+        elif LazyType["PIL.Image.Image"]("PIL.Image.Image").isinstance(obj):
             image = obj
         else:
             raise InternalServerError(
--- a/bentoml/_internal/io_descriptors/image.py
+++ b/bentoml/_internal/io_descriptors/image.py
@@ -1,5 +1,4 @@
 import io
-import sys
 import typing as t
 from typing import TYPE_CHECKING
 from urllib.parse import quote
@@ -7,6 +6,8 @@ from urllib.parse import quote
 from starlette.requests import Request
 from multipart.multipart import parse_options_header
 from starlette.responses import Response
+
+from bentoml._internal.types import TypeRef
 
 from .base import ImageType
 from .base import IODescriptor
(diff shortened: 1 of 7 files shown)


bentoml/_internal/io_descriptors/image.py
```diff
@@ -150,9 +144,9 @@ async def from_http_request(self, request: Request) -> ImageType:
         return PIL.Image.open(io.BytesIO(bytes_))
 
     async def to_http_response(self, obj: ImageType) -> Response:
-        if isinstance(obj, np.ndarray):
+        if TypeRef["numpy.typing.NDArray[t.Any]"]("numpy.ndarray").isinstance(obj):
             image = PIL.Image.fromarray(obj, mode=self._pilmode)
-        elif isinstance(obj, PIL.Image.Image):  # type: ignore
+        elif TypeRef["PIL.Image.Image"]("PIL.Image.Image").isinstance(obj):
             image = obj
```
bojiang, line 150: example:
no `type: ignore`; lazy import; works both for runtime and type checkers.

bentoml/_internal/types.py
```diff
@@ -43,6 +43,130 @@
...
+    >>>
+    >>> HANDLER_MAP[TypeRef(numpy.ndarray)]](array)
+    >>> TypeRef("numpy.ndarray") == numpy.ndarray
+    """
+
+    @t.overload
+    def __init__(self, module_or_cls: str, qualname: str) -> None:
+        """TypeRef("numpy", "ndarray")"""
+
+    @t.overload
```
aarnphm, line 92: I think this will throw the pylint error
```suggestion
    @overload
```

bojiang: error happens after your patch
```
bentoml/_internal/types.py:92:6: F821 undefined name 'overload'
bentoml/_internal/types.py:97:5: F811 redefinition of unused '__init__' from line 93
```

bentoml/_internal/types.py
```diff
@@ -43,6 +43,130 @@
...
+    >>>
+    >>> HANDLER_MAP[TypeRef(numpy.ndarray)]](array)
+    >>> TypeRef("numpy.ndarray") == numpy.ndarray
+    """
+
+    @t.overload
+    def __init__(self, module_or_cls: str, qualname: str) -> None:
+        """TypeRef("numpy", "ndarray")"""
+
+    @overload
```
bojiang, line 92:
```suggestion
    @t.overload
```

bentoml/_internal/io_descriptors/image.py
```diff
@@ -16,11 +17,15 @@
 from ...exceptions import InternalServerError
 
 if TYPE_CHECKING:
-    import PIL
-    import numpy as np
     import PIL.Image
+    import numpy.typing
+
+    NDArray = numpy.typing.NDArray[t.Any]
```
aarnphm, line 23: I don't know if this would work for numpy < 1.20

parano: it doesn’t have to? We can fix a numpy version for dev env for type checking

aarnphm: noted

yetone (approved):

PR #2161, opened 2022-01-07
feat: add size_bytes for bento and model

yetone:


bentoml/_internal/cli/yatai.py
```diff
@@ -1,12 +1,14 @@
+import logging
+
 import click
 
 from bentoml.exceptions import CLIException
-from bentoml._internal.cli.click_utils import _echo
-from bentoml._internal.cli.click_utils import CLI_COLOR_SUCCESS
 from bentoml._internal.yatai_rest_api_client.config import add_context
```
aarnphm, line 6: oo can you change everything under `_internal` to relative imports?

yetone: Why do we use relative import? My impression is that relative import has always been a bad practice. First, relative import is very low readability and basically unreadable in review PR scenarios, second, relative import introduces confusion when changing the directory structure, e.g. relative import works fine but imports something else, which is an implicitly catastrophic bug, and finally, it adds some mental burden because relative import must be contextualized to make sense. This is why PEP 328 was created when Python 2 had implicit relative import to avoid the relative import problem

aarnphm: Relative imports just makes ridiculously long imports shorter, imo it is a lot clearer that way. I do agree that relative imports would make it harder to maintain.

```
second, relative import introduces confusion when changing the directory structure, e.g. relative import works fine but imports something else, which is an implicitly catastrophic bug
```
I don't really understand the context of this. I assume that since we only do relative imports under `_internal`, if we ever need to imports anything from `_internal` for say when you are writting test, then we can still do absolute imports there?

yetone: for example:

```
- a
    - internal
        - b
            c.py
        - utils.py
    - utils.py
```

if `c.py` has this content:

```python
from ..utils import *
```

if `c.py` move to `internal` directory, the import will work fine but import another utils

PR #2167, opened 2022-01-12
feat: add OpenTelemetry middleware

bojiang:
## Description

TODO:
* test with external exporters enabled like zipkin or jaeger
## Checklist:


bentoml/_internal/server/service_app.py
```diff
@@ -104,6 +105,34 @@ def log_exception(request: "Request", exc_info: t.Any) -> None:
...
+        if span is None:
+            return None
+        return span.get_span_context().span_id
+
+    @property
+    def request_id(self) -> t.Optional[int]:
+        return self.request_id_var.get()
+
+
+ServiceContext = ServiceContextClass()
```
bojiang, line 133: @ssheng Here is the proxy of tracing context. The sampler is setten as always record (but not always export)

bentoml/_internal/configuration/containers.py
```diff
@@ -257,24 +233,31 @@ def tracer(
 
 
 @dataclass
-class BentoServerContainerClass:
+class DeploymentContainerClass:
 
     bentoml_container = BentoMLContainer
-    config = bentoml_container.config.bento_server
+    config = bentoml_container.config
```
bojiang, line 239: note: the tracing config should be and has been in the outer level of the API server. Runners share the same config.

bojiang: TODO: move metric config out of API server config as well?

bentoml/_internal/server/service_app.py
ssheng: LGTM. 👍

bentoml/_internal/utils/telemetry.py
```diff
@@ -0,0 +1,65 @@
...
+    return parent_span_context.trace_state
+
+
+class TraceIdRatioBasedAlwaysRecording(TraceIdRatioBased):
+    """
+    A trace Sampler that:
+    * always recording (so that we can get the trace_id)
+    * respect the parent's trace_state
+    * ratio sampling
+    """
```
ssheng, line 32: I need to think about this more but we might be more fitting to use `ParentBased` instead of `TraceIdRatioBased` here. If BentoML is a part of a bigger trace, the sampling decision might already be made at one of the earlier services. BentoML should respect that trace decision. If BentoML API server is the root, we can then define and leverage our own sample rate. Let's discuss more tomorrow.

bojiang: Actually we are using ParrentBasedAndTraceIdRatioBasedAlwaysRecording

PR #2168, opened 2022-01-12
feat: shorten versions in bentoml list

sauyon:


bentoml/_internal/cli/bento_management.py
```diff
@@ -98,6 +99,11 @@ def list_bentos(bento_name, output, no_trunc):
                 bentos, key=lambda x: x.info.creation_time, reverse=True
             )
         ]
+
+        if not no_trunc:
```
sauyon, line 103: Why is the `no_trunc` option not a flag? `bentoml list --no-trunc` seems more natural to me than `bentoml list --no-trunc true`

yetone: @sauyon I think it should be a flag

bentoml/_internal/utils/trie.py
```diff
@@ -0,0 +1,46 @@
+import typing as t
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    TrieDataType: t.TypeAlias = t.Union[str, t.Dict[str, "Trie"]]
```
aarnphm, line 5:
```suggestion
    TrieDataType: "t.TypeAlias" = "t.Union[str, t.Dict[str, Trie]]"
```

sauyon: I'd tried this but because `TypeAlias` is special you can't quote it. Just found out you *can* quote the alias itself, though, so I'll do that.

sauyon (commented):
I think I might retool this to associate the shortening with the store instead since that seems a lot more reasonable...

bojiang (approved):

parano:
discussed offline, we will not shorten the version in UI, but actually shorten the generated version str instead

pamobazan12 (commented):
confirmo

PR #2175, opened 2022-01-15
fix(internal): shorten tag version

sauyon:


sauyon:
Ok, this should fix it...

parano:
Saw the following error in the python3.6 e2e test, @bojiang could you help take a look?

```
AttributeError: module 'asyncio' has no attribute 'run'
159
```

Merging now since the error is not related to this PR.

PR #2205, opened 2022-01-25
feature(cli): import/export bento/model

larme:
## Checklist:


bentoml/_internal/store.py
```diff
@@ -162,3 +163,65 @@ def delete(self, tag: t.Union[str, Tag]) -> None:
...
+
+        item_path = _tag.path()
+        item_sys_path = self._fs.getsyspath(item_path)
+
+        if not self._fs.exists(item_path):
+            raise BentoMLException(
+                f"Item '{_tag}' does not exist in the store {self._fs}"
+            )
+        with tarfile.open(export_path, "w", fileobj=fileobj) as tar:
+            tar.add(item_sys_path, arcname=arcname)
```
larme, line 191: this will over an existing tar file.

Should we ask user confirmation?

larme: seems that docker just overwrite the existing file

bentoml/_internal/cli/bento_management.py
```diff
@@ -168,27 +168,32 @@ def delete_target(target):
...
     def export(bento_tag, out_file):
         """Export Bento files to a tar file
 
         bentoml export FraudDetector:latest > my_bento.tar
         bentoml export FraudDetector:20210709_DE14C9 ./my_bento.tar
         """
-        pass
+        if not _is_valid_bento_tag(bento_tag):
+            raise click.BadParameter(
+                'Bad formatting. Please present a valid bento bundle "name:version" tag'
```
sauyon, line 181:
```suggestion
                f'Invalid tag {bento_tag}. Please use a valid "name:version" tag'
```

bentoml/_internal/cli/model_management.py
```diff
@@ -170,29 +170,34 @@ def delete_target(target):
...
     def export(model_tag, out_file):
         """Export Model files to a tar file
 
         bentoml model export FraudDetector:latest > my_model.tar
         bentoml model export FraudDetector:20210709_DE14C9 ./my_model.tar
         """
-        pass
+        if not _is_valid_bento_tag(model_tag):
+            raise click.BadParameter(
+                'Bad formatting. Please present a valid model "name:version" tag'
```
sauyon, line 183:
```suggestion
                f'Invalid tag {bento_tag}. Please use a valid "name:version" tag'
```

bentoml/_internal/store.py
```diff
@@ -162,3 +163,65 @@ def delete(self, tag: t.Union[str, Tag]) -> None:
                 # otherwise, update the latest version
                 assert new_latest.tag.version is not None
                 self._fs.writetext(_tag.latest_path(), new_latest.tag.version)
+
+    def export_tar(
```
sauyon, line 167: Can we not just do `self.get(tag).export("tar://" + path)`?

larme: @sauyon in the cli part we are passing a file object instead of path to support both real file and standard out. I think `fs` cannot support this use case?

bentoml/_internal/cli/model_management.py
```diff
@@ -170,29 +170,34 @@ def delete_target(target):
...
+        "model_path", type=click.File("rb"), default=sys.stdin.buffer, required=False
     )
     def import_model(model_path):
         """Export Model files to a tar file
 
         bentoml model import < ./my_model.tar
         bentoml model import ./my_model.tar
         """
-        pass
+        model_store.import_tar(fileobj=model_path)
```
parano, line 200: could we add an output message here? e.g.:

```
        model = model_store.import_tar(fileobj=model_path)
        logger.info(f'{model} imported from "{model_path}"')
```

And same for the export function?

bentoml/_internal/cli/model_management.py
```diff
@@ -170,29 +170,34 @@ def delete_target(target):
     @model_cli.command(help="Export Model to a tar file")
```
parano, line 170: could you add an sample CLI command in the help message? it would be great to show user an example command in the help message of `bentoml export --help`

bentoml/_internal/cli/model_management.py
parano: same for the Bento import/export

sotte:
I guess once https://github.com/bentoml/BentoML/pull/2237 is merged (adding more advanced import/export including s3), this PR will also offer import/export via the CLI to s3 and similar. Is that correct/the goal of this PR?

parano:
> I guess once #2237 is merged (adding more advanced import/export including s3), this PR will also offer import/export via the CLI to s3 and similar. Is that correct/the goal of this PR?

Yes that's correct!  #2237 will introduce APIs to export Bento/Model to tar file, directly to local or s3 path, as well as importing it

PR #2207, opened 2022-01-25
test: fix flaky framework ci tests

aarnphm:

## Description

With the newest updates from starlette require typing_extensions to be >=3.10, which is why transformers tests failed.

easyocr problem (for reference): https://github.com/JaidedAI/EasyOCR/issues/633

paddlepaddle problems is due to setuptools > 60, solved with https://github.com/pypa/setuptools/issues/2993#issuecomment-1003765389

disable ONNXMLir tests for now due too way too much overhead in setting up the library and fixes all the bugs with the recent updates :)

There was a a bug when developing transformers i didn't notice, is that for cases that uses a `FeatureExtractor` class, internal transformers API load model differently, which causes CI to fail. Should be fix and applicable for most of the popular usecase with transformers now.

## Checklist:


bentoml/_internal/ext_typing.py
```diff
@@ -10,6 +10,20 @@
...
+    from transformers.pipelines.base import Pipeline as TransformersPipeline
+    from transformers.modeling_tf_utils import TFPreTrainedModel
+    from transformers.tokenization_utils import PreTrainedTokenizer
+    from transformers.modeling_flax_utils import FlaxPreTrainedModel
+    from transformers.tokenization_utils_fast import PreTrainedTokenizerFast
+    from transformers.feature_extraction_sequence_utils import (
+        SequenceFeatureExtractor as PreTrainedFeatureExtractor,
+    )
+
+    TransformersModelType = t.Union[
```
parano, line 23: any reason not to put these in the transformer module?

tests/integration/frameworks/test_onnxmlir_impl.py
```diff
@@ -70,6 +66,8 @@ def compile_model(convert_to_onnx, tmpdir):
         cwd=onnx_mlir_loc,
     )
     stdout, stderr = docker_proc.communicate()
+    print(stdout)
```
parano, line 69: remove?

bentoml/_internal/frameworks/transformers.py
```diff
@@ -23,29 +23,28 @@
...
-    from transformers import PreTrainedTokenizer
-    from transformers import PreTrainedTokenizerFast
-    from transformers.feature_extraction_utils import PreTrainedFeatureExtractor
-    from transformers.models.auto.auto_factory import _BaseAutoModelClass
+    from transformers.configuration_utils import PretrainedConfig
+    from transformers.models.auto.auto_factory import (
+        _BaseAutoModelClass,  # type: ignore[reportPrivateUsage]
+    )
 
+    from .. import ext_typing as ext
```
parano, line 39: let's move the transformers type definition here

bentoml/_internal/frameworks/transformers.py
```diff
@@ -63,10 +62,8 @@
 
 MODULE_NAME = "bentoml.transformers"
 
-try:
-    from huggingface_hub import HfFolder
-except ImportError:
-    HfFolder = None
+# `save` should only be used when training or working with a customized pretrained transformers model
```
parano, line 65: remove? or move to the `save` method doc string

bentoml/_internal/frameworks/transformers.py
```diff
@@ -415,127 +452,148 @@ def _save(
...
-            #  a Tokenizer instance, then save it to path.
-            try:
-                _tokenizer_inst = AutoTokenizer.from_pretrained(model_identifier)
-                _tokenizer_inst.save_pretrained(_model.path)
-                _model.info.options["tokenizer"] = type(_tokenizer_inst).__name__
-            except ValueError:
-                _model.info.options["tokenizer"] = "na"
+            _ = _tokenizer_inst.save_pretrained(_model.path)  # type: ignore[reportUnknownMemberType]
+            _model.info.options["tokenizer"] = type(_tokenizer_inst).__name__
+        except (ValueError, KeyError):
```
parano, line 569: Is this something we can refactor to a `if else` check?

scripts/ci/config.yml
```diff
@@ -96,7 +97,8 @@ mlflow:
   override_name_or_path: mlflow
   dependencies:
     - "mlflow"
-    - "scikit-learn"
+  external_scripts: |
+    python -m pip install -U scikit-learn
```
parano, line 101: what's the reason for moving this change?

aarnphm: I think the version for scikit-learn from mlflow is older and it doesn't have some of the internal library

bentoml/_internal/frameworks/transformers.py
aarnphm: I'm still looking into this changes from transformers to make it a lot more simple. Had a long demo with @bojiang  this morning about this whole segment and found some bugs when we tried to load the a transformers model back.

bentoml/_internal/frameworks/transformers.py
aarnphm: I will move it to docstring

scripts/ci/config.yml
```diff
@@ -206,6 +210,8 @@ transformers:
     - "tensorflow==2.5.0"
     - "importlib_metadata"
     - "Pillow==8.4.0"
+  external_scripts: |
+    python -m pip install "typing_extensions>=3.10"
```
parano, line 214: can we just put `"typing_extensions>=3.10"` in the dependencies list above?

aarnphm: it would resolve in pip conflict

bojiang:
https://github.com/bentoml/BentoML/runs/4975985185?check_suite_focus=true#step:7:267

bojiang:
Transformers is a hard lib. There is a lot of coupling in the implementation of Transformers now.
It's ideal to only rely on the public API of transformers, also leave the details to it.

The cases are complex. Merging it now.

bojiang (approved):

PR #2210, opened 2022-01-25
fix: handle crash when runners return a generator instead of a list

eob:
## Description

Some runners (e.g. SpaCy) return a generator object from their batch run instead of a list (which is expected). This causes an error when the `run` method blindly indexes into the zeroth-element of the object: permitted in a list, but not a generator. In production, it also results in a "can not pickle generator" error.

Since BentoML has a wide diversity of authors & runner implementations, I think it's best to handle this defensively instead of frame it exclusively as a problem for the runner to fix (e.g. the SpaCy runner).

This PR is based off of the `1.0.0a2` tag and simply checks if the result is a generator before attempting to index into the zeroth element. If it's a generator, it converts it to the intended `list` type.

I think it's safest to handle this defensively: 

## Motivation and Context

This makes BentoML core more robust to anticipated errors in framework-specific runner implementations.

## How Has This Been Tested?

My code no longer crashes as it was before. Not a great test, but it's a relatively simple two-line fix..

## Checklist:

### DIFF
--- a/bentoml/_internal/runner/local.py
+++ b/bentoml/_internal/runner/local.py
@@ -1,3 +1,4 @@
+from types import GeneratorType
 import typing as t
 
 from .runner import Runner
@@ -40,6 +41,11 @@ class LocalRunner(RunnerImpl):
                 batch_result,
                 batch_axis=self._runner.batch_options.output_batch_axis,
             )
+            if isinstance(single_results, GeneratorType):
+                # A runner may mistakenly return a generator, causing 
+                # indexing & pickling errors downstream.
+                single_results = list(single_results)
+
             return single_results[0]
 
         if isinstance(self._runner, SimpleRunner):
--- a/bentoml/_internal/runner/local.py
+++ b/bentoml/_internal/runner/local.py
@@ -1,4 +1,3 @@
-from types import GeneratorType
 import typing as t
 
 from .runner import Runner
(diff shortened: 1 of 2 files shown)


bentoml/_internal/runner/local.py
```diff
@@ -40,6 +41,11 @@ def run(self, *args: t.Any, **kwargs: t.Any) -> t.Any:
                 batch_result,
                 batch_axis=self._runner.batch_options.output_batch_axis,
             )
+            if isinstance(single_results, GeneratorType):
```
parano, line 44: It would be great to move this check to the `DefaultContainer#batch_to_singles`, I think it is nice if the default Data Container can handle generator type.  For other Runner data container types(Numpy, Pandas, etc) we probably won't need this check here.

Thanks @eob for the PR! We haven't got around to document the design of Runner and DataContainer yet, basically Runner class implements a `run_batch` method, allow runner instance to invoke `run` or `run_batch` call. These calls will be dynamically batched, in order to better utilize hardware concurrency.  In order to do this, BentoML needs to know how to group a batch of `run` inputs into a batch data type for the `_run_batch` method, and the DataContainer class is meant for definition that process for each data type that Runner supports.

parano:
the code_style_check CI is currently failing, you may need to run `make format` locally to auto-format your code change

eob:
@parano thanks for the feedback & comments. 

I'll look at the `DefaultContainer#batch_to_singles` relocation -- since posting this PR I've realized it doesn't fix all the code paths that lead to the bug, and I've been trying to find the best spot to place it to catch them all. Maybe `DefaultContainer#batch_to_singles` is it.

Will run the formatter before resubmitting.

Will get back to you! 🙌

eob:
Hi @parano --

OK. This PR:
1. Has the code formatter applied, and
2. Relocates the fix to the container classes

I had to add it to several places to get rid of all the exceptions I was seeing with SpaCy. It feels a bit like Generator whack-a-mole to me... Happy to refactor it if there's a better way; I just don't know what it is since I'm not familiar with the BentoML codebase. :)

PR #2208, opened 2022-01-25
docs: update the quickstart guide

sauyon:

PR #2212, opened 2022-01-25
docs: fix import bentoml

dbulgakov7:
Add "mport bentoml" to Getting Started - Save Models


sauyon:
Thanks, nice spot! This should be covered in #2208.

Issue #2215, opened 2022-01-26
model on cpu/gpu (ver.1.0.0a2)


jjerry-k:
**Is your feature request related to a problem? Please describe.**
I used pytorch 1.10, bentoml 1.0.0a2.
I trained a model using GPU and save using bentoml.pytorch.save.
When I load a model using CPU, runner didn't use CPU. 

If I saved CPU model, runner only uses CPU.
If I saved GPU model, runner only uses GPU.

**Describe the solution you'd like**
Need to modify this line [link](https://github.com/bentoml/BentoML/blob/9d803c30d0576df872d48d72648f6eab2027d345/bentoml/_internal/frameworks/pytorch.py#L95) to set a model to the desired device using device_id.
ex) `return cloudpickle.load(file).to(torch.device(device_id))`

**Describe alternatives you've considered**
Nope

**Additional context**
Original code & Test result
![before_code](https://user-images.githubusercontent.com/25097024/151182606-3be86939-fbf5-4642-ba37-3afeaf120f7e.png)
![before](https://user-images.githubusercontent.com/25097024/151182600-7a0a41f3-5563-4a87-8588-471eecfde314.png)

Modified code & Test result
![after_code](https://user-images.githubusercontent.com/25097024/151182796-3f461195-7c87-4ca3-971e-806b301df0db.png)
![after](https://user-images.githubusercontent.com/25097024/151182788-95b894a4-e21d-4928-9cc3-cc87b7724db4.png)

larme:
Thank you for reporting the issue.

We will push a quick fix for this problem.

aarnphm:
we merged this on our main branch. Can you try installing `1.0.0a3`? Let me know if you still have any other question.

jjerry-k:
@aarnphm 

I tried model loading across device_id.
First of all, I have no problem about model. (bentoml.pytorch.load)
But, I have another problem in loading runner. (bentoml.pytorch.load_runner)
I'll report issue about it.
Thank you!

PR #2216, opened 2022-01-26
fix: pytorch load correctly address device_id

larme:
addressing #2215

## Checklist:

### DIFF
--- a/bentoml/_internal/frameworks/pytorch.py
+++ b/bentoml/_internal/frameworks/pytorch.py
@@ -52,7 +52,7 @@ def _is_gpu_available() -> bool:  # pragma: no cover
 
 @inject
 def load(
-    tag: Tag,
+    tag: t.Union[Tag, str],
     device_id: t.Optional[str] = "cpu",
     model_store: "ModelStore" = Provide[BentoMLContainer.model_store],
 ) -> _ModelType:
@@ -86,13 +86,27 @@ def load(
                 f"Model {tag} was saved with module {bentoml_model.info.module}, failed loading with {MODULE_NAME}."
             )
     weight_file = bentoml_model.path_of(f"{SAVE_NAMESPACE}{PT_EXT}")
-    # TorchScript Models are saved as zip files
-    if zipfile.is_zipfile(weight_file):
+    model_format = bentoml_model.info.options.get("_model_format")
+    # backward compatibility
+    if not model_format:
+        if zipfile.is_zipfile(weight_file):
+            model_format = "torchscript:v1"
+        else:
+            model_format = "cloudpickle:v1"
+
+    if model_format == "torchscript:v1":
         model: "torch.jit.ScriptModule" = torch.jit.load(weight_file, map_location=device_id)  # type: ignore[reportPrivateImportUsage] # noqa: LN001
-        return model
-    else:
+    elif model_format == "cloudpickle:v1":
         with Path(weight_file).open("rb") as file:
-            return cloudpickle.load(file)
+            model: "torch.nn.Module" = cloudpickle.load(file).to(device_id)
+    elif model_format == "torch.save:v1":
+        with Path(weight_file).open("rb") as file:
+            model: "torch.nn.Module" = torch.load(file, map_location=device_id)
+    else:
+        raise BentoMLException(f"Unknown model format {model_format}")
+
+    model.eval()
+    return model
 
 
 @inject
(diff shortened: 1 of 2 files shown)


bentoml/_internal/frameworks/pytorch.py
```diff
@@ -86,13 +86,27 @@ def load(
...
         model: "torch.jit.ScriptModule" = torch.jit.load(weight_file, map_location=device_id)  # type: ignore[reportPrivateImportUsage] # noqa: LN001
-        return model
-    else:
+    elif model_format == "cloudpickle:v1":
         with Path(weight_file).open("rb") as file:
-            return cloudpickle.load(file)
+            model: "torch.nn.Module" = cloudpickle.load(file).to(device_id)
+    elif model_format == "torch.save:v1":
+        with Path(weight_file).open("rb") as file:
+            model: "torch.nn.Module" = torch.load(file, map_location=device_id)
```
larme, line 104: we cannot use `torch.load(file, pickle_module=cloudpickle` because `cloudpickle` does not implement `Unpickler` class. However data (including class/lambda/function) pickled by `cloudpickle` should be unpickled by `pickle` module so this line still works

bentoml/_internal/frameworks/pytorch.py
```diff
@@ -170,10 +184,12 @@ def forward(self, inputs):
     )
     weight_file = _model.path_of(f"{SAVE_NAMESPACE}{PT_EXT}")
     if isinstance(model, torch.jit.ScriptModule):  # type: ignore[reportPrivateUsage]
+        _model.info.options["_model_format"] = "torchscript:v1"
```
larme, line 187: should we save this information in `ModelInfo.options` or `ModelInfo.context`?

bentoml/_internal/frameworks/pytorch.py
```diff
@@ -86,13 +86,27 @@ def load(
...
         with Path(weight_file).open("rb") as file:
-            return cloudpickle.load(file)
+            model: "torch.nn.Module" = cloudpickle.load(file).to(device_id)
+    elif model_format == "torch.save:v1":
+        with Path(weight_file).open("rb") as file:
+            model: "torch.nn.Module" = torch.load(file, map_location=device_id)
+    else:
+        raise BentoMLException(f"Unknown model format {model_format}")
+
+    model.eval()
```
aarnphm, line 108: should we set the model to `eval()` mode here?

bentoml/_internal/frameworks/pytorch.py
aarnphm: `options` seems find for me

bentoml/_internal/frameworks/pytorch.py
```diff
@@ -170,10 +184,12 @@ def forward(self, inputs):
     )
     weight_file = _model.path_of(f"{SAVE_NAMESPACE}{PT_EXT}")
     if isinstance(model, torch.jit.ScriptModule):  # type: ignore[reportPrivateUsage]
+        _model.info.options["_model_format"] = "torchscript:v1"
         torch.jit.save(model, weight_file)  # type: ignore[reportUnknownMemberType]
     else:
+        _model.info.options["_model_format"] = "torch.save:v1"
```
aarnphm, line 190: Hmm on a second thought, this is only our internal format right? I don't think with the current implementation allows us to modify `context` at bento model creation time

bentoml/_internal/frameworks/pytorch.py
larme: or should we do it in `_PyTorchRunner._setup()`? I think we should do `eval()` for user somewhere in our codes?

aarnphm: there will be use-case where users load back, they want to continue train, so i think we can put `eval` under runner instead of here

bentoml/_internal/frameworks/pytorch.py
aarnphm: chatted with Chaoyu, lets put it under context

larme: done

bentoml/_internal/frameworks/pytorch.py
larme: moved `eval` to `_setup()`

bentoml/_internal/frameworks/pytorch.py
bojiang: Can we adapt the cloudpickle to support Unpickler?

aarnphm (approved):

PR #2237, opened 2022-02-02
Implement more sophisticated import / export APIs

sauyon:
This is just the export implementation; will add tests and import etc. later but putting this up for review so we can talk about how this should work.

Currently, the API supports:
 - `bento.export("")` - export the bento to `name-version.bento` in the current directory. this isn't a default because it feels weird for the API to have a default; arguably this should be handled by the CLI.
 - `bento.export("dir")` - if `dir` is a directory, will output to `dir/name-version.bento`.
 - `bento.export("dir/")` - even if `dir` is not a directory, try to output to `dir/name-version.bento`. note this doesn't actually try to create the folder, so save for weird filesystems where directories don't need to be created this will error if `dir` doesn't exist.
 - `bento.export("/path/to/filename")` - if filename does not have a supported extension, will create `filename.bento`.
 - `bento.export("/path/to/filename.<supported extension>")` - will use the relevant format and not append anything to the filename.

For any of those, if prefixed with supported `FS` prefixes like `s3://` or `ftp://`, will work the same way. The only exception is that for those types no check that the given path already exists as a directory is performed, so the `bento.export("dir")` form will create `dir.bento` instead of `dir/name-version.bento`.

### DIFF
--- a/bentoml/_internal/bento/bento.py
+++ b/bentoml/_internal/bento/bento.py
@@ -82,21 +82,30 @@ python:
 @attr.define(repr=False, auto_attribs=False)
 class Bento(StoreItem):
     _tag: Tag = attr.field()
-    _fs: "FS" = attr.field()
+    __fs: "FS" = attr.field()
 
     info: "BentoInfo"
 
     _model_store: ModelStore
     _doc: t.Optional[str] = None
 
-    @_fs.validator
+    @staticmethod
+    def _export_ext() -> str:
+        return "bento"
+
+    @__fs.validator
     def check_fs(self, _attr: t.Any, new_fs: "FS"):
-        new_fs.makedir("models", recreate=True)
+        try:
+            new_fs.makedir("models", recreate=True)
+        except fs.errors.ResourceReadOnly:
+            # when we import a tarfile, it will be read-only, so just skip the step where we create
+            # the models folder.
+            pass
         self._model_store = ModelStore(new_fs.opendir("models"))
 
     def __init__(self, tag: Tag, bento_fs: "FS", info: "BentoInfo"):
         self._tag = tag
-        self._fs = bento_fs
+        self.__fs = bento_fs
         self.check_fs(None, bento_fs)
         self.info = info
         self.validate()
@@ -104,6 +113,10 @@ class Bento(StoreItem):
     @property
     def tag(self) -> Tag:
         return self._tag
+
+    @property
+    def _fs(self) -> "FS":
+        return self.__fs
 
     @classmethod
     @inject
@@ -298,13 +311,9 @@ class Bento(StoreItem):
             out_fs = fs.open_fs(bento_path, create=True, writeable=True)
             fs.mirror.mirror(self._fs, out_fs, copy_if_newer=False)
             self._fs.close()
-            self._fs = out_fs
+            self.__fs = out_fs
 
         return self
-
-    def export(self, path: str):
-        out_fs = fs.open_fs(path, create=True, writeable=True)
-        fs.mirror.mirror(self._fs, out_fs, copy_if_newer=False)
 
     def validate(self):
         return self._fs.isfile(BENTO_YAML_FILENAME)
--- a/bentoml/_internal/bento/bento.py
+++ b/bentoml/_internal/bento/bento.py
@@ -277,13 +277,15 @@ class Bento(StoreItem):
             with item_fs.open(BENTO_YAML_FILENAME, "r", encoding="utf-8") as bento_yaml:
                 info = BentoInfo.from_yaml_file(bento_yaml)
         except fs.errors.ResourceNotFound:
-            logger.warning(f"Failed to import Bento from {item_fs}.")
-            raise BentoMLException("Failed to create Bento because it was invalid")
+            raise BentoMLException(
+                f"Failed to load bento because it does not contain a '{BENTO_YAML_FILENAME}'"
+            )
 
         res = cls(info.tag, item_fs, info)
         if not res.validate():
-            logger.warning(f"Failed to import Bento from {item_fs}.")
-            raise BentoMLException("Failed to create Bento because it was invalid")
+            raise BentoMLException(
+                f"Failed to create bento because it contains an invalid '{BENTO_YAML_FILENAME}'"
+            )
 
         res._flushed = True
         return res
--- a/bentoml/_internal/bento/bento.py
+++ b/bentoml/_internal/bento/bento.py
@@ -84,10 +84,12 @@ class Bento(StoreItem):
     _tag: Tag = attr.field()
     __fs: "FS" = attr.field()
 
-    info: "BentoInfo"
+    _info: "BentoInfo"
 
     _model_store: ModelStore
     _doc: t.Optional[str] = None
+
+    _flushed: bool = False
 
     @staticmethod
     def _export_ext() -> str:
@@ -117,6 +119,16 @@ class Bento(StoreItem):
     @property
     def _fs(self) -> "FS":
         return self.__fs
+
+    @property
+    def info(self) -> "BentoInfo":
+        self._flushed = False
+        return self._info
+
+    @info.setter
+    def info(self, new_info: "BentoInfo"):
+        self._flushed = False
+        self._info = new_info
 
     @classmethod
     @inject
@@ -273,6 +285,7 @@ class Bento(StoreItem):
             logger.warning(f"Failed to import Bento from {item_fs}.")
             raise BentoMLException("Failed to create Bento because it was invalid")
 
+        res._flushed = True
         return res
 
     @property
@@ -283,8 +296,13 @@ class Bento(StoreItem):
         return self._fs.getsyspath(item)
 
     def flush_info(self):
+        if self._flushed:
+            return
+
         with self._fs.open(BENTO_YAML_FILENAME, "w") as bento_yaml:
             self.info.dump(bento_yaml)
+
+        self._flushed = True
 
     @property
     def doc(self) -> str:
@@ -303,6 +321,8 @@ class Bento(StoreItem):
     def save(
         self, bento_store: "BentoStore" = Provide[BentoMLContainer.bento_store]
     ) -> "Bento":
+        self.flush_info()
+
         if not self.validate():
             logger.warning(f"Failed to create Bento for {self.tag}, not saving.")
             raise BentoMLException("Failed to save Bento because it was invalid")
@@ -314,6 +334,28 @@ class Bento(StoreItem):
             self.__fs = out_fs
 
         return self
+
+    def export(
+        self,
+        path: str,
+        output_format: t.Optional[str] = None,
+        *,
+        protocol: t.Optional[str] = None,
+        user: t.Optional[str] = None,
+        passwd: t.Optional[str] = None,
+        params: t.Optional[t.Dict[str, str]] = None,
+        subpath: t.Optional[str] = None,
+    ) -> str:
+        self.flush_info()
+        return super().export(
+            path,
+            output_format,
+            protocol=protocol,
+            user=user,
+            passwd=passwd,
+            params=params,
+            subpath=subpath,
+        )
 
     def validate(self):
         return self._fs.isfile(BENTO_YAML_FILENAME)
@@ -337,6 +379,8 @@ class BentoInfo:
     models: t.List[Tag]  # TODO: populate with model & framework info
     bentoml_version: str = BENTOML_VERSION
     creation_time: datetime = attr.field(factory=lambda: datetime.now(timezone.utc))
+
+    _flushed: bool = False
 
     def __attrs_post_init__(self):
         self.validate()
--- a/bentoml/_internal/exportable.py
+++ b/bentoml/_internal/exportable.py
@@ -1,0 +1,364 @@
+import os
+import re
+import typing as t
+import logging
+import urllib.parse
+from abc import ABC
+from abc import abstractmethod
+
+import fs
+import fs.copy
+import fs.mirror
+import fs.opener
+import fs.tempfs
+from fs.base import FS
+
+T = t.TypeVar("T")
+
+
+uriSchemeRe = re.compile(r".*[^\\](?=://)")
+
+
+class Exportable(ABC):
+    @staticmethod
+    @abstractmethod
+    def _export_ext() -> str:
+        raise NotImplementedError
+
+    @property
+    @abstractmethod
+    def _export_name(self) -> str:
+        raise NotImplementedError
+
+    @property
+    @abstractmethod
+    def _fs(self) -> FS:
+        raise NotImplementedError
+
+    @classmethod
+    @abstractmethod
+    def from_fs(cls: t.Type[T], item_fs: FS) -> T:
+        raise NotImplementedError
+
+    @classmethod
+    def guess_format(cls, path: str) -> str:
+        _, ext = fs.path.splitext(path)
+
+        ext = ext if ext == "" else ext[1:]
+        if ext in [cls._export_ext(), "gz", "xz", "bz2", "tar", "zip"]:
+            return ext
+        else:
+            return cls._export_ext()
+
+    @classmethod
+    def import_from(
+        cls: t.Type[T],
+        path: str,
+        input_format: t.Optional[str] = None,
+        *,
+        protocol: t.Optional[str] = None,
+        user: t.Optional[str] = None,
+        passwd: t.Optional[str] = None,
+        params: t.Optional[t.Dict[str, str]] = None,
+        subpath: t.Optional[str] = None,
+    ) -> T:
+        try:
+            parsedurl = fs.opener.parse(path)
+        except fs.opener.errors.ParseError:
+            if protocol is None:
+                protocol = "osfs"
+                resource = path if os.sep == "/" else path.replace(os.sep, "/")
+        else:
+            if any(v is not None for v in [protocol, user, passwd, params, subpath]):
+                raise ValueError(
+                    "An FS URL was passed as the output path; all additional information should be passed as part of the URL."
+                )
+
+            protocol = parsedurl.protocol
+            user = parsedurl.username
+            passwd = parsedurl.password
+            resource = parsedurl.resource
+            params = parsedurl.params
+            subpath = parsedurl.path
+
+        if protocol not in fs.opener.registry.protocols:
+            if protocol == "s3":
+                raise ValueError(
+                    "Tried to open an S3 url but the protocol is not registered; did you 'pip install fs-s3fs'?"
+                )
+            else:
+                raise ValueError(
+                    f"Unknown or unsupported protocol {protocol}. Some supported protocols are 'ftp', 's3', and 'osfs'."
+                )
+
+        isOSPath = protocol in [
+            "temp",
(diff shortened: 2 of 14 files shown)


bentoml/_internal/bento/bento.py
```diff
@@ -302,9 +302,111 @@ def save(
 
         return self
 
-    def export(self, path: str):
-        out_fs = fs.open_fs(path, create=True, writeable=True)
+    Format: TypeAlias = t.Literal["bento", "gz", "xz", "bz2", "tar", "zip", "folder"]
```
parano, line 305: The CI is failing because of missing import in this file:
```
E0602:undefined-variable [305:12] Bento: Undefined variable 'TypeAlias'
13
E0602:undefined-variable [306:17] Bento: Undefined variable 'TypeAlias'
14
E0602:undefined-variable [307:18] Bento: Undefined variable 're'
15
E0602:undefined-variable [340:27] Bento.export: Undefined variable 'hasSchemeRe'
```

bentoml/_internal/bento/bento.py
```diff
@@ -302,9 +302,111 @@ def save(
...
+    Destination: TypeAlias = t.Literal["fs", "s3", "ftp"]
+    hasSchemeRe = re.compile(r".*[^\\]://")
+
+    def export(
+        self,
+        path: str,
+        dest: t.Optional[Destination] = None,
+        output_format: t.Optional[Format] = None,
+        *,
+        user: t.Optional[str] = None,
```
parano, line 315: rename to `ftp_user` and `ftp_passwd` to be more clear to users?

aarnphm:
>     * `bento.export("")` - export the bento to `name-version.bento` in the current directory. this isn't a default because it feels weird for the API to have a default; arguably this should be handled by the CLI.
`bento.export()`?
 
>     * `bento.export("dir/")` - even if `dir` is not a directory, try to output to `dir/name-version.bento`. note this doesn't actually try to create the folder, so save for weird filesystems where directories don't need to be created this will error if `dir` doesn't exist.
should we create the directory if it doesn't exist? I think we should just enforce this behaviour to export to a directory.
 
>     * `bento.export("/path/to/filename.<supported extension>")` - will use the relevant format and not append anything to the filename.
For these supported extensions would it just be tarball related? There might be cases where users might try to export the file to have `.pt` extension. Also I think if we can support json export that could be powerful

sauyon:
> > ```
> > * `bento.export("")` - export the bento to `name-version.bento` in the current directory. this isn't a default because it feels weird for the API to have a default; arguably this should be handled by the CLI.
> > ```
> 
> `bento.export()`?

I still think it's weird for the API to have a default that writes to the CWD, so I'm more in favor of nixing this entirely.

> > ```
> > * `bento.export("dir/")` - even if `dir` is not a directory, try to output to `dir/name-version.bento`. note this doesn't actually try to create the folder, so save for weird filesystems where directories don't need to be created this will error if `dir` doesn't exist.
> > ```
> 
> should we create the directory if it doesn't exist? I think we should just enforce this behaviour to export to a directory.

The reason I chose not to create the directory is because that's generally what tools like this do.

I think not enforcing this is most likely generally fine; while it makes the documentation slightly harder I think in the vast majority of cases where the user specifies a directory that already exists they mean to output into that directory.

My philosophy here was to make it as complex as it needs to be in order to do what the user likely intended when they wrote the command. That said, I think maybe the API shouldn't try to do all this complex stuff and it should be handled by the CLI.

> > ```
> > * `bento.export("/path/to/filename.<supported extension>")` - will use the relevant format and not append anything to the filename.
> > ```
> 
> For these supported extensions would it just be tarball related? There might be cases where users might try to export the file to have `.pt` extension. Also I think if we can support json export that could be powerful

At the moment the only supported extensions are tarfiles (`gz`, `bz2`, `xz`, `bento` [equivalent to `tar.xz`, though that is subject to change]) and zipfiles, but there's nothing stopping us from implementing more formats inside of `_compress`.

bentoml/_internal/bento/bento.py
sauyon: This is actually also applicable elsewhere so I'm moving these to work more generally.

bentoml/_internal/bento/bento.py
```diff
@@ -302,9 +302,111 @@ def save(
...
+
+        # try to guess output format if it hasn't been passed
+        if output_format is None:
+            _, ext = fs.path.splitext(path)
+            if ext in ["bento", "gz", "xz", "bz2", "tar", "zip"]:
+                output_format = ext
+            else:
+                output_format = "bento"
+
+        if path == "" or (not output_format == "folder" and path[-1] == "/"):
```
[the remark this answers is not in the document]
sauyon: For the future: we discussed this and are going to try using `beartype`.

bentoml/_internal/bento/bento.py
```diff
@@ -88,23 +88,45 @@ def _Bento_set_fs(bento: "Bento", _: t.Any, new_fs: "FS") -> "FS":
 @attr.define(repr=False, auto_attribs=False)
 class Bento(StoreItem):
     _tag: Tag = attr.field()
-    _fs: "FS" = attr.field(on_setattr=_Bento_set_fs)
+    __fs: "FS" = attr.field(on_setattr=_Bento_set_fs)
```
larme, line 91: why we want this renaming for _fs while other private member still use single underscore?

bentoml/_internal/bento/bento.py
```diff
@@ -88,23 +88,45 @@ def _Bento_set_fs(bento: "Bento", _: t.Any, new_fs: "FS") -> "FS":
...
     def tag(self) -> Tag:
         return self._tag
 
+    @property
+    def _fs(self) -> "FS":
+        return self.__fs
+
+    @property
+    def info(self) -> "BentoInfo":
+        self._flushed = False
```
larme, line 118: Why accessing `.info` will mark self dirty? Incase we use `bento.info['options'] = xxx`?

bentoml/_internal/models/model.py
```diff
@@ -173,25 +191,51 @@ def path_of(self, item: str) -> str:
...
     def flush_custom_objects(self):
+        if self._flushed:
+            return
+
         # pickle custom_objects if it is not None and not empty
         if self.custom_objects:
             with self._fs.open(CUSTOM_OBJECTS_FILENAME, "wb") as cofile:
                 cloudpickle.dump(self.custom_objects, cofile)
 
+        self._flushed = True
```
larme, line 211: Is it possible that both `custom_objects` and `info` is dirty and we only flush `custom_objects` but marked self flushed, so the `info` part change is lost?

bentoml/_internal/exportable.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def _compress(self, path: str, output_format: str):
+        if output_format in ["gz", "xz", "bz2", "tar", self._export_ext]:
+            import fs.tarfs
+
+            compression = output_format
+            if compression == "tar":
+                compression = None
+            if compression == self._export_ext:
+                compression = "xz"
```
larme, line 220: Is format like "gz" and "bz2" handled?

larme (changes requested):

bentoml/_internal/bento/bento.py
aarnphm: actually we shouldn't use double underscore

sauyon: The reason for the double underscore is that `_fs` is now a method.

bentoml/_internal/exportable.py
```diff
@@ -0,0 +1,232 @@
...
+            else:
+                resource, subpath = fs.path.split(resource)
+
+        # as a special case, if the output format is the default we will always append the relevant
+        # extension if the output filename does not have it
+        if (
+            output_format == self._export_ext
+            and fs.path.splitext(subpath)[1] != "." + self._export_ext
+        ):
+            print(
```
aarnphm, line 160: should we do `logger` instead of `print`?

sauyon: 👍 done.

bentoml/bentos.py
```diff
@@ -54,11 +54,17 @@ def delete(
...
+        path: str,
+        input_format: t.Optional[str] = None,
+        *,
+        protocol: t.Optional[str] = None,
+        user: t.Optional[str] = None,
+        passwd: t.Optional[str] = None,
+        params: t.Optional[t.Dict[str, str]] = None,
+        subpath: t.Optional[str] = None,
     _bento_store: "BentoStore" = Provide[BentoMLContainer.bento_store],
 ) -> Bento:
```
parano, line 66: add some docstring here to explain the usage?

bentoml/models.py
```diff
@@ -44,9 +44,17 @@ def delete(
...
+        *,
+        protocol: t.Optional[str] = None,
+        user: t.Optional[str] = None,
+        passwd: t.Optional[str] = None,
+        params: t.Optional[t.Dict[str, str]] = None,
+        subpath: t.Optional[str] = None,
+    _model_store: "ModelStore" = Provide[BentoMLContainer.model_store]
 ) -> Model:
-    return Model.from_fs(fs.open_fs(path)).save(_model_store)
+    return Model.import_from(path, input_format, protocol=protocol, user=user, passwd=passwd, params=params, subpath=subpath).save(_model_store)
```
parano, line 57: same here - also in this case it's especially confusing - user may think they can import any model files, e.g. a Tensorflow SavedModel or a sklearn pickle file, we should explain that this only works for models exported from BentoML

bentoml/_internal/bento/bento.py
```diff
@@ -82,21 +82,30 @@ def get_default_bento_readme(svc: "Service"):
...
+        return "bento"
+
+    @__fs.validator
     def check_fs(self, _attr: t.Any, new_fs: "FS"):
-        new_fs.makedir("models", recreate=True)
-        self._model_store = ModelStore(new_fs.opendir("models"))
+        try:
+            new_fs.makedir("models", recreate=True)
+            self._model_store = ModelStore(new_fs.opendir("models"))
+        except fs.errors.ResourceReadOnly:
```
parano, line 103: just curious, what's the reason for the `fs.errors.ResourceReadOnly` check here?

bentoml/_internal/bento/bento.py
parano: Should we allow user to change the `info` at all? or is it only used internally?   It seems the only way user can access to a Bento instance is from `get`, `list` and `build`, where we probably don't want user to change the `info`

bentoml/_internal/exportable.py
```diff
@@ -0,0 +1,363 @@
...
+            subpath = parsedurl.path
+
+        if protocol not in fs.opener.registry.protocols:
+            if protocol == "s3":
+                raise ValueError(
+                    "Tried to open an S3 url but the protocol is not registered; did you 'pip install fs-s3fs'?"
+                )
+            else:
+                raise ValueError(
+                    f"Unknown or unsupported protocol {protocol}. Some supported protocols are ftp, s3, and osfs."
```
parano, line 90:
```suggestion
                    f"Unknown or unsupported protocol {protocol}. Some supported protocols are ftp, s3, and OS Filesystem."
```

suggest this change in error message since "osfs" probably don't mean anything to the end user

bentoml/_internal/exportable.py
```diff
@@ -0,0 +1,363 @@
...
+            "usercache",
+            "userlog",
+        ]
+
+        isCompressedPath = protocol in ["tar", "zip"]
+
+        if input_format is not None:
+            if isCompressedPath:
+                raise ValueError(
+                    f"A {protocol} path was passed along with an output format ({input_format}); only pass one or the other."
```
parano, line 109:
```suggestion
                    f"A {protocol} path was passed along with an input format ({input_format}); only pass one or the other."
```
```suggestion
                    f"A {protocol} path was passed along with an output format ({input_format}); only pass one or the other."
```

bentoml/_internal/exportable.py
```diff
@@ -0,0 +1,363 @@
...
+        if ext in [cls._export_ext(), "gz", "xz", "bz2", "tar", "zip"]:
+            return ext
+        else:
+            return cls._export_ext()
+
+    @classmethod
+    def import_from(
+        cls: t.Type[T],
+        path: str,
+        input_format: t.Optional[str] = None,
```
parano, line 56: what do you think about changing this type to a `t.Union[None, t.Literal[ SUPPORT_FORMAT_LIST ]]`? which would serve as a good documentation for supported formats, and similarly for the `protocol` below

parano: nvm, I think documenting it at the top-level user facing API is good enough

bentoml/_internal/bento/bento.py
sauyon: If the underlying FS is read-only (e.g. if the user imported a tarfile) the `makedir` call will fail but we still want to try to open the model store.

bentoml/_internal/exportable.py
sauyon: Ah, the protocol is supposed to be `"osfs"`, because that's the name of the FS protocol. Maybe I should add quotes?

bentoml/models.py
sauyon: I'll rename it to `import_bento_model`, maybe?

bentoml/_internal/bento/bento.py
sauyon: I think freezing it is probably correct. Should we do the same for models?

bentoml/_internal/exportable.py
parano: got it, sounds good!

bentoml/_internal/bento/bento.py
parano: Let's add an inline comment here?

bentoml/models.py
parano: I feel adding some docstring here should be good enough, I'm afraid `import_bento_model` may lead to users thinking it must be models packed in a Bento.  

One thing missing here is we don't seem to check if the path being imported is a valid exported model, maybe it should check that a `bentomodel.yaml` file exist in the path before importing? That way we can show a good error message telling user that import target must be a model exported from BentoML model store.

bentoml/_internal/bento/bento.py
parano: Yes I also think we should be freezing it here

parano: About models, it depends on the model creation Api we expose to users.

Currently we have:
```
model = Model.create()
# dump model files to model.path
# assign metadata to model.metadata
model.save()
```
In this case, it must be mutable so user can set metadata, options, and custom_objects.

I think one of the idea we explored a bit earlier was to provide a `with Model.create() as model:` block, in that case, we can freeze everything after the `with` block.

sauyon: Will do this in a separate PR.

parano:
Looks like all windows tests are failing now :\
```
>       assert from_fs_model_str == f'Model(tag="{model_b.tag}", path="{save_path}")'
[214](https://github.com/bentoml/BentoML/runs/5396624595?check_suite_focus=true#step:8:214)
E       assert 'Model(tag="t...2lgbwyaan\\")' == 'Model(tag="t...u2lgbwyaan/")'
[215](https://github.com/bentoml/BentoML/runs/5396624595?check_suite_focus=true#step:8:215)
E         Skipping 147 identical leading characters in diff, use -v to show
[216](https://github.com/bentoml/BentoML/runs/5396624595?check_suite_focus=true#step:8:216)
E         - u2lgbwyaan/")
[217](https://github.com/bentoml/BentoML/runs/5396624595?check_suite_focus=true#step:8:217)
E         ?           ^
[218](https://github.com/bentoml/BentoML/runs/5396624595?check_suite_focus=true#step:8:218)
E         + u2lgbwyaan\")
[219](https://github.com/bentoml/BentoML/runs/5396624595?check_suite_focus=true#step:8:219)
E         ?           ^
```

bentoml/_internal/bento/bento.py
parano: sounds good!

bentoml/_internal/bento/bento.py
```diff
@@ -252,14 +277,17 @@ def from_fs(cls, item_fs: "FS") -> "Bento":
...
+                f"Failed to load bento because it does not contain a '{BENTO_YAML_FILENAME}'"
+            )
 
         res = cls(info.tag, item_fs, info)
         if not res.validate():
-            logger.warning(f"Failed to import Bento from {item_fs}.")
-            raise BentoMLException("Failed to create Bento because it was invalid")
+            raise BentoMLException(
+                f"Failed to create bento because it contains an invalid '{BENTO_YAML_FILENAME}'"
+            )
```
parano, line 288: 👍

bentoml/models.py
```diff
@@ -44,19 +43,155 @@ def delete(
...
 ) -> Model:
-    return Model.from_fs(fs.open_fs(path)).save(_model_store)
+    """
+    Import a bento model exported with :code:`bentoml.models.export_model`. To import a model saved
+    with a framework, see the :code:`save` function under the relevant framework, e.g.
+    :code:`bentoml.sklearn.save`.
+
+    Examples:
+
+    .. code-block:: python
```
parano, line 63:
```suggestion
    .. code-block:: python
    
```
This fixed the formatting, currently this code block is not showing up: https://bentoml--2237.org.readthedocs.build/en/2237/api/stores.html

bentoml/models.py
```diff
@@ -44,19 +43,155 @@ def delete(
...
+    subpath: t.Optional[str] = None,
     _model_store: "ModelStore" = Provide[BentoMLContainer.model_store],
-):
+) -> str:
+    """
+    Export a BentoML model.
+
+    Examples:
+
+    .. code-block:: python
```
parano, line 136:
```suggestion
    .. code-block:: python
    
```

bentoml/bentos.py
```diff
@@ -55,21 +54,153 @@ def delete(
...
+    subpath: t.Optional[str] = None,
     _bento_store: "BentoStore" = Provide[BentoMLContainer.bento_store],
-):
+) -> str:
+    """
+    Export a bento.
+
+    Examples:
+
+    .. code-block:: python
```
parano, line 144:
```suggestion
    .. code-block:: python
    
```

bentoml/bentos.py
```diff
@@ -55,21 +54,153 @@ def delete(
...
+    subpath: t.Optional[str] = None,
     _bento_store: "BentoStore" = Provide[BentoMLContainer.bento_store],
 ) -> Bento:
-    return Bento.from_fs(fs.open_fs(path))
+    """
+    Import a bento.
+
+    Examples:
+
+    .. code-block:: python
```
parano, line 71:
```suggestion
    .. code-block:: python
    
```

parano:
Looks like docstring issue has been fixed now: https://bentoml--2237.org.readthedocs.build/en/2237/api/stores.html#bento-store

@sauyon could you take a look at the windows issue? everything else looks great

sauyon:
I think the Windows issues was just me hardcoding `"/"` in the tests instead of the correct `os.path.sep`, so hopefully things should be good now.

PR #2253, opened 2022-02-05
fix: typing issues with xgboost

aarnphm:

Include a `FrozenDict` implementation for freezing default dictionary for better typing

### DIFF
--- a/bentoml/_internal/frameworks/xgboost.py
+++ b/bentoml/_internal/frameworks/xgboost.py
@@ -11,14 +11,15 @@ from bentoml import Runner
 from bentoml.exceptions import BentoMLException
 from bentoml.exceptions import MissingDependencyException
 
+from ..types import LazyType
+from ..types import FrozenDict
 from ..models import JSON_EXT
 from ..models import SAVE_NAMESPACE
 from ..utils.pkg import get_pkg_version
 from ..configuration.containers import BentoMLContainer
 
 if TYPE_CHECKING:
-    import pandas as pd
-
+    from .. import ext_typing as ext
     from ..models import ModelStore
 
 try:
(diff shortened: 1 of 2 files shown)


bentoml/_internal/types.py
```diff
@@ -268,6 +268,23 @@ def _format_dt(d: t.Union[datetime, str], _: type) -> datetime:
 cattr.register_structure_hook(datetime, _format_dt)  # type: ignore[misc]
 
 
+class FrozenDict(t.Dict[str, t.Any]):
```
bojiang, line 271: This solution actually breaks the consensus of typing. We devs see the arg type is Dict, they will assume that they can do `dict.update` on it.

bentoml/_internal/frameworks/xgboost.py
```diff
@@ -78,9 +79,9 @@ def _get_model_info(
 @inject
 def load(
     tag: t.Union[str, Tag],
-    booster_params: t.Optional[t.Dict[str, t.Union[str, int]]] = None,
+    booster_params: t.Dict[str, t.Union[str, int]] = FrozenDict(),
```
bojiang, line 82: This solution actually breaks the consensus of typing. We devs see an arg which is Dict, they will assume that they can do `dict.update` on it.

bentoml/_internal/frameworks/xgboost.py
```diff
@@ -116,10 +119,10 @@ def load(
 @inject
 def save(
     name: str,
-    model: "xgb.core.Booster",
+    model: "xgb.Booster",
     *,
-    booster_params: t.Optional[t.Dict[str, t.Union[str, int]]] = None,
-    metadata: t.Optional[t.Dict[str, t.Any]] = None,
+    booster_params: t.Dict[str, t.Union[str, int]] = FrozenDict(),
```
parano, line 124: FrozenDict will break `_setup_booster_params` below, having default value for non primitive types is generally not recommended

bentoml/_internal/frameworks/xgboost.py
aarnphm: but the idea of FrozenDict is that `update` or all ops is forbidden if the given parameters is not given?

aarnphm:
chatted offline with Bojiang.

PR #2254, opened 2022-02-05
refactor(tensorflow): separate V1 and V2 impl

aarnphm:
refactor `ext_typings` to its own modules for typing use case
wip tensorflow types, and make `_internal/utils/tensorflow.py` more clear with type behaviour
separate tensorflow_v1 and tensorflow_v2


bentoml/_internal/external_typing/tensorflow.py
```diff
@@ -0,0 +1,356 @@
+# type: ignore[reportMissingTypeStubs]
+import os
```
bojiang, line 2: @aarnphm For TensorFlow which is not typed officially, we may just implement the pyi file for it.

bentoml/_internal/external_typing/__init__.py
```diff
@@ -0,0 +1,42 @@
+from typing import TYPE_CHECKING
```
bojiang, line 1: We should keep stubs in `<repo_root>/typing/`

bentoml/_internal/external_typing/tensorflow.py
```diff
@@ -0,0 +1,356 @@
+# type: ignore[reportMissingTypeStubs]
```
bojiang, line 1: We should keep stubs in `<repo_root>/typing/`

aarnphm:
code style check depends on typing refactor, will be tackled in a different PR.

aarnphm:
@bojiang ready now

bentoml/_internal/configuration/containers.py
```diff
@@ -24,7 +24,7 @@
 if TYPE_CHECKING:
     from multiprocessing.synchronize import Lock as SyncLock  # noqa: F401
 
-    from .. import ext_typing as ext
+    from .. import external_typing as ext  # noqa # pylint: disable=unused-import
```
parano, line 27: why do we need `disable=unused-import` here?

parano: does the next line that imports `PrometheusClient` need it as well? is there a way to disable this check for TYPE_CHECKING related imports?

aarnphm: pep8 was complaining about this

aarnphm: I think we already disable this for `TYPE_CHECKING`, but the pep8 bot still seem to complain

bojiang: It's an example of a false positive raised by flake8. That's why we abandon flake8. It has many overlaps with pylint, and many differences are false positives.
We are not using flake8 as the linter anymore. So we can just remove those #noqa #pylint things.

> is there a way to disable this check for TYPE_CHECKING related imports?
That's the behavior of pylint now.

We can turn off the flake8 checking in our editors now.

bentoml/_internal/frameworks/tensorflow_v1.py
```diff
@@ -72,130 +49,65 @@
...
-        return loaded
-
-
 @inject
 def load(
-    tag: Tag,
+    tag: t.Union[str, Tag],
     tfhub_tags: t.Optional[t.List[str]] = None,
-    tfhub_options: t.Optional[t.Any] = None,
+    tfhub_options: t.Optional["tf_ext.SaveOptions"] = None,
```
bojiang, line 107: maybe just `options`? They are tf.saved_model.options.

bojiang: the same for the `tfhub_tags`

bentoml/_internal/frameworks/tensorflow_v1.py
```diff
@@ -611,72 +527,50 @@ def required_models(self) -> t.List[Tag]:
...
-                    tf.Tensor,
-                ]
-            ) -> tf.Tensor:
-                if not isinstance(item, tf.Tensor):
-                    return tf.convert_to_tensor(item, dtype=tf.float32)
+            def _mapping(item: "TFArgType") -> "tf_ext.TensorLike":
+                if not LazyType["tf_ext.TensorLike"]("tf.Tensor").isinstance(item):
+                    return t.cast(
+                        "tf_ext.TensorLike",
+                        tf.convert_to_tensor(item, dtype=tf.float32),
```
bojiang, line 560: why float32 here?

bentoml/_internal/frameworks/tensorflow_v1.py
```diff
@@ -250,58 +162,64 @@ def load(
             " the functionalities of the given model, set `load_as_wrapper=False`"
             " will return a SavedModel object."
         )
-        assert all(
-            i is not None for i in [hub, resolve, native_module]
-        ), MissingDependencyException(
-            "`tensorflow_hub` is required to load a tfhub module."
-        )
+
+        if hub is None:
```
bojiang, line 166: do we need tf_hub here, since we are just loading the model from model store?

bentoml/_internal/frameworks/tensorflow_v1.py
aarnphm: because we don't want to do float64

bentoml/_internal/frameworks/tensorflow_v1.py
aarnphm: we do, because if the model is imported from tfhub we need to know that

bentoml/_internal/frameworks/tensorflow_v1.py
aarnphm: oh i see

aarnphm: so should we just not specify the dtype then?

bojiang (approved):

PR #2260, opened 2022-02-08
Docs/getting started

timliubentoml:
## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,14 +8,14 @@
 [![documentation_status](https://readthedocs.org/projects/bentoml/badge/?version=latest)](https://docs.bentoml.org/)
 [![join_slack](https://badgen.net/badge/Join/BentoML%20Slack/cyan?icon=slack)](https://join.slack.bentoml.org)
 
-BentoML let you create machine learning powered prediction service in minutes and bridges the gap between data science and DevOps.
+BentoML let you create a machine learning powered prediction service in minutes and bridges the gap between data science and DevOps.
+
+👉 [Pop into our Slack community!](https://join.slack.bentoml.org) We're happy to help with any issue you face or even just to meet you and hear what you're working on :)
 
 __The BentoML version 1.0 is around the corner.__ For stable release version 0.13, see
 the [0.13-LTS branch](https://github.com/bentoml/BentoML/tree/0.13-LTS). Version 1.0 is 
 under active development, you can be of great help by testing out the preview release, 
 reporting issues, contribute to the documentation and create sample gallery projects.
-
-👉 [Join our Slack community](https://join.slack.bentoml.org) to receive project updates and get involved with latest development.
 
 ## Why BentoML ##
 
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,5 +1,9 @@
 ![bentoml-docker](tools/bentoml-docker.png)
 
+BentoML let you create a machine learning powered prediction service in minutes and bridges the gap between data science and DevOps.
+
+👉 [Pop into our Slack community!](https://join.slack.bentoml.org) We're happy to help with any issue you face or even just to meet you and hear what you're working on :)
+
 ## Table of Content
 - [Overview](#overview)
 - [Basic Usage](#usage)
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -16,12 +16,12 @@ Unified Model Serving Framework
 
 BentoML let you create machine learning powered prediction service in minutes and bridges the gap between data science and DevOps.
 
+👉 `Pop into our Slack community! <https://join.slack.bentoml.org>`_ We're happy to help with any issue you face or even just to meet you and hear what you're working on :)
+
 The BentoML version 1.0 is around the corner. For stable release version 0.13, see
 the `0.13-LTS branch <https://github.com/bentoml/BentoML/tree/0.13-LTS>`_. Version 1.0 is
 under active development, you can be of great help by testing out the preview release,
 reporting issues, contribute to the documentation and create sample gallery projects.
-
-👉 `Join our Slack community <https://join.slack.bentoml.org>`_ to receive project updates and get involved with latest development.
 
 Why BentoML
 -----------
--- a/docs/source/quickstart.rst
+++ b/docs/source/quickstart.rst
@@ -3,16 +3,17 @@
 Getting Started
 ===============
 
-A "Bento" is a collection of models and methods which is packaged and optimized to run your Machine Learning workflows in your choice of either local or cloud service.
+In this guide we will show you how to create a local web service for your machine learning model(s). Then we will package that web service into a self contained package (Bento) which is ready for production deployment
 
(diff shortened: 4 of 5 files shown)


docs/source/index.rst
```diff
@@ -16,13 +16,13 @@ Unified Model Serving Framework
 
 BentoML let you create machine learning powered prediction service in minutes and bridges the gap between data science and DevOps.
 
+👉 `Pop into our Slack community! <https://join.slack.bentoml.org>`_ We're happy to help with any issue you face or even just to meet you and hear what you're working on :)
```
yubozhao, line 19: love it

parano, line 19: could we also make this change to the README file https://github.com/bentoml/BentoML/blob/main/README.md and the docker hub readme template file here: https://github.com/bentoml/BentoML/blob/main/docker/templates/docs/README.md.j2 - this docker template file will be rendered on the docker hub page here:  https://hub.docker.com/r/bentoml/bento-server

docs/source/quickstart.rst
```diff
@@ -167,6 +182,10 @@ To build a Bento, first create a :code:`bentofile.yaml` in your project director
 
 Next, use the :code:`bentoml build` CLI command in the same directory to build a bento.
 
+Please ensure that:
+1. The "service" name is the same as the python file name where you're service is declared
```
parano, line 186: this is not required by bentoml, but seem like a good idea to recommend users to do this as a contention

docs/source/quickstart.rst
```diff
@@ -167,6 +182,10 @@ To build a Bento, first create a :code:`bentofile.yaml` in your project director
...
+2. That you are instantiating your load runners as part of the main path of execution in your python file, not inside of a function. This is because BentoML inspects the code to determine which models to save with the Bento.
```
parano, line 187: this is already enforced by the API design, since the service instance creation: `bentoml.Service(name, runners=[ .. ])` will need a list of runner instances.

docs/source/index.rst
timliubentoml: ah, nice, I didn't know that. cool yea i'll do it

docs/source/quickstart.rst
timliubentoml: ok, because this was something weird that wasn't working for me unless the service name was the same as the python file name and it kind of messed me up a little. I must have missed something about how these 2 go together

docs/source/quickstart.rst
```diff
@@ -105,6 +118,8 @@ In this example, we defined the input and output type to be :code:`numpy.ndarray
 :code:`pandas.DataFrame` and :code:`PIL.image` are also supported. To see all supported options, see
 :ref:`API and IO Descriptors <api-io-descriptors>`.
 
+BentoML uses a performance optimized version of https://www.starlette.io/ in order to serve your model efficiently at scale.
```
parano, line 121:
```suggestion
BentoML uses two of the fastest Python web framework `starlette https://www.starlette.io/`_ and `uvicorn https://www.uvicorn.org`, in order to serve your model efficiently at scale.
```

* https://www.techempower.com/benchmarks/#section=data-r17&hw=ph&test=fortune&l=zijzen-1

docs/source/quickstart.rst
```diff
@@ -75,7 +88,7 @@ Define and Debug Services
 -------------------------
 
 Services are the core components of BentoML, where the serving logic is defined. With the model
-saved in the model store, we can define the :ref:`service <service-definition-page>` by creating a
+saved in the model store, we can define the :ref:`web service <service-definition-page>` by creating a
```
parano, line 91:
```suggestion
saved in the model store, we can define the :ref:`service <service-definition-page>` by creating a
```

Just "service"? since it's called `bentoml.Service` in the API?

docs/source/quickstart.rst
timliubentoml: hmmm... yea I guess so... It seems a little weird to me that you have to instantiate the runner (even though the model isn't loaded) to go into something static like an annotation. But yea, I see what you're saying. Didn't know if this was a common python convention or not.

docs/source/quickstart.rst
timliubentoml: ah right. I guess I was trying to clarify that what you're getting is a web service. There's no other types of services that you can instantiate right?

docs/source/index.rst
timliubentoml: done

docs/source/quickstart.rst
parano: @timliubentoml we should probably mention the runner is lazily loaded on the runner process?

README.md
```diff
@@ -8,15 +8,15 @@
 [![documentation_status](https://readthedocs.org/projects/bentoml/badge/?version=latest)](https://docs.bentoml.org/)
 [![join_slack](https://badgen.net/badge/Join/BentoML%20Slack/cyan?icon=slack)](https://join.slack.bentoml.org)
 
-BentoML let you create machine learning powered prediction service in minutes and bridges the gap between data science and DevOps.
+BentoML let you create a machine learning powered prediction service in minutes and bridges the gap between data science and DevOps.
```
aarnphm, line 11: Should we do something like this?
```suggestion
BentoML enables users to create a machine learning powered prediction service in minutes and bridges the gap between data science and DevOps.
```

docker/README.md
```diff
@@ -1,5 +1,9 @@
 ![bentoml-docker](tools/bentoml-docker.png)
 
+BentoML let you create a machine learning powered prediction service in minutes and bridges the gap between data science and DevOps.
```
aarnphm, line 3: cc same changes in `README.md`.

README.md
sauyon: I think "let's you" is fine, FWIW.

Maybe something like "BentoML is a tool that allows the creation of..." would be more in line with standard wording.

docs/source/quickstart.rst
timliubentoml: the current version I'm about to push completely removes this, and rather annotates the bentofile.yaml and makes a couple of these things slightly more clear. For #2, as you said, based on convention it would be difficult to misconfigure, so maybe leave that clarification out for now. Or add an extra comment on the code that does it to clarify

README.md
timliubentoml: Yea, I'm also in favor of using the 2nd person in general

docker/README.md
```diff
@@ -1,5 +1,9 @@
 ![bentoml-docker](tools/bentoml-docker.png)
 
+BentoML enables users to create a machine learning powered prediction service in minutes and bridges the gap between data science and DevOps.
```
parano, line 3: @timliubentoml this is actually a developer facing API explaining how to add new docker base image, the user facing one is a template file located at `docker/templates/docs/README.md.j2`: https://github.com/bentoml/BentoML/blob/main/docker/templates/docs/README.md.j2

parano:
LGTM!

PR #2272, opened 2022-02-15
feat(internal): File I/O descriptor (re-)implementation

sauyon:
Currently contains stuff with `path` that I think I will probably remove because it's a bit of a pain to deal with so far.


sauyon:
Pushed the current implementation.

I've ended up returning a `FileLike` anyway because there was no other way to assign the `name` field, but it's now just a thin wrapper around the underlying object.

I removed the URL handling bits of it too because they didn't seem like they were being used, and I had security concerns about them anyway.

aarnphm (commented):
LGTM, also we should just start putting `from __future__ import annotations` per file we touch :)

PR #2279, opened 2022-02-16
refactor(internal): make the bento fs onsetattr a validator

sauyon:
Seems ever so slightly nicer.

### DIFF
--- a/bentoml/_internal/bento/bento.py
+++ b/bentoml/_internal/bento/bento.py
@@ -79,25 +79,26 @@ python:
     return doc
 
 
-def _Bento_set_fs(bento: "Bento", _: t.Any, new_fs: "FS") -> "FS":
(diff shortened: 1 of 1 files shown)


bentoml/_internal/bento/bento.py
```diff
@@ -79,25 +79,26 @@ def get_default_bento_readme(svc: "Service"):
...
+    _fs: "FS" = attr.field()
 
     info: "BentoInfo"
 
     _model_store: ModelStore
     _doc: t.Optional[str] = None
 
+    @_fs.validator
+    def check_fs(self, _attr: t.Any, new_fs: "FS"):
+        print(f"TYPE: {type(attr)}")
```
parano, line 94:
```suggestion
```

PR #2288, opened 2022-02-18
refactor(sdk&framework): [Broken] framework runners

bojiang:
## Checklist:


aarnphm:
we are removing `load_runner` ?

bentoml/_internal/frameworks/detectron.py
```diff
@@ -3,6 +3,7 @@
 from typing import TYPE_CHECKING
 
 import numpy as np
+from torch._C import device
```
parano, line 6: should this be in the conditional import `try` block?

bentoml/_internal/frameworks/keras.py
```diff
@@ -403,11 +377,8 @@ def load_runner(
     *,
     predict_fn_name: str = "predict",
     device_id: str = "CPU:0",
-    predict_kwargs: t.Optional[t.Dict[str, t.Any]] = None,
+    partial_kwargs: t.Optional[t.Dict[str, t.Any]] = None,
```
parano, line 380: also update in the doctoring below

bentoml/_internal/frameworks/catboost.py
```diff
@@ -33,6 +25,16 @@
         """
     )
 
+if TYPE_CHECKING:
+    from .. import external_typing as ext
+    from ..models import ModelStore
+
+    CatBoostModel = t.Union[
```
parano, line 32: great refactoring 👍

bentoml/_internal/frameworks/catboost.py
```diff
@@ -227,59 +215,43 @@ def save(
     return _model.tag
 
 
-class _CatBoostRunner(Runner):
+class _CatBoostRunner(BaseModelRunner):
     @inject
     def __init__(
         self,
-        tag: Tag,
+        tag: t.Union[str, Tag],
```
parano, line 222: Should we make everything below kwarg only?

```suggestion
        tag: t.Union[str, Tag],
        *,
```

bentoml/_internal/frameworks/catboost.py
```diff
@@ -95,9 +89,7 @@ def load(
     tag: t.Union[str, Tag],
     model_params: t.Optional[t.Dict[str, t.Union[str, int]]] = None,
     model_store: "ModelStore" = Provide[BentoMLContainer.model_store],
```
parano, line 91: I saw that we removed `model_store` inject from `load_runner`, should we also do that for `save` and `load`?

I think it makes more sense to just use public API under `bentoml.models`  in our framework modules - I don't think allowing user to customize model_store for each `save` or `load` call is a valid use case, given that 1.0 will bring in a pretty powerful export/import feature that @sauyon is working on.

bentoml/_internal/frameworks/common/model_runner.py
```diff
@@ -0,0 +1,58 @@
...
+    @inject
+    def _model_info(
+        self,
+        model_store: "ModelStore" = Provide[BentoMLContainer.model_store],
+    ):
+        return model_store.get(self._tag)
+
+    @property
+    def default_name(self) -> str:
+        return Tag.from_taglike(self._tag).name
```
parano, line 36: any reason not to do `self._tag = Tag.from_taglike(tag)` in `__init__`?

bentoml/_internal/frameworks/common/model_runner.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class BaseModelSimpleRunner(SimpleRunner):
+    def __init__(
+        self,
+        tag: t.Union[str, Tag],
+        name: t.Optional[str] = None,
+    ):
+        super().__init__(name)
+        self._tag = tag
```
parano, line 50: same here

bentoml/_internal/runner/runner.py
```diff
@@ -91,25 +96,27 @@ class BatchOptions:
...
-        if not name.isidentifier():
-            name = VARNAME_RE.sub("_", name)
-        self.name = name
-        self.resource_quota = ResourceQuota(
-            **(resource_quota if resource_quota else {})
-        )
-        self.batch_options = BatchOptions(**(batch_options if batch_options else {}))
-        self._impl_provider = SingletonFactory(create_runner_impl, self)
+    """
+    This class should not be implemented directly. Instead, implement the SimpleRunner or Runner.
```
parano, line 100: This class will not be exposed to user right?

```suggestion
    Base class for SimpleRunner or Runner.
```

bentoml/_internal/runner/runner.py
```diff
@@ -119,23 +126,45 @@ def num_replica(self) -> int:
...
 
+    @final
     async def async_run_batch(self, *args: t.Any, **kwargs: t.Any) -> t.Any:
         return await self._impl.async_run_batch(*args, **kwargs)
 
+    @final
     def run(self, *args: t.Any, **kwargs: t.Any) -> t.Any:
         return self._impl.run(*args, **kwargs)
 
+    @final
```
parano, line 167: 👍

bentoml/_internal/runner/runner.py
```diff
@@ -208,6 +237,8 @@ class RunnerState(enum.IntEnum):
     INIT = 0
     SETTING = 1
     SET = 2
+    SHUTIING_DOWN = 3
+    SHUTDOWN = 4
```
parano, line 241: Not related in this PR, suggest renaming the two states above:
```
UNINITILIZED = 0
INITIALIZING = 1
READY = 2
```

And should it also have an `ERROR` state? in case `_setup` failed in runner_app?

bentoml/_internal/frameworks/catboost.py
bojiang: Let's do that in another PR

bentoml/_internal/runner/runner.py
bojiang: The names reflect our `setup` and `shutdown` API.

bojiang: But `READY` lgtm

PR #2292, opened 2022-02-19
chore(transformers): returns pipelines automatically

aarnphm:

`load` should return the `pipelines` when `pipeline` is saved with `save`. Before `load` will return `model` and `tokenizer`, and users can load those into pipelines. This could be inconsistent as pipeline tasks will depend on the users to decide what to do with those models. This PR solves this problem:
```python
import transformers
import bentoml

pipeline = transformers.pipeline('text-classification', from_tf=True)
tag = bentoml.transformers.save('distilbert', pipeline)

...
loaded = bentoml.transformers.load(tag, from_tf=True) # type: transformers.pipelines.base.Pipeline
```

## Motivation and Context

Before this PR:
```python
import transformers
import bentoml

pipeline = transformers.pipeline('text-classification', from_tf=True)
tag = bentoml.transformers.save('distilbert', pipeline)

...
model, tokenizer = bentoml.transformers.load(tag)
# the new pipeline might have a different tasks, and is not consistent with our save pipeline.
new_pipeline = transformers.pipeline('text-generation', model=model, tokenizer=tokenizer) 
```
Also now accepts passthrough kwargs for corresponding class with `load`:
```python
config, model, tokenizer = bentoml.transformers.load(tag, return_config=True, model_kwargs={"eos_token_id": "<EOS>"}, tokenizer_kwargs={"use_fast": False}, config_kwargs = {"labels":"..."}
``` 
This is similar to how users can define kwargs for model when loading pipeline, so no learning curve needed. By separating options we can pass those kwargs to the correct loaded module.

## Checklist:


aarnphm:
Will wait for #2288 before.

bentoml/_internal/frameworks/transformers.py
```diff
@@ -18,7 +18,8 @@
 from ..utils.pkg import get_pkg_version
 from ..configuration.containers import BentoMLContainer
 
-logger = logging.getLogger(__name__)
+logging.captureWarnings(True)
```
parano, line 21: what is this for?

bentoml/_internal/frameworks/transformers.py
```diff
@@ -18,7 +18,8 @@
...
+logger = logging.getLogger("bentoml")
```
parano, line 22: let's use `__name__`

bentoml/_internal/frameworks/transformers.py
aarnphm: redirect warnings capture from `jax` to logger

aarnphm:
on hold for now for logging discussion

aarnphm:
@bojiang ready now

PR #2301, opened 2022-02-23
feat: support for parquet ser/deser for PandasDataFrame

mqk:
## Description
This PR supports creating a PandasDataFrame API specification that uses parquet de/serialization, rather than JSON, to translate between the inputs/outputs of a service and the internal pandas DataFrame.

Assume you have a pandas DataFrame `df` loaded in memory. If you define an API in a bento service like this:
```
@svc.api(input=PandasDataFrame(from_parquet=True), output=PandasDataFrame())
def score(df: pd.DataFrame) -> pd.DataFrame:
    return pd.DataFrame(runner.run_batch(df), index=df.index, columns=['score'])
```
Then you can make an API request that passes in binary data in parquet format, rather than JSON text:
```
data = io.BytesIO()
df.to_parquet(data)
response = requests.post("http://127.0.0.1:5000/score", data=data.getvalue())
```

[You might also want get back parquet binary data, in which case you could set `output=PandasDataFrame(to_parquet=True)`. I have found this to be less useful in practice, but there's support for it in this PR.]

## Motivation and Context

Parquet ser/deser is _significantly_ faster than JSON ser/deser, and this change allows API post requests with large payloads (100k row, 1,500 column dataframe in this example) to complete in ~10 seconds with parquet, rather than taking ~200 seconds with JSON. (The vast majority of time in the latter is spent converting the pandas DataFrame to JSON to make the POST request and then converting it back into a pandas DataFrame inside the service. Here are some basic benchmarks:

* JSON:
  ```
  INFO:score_json:Processing 1 files with 1 cores and a chunksize of 100000.
  INFO:score_json:Reading chunk #0 took 16.37 seconds. shape: (100000, 1664)   # This is reading a CSV from disk. (I don't have control over that.)
  INFO:score_json:Data serialization took 65.64 seconds.
  INFO:score_json:BentoML request took 147.94 seconds.
  ```
* Parquet:
  ```
  INFO:score_parquet:Processing 1 files with 1 cores and a chunksize of 100000.
  INFO:score_parquet:Reading chunk #0 took 19.33 seconds. shape: (100000, 1664)
  INFO:score_parquet:Data serialization took 4.26 seconds.  # This is >15x faster!
  INFO:score_parquet:BentoML request took 8.19 seconds.  # This is 18x faster!
  ```

This was briefly discussed in a slack thread in #bentoml-users from [Feb. 15, 1:39 PM](https://bentoml.slack.com/archives/CKRANBHPH/p1644961183324039). As @parano mentioned in that slack convo, the batch scoring API is currently being developed. It's totally possible that the `batch_job` way of defining a batch API will obviate the need for making network API calls with large payloads, but I figured this might nevertheless be a useful addition to BentoML. I have been using it successfully myself.

## How Has This Been Tested?

I didn't see an obvious place that I could add a unit test for this change. If we could add a parquet-based endpoint to http://bentoml.ai, then I could add a test to tests/e2e/bento_server_general_features_sync/tests/test_io.py.

## Checklist:

If you decide that this PR is worth merging, then I will be happy to add documentation, and add tests (as discussed above).

### DIFF
--- a/bentoml/_internal/io_descriptors/pandas.py
+++ b/bentoml/_internal/io_descriptors/pandas.py
@@ -356,9 +356,7 @@ class PandasDataFrame(IODescriptor["ext.PdDataFrame"]):
         if serialization_format is SerializationFormat.JSON:
             resp = obj.to_json(orient=self._orient)  # type: ignore[arg-type]
         elif serialization_format is SerializationFormat.PARQUET:
-            resp = io.BytesIO()
-            obj.to_parquet(resp)
-            resp = resp.getvalue()
+            resp = obj.to_parquet()
         elif serialization_format is SerializationFormat.CSV:
             resp = obj.to_csv()
         else:
--- a/bentoml/_internal/io_descriptors/pandas.py
+++ b/bentoml/_internal/io_descriptors/pandas.py
@@ -222,6 +222,8 @@ class PandasDataFrame(IODescriptor["ext.PdDataFrame"]):
         :obj:`~bentoml._internal.io_descriptors.IODescriptor`: IO Descriptor that `pd.DataFrame`.
     """
 
+    has_parquet = _HAS_PARQUET
+
     def __init__(
         self,
         orient: "ext.DataFrameOrient" = "records",
@@ -247,11 +249,7 @@ class PandasDataFrame(IODescriptor["ext.PdDataFrame"]):
 
     def openapi_request_schema(self) -> t.Dict[str, t.Any]:
         """Returns OpenAPI schema for incoming requests"""
-        return {
-            f'request.headers["content-type"] (default: {self._default_format.mime_type})': {
-                "schema": self.openapi_schema_type()
-            }
-        }
+        return {self._default_format.mime_type: {"schema": self.openapi_schema_type()}}
 
     def openapi_responses_schema(self) -> t.Dict[str, t.Any]:
         """Returns OpenAPI schema for outcoming responses"""
--- a/bentoml/_internal/io_descriptors/pandas.py
+++ b/bentoml/_internal/io_descriptors/pandas.py
@@ -24,7 +24,7 @@ if TYPE_CHECKING:
 else:
     pd = LazyLoader("pd", globals(), "pandas")
 
-sys.version_info[0] == 3 and sys.version_info[1] >= 8:
+if sys.version_info[0] == 3 and sys.version_info[1] >= 8:
     from typing import Literal
 else:
     from typing_extensions import Literal
--- a/bentoml/_internal/io_descriptors/pandas.py
+++ b/bentoml/_internal/io_descriptors/pandas.py
@@ -2,6 +2,7 @@ import io
 import sys
 import typing as t
 import logging
+import importlib.util
 from enum import Enum
 from typing import TYPE_CHECKING
 
@@ -30,19 +31,14 @@ else:
 
 logger = logging.getLogger(__name__)
 
-
 # Check for parquet support
-_HAS_PARQUET = True
-try:
-    import pyarrow  # noqa: F401
-except ModuleNotFoundError:
-    try:
-        import fastparquet  # noqa: F401
-    except ModuleNotFoundError:
-        logger.warning(
-            "Neither pyarrow nor fastparquet packages found. Parquet de/serialization will not be available."
-        )
-        _HAS_PARQUET = False
+if importlib.util.find_spec("pyarrow") is not None:
+    _PARQUET_ENGINE = "pyarrow"
+elif importlib.util.find_spec("fastparquet") is not None:
+    _PARQUET_ENGINE = "fastparquet"
+else:
+    logger.warning("Neither pyarrow nor fastparquet packages found. Parquet de/serialization will not be available.")
+    _PARQUET_ENGINE = None
 
 
 def _infer_type(item: str) -> str:  # pragma: no cover
@@ -108,7 +104,7 @@ def _infer_serialization_format_from_request(
 
 
 def _validate_serialization_format(serialization_format: SerializationFormat):
-    if (serialization_format is SerializationFormat.PARQUET) and (not _HAS_PARQUET):
+    if serialization_format is SerializationFormat.PARQUET and _PARQUET_ENGINE is None:
         raise MissingDependencyException(
             "Parquet serialization is not available. Try installing pyarrow or fastparquet first."
         )
@@ -228,7 +224,7 @@ class PandasDataFrame(IODescriptor["ext.PdDataFrame"]):
         :obj:`~bentoml._internal.io_descriptors.IODescriptor`: IO Descriptor that `pd.DataFrame`.
     """
 
-    has_parquet = _HAS_PARQUET
+    parquet_engine = _PARQUET_ENGINE
 
     def __init__(
         self,
@@ -299,6 +295,7 @@ class PandasDataFrame(IODescriptor["ext.PdDataFrame"]):
         elif serialization_format is SerializationFormat.PARQUET:
             res = pd.read_parquet(  # type: ignore[arg-type]
                 io.BytesIO(obj),
+                engine=parquet_engine,
             )
         elif serialization_format is SerializationFormat.CSV:
             res = pd.read_csv(  # type: ignore[arg-type]
@@ -360,7 +357,7 @@ class PandasDataFrame(IODescriptor["ext.PdDataFrame"]):
         if serialization_format is SerializationFormat.JSON:
             resp = obj.to_json(orient=self._orient)  # type: ignore[arg-type]
         elif serialization_format is SerializationFormat.PARQUET:
-            resp = obj.to_parquet()
+            resp = obj.to_parquet(engine=parquet_engine)
         elif serialization_format is SerializationFormat.CSV:
             resp = obj.to_csv()
         else:
--- a/bentoml/_internal/io_descriptors/pandas.py
+++ b/bentoml/_internal/io_descriptors/pandas.py
@@ -155,6 +155,12 @@ class PandasDataFrame(IODescriptor["ext.PdDataFrame"]):
             Whether to enforce a certain shape. If `enforce_shape=True` then `shape`
             must be specified
 
+        from_parquet (`bool`, `optional`, default to :code:`False`):
+            Read DataFrame from binary parquet format, rather than from JSON.
+
+        to_parquet (`bool`, `optional`, default to :code:`False`):
+            Return DataFrame in binary parquet format, rather than JSON.
+
     Returns:
         :obj:`~bentoml._internal.io_descriptors.IODescriptor`: IO Descriptor that `pd.DataFrame`.
     """
@@ -168,6 +174,8 @@ class PandasDataFrame(IODescriptor["ext.PdDataFrame"]):
         enforce_dtype: bool = False,
         shape: t.Optional[t.Tuple[int, ...]] = None,
         enforce_shape: bool = False,
+        from_parquet: bool = False,
+        to_parquet: bool = False,
     ):
         self._orient = orient
         self._columns = columns
@@ -176,18 +184,21 @@ class PandasDataFrame(IODescriptor["ext.PdDataFrame"]):
         self._enforce_dtype = enforce_dtype
         self._shape = shape
         self._enforce_shape = enforce_shape
-        self._mime_type = "application/json"
+        self._mime_type_request = "application/octet-stream" if from_parquet else "application/json"
+        self._mime_type_response = "application/octet-stream" if to_parquet else "application/json"
+        self._from_parquet = from_parquet
+        self._to_parquet = to_parquet
 
     def openapi_schema_type(self) -> t.Dict[str, t.Any]:
         return _schema_type(self._dtype)
 
     def openapi_request_schema(self) -> t.Dict[str, t.Any]:
         """Returns OpenAPI schema for incoming requests"""
-        return {self._mime_type: {"schema": self.openapi_schema_type()}}
+        return {self._mime_type_request: {"schema": self.openapi_schema_type()}}
 
     def openapi_responses_schema(self) -> t.Dict[str, t.Any]:
         """Returns OpenAPI schema for outcoming responses"""
-        return {self._mime_type: {"schema": self.openapi_schema_type()}}
+        return {self._mime_type_response: {"schema": self.openapi_schema_type()}}
 
     async def from_http_request(self, request: Request) -> "ext.PdDataFrame":
         """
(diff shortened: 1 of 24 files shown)


aarnphm:
I'm wondering whether if we support this, we would need to add parquet backend as a requirements for bentoml

bentoml/_internal/io_descriptors/pandas.py
```diff
@@ -168,6 +174,8 @@ def __init__(
         enforce_dtype: bool = False,
         shape: t.Optional[t.Tuple[int, ...]] = None,
         enforce_shape: bool = False,
+        from_parquet: bool = False,
```
aarnphm, line 177: just a thought, maybe instead of having `from_parquet` and `to_parquet`, lets just do `parquet: bool=True|False`? It seems to me that these two fields are similar, so instead of having duplicates field we can just have one?

bentoml/_internal/io_descriptors/pandas.py
```diff
@@ -176,18 +184,21 @@ def __init__(
...
+        self._from_parquet = from_parquet
+        self._to_parquet = to_parquet
 
     def openapi_schema_type(self) -> t.Dict[str, t.Any]:
         return _schema_type(self._dtype)
 
     def openapi_request_schema(self) -> t.Dict[str, t.Any]:
         """Returns OpenAPI schema for incoming requests"""
-        return {self._mime_type: {"schema": self.openapi_schema_type()}}
+        return {self._mime_type_request: {"schema": self.openapi_schema_type()}}
```
aarnphm, line 197: reflect on previous comments

bentoml/_internal/io_descriptors/pandas.py
```diff
@@ -168,6 +174,8 @@ def __init__(
         enforce_dtype: bool = False,
         shape: t.Optional[t.Tuple[int, ...]] = None,
         enforce_shape: bool = False,
+        from_parquet: bool = False,
+        to_parquet: bool = False,
```
aarnphm, line 178: just a thought, maybe instead of having `from_parquet` and `to_parquet`, lets just do `parquet: bool=True|False`? It seems to me that these two fields are similar, so instead of having duplicates field we can just have one?

aarnphm: also should we make it a behaviour to `enforce_parquet`?

bentoml/_internal/io_descriptors/pandas.py
```diff
@@ -176,18 +184,21 @@ def __init__(
         self._enforce_dtype = enforce_dtype
         self._shape = shape
         self._enforce_shape = enforce_shape
-        self._mime_type = "application/json"
+        self._mime_type_request = "application/octet-stream" if from_parquet else "application/json"
+        self._mime_type_response = "application/octet-stream" if to_parquet else "application/json"
+        self._from_parquet = from_parquet
+        self._to_parquet = to_parquet
```
aarnphm, line 190: reflect on above comments.

bentoml/_internal/io_descriptors/pandas.py
```diff
@@ -176,18 +184,21 @@ def __init__(
...
 
     def openapi_request_schema(self) -> t.Dict[str, t.Any]:
         """Returns OpenAPI schema for incoming requests"""
-        return {self._mime_type: {"schema": self.openapi_schema_type()}}
+        return {self._mime_type_request: {"schema": self.openapi_schema_type()}}
 
     def openapi_responses_schema(self) -> t.Dict[str, t.Any]:
         """Returns OpenAPI schema for outcoming responses"""
-        return {self._mime_type: {"schema": self.openapi_schema_type()}}
+        return {self._mime_type_response: {"schema": self.openapi_schema_type()}}
```
aarnphm, line 201: reflect on previous comments

bentoml/_internal/io_descriptors/pandas.py
```diff
@@ -268,6 +290,8 @@ def from_sample(
         apply_column_names: bool = True,
         enforce_shape: bool = True,
         enforce_dtype: bool = False,
+        from_parquet: bool = False,
+        to_parquet: bool = False,
```
aarnphm, line 294: refers to above argument.

parano:
Thank you for the PR @mqk! I think this is definitely valuable feature, especially considering we are also adding support for async endpoints in Yatai, where user can post a long-running inference job without client waiting. And in those type of workloads, client may need to send a bigger DataFrame input to the model server. Ideally we want to support something like Apache Arrow flight for this purpose, but I think parquet is definitely a step up from JSON, in BentoML's target use cases.

bentoml/_internal/io_descriptors/pandas.py
parano: I have a similar feedback here, since `PandasDataFrame` is design to be used for both input and output of an API endpoint, a `to_parquet` parameter would not make sense when the instance is being passed to API's input parameter. Also considering we may want to add support for csv or arrow as additional wired format, I'd suggest to make it `format=json|parquet|csv`, that way it makes the API easier to extend in the future.

parano: Or we could make it `default_format=json|parquet|csv`, indicating the default wire format to use. And we can decide how to parse the client request dynamically, based on the content-type header. That way, one API endpoint can support both JSON and parquet simultaneously, it's just that client side need to specify a header, indicating the format. What do you think?

parano:
> I'm wondering whether if we support this, we would need to add parquet backend (such as`fastparquet>=0.1.4`) as a requirements for bentoml. Thanks for making the PR.

I think this will require either pyarrow or fastparquet backend, however we may not want to add that as a required dependency for BentoML. I'm inclined to just documenting this as an optional dependency, that user who needs this feature will need to add `pyarrow` or `fastparquet` to the python packages list. And in the PandasDataFrame class implementation, we should try to catch the ImportError when the format option is set to parquet or client request header content type is set to parquet.

bentoml/_internal/io_descriptors/pandas.py
aarnphm: I think having the options to pass the wire format from content-type header is very powerful. `default_format` seems the nicest of all.

aarnphm:
> I think this will require either pyarrow or fastparquet backend, however we may not want to add that as a required dependency for BentoML. I'm inclined to just documenting this as an optional dependency, that user who needs this feature will need to add `pyarrow` or `fastparquet` to the python packages list.
Agreed

bentoml/_internal/io_descriptors/pandas.py
mqk: Good suggestions! I've given the `default_format` a shot. Let me know what you think.

bentoml/_internal/io_descriptors/pandas.py
```diff
@@ -176,18 +240,22 @@ def __init__(
...
+        self._default_format = SerializationFormat[default_format.upper()]
 
     def openapi_schema_type(self) -> t.Dict[str, t.Any]:
         return _schema_type(self._dtype)
 
     def openapi_request_schema(self) -> t.Dict[str, t.Any]:
         """Returns OpenAPI schema for incoming requests"""
-        return {self._mime_type: {"schema": self.openapi_schema_type()}}
+        return {
+            f'request.headers["content-type"] (default: {self._default_format.mime_type})': {
```
mqk, line 251: I was a little unsure how to handle the dynamic `content-type` here. I figured that since the key of the outer dict is `str`, I could put anything in there, so I chose something descriptive.

bentoml/_internal/io_descriptors/pandas.py
```diff
@@ -184,21 +241,19 @@ def __init__(
...
+        self._default_format = SerializationFormat[default_format.upper()]
 
     def openapi_schema_type(self) -> t.Dict[str, t.Any]:
         return _schema_type(self._dtype)
 
     def openapi_request_schema(self) -> t.Dict[str, t.Any]:
         """Returns OpenAPI schema for incoming requests"""
-        return {self._mime_type_request: {"schema": self.openapi_schema_type()}}
+        return {f"request.headers[\"content-type\"] (default: {self._default_format.mime_type})":
+                    {"schema": self.openapi_schema_type()}}
```
aarnphm:
```suggestion
        return {self._default_format.mime_type: {"schema": self.openapi_schema_type()}}
```

aarnphm (commented):
Hi Michael,  ty for your PR. Can you add a test under https://github.com/bentoml/BentoML/blob/main/tests/e2e/bento_server_general_features_sync/tests/test_io.py and also one for OpenAPI under https://github.com/bentoml/BentoML/blob/main/tests/unit/_internal/io/test_openapi_schema.py?

bentoml/_internal/io_descriptors/pandas.py
aarnphm: I don't think this follows OpenAPI convention. Refers to [here](https://swagger.io/docs/specification/media-types/)

bojiang (approved):

bentoml/_internal/io_descriptors/pandas.py
mqk: Are we ok with using `_default_format.mime_type` here even though it can be overridden by the request headers["content-type"]?

aarnphm: I think so.

mqk:
The format check fails with:
```
Run make ci-format
Running black format check...
would reformat bentoml/_internal/server/__init__.py

Oh no! 💥 💔 💥
1 file would be reformatted, 22[9](https://github.com/bentoml/BentoML/runs/5327197713?check_suite_focus=true#step:9:9) files would be left unchanged.
[FAIL] black format check failed
make: *** [Makefile:31: ci-black] Error 1
Make sure to run `make format`
Error: Process completed with exit code 2.
```
However, I cannot reproduce that locally. `make ci-format` completes without errors for me:
```
(bentoml-1.0) ➜  bentoML git:(parquet-dataframe) ✗ make ci-format                                                                                                                                               [16:27:34]
Running black format check...
All done! ✨ 🍰 ✨
228 files would be left unchanged.
[PASS] black format check passed!
Running isort format check...
Skipped 13 files
[PASS] isort format check passed!
```

aarnphm:
can you run `make format` and make sure `black==22.1.0`? we recently lock `black` to the stable releases

mqk:
Yes, that's what I've done. That's how I discovered the isort issue that I fixed the last commit.
```
(bentoml-1.0) ➜  bentoML git:(parquet-dataframe) ✗ black --version                                                                                                                                              [16:40:23]
black, 22.1.0 (compiled: yes)
(bentoml-1.0) ➜  bentoML git:(parquet-dataframe) ✗ make ci-black                                                                                                                                                [16:41:01]
Running black format check...
All done! ✨ 🍰 ✨
228 files would be left unchanged.
[PASS] black format check passed!
```

bentoml/_internal/io_descriptors/pandas.py
```diff
@@ -20,10 +23,28 @@
 else:
     pd = LazyLoader("pd", globals(), "pandas")
 
+if sys.version_info >= (3, 8):
```
aarnphm, line 26:
```suggestion
sys.version_info[0] == 3 and sys.version_info[1] >= 8:
```

bentoml/_internal/io_descriptors/pandas.py
```diff
@@ -20,10 +23,28 @@
...
+try:
+    import pyarrow  # noqa: F401
+except ModuleNotFoundError:
+    try:
+        import fastparquet  # noqa: F401
+    except ModuleNotFoundError:
+        logger.warning(
+            "Neither pyarrow nor fastparquet packages found. Parquet de/serialization will not be available."
+        )
+        _HAS_PARQUET = False
```
aarnphm, line 45:
```suggestion
# Check for parquet support
import importlib.util
if importlib.util.find_spec("pyarrow") is not None:
    engine = "pyarrow"
elif import.util.find_spec("fastparquet") is not None:
    engine = "fastparquet"
else:
    logger.warning("Neither pyarrow nor fastparquet packages found. Parquet de/serialization will not be available.")
    engine = ""
```

bentoml/_internal/io_descriptors/pandas.py
```diff
@@ -52,6 +73,47 @@ def _schema_type(
...
+        return default_format
+    else:
+        logger.warning(
+            f"Content-type not specified, falling back to {default_format} serialization format."
+        )
+        return default_format
+
+
+def _validate_serialization_format(serialization_format: SerializationFormat):
+    if (serialization_format is SerializationFormat.PARQUET) and (not _HAS_PARQUET):
```
aarnphm, line 111:
```suggestion
    if serialization_format is SerializationFormat.PARQUET and engine is None:
```

bentoml/_internal/io_descriptors/pandas.py
```diff
@@ -204,18 +276,40 @@ async def from_http_request(self, request: Request) -> "ext.PdDataFrame":
...
+        if serialization_format is SerializationFormat.JSON:
+            res = pd.read_json(  # type: ignore[arg-type]
+                io.BytesIO(obj),
+                dtype=self._dtype,  # type: ignore[arg-type]
+                orient=self._orient,
+            )
+        elif serialization_format is SerializationFormat.PARQUET:
+            res = pd.read_parquet(  # type: ignore[arg-type]
+                io.BytesIO(obj),
+            )
```
aarnphm, line 302:
```suggestion
            res = pd.read_parquet(  # type: ignore[arg-type]
                io.BytesIO(obj),
                engine=engine
            )
```

bentoml/_internal/io_descriptors/pandas.py
```diff
@@ -248,17 +342,33 @@ async def to_http_response(self, obj: "pd.DataFrame") -> Response:
...
         if not LazyType["ext.PdDataFrame"](pd.DataFrame).isinstance(obj):
             raise InvalidArgument(
                 f"return object is not of type `pd.DataFrame`, got type {type(obj)} instead"
             )
-        resp = obj.to_json(orient=self._orient)  # type: ignore[arg-type]
-        return Response(resp, media_type=MIME_TYPE_JSON)
+        if serialization_format is SerializationFormat.JSON:
+            resp = obj.to_json(orient=self._orient)  # type: ignore[arg-type]
+        elif serialization_format is SerializationFormat.PARQUET:
+            resp = obj.to_parquet()
```
aarnphm, line 363:
```suggestion
            resp = obj.to_parquet(engine=engine)
```

mqk:
@aarnphm Nice, I like your suggestion of how to handle the parquet detection. One question though: I need to know whether parquet is available in the e2e test test_io.py. In your suggestion `engine` is a module-level variable and since it's in a private path (`_internal`) I don't think I should import it directly like that from test_io.py. My solution is to name the module-level variable `_PARQUET_ENGINE` (leading underscore to indicate that it's "private" to this module, and all-caps because it's a global variable in this module) and then copy that into a class variable `PandasDataFrame.parquet_engine`, which I can now access in test_io.py by importing `bentoml.io.PandasDataFrame`. That seems like a good solution to me, but let me know if you feel otherwise.

aarnphm:
> @aarnphm Nice, I like your suggestion of how to handle the parquet detection. One question though: I need to know whether parquet is available in the e2e test test_io.py. In your suggestion `engine` is a module-level variable and since it's in a private path (`_internal`) I don't think I should import it directly like that from test_io.py. My solution is to name the module-level variable `_PARQUET_ENGINE` (leading underscore to indicate that it's "private" to this module, and all-caps because it's a global variable in this module) and then copy that into a class variable `PandasDataFrame.parquet_engine`, which I can now access in test_io.py by importing `bentoml.io.PandasDataFrame`. That seems like a good solution to me, but let me know if you feel otherwise.

I think the concept of parquet engine is known with DataFrame, so I think the solution for having a `parquet_engine` under PandasDataFrame makes sense to me, and it also seems nice to have.

aarnphm:
I think there is a bug with our CI. otherwise everything else LGTM

aarnphm (approved):

aarnphm (approved):

Issue #2230, opened 2022-01-29
bentoml.pytorch.load_runner using cpu/gpu (ver 1.0.0a3)


jjerry-k:
**Describe the bug**
Hi, I'm Jerry!
Thank you for the quick response to the last report!
After that, I have another problem.

First of all, This is my model code.
```python
class Model(nn.Module):
    """
    Thi is example code.
    Customize your self.
    """
    def __init__(self, base_model_name="alexnet", num_classes=10, freeze=True):
        super(Model, self).__init__()

        assert base_model_name in available_models, f"Please Check available pretrained model list at https://rwightman.github.io/pytorch-image-models/results/"
        
        self.base_model_name = base_model_name
        self.num_classes = num_classes
        self.freeze = freeze
        
        self.model = timm.create_model(self.base_model_name, 
                                        pretrained=True, 
                                        num_classes=num_classes)
        
        if freeze: 
            for layer_name, ops in self.model.named_children():
                if layer_name != 'classifier':
                    ops.requires_grad = False
        
    def forward(self, x):
        x = self.model(x)
        return x

    def predict(self, x):
        self.eval()
        with torch.no_grad():
            raw_output = self(x)
            _, pred = torch.max(raw_output, 1)
            return pred
```

When I use bentoml.pytorch.load, `predict function in model` is working in every case.
![case_cpu](https://user-images.githubusercontent.com/25097024/151649478-4766f852-e525-4aa6-8301-3b5a417cb65c.png)
![case_gpu](https://user-images.githubusercontent.com/25097024/151649482-cd44db02-4378-4677-9dff-61a9580b58c6.png)

When I use bentoml.pytorch.load_runner with cuda, predict function is not working in runner.
Because of DataParallel model..
[code link](https://github.com/bentoml/BentoML/blob/1efd77c609b3fc2e153436c57db9b3707608c894/bentoml/_internal/frameworks/pytorch.py#L259)

This are example images
![runner_case_cpu](https://user-images.githubusercontent.com/25097024/151649822-689fce1f-58a7-4762-be17-307eb6cbe831.png)
![runner_case_gpu](https://user-images.githubusercontent.com/25097024/151649826-996862b2-ff9c-4adb-9ce7-629853703acf.png)

jjerry-k:
As I know, Data parallel model can't use custom method like predict!

aarnphm:
[...]

PR #2317, opened 2022-02-25
refactor(framework): pytorch splitting refactor

larme:
## Checklist:

### DIFF
--- a/.github/workflows/frameworks.yml
+++ b/.github/workflows/frameworks.yml
@@ -145,13 +145,20 @@ jobs:
             pytorch:
               - *runner
               - bentoml/pytorch.py
+              - bentoml/_internal/frameworks/common/pytorch.py
               - bentoml/_internal/frameworks/pytorch.py
               - tests/integration/frameworks/test_pytorch_impl.py
+            torchscript:
+              - *runner
+              - bentoml/torchscript.py
+              - bentoml/_internal/frameworks/common/pytorch.py
+              - bentoml/_internal/frameworks/torchscript.py
+              - tests/integration/frameworks/test_torchscript_impl.py
             pytorch_lightning:
               - *runner
               - bentoml/pytorch.py
               - bentoml/pytorch_lightning.py
-              - bentoml/_internal/frameworks/pytorch.py
+              - bentoml/_internal/frameworks/common/pytorch.py
               - bentoml/_internal/frameworks/pytorch_lightning.py
               - tests/integration/frameworks/test_pytorch_lightning_impl.py
             sklearn:
@@ -827,6 +834,50 @@ jobs:
           directory: ./
           files: ./pytorch.xml
           flags: pytorch
+          verbose: true
+
+  torchscript_integration_tests:
+    needs: diff
+    if: ${{ (github.event_name == 'pull_request' && needs.diff.outputs.torchscript == 'true') || github.event_name == 'push' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0 # fetch all tags and branches
+      - name: Setup python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+
+      - name: Get pip cache dir
+        id: cache-dir
+        run: |
+          echo ::set-output name=dir::$(pip cache dir)
+
+      - name: Cache pip dependencies
+        uses: actions/cache@v2
+        id: cache-pip
+        with:
+          path: ${{ steps.cache-dir.outputs.dir }}
+          key: ${{ runner.os }}-tests-${{ hashFiles('requirements/tests-requirements.txt') }}
+
+
+      - name: Install dependencies
+        run: |
+          pip install .
+          pip install -r requirements/tests-requirements.txt
+
+      - name: Run tests and generate coverage report
+        run: make tests-torchscript
+        shell: bash
+
+      - name: Upload test coverage to Codecov
+        if: ${{ needs.diff.outputs.torchscript == 'true' }}
+        uses: codecov/codecov-action@v2
+        with:
+          directory: ./
+          files: ./torchscript.xml
+          flags: torchscript
           verbose: true
 
   pytorch_lightning_integration_tests:
--- a/.github/workflows/frameworks.yml
+++ b/.github/workflows/frameworks.yml
@@ -40,6 +40,7 @@ jobs:
       statsmodels: ${{ steps.filter.outputs.statsmodels}}
       tf1: ${{ steps.filter.outputs.tf1 }}
       tf2: ${{ steps.filter.outputs.tf2 }}
+      torchscript: ${{ steps.filter.outputs.torchscript }}
       transformers: ${{ steps.filter.outputs.transformers }}
       xgboost: ${{ steps.filter.outputs.xgboost }}
     steps:
--- a/bentoml/_internal/frameworks/common/pytorch.py
+++ b/bentoml/_internal/frameworks/common/pytorch.py
@@ -19,7 +19,7 @@ from ...runner.container import DataContainer
 from ...runner.container import DataContainerRegistry
 
 if t.TYPE_CHECKING:
-    import pytorch_lightning as pl
+    import pytorch_lightning as pl  # disable=unused-import
 
 try:
     import numpy as np
--- a/bentoml/_internal/frameworks/common/pytorch.py
+++ b/bentoml/_internal/frameworks/common/pytorch.py
@@ -19,24 +19,24 @@ from ...runner.container import Payload
 from ...runner.container import DataContainer
 from ...runner.container import DataContainerRegistry
 
-if TYPE_CHECKING:
-    import pytorch_lightning as pl
-    from ... import external_typing as ext
-
 try:
     import numpy as np
     import torch
     import torch.nn.parallel as parallel
 except ImportError:  # pragma: no cover
     raise MissingDependencyException(
-        """
-        torch is required in order to use module `bentoml.pytorch`.
-         Instruction: Refers to https://pytorch.org/get-started/locally/
-         to setup PyTorch correctly.
-        """  # noqa
+        """\
+        torch is required in order to use module `bentoml.pytorch`,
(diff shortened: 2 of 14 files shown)


bentoml/_internal/frameworks/common/pytorch.py
```diff
@@ -0,0 +1,185 @@
...
+except ImportError:  # pragma: no cover
+    raise MissingDependencyException(
+        """
+        torch is required in order to use module `bentoml.pytorch`.
+         Instruction: Refers to https://pytorch.org/get-started/locally/
+         to setup PyTorch correctly.
+        """  # noqa
+    )
+
+_ModelType = t.Union["torch.nn.Module", "torch.jit.ScriptModule", "pl.LightningModule"]  # type: ignore[reportPrivateUsage]
```
aarnphm, line 37: Lets do this under `TYPE_CHECKING`
```suggestion
ModelType = t.Union[torch.nn.Module, torch.jit.ScriptModule, pl.LightningModule]
```

bentoml/_internal/frameworks/common/pytorch.py
```diff
@@ -0,0 +1,185 @@
...
+        name: t.Optional[str] = None,
+    ):
+        super().__init__(tag=tag, name=name)
+
+        self._predict_fn_name = predict_fn_name
+        self._partial_kwargs = partial_kwargs or dict()
+
+        self._predict_fn: t.Callable[..., torch.Tensor]
+        self._no_grad_context: t.Optional[contextlib.ExitStack] = None
+        self._model: t.Optional[_ModelType] = None
```
aarnphm, line 57:
```suggestion
        self._model: t.Optional[ModelType] = None
```

bentoml/_internal/frameworks/common/pytorch.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def _shutdown(self) -> None:
+        if self._no_grad_context is not None:
+            self._no_grad_context.close()
+            self._no_grad_context = None
+
+    def _run_batch(
+        self,
+        *args: t.Union["np.ndarray[t.Any, np.dtype[t.Any]]", torch.Tensor],
+        **kwargs: t.Union["np.ndarray[t.Any, np.dtype[t.Any]]", torch.Tensor],
```
aarnphm, line 127:
```suggestion
        *args: t.Union["ext.NpNDArray", torch.Tensor],
        **kwargs: t.Union["ext.NpNDArray", torch.Tensor],
```

bentoml/_internal/frameworks/common/pytorch.py
```diff
@@ -0,0 +1,185 @@
...
+    ) -> torch.Tensor:
+
+        params = Params[t.Union["np.ndarray[t.Any, np.dtype[t.Any]]", torch.Tensor]](
```
aarnphm, line 130:
```suggestion
        params = Params[t.Union["ext.NpNDArray", torch.Tensor]](
```

bentoml/_internal/frameworks/common/pytorch.py
```diff
@@ -0,0 +1,185 @@
...
+            *args, **kwargs
+        )
+
+        def _mapping(
+            item: t.Union["np.ndarray[t.Any, np.dtype[t.Any]]", torch.Tensor]
```
aarnphm, line 135:
```suggestion
            item: t.Union["ext.NpNDArray", torch.Tensor]
```

bentoml/_internal/frameworks/common/pytorch.py
```diff
@@ -0,0 +1,185 @@
...
+        ) -> torch.Tensor:
+            if isinstance(item, np.ndarray):
```
aarnphm, line 137:
```suggestion
            if LazyType["ext.NpNDArray"]("np.ndarray").isinstance(item):
```

bentoml/_internal/frameworks/pytorch.py
```diff
@@ -43,8 +33,6 @@
         """  # noqa
     )
```
aarnphm, line 34: I don't think we need this block anymore since you already do an import check under `.common.pytorch`

bentoml/_internal/frameworks/pytorch.py
```diff
@@ -122,7 +102,7 @@ def save(
     Args:
         name (:code:`str`):
             Name for given model instance. This should pass Python identifier check.
-        model (:code:`Union[torch.nn.Module, torch.jit.ScriptModule]`):
+        model (:code:`Uniontorch.nn.Module`):
```
aarnphm, line 105:
```suggestion
        model (:code:`torch.nn.Module`):
```

aarnphm (commented):
There are still some small and quick fix needed

aarnphm:
You also have to add flags for `torchscript` to `codecov.yml`

bentoml/_internal/frameworks/common/pytorch.py
```diff
@@ -0,0 +1,185 @@
...
+from ...utils.pkg import get_pkg_version
+from .model_runner import BaseModelRunner
+from ....exceptions import MissingDependencyException
+from ...runner.utils import Params
+from ...runner.container import Payload
+from ...runner.container import DataContainer
+from ...runner.container import DataContainerRegistry
+
+if t.TYPE_CHECKING:
+    import pytorch_lightning as pl  # disable=unused-import
```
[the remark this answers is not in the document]
larme: I wonder why we want to do that?

bentoml/_internal/frameworks/pytorch.py
larme: I'm thinking about just keeping these blocks in case one day we remove torch import from `.common.pytorch`. What's your opinion? @aarnphm

aarnphm: Hmm why would we do so? I don't think we would ever do such things right?

aarnphm: Screenshot 2022-02-26 at 17 43 02

aarnphm:
the PyTorchContainer is failing the test, and for some reason the torchscript integration test is not running ?

tests/integration/frameworks/test_pytorch_impl.py
```diff
@@ -7,87 +7,42 @@
...
+    tag = models()
+    bentoml_model = bentoml.models.get(tag)
+    assert_have_file_extension(bentoml_model.path, ".pt")
+    assert bentoml_model.info.context.get("model_format") == "torch.save:v1"
 
     pytorch_loaded: nn.Module = bentoml.pytorch.load(tag)
     assert predict_df(pytorch_loaded, test_df) == 5.0
 
 
 @pytest.mark.gpus
```
aarnphm, line 39: actually this mark will only run when we pass `--gpus` to pytest
```bash
pytest -vvv ... --gpus
```
so your cpu test would not be tested here

.github/workflows/frameworks.yml
```diff
@@ -145,13 +145,20 @@ jobs:
             pytorch:
               - *runner
               - bentoml/pytorch.py
+              - bentoml/_internal/frameworks/common/pytorch.py
               - bentoml/_internal/frameworks/pytorch.py
               - tests/integration/frameworks/test_pytorch_impl.py
+            torchscript:
```
aarnphm, line 151: needs to add
```
torchscript: ${{ steps.filter.outputs.torchscript }}
```
under `jobs.diff.outputs` as well

larme:
> the PyTorchContainer is failing the test, and for some reason the torchscript integration test is not running ?

Let's fix this after `batch_axis` is redone. Current fix is to disable testing for `batch_axis=1`

bentoml/_internal/frameworks/pytorch.py
larme: I made a change to import torch from .common.pytorch directly @aarnphm

bentoml/_internal/frameworks/common/pytorch.py
```diff
@@ -0,0 +1,183 @@
...
+        *args: t.Union["ext.NpNDArray", torch.Tensor],
+        **kwargs: t.Union["ext.NpNDArray", torch.Tensor],
+    ) -> torch.Tensor:
+
+        params = Params[t.Union["ext.NpNDArray", torch.Tensor]](*args, **kwargs)
+
+        def _mapping(item: t.Union["ext.NpNDArray", torch.Tensor]) -> torch.Tensor:
+            if LazyType["ext.NpNDArray"]("numpy.ndarray").isinstance(item):
+                item = torch.Tensor(item, device=self._device_id)
+            return item
```
bojiang, line 137: If it's already a tensor, we need to call `to(device)` on it as well, right?

aarnphm: init with `self._device_id` so no need to call `to(device)`

aarnphm (approved):

bentoml/_internal/frameworks/common/pytorch.py
larme: @aarnphm @bojiang  I think you mean in `self._configure()` we have set default tensor storage type? But if the tensor is deserialized from DataContainer payload which is serialized in another machine and these 2 machine have mismatched device, then the tensor won't be converted to the proper device automatically. Also in my simple benchmark `to(device)` won't cost a lot if `from_device` and `to_device` is the same

Here's my simple demo:

https://gist.github.com/larme/7640f2d97d45c92946f1573b3eaaa0be

aarnphm: I think calling `to` shouldn't affect perf that much. I think we just need to make sure that device is consistent across, so that https://github.com/bentoml/BentoML/issues/2230 won't happen.

aarnphm:
paddlepaddle test will be addressed in a separate PR, seems like it just randomly fails all the time when using paddlehub module.

bentoml/_internal/frameworks/common/pytorch.py
larme: I think this change will make device more consistent?

aarnphm: sounds good then

bojiang (approved):

PR #2318, opened 2022-02-27
internal: usage tracking

aarnphm:

## Description

Implements usage tracking. Refactor some CLI to make sure tracking works with 1.0 and respect `BENTOML_DO_NOT_TRACK`

## Motivation and Context

check under `_internal/utils/analytics`

## How Has This Been Tested?

TODO:

## Checklist:


bentoml/_internal/server/__init__.py
```diff
@@ -107,12 +146,35 @@ def serve_development(
...
+    try:
+        tracking_thread.start()
+        arbiter.start(
+            cb=lambda _: logger.info(  # type: ignore
+                f'Starting development BentoServer from "{bento_identifier}" '
+                f"running on http://{host}:{port} (Press CTRL+C to quit)"
+            ),
+        )
+    finally:
+        stop_thread_event.set()
```
bojiang, line 168: There're two things here:
* we told the thread to stop but don't wait for it. suggestion: use thread.join() after this line.
* too many lines here in the caller. I think we may shield this logic in a custom context manager, like this 
```
with schedule_track(**options).start():
  arbiter.start(...)
```

bentoml/_internal/utils/analytics/usage_stats.py
```diff
@@ -0,0 +1,216 @@
...
+        event_type=event_type, event_pid=event_pid, event_properties=event_properties
+    )
+
+    loop = asyncio.get_running_loop()
+    with ThreadPoolExecutor(max_workers=1) as executor:
+        await loop.run_in_executor(executor, send_usage_event, payload, timeout, uri)
+
+
+@slient
+async def scheduled_track_async(
```
bojiang, line 197: This is not used, right?

bentoml/_internal/utils/analytics/usage_stats.py
```diff
@@ -0,0 +1,216 @@
...
+                BENTO_SERVE_SCHEDULED_TRACK_EVENT_TYPE,
+                event_pid,
+                event_properties=event_properties,
+            )
+
+    t = threading.Thread(target=loop, daemon=True)
+    return t, stop_event
+
+
+async def track_async(
```
bojiang, line 175: This is not used, right?

bentoml/_internal/utils/analytics/usage_stats.py
```diff
@@ -0,0 +1,216 @@
...
+
+@slient
+def scheduled_track(
+    event_pid: int,
+    event_properties: t.Optional[t.Dict[str, t.Any]] = None,
+    interval: int = get_usage_stats_interval_seconds(),
+) -> t.Tuple[threading.Thread, threading.Event]:
+    stop_event = threading.Event()
+
+    def loop() -> None:
```
bojiang, line 163: An endless loop returns typing.NoReturn

bentoml/_internal/utils/analytics/usage_stats.py
aarnphm: ohh i forgot to remove

bentoml/_internal/server/__init__.py
aarnphm: - thread cannot join since this is not a poolexecutor.
- thread is not a context manager since it doesn't have `__enter__`, so cannot start with a `with` block.

README.md
```diff
@@ -55,21 +56,26 @@ There are many ways to contribute to the project:
 
 ### Usage Reporting ###
 
-BentoML by default collects anonymous usage data using [Amplitude](https://amplitude.com/). 
-It only collects BentoML library's own actions and parameters, no user or model data will be collected. 
-Here is the [code](https://github.com/bentoml/BentoML/blob/main/bentoml/_internal/utils/usage_stats.py) that does it.
+By default, BentoML anonymously collect usage data, and here's the [code](./bentoml/_internal/utils/analytics/usage_stats.py) for it. We rely heavily of
```
parano, line 59: BentoML collects anonymous usage data that helps our team to improve the product. Only BentoML's internal API calls and CLI commands are being reported and we striped out any potentially sensitive information as much as possible, we will never collect user code, model data, model name or stacktrace. Here's the [code] for usage tracking.

You can opt-out of usage tracking by the `--do-not-track` CLI option:
```bash
> bentoml [command] --do-not-track
```

Or by setting environment variable `BENTOML_DO_NOT_TRACK=True`:
```bash
> export BENTOML_DO_NOT_TRACK=True
```

aarnphm: done

README.md
```diff
@@ -55,21 +56,24 @@ There are many ways to contribute to the project:
 
 ### Usage Reporting ###
 
-BentoML by default collects anonymous usage data using [Amplitude](https://amplitude.com/). 
-It only collects BentoML library's own actions and parameters, no user or model data will be collected. 
-Here is the [code](https://github.com/bentoml/BentoML/blob/main/bentoml/_internal/utils/usage_stats.py) that does it.
+ We rely heavily of
```
parano, line 59:
```suggestion
```

README.md
```diff
@@ -55,21 +56,23 @@ There are many ways to contribute to the project:
...
-    > bentoml [command] --do-not-track
-
-You can also opt-out via setting environment variable `BENTOML_DO_NOT_TRACK=True`
-
-    > export BENTOML_DO_NOT_TRACK=True
 
 ### License ###
 
-[Apache License 2.0](https://github.com/bentoml/BentoML/blob/main/LICENSE)
+[Apache License 2.0](./LICENSE)
```
parano, line 76: can we change this back, this doesn't work on the pypi page

bentoml/_internal/bento/bento.py
```diff
@@ -334,6 +340,21 @@ def save(
             fs.mirror.mirror(self._fs, out_fs, copy_if_newer=False)
             self._fs.close()
             self.__fs = out_fs
+            event_properties = {
+                "bento_tag": str(self.info.tag),
+                "bento_creation_timestamp": self.info.creation_time.isoformat(),
+                "bento_size": human_readable_size(calc_dir_size(bento_path)),
```
parano, line 346: no need for `human_readable_size` here, it will make ETL and analytic harder

parano: I'd recommend making it `bento_size_in_bytes` with an integer type

bentoml/_internal/bento/bento.py
```diff
@@ -334,6 +340,21 @@ def save(
...
+                "models_tags": [str(i) for i in self.info.models],
+                "models_type": [
+                    model_store.get(i).info.module for i in self.info.models
+                ],
+            }
+
+        track(
+            BENTO_BUILD_TRACK_EVENT_TYPE,
+            event_pid=os.getpid(),
```
parano, line 355: why do we want to pid here?

bentoml/_internal/cli/__init__.py
```diff
@@ -1,18 +1,32 @@
...
 from .containerize import add_containerize_command
+from ..configuration import is_pypi_installed_bentoml
 from .bento_management import add_bento_management_commands
 from .model_management import add_model_management_commands
 
 
+def _rich_callback():
+    import sys
+
+    if sys.stdout.isatty() and not is_pypi_installed_bentoml():
```
parano, line 17: does `sys.stdout.isatty` return True or False in a Jupyter notebook environment?

And why does this has to do with `is_pypi_installed_bentoml`?  I think regardless we should show a nice stacktrace when user is in an interactive mode, either a jupyter notebook or python session.

bentoml/_internal/cli/bento_server.py
```diff
@@ -13,7 +14,7 @@
 
 
 def add_serve_command(cli: click.Group) -> None:
-    @cli.command()
+    @cli.command(help="Serve a Bento")
```
parano, line 17: I think with `click`, this actually will override the docstring below in the `--help` message, is that correct?  if so, let's remove it

bentoml/_internal/cli/click_utils.py
```diff
@@ -1,66 +1,55 @@
...
 import functools
+from typing import TYPE_CHECKING
 
 import click
 from click import ClickException
 
 from ...exceptions import BentoMLException
 
-# from bentoml import configure_logging
+# from ..log import configure_logging
```
parano, line 14: remove?

bentoml/_internal/utils/analytics/usage_stats.py
```diff
@@ -0,0 +1,170 @@
...
+    # Returns True if and only if the environment variable is defined and has value True.
+    # The function is cached for better performance.
+    return os.environ.get(BENTOML_DO_NOT_TRACK, str(False)).lower() == "true"
+
+
+@lru_cache(maxsize=1)
+def get_usage_stats_interval_seconds() -> int:
+    # The interval for getting usage stats for a serve is 12 hours.
+    # cached for better performance.
+    return int(os.environ.get(BENTOML_USAGE_REPORT_INTERVAL_SECONDS, 12 * 60 * 60))
```
parano, line 60: why do we need this env var?

bentoml/_internal/utils/analytics/usage_stats.py
```diff
@@ -0,0 +1,170 @@
...
+) -> t.Dict[str, t.Any]:
+    """
+    event_type(str): type of event (model_save, bento_build, etc.)
+    event_pid(int): retrieved with os.getpid() for the given event
+    event_properties(dict(str, Any)): custom properties of a given event
+    """
+    if event_properties is None:
+        event_properties = {}
+
+    # TODO: add yatai_user email and yatai_version when available
```
parano, line 115: I believe this is available now, could you take a look and discuss with @yubozhao ?

bentoml/_internal/utils/analytics/usage_stats.py
```diff
@@ -0,0 +1,170 @@
...
+        return client_id
+    return {}
+
+
+def get_hardware_usage(pid: int) -> t.Dict[str, t.Any]:  # pragma: no cover
+    proc = psutil.Process(pid)
+    with proc.oneshot():
+        return {
+            "memory_usage_percent": proc.memory_percent(),
+            "total_memory(MB)": psutil.virtual_memory().total / 1024**2,
```
parano, line 95: suggest rename: `total_memory_in_mb`

bentoml/_internal/utils/analytics/usage_stats.py
```diff
@@ -0,0 +1,170 @@
...
+            "num_threads": proc.num_threads(),
+        }
+
+
+@inject
+def get_payload(
+    event_type: str,
+    event_pid: int,
```
parano, line 103: why do we need to pass pid here? can it get it within this function?

bentoml/_internal/bento/bento.py
```diff
@@ -334,6 +340,21 @@ def save(
             fs.mirror.mirror(self._fs, out_fs, copy_if_newer=False)
             self._fs.close()
             self.__fs = out_fs
+            event_properties = {
```
parano, line 343: should we use an attrs class to represent each event type's properties?  we already have cattrs for JSON serialization, and it provides type checking and validation for us

bentoml/_internal/server/__init__.py
```diff
@@ -107,12 +146,35 @@ def serve_development(
...
+
+    atexit.register(
```
parano, line 170: move this to before the `arbiter.start` at l158

bentoml/_internal/utils/analytics/usage_stats.py
```diff
@@ -0,0 +1,170 @@
...
+        event_type=event_type, event_pid=event_pid, event_properties=event_properties
+    )
+    send_usage_event(payload, timeout=timeout, uri=uri)
+
+
+@slient
+def scheduled_track(
+    event_pid: int,
+    event_properties: t.Optional[t.Dict[str, t.Any]] = None,
+    interval: int = get_usage_stats_interval_seconds(),
```
parano, line 157: let's just fix this to BENTOML_USAGE_REPORT_INTERVAL_SECONDS, I don't think we should allow user to customize this. Fixing it also simplifies our analytic query.

bentoml/_internal/cli/__init__.py
aarnphm: I don't think we want to show traceback in non-development mode. This is more for us when developing CLI, Users can still see all the traceback from their end of the code.

`sys.stdout.isatty` return False in a jupyter notebook environment.

bentoml/_internal/cli/bento_server.py
aarnphm: will do

bentoml/_internal/cli/click_utils.py
aarnphm: --quiet actually doesn't work here just yet.

bentoml/_internal/utils/analytics/usage_stats.py
```diff
@@ -0,0 +1,170 @@
...
+if TYPE_CHECKING:
+    P = t.ParamSpec("P")
+    T = t.TypeVar("T")
+    AsyncFunc = t.Callable[P, t.Coroutine[t.Any, t.Any, t.Any]]
+
+logger = logging.getLogger(__name__)
+
+BENTOML_DO_NOT_TRACK = "BENTOML_DO_NOT_TRACK"
+BENTOML_TRACKING_URL = "https://t.bentoml.com"
+BENTOML_USAGE_REPORT_INTERVAL_SECONDS = "__BENTOML_USAGE_REPORT_INTERVAL_SECONDS"
```
parano, line 33: instead of using env var, you can always monkey patching a constant in a module

bentoml/_internal/utils/analytics/usage_stats.py
aarnphm: imo we should allow users to configure a period that they want to send the usage tracking if they want to do so

bentoml/_internal/utils/analytics/usage_stats.py
parano: see you want to set it in testing, made a suggestion above to patch the constant variable in testing instead

bentoml/_internal/utils/ngrok.py
```diff
@@ -44,10 +43,10 @@ def _print_url():
             if data["tunnels"]:
                 tunnel = data["tunnels"][0]
                 logger.info(
-                    " * Ngrok running at: %s",
+                    "Ngrok running at: %s",
```
parano, line 46: why remove `* ` here? it is meant to be consistent with other Uvicorn logs I believe.

bentoml/_internal/utils/ngrok.py
```diff
@@ -44,10 +43,10 @@ def _print_url():
...
                     tunnel["public_url"].replace("https://", "http://"),
                 )
-                logger.info(" * Traffic stats available on http://127.0.0.1:4040")
+                logger.info("Traffic stats available on http://127.0.0.1:4040")
```
parano, line 49: same here

bentoml/_internal/utils/ngrok.py
```diff
@@ -82,16 +81,21 @@ def _download_ngrok(ngrok_path):
...
     else:
         raise Exception(f"{system} is not supported")
-    download_path = _download_file(url)
+    download_path = download_file(url)
     with zipfile.ZipFile(download_path, "r") as zip_ref:
         zip_ref.extractall(ngrok_path)
 
 
-def _download_file(url):
+def download_file(url: str) -> str:
```
parano, line 89: why renaming these functions? I think they are clearly only internal to this module and not meant to be used by other BentoML modules, l'd suggest we keep the leading `_` here

docs/source/index.rst
```diff
@@ -62,22 +62,23 @@ There are many ways to contribute to the project:
 Usage Reporting
 ---------------
 
-BentoML by default collects anonymous usage data using `Amplitude <https://amplitude.com/>`_. 
-It only collects BentoML library's own actions and parameters, no user or model data will be collected. 
-Here is the `code <https://github.com/bentoml/BentoML/blob/main/bentoml/utils/usage_stats.py>`_ that does it.
+By default, BentoML anonymously collect usage data, and here's the `code <https://github.com/bentoml/BentoML/blob/main/bentoml/_internal/utils/analytics/usage_stats.py>`_ for it. We rely heavily of
```
parano, line 65: let's make this consistent with readme.md

bentoml/_internal/utils/analytics/usage_stats.py
aarnphm: I think we can, let me update that

bentoml/_internal/server/__init__.py
```diff
@@ -131,6 +193,33 @@ def serve_production(
...
+    else:
+        # In this case serving from a file/directory with --production
+        event_properties = {
+            "production": True,
+            "bento_tag": bento_identifier,
+            **serve_info,
+        }
+
+    track(
+        BENTO_SERVE_TRACK_EVENT_TYPE,
```
parano, line 219: to make it easier for analytics, can we separate serve_production and serve_development to two different event types?

bentoml/_internal/utils/analytics/usage_stats.py
aarnphm: Idk what to use between `getpid` and getppid`?

bentoml/_internal/bento/bento.py
aarnphm: We can

bentoml/_internal/bento/bento.py
```diff
@@ -334,6 +340,21 @@ def save(
...
+                "bento_tag": str(self.info.tag),
+                "bento_creation_timestamp": self.info.creation_time.isoformat(),
+                "bento_size": human_readable_size(calc_dir_size(bento_path)),
+                "models_tags": [str(i) for i in self.info.models],
```
parano, line 347: model_tags, model_types?

bentoml/_internal/cli/__init__.py
```diff
@@ -1,18 +1,32 @@
...
+        from rich.traceback import install
+
+        install(suppress=[click])
+
+
 def create_bentoml_cli():
-    @click.group(cls=BentoMLCommandGroup)
-    @click.version_option(version=__version__)
+    _rich_callback()
+    CONTEXT_SETTINGS = {"help_option_names": ("-h", "--help")}
```
parano, line 26: isn't this the default?

bentoml/_internal/bento/bento.py
aarnphm: done

bentoml/_internal/bento/bento.py
aarnphm: pid is a common field that we want to track?

bentoml/_internal/utils/analytics/usage_stats.py
aarnphm: done

bentoml/_internal/server/__init__.py
aarnphm: done

bentoml/_internal/utils/ngrok.py
aarnphm: I think that is the format before, where we don't use uvicorn logs

bentoml/_internal/utils/ngrok.py
aarnphm: will do

bentoml/_internal/cli/__init__.py
aarnphm: nope

bentoml/_internal/cli/__init__.py
```diff
@@ -11,8 +11,14 @@
 
 
 def create_bentoml_cli():
-    @click.group(cls=BentoMLCommandGroup)
-    @click.version_option(version=__version__)
+    # add traceback when in interactive shell for development
```
parano, line 14:
```suggestion
    # exclude traceback from the click module
```

bentoml/_internal/cli/bento_management.py
```diff
@@ -191,7 +195,7 @@ def delete_target(target):
         default="",
         required=False,
     )
-    def export(bento_tag, out_path):
+    def export(bento_tag: str, out_file: "BytesIO") -> None:
```
parano, line 198: why changing the type here? it should be out_path: str

bentoml/_internal/server/__init__.py
aarnphm: we actually have a `bento_serve_scheduled_track` and `bento_serve_init_track`

bentoml/_internal/utils/analytics/usage_stats.py
aarnphm: noted

bentoml/_internal/bento/bento.py
aarnphm: we want to track the model tag and types of the model in each bentos?

bentoml/_internal/cli/click_utils.py
```diff
@@ -110,65 +103,85 @@ def wrapper(quiet, verbose, config, *args, **kwargs):
...
-        def wrapper(func):
-            # add common parameters to command
+        def wrapper(func: "t.Callable[P, t.Any]") -> click.Command:
+            # add common parameters to command.
             func = BentoMLCommandGroup.bentoml_common_params(func)
             # Send tracking events before command finish.
             func = BentoMLCommandGroup.bentoml_track_usage(func, self, **kwargs)
-            # If BentoMLException raise ClickException instead before exit
+            # If BentoMLException raise ClickException instead before exit.
+            # NOTE: Always call this function last for type checker to work.
```
parano, line 184: why is that?

bentoml/_internal/cli/click_utils.py
```diff
@@ -181,18 +194,9 @@ def wrapper(func):
         return wrapper
 
 
-def _is_valid_bento_tag(value):
+def is_valid_bento_tag(value: str) -> bool:
```
parano, line 197: should we just use the Tag class's validation logic here?

aarnphm: yep we should

bentoml/_internal/cli/click_utils.py
aarnphm: because of type casting

aarnphm: click dynamically inject `__click_params__` into a function, so there aren't any class we can use for type hint, thus the wrapper on L22

bentoml/_internal/server/__init__.py
bojiang: @aarnphm I mean, to write your custom context manager, making use of try...except and event

bojiang:
```python
# utils/tracking.py

@contextlib.contextmanager
def server_tracking(...):
    tracking_thread, stop_thread_event = scheduled_track(
        current_pid,
        event_properties=get_scheduled_event_properties(
            production=False,
            bento_identifier=bento_identifier,
            serve_info=serve_info,
        ),
    )
    try:
        tracking_thread.start()
        yield
    finally:
        stop_thread_event.set()
        tracking_thread.join()

# caller
with server_tracking():
        arbiter.start(
            cb=lambda _: logger.info(  # type: ignore
                f'Starting development BentoServer from "{bento_identifier}" '
                f"running on http://{host}:{port} (Press CTRL+C to quit)"
            ),
        )
```

bojiang: @aarnphm https://docs.python.org/3/library/threading.html#threading.Thread.join

bentoml/_internal/bento/bento.py
```diff
@@ -321,19 +324,32 @@ def creation_time(self) -> datetime:
...
         with bento_store.register(self.tag) as bento_path:
             out_fs = fs.open_fs(bento_path, create=True, writeable=True)
             fs.mirror.mirror(self._fs, out_fs, copy_if_newer=False)
             self._fs.close()
             self.__fs = out_fs
+            track(
+                event_properties=BentoBuildEvent(
+                    bento_tag=self.info.tag,
+                    bento_creation_timestamp=self.info.creation_time,
+                    bento_size_in_kb=calc_dir_size(bento_path),
```
parano, line 346: add a `models_size_in_kb` here? with the dir size of `os.path.join(bento_path, "models")`

bentoml/_internal/bento/bento.py
```diff
@@ -321,19 +324,32 @@ def creation_time(self) -> datetime:
...
+            raise BentoMLException("Failed to save Bento because it was invalid.")
 
         with bento_store.register(self.tag) as bento_path:
             out_fs = fs.open_fs(bento_path, create=True, writeable=True)
             fs.mirror.mirror(self._fs, out_fs, copy_if_newer=False)
             self._fs.close()
             self.__fs = out_fs
+            track(
+                event_properties=BentoBuildEvent(
+                    bento_tag=self.info.tag,
```
parano, line 344: remove bento_tag here? it might be sensitive for some users

bentoml/_internal/bento/bento.py
```diff
@@ -321,19 +324,32 @@ def creation_time(self) -> datetime:
...
+                    bento_creation_timestamp=self.info.creation_time,
+                    bento_size_in_kb=calc_dir_size(bento_path),
+                    model_tags=self.info.models,
```
parano, line 347: change to `num_of_models`?

parano: similarly, tags might be sensitive in some cases

bentoml/_internal/bento/bento.py
aarnphm: ok

bentoml/_internal/bento/bento.py
aarnphm: I don't think model_tags would have anything sensitive?

bentoml/_internal/bento/bento.py
aarnphm: done

bentoml/_internal/server/__init__.py
```diff
@@ -131,6 +191,30 @@ def serve_production(
...
+            model_types=[
+                bento._model_store.get(i).info.module for i in bento.info.models
+            ],
+        )
+    else:
+        # In this case serving from a file/directory with --production
+        event_properties = BentoServeProductionOnStartupEvent(
+            serve_id=serve_info.serve_id,
+            serve_creation_timestamp=serve_info.serve_creation_timestamp,
+            bento_identifier=bento_identifier,
```
parano, line 209: remove `bento_identifier` here and above - it contains user file name, or bento name

bentoml/_internal/server/__init__.py
```diff
@@ -131,6 +191,30 @@ def serve_production(
...
     working_dir = os.path.realpath(os.path.expanduser(working_dir))
     svc = load(bento_identifier, working_dir=working_dir)
+    serve_info = get_serve_info()
+
+    # Track first time
+    if svc.bento is not None:
+        bento = svc.bento
+        event_properties = BentoServeProductionOnStartupEvent(
+            serve_id=serve_info.serve_id,
+            serve_creation_timestamp=serve_info.serve_creation_timestamp,
```
parano, line 196: serve_creation_timestamp => serve_start_timestamp ?

bentoml/_internal/server/__init__.py
```diff
@@ -131,6 +191,30 @@ def serve_production(
 ) -> None:
     working_dir = os.path.realpath(os.path.expanduser(working_dir))
     svc = load(bento_identifier, working_dir=working_dir)
+    serve_info = get_serve_info()
```
parano, line 189: move L194-L217 to a help function `track_serve(svc)`

bentoml/_internal/server/__init__.py
```diff
@@ -1,21 +1,33 @@
...
-from bentoml._internal.utils.uri import path_to_uri
-from bentoml._internal.utils.circus import create_standalone_arbiter
 
+from ..utils import reserve_free_port
+from ..utils.uri import path_to_uri
+from ..utils.circus import create_standalone_arbiter
+from ..utils.analytics import track
+from ..utils.analytics import get_serve_info
+from ..utils.analytics import scheduled_track
+from ..utils.analytics import BentoServeProductionOnStartupEvent
```
parano, line 25: Suggest renaming events:
```
ServeStartEvent
ServeUpdateEvent
ServeEndEvent
ServeDevStartEvent
ServeDevUpdateEvent
ServeDevEndEvent
```

bentoml/_internal/server/__init__.py
```diff
@@ -107,14 +148,33 @@ def serve_development(
...
-            f"running on http://{host}:{port} (Press CTRL+C to quit)"
-        ),
-    )
+    try:
+        tracking_thread.start()
+        atexit.register(
+            track,
+            event_properties=BentoServeDevelopmentOnShutdownEvent(
+                serve_id=serve_info.serve_id,
+                bento_identifier=bento_identifier,
```
parano, line 163: remove bento_identifier

bentoml/_internal/server/__init__.py
aarnphm: done

bentoml/_internal/server/__init__.py
aarnphm: serve_started_timestamp

bentoml/_internal/utils/analytics/schemas.py
```diff
@@ -0,0 +1,211 @@
...
+    else:
+        return "local"
+
+
+@attrs.define
+class BentoServeDevelopmentOnStartupEvent(EventMeta):
+    serve_id: str
+    serve_creation_timestamp: datetime
+    bento_identifier: str
+    serve_location: str = attrs.field(init=False)
```
parano, line 145: what is this for?  I think it is very clear that when all the `bento` fields are empty, it is served from bento, otherwise from a directory

bentoml/_internal/utils/analytics/schemas.py
```diff
@@ -0,0 +1,211 @@
...
+@attrs.define
+class CliEvent(EventMeta):
+
+    command_group: str
+    command_name: str
+    error_type: str = ""
+    error_message: str = ""
+
+    return_code: int = 0
+    duration: float = 0
```
parano, line 104: change to `duration_in_seconds` or `duration_in_ms`?

bentoml/_internal/utils/analytics/schemas.py
```diff
@@ -0,0 +1,211 @@
...
+            self.memory_usage_percent = proc.memory_percent()
+            self.total_memory_in_kb = convert_to_kb(psutil.virtual_memory().total)  # type: ignore
+            self.num_threads = proc.num_threads()
+
+
+class EventMeta(ABC):
+    @property
+    def _track_event_name(self):
+        return "_".join(
+            map(str.lower, re.findall(r"[A-Z][^A-Z]*", self.__class__.__name__))
```
parano, line 91: also remove the trailing "event"?   so ModelSaveEvent will have name `model_save`?

bentoml/_internal/utils/analytics/schemas.py
```diff
@@ -0,0 +1,211 @@
...
+    model_size_in_kb: float = attrs.field(converter=convert_to_kb)
+
+
+@attrs.define
+class BentoBuildEvent(EventMeta):
+    bento_tag: Tag
+    bento_creation_timestamp: datetime
+    bento_size_in_kb: float = attrs.field(converter=convert_to_kb)
+
+    model_tags: t.List[Tag] = attrs.field(
```
parano: change to num_of_models, also add `num_of_runners` here

bentoml/_internal/utils/analytics/README.md
```diff
@@ -0,0 +1,27 @@
+## BentoML Analytics Tracking
```
parano, line 1: remove this file

bentoml/_internal/server/__init__.py
aarnphm: I think `ServeProductionStartEvent` would be more explicit?

bentoml/_internal/utils/analytics/schemas.py
aarnphm: noted

bentoml/_internal/bento/bento.py
parano: can we track this in the CLI part?  this will miss all the cases where build failures. e.g. if `Bento.create` failed, it doesn't hit the `Bento#save` method at all.  It's also nice to know the duration of this CLI command run.

bentoml/_internal/utils/analytics/schemas.py
aarnphm: duration_in_ms

bentoml/_internal/bento/bento.py
```diff
@@ -321,19 +324,32 @@ def creation_time(self) -> datetime:
...
             logger.warning(f"Failed to create Bento for {self.tag}, not saving.")
-            raise BentoMLException("Failed to save Bento because it was invalid")
+            raise BentoMLException("Failed to save Bento because it was invalid.")
 
         with bento_store.register(self.tag) as bento_path:
             out_fs = fs.open_fs(bento_path, create=True, writeable=True)
             fs.mirror.mirror(self._fs, out_fs, copy_if_newer=False)
             self._fs.close()
             self.__fs = out_fs
+            track(
```
parano, line 342: let's move this to the CLI code, where we can track when build failed

bentoml/_internal/utils/analytics/schemas.py
aarnphm: done

bentoml/_internal/cli/bento_management.py
```diff
@@ -262,8 +268,20 @@ def push(bento_tag: str, force: bool):
...
 
-        build_bentofile(bentofile, build_ctx=build_ctx, version=version)
+        start = time.time_ns()
+        bento = build_bentofile(bentofile, build_ctx=build_ctx, version=version)
+        duration = time.time_ns() - start
+
+        event_properties = ctx.custom_event_mapping["build"](bento=bento)
+        injected_payload = {
+            "event_properties": {
+                "duration_in_ms": duration / 1e6,
```
parano, line 283:
```suggestion
                "duration_in_ms": int(duration / 1e6),
```

bentoml/_internal/bento/bento.py
```diff
@@ -166,6 +166,7 @@ def create(
         ctx_fs = fs.open_fs(build_ctx)
 
         model_tags = build_config.additional_models
+        runners_info = {k: type(runners).__name__ for k, runners in svc.runners.items()}
```
parano, line 169:
```suggestion
```

bentoml/_internal/utils/analytics/usage_stats.py
```diff
@@ -0,0 +1,137 @@
...
+) -> t.Optional[requests.Response]:
+    if do_not_track():
+        return
+    payload = get_payload(
+        event_properties=event_properties, injected_payload=injected_payload
+    )
+    return requests.post(uri, json=payload, timeout=timeout)
+
+
+@slient
```
parano, line 119: this will silent `yield` as well?

parano: remove it here

setup.cfg
```diff
@@ -19,6 +19,7 @@ classifiers =
     Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
     Programming Language :: Python :: 3.9
+    Programming Language :: Python :: 3.10
```
parano, line 22: remove, not in this PR

setup.cfg
```diff
@@ -103,20 +104,20 @@ formats = gztar
...
     bentoml/testing/*
     bentoml/__main__.py
-    bentoml/build_utils.py
     bentoml/io.py
 
 [coverage:report]
 show_missing = true
 precision = 2
 omit =
+    bentoml/_internal/utils/analytics/get_event_tracking.py
```
parano, line 117: no need to exclude it here

setup.cfg
parano, line 22:
```suggestion
```

tests/unit/_internal/utils/test_analytics.py
```diff
@@ -0,0 +1,94 @@
...
+            "yatai_user_email": Or(str, None),
+        },
+        "event_properties": {
+            "module": str,
+            "model_tag": str,
+            "model_creation_timestamp": str,
+            "model_size_in_kb": Or(float, int),
+        },
+        "session_id": str,
+        "event_type": And(str, _is_lower),
```
parano, line 39:
```suggestion
        "event_type": And(str, str.islower),
```

tests/unit/_internal/utils/test_analytics.py
```diff
@@ -0,0 +1,94 @@
...
+from schema import And
+from schema import Schema
+
+import bentoml._internal.utils.analytics as analytics_lib
+from bentoml._internal.types import Tag
+
+if TYPE_CHECKING:
+    from _pytest.monkeypatch import MonkeyPatch
+
+_is_lower: t.Callable[[str], bool] = lambda s: s.islower()
```
parano, line 16:
```suggestion
```

bentoml/_internal/server/__init__.py
```diff
@@ -107,14 +179,26 @@ def serve_development(
...
+    ensure_prometheus_dir()
+
+    with scheduled_track(event_properties):
+        atexit.register(
+            track,
+            event_properties=ServeDevEndEvent(
+                serve_id=serve_info.serve_id,
+                triggered_at=datetime.now(timezone.utc),
+            ),
+        )
```
parano, line 195: let's put the End event also in the `scheduled_track` call to make it cleaner?

bentoml/_internal/utils/analytics/usage_stats.py
```diff
@@ -0,0 +1,137 @@
...
+@contextlib.contextmanager
+def scheduled_track(
+    event_properties: EventMeta,
+    interval: int = BENTOML_USAGE_REPORT_INTERVAL_SECONDS,
```
parano, line 123: remove it from parameter, just use the constant below

bentoml/_internal/server/__init__.py
```diff
@@ -107,14 +179,26 @@ def serve_development(
...
-    _ensure_prometheus_dir()
-
-    arbiter.start(
-        cb=lambda _: logger.info(  # type: ignore
-            f'Starting development BentoServer from "{bento_identifier}" '
-            f"running on http://{host}:{port} (Press CTRL+C to quit)"
-        ),
+    event_properties = ServeDevUpdateEvent(
+        serve_id=serve_info.serve_id,
+        triggered_at=datetime.now(timezone.utc),
```
parano, line 184: all update event will have the same `triggered_at`?

bentoml/_internal/cli/yatai.py
```diff
@@ -35,6 +37,7 @@ def login(endpoint: str, api_token: str) -> None:
         if user is None:
             raise CLIException("current user is not found")
 
+        user = yatai_rest_client.get_current_user()
```
parano, line 40: what is this for?

bentoml/_internal/bento/bento.py
parano: we don't want to include the model tag

bentoml/_internal/cli/bento_management.py
```diff
@@ -262,8 +263,8 @@ def push(bento_tag: str, force: bool):
     @click.argument("build_ctx", type=click.Path(), default=".")
     @click.option("-f", "--bentofile", type=click.STRING, default="bentofile.yaml")
     @click.option("--version", type=click.STRING, default=None)
-    def build(build_ctx, bentofile, version):
+    def build(build_ctx: str, bentofile: str, version: str) -> None:
         if sys.path[0] != build_ctx:
             sys.path.insert(0, build_ctx)
 
-        build_bentofile(bentofile, build_ctx=build_ctx, version=version)
+        return build_bentofile(bentofile, build_ctx=build_ctx, version=version)
```
timliubentoml, line 270: just a nitpick, maybe this is a python convention, but is the "return" intended on a method which returns -> None?

bentoml/_internal/utils/analytics/schemas.py
```diff
@@ -0,0 +1,223 @@
...
+
+    if os.path.exists(CLIENT_ID_PATH):
+        with open(CLIENT_ID_PATH, "r", encoding="utf-8") as f:
+            client_info = yaml.safe_load(f)
+        return ClientInfo(**client_info)
+    else:
+
+        def create_client_id() -> t.Dict[str, str]:
+            # returns an unique client_id and timestamp in ISO format
+            uniq = uuid.uuid1().bytes
```
timliubentoml, line 63: maybe if this wasn't a deliberate decision uuid4 might be more appropriate? for a pure random uuid

bentoml/_internal/cli/bento_management.py
aarnphm: Usually, if a function returns None, we shouldn't return. In python this is acceptable:
```python
def func():
    test = "hello"
    return test

func()
```
However, this is a CLI command, users wouldn't be able to access the return value. 

`return` here is a bit of a hack. Since this is a click command, we are wrapping every click function through `BentoMLCommandGroup` under `click_utils.py`. You can check under `track_usage` commands is that we actually use this return value for tracking.

bentoml/_internal/utils/analytics/schemas.py
```diff
@@ -0,0 +1,223 @@
...
+            client_id = hmac.new(uniq, digestmod=hashlib.blake2s).hexdigest()
```
timliubentoml, line 64: I don't think this is a big deal, but why are we doing this?

bentoml/_internal/utils/analytics/cli_events.py
```diff
@@ -0,0 +1,36 @@
...
+    cmd_name: str,
+    return_value: "t.Optional[Bento]",
+) -> BentoBuildEvent:  # pragma: no cover
+    if return_value is not None:
+        bento = return_value
+        return BentoBuildEvent(
+            cmd_group=cmd_group,
+            cmd_name=cmd_name,
+            bento_creation_timestamp=bento.info.creation_time,
+            bento_size_in_kb=calc_dir_size(bento.path_of("/")),
```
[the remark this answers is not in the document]
aarnphm: it should be fast

bentoml/_internal/utils/analytics/schemas.py
aarnphm: I was learning about this earlier. We definitely don't have to do this

bentoml/_internal/utils/analytics/usage_stats.py
```diff
@@ -0,0 +1,177 @@
...
+        yield
+        return
+
+    _track_serve_init(svc, production)
+
+    stop_event = threading.Event()
+
+    @slient
+    def loop() -> t.NoReturn:  # type: ignore
+        while not stop_event.wait(SERVE_USAGE_TRACKING_INTERVAL_SECONDS):
```
timliubentoml, line 160: cool! this seems pretty straightforward

README.md
```diff
@@ -1,6 +1,7 @@
 [<img src="https://raw.githubusercontent.com/bentoml/BentoML/main/docs/source/_static/img/bentoml-readme-header.jpeg" width="600px" margin-left="-5px">](https://github.com/bentoml/BentoML)
+<br>
```
sauyon, line 2: This doesn't seem like it's doing much? Maybe add another one?

README.md
```diff
@@ -1,6 +1,7 @@
...
 
-# Unified Model Serving Framework  [![Tweet](https://img.shields.io/twitter/url/http/shields.io.svg?style=social)](https://twitter.com/intent/tweet?text=BentoML:%20The%20Unified%20Model%20Serving%20Framework%20&url=https://github.com/bentoml&via=bentomlai&hashtags=mlops,bentoml)
+# 🍱 BentoML: Unified Model Serving Framework  [![Tweet](https://img.shields.io/twitter/url/http/shields.io.svg?style=social)](https://twitter.com/intent/tweet?text=BentoML:%20The%20Unified%20Model%20Serving%20Framework%20&url=https://github.com/bentoml&via=bentomlai&hashtags=mlops,bentoml)
```
sauyon, line 4:
```suggestion
# 🍱 BentoML: The Unified Model Serving Framework  [![Tweet](https://img.shields.io/twitter/url/http/shields.io.svg?style=social)](https://twitter.com/intent/tweet?text=BentoML:%20The%20Unified%20Model%20Serving%20Framework%20&url=https://github.com/bentoml&via=bentomlai&hashtags=mlops,bentoml)
```

README.md
```diff
@@ -55,18 +56,20 @@ There are many ways to contribute to the project:
...
 ### Usage Reporting ###
 
-BentoML by default collects anonymous usage data using [Amplitude](https://amplitude.com/). 
-It only collects BentoML library's own actions and parameters, no user or model data will be collected. 
-Here is the [code](https://github.com/bentoml/BentoML/blob/main/bentoml/_internal/utils/usage_stats.py) that does it.
+BentoML collects anonymous usage data that helps our team to improve the product.
+Only BentoML's internal API calls and CLI commands are being reported and we striped out any potentially 
+sensitive information as much as possible, we will never collect user code, model data, model name or stacktrace. 
+Here's the [code](./bentoml/_internal/utils/analytics/usage_stats.py) for usage tracking.
+You can opt-out of usage tracking by the `--do-not-track` CLI option:
```
sauyon, line 63:
```suggestion
Only BentoML's internal API calls and CLI commands are being reported. We strip out as much potentially 
sensitive information as possible, and we will never collect user code, model data, model names, or stack traces.
Here's the [code](./bentoml/_internal/utils/analytics/usage_stats.py) for usage tracking.
You can opt-out of usage tracking by the `--do-not-track` CLI option:
```

bentoml/_internal/bento/bento.py
```diff
@@ -373,14 +381,66 @@ def __init__(self, base_path: t.Union[PathType, "FS"]):
...
+    def from_inference_api(cls, api: "InferenceAPI") -> "ApiInfo":
+        return cls(
+            name=api.name,
+            input_type=api.input.__class__.__name__,
+            output_type=api.output.__class__.__name__,
+        )
+
+
+@attr.define
+class ModelInfo:
```
sauyon, line 420: Maybe we should rename this because it's confusing to have this and `models.ModelInfo` be different things?

bentoml/_internal/bento/bento.py
```diff
@@ -418,7 +480,10 @@ def from_yaml_file(cls, stream: t.IO[t.Any]) -> "BentoInfo":
         del yaml_content["version"]
 
         try:
-            return cattr.structure(yaml_content, cls)  # type: ignore[attr-defined]
+            # type: ignore[attr-defined]
+            print(yaml_content)
+            print(cattr.structure(yaml_content, cls))
```
sauyon, line 485: Debug prints?

bentoml/_internal/cli/bento_management.py
```diff
@@ -15,27 +17,26 @@
...
+    ctx: "click.Context", params: "click.Parameter", value: t.Any
+) -> t.Any:  # pylint: disable=unused-argument
     if value is None:
         return value
     delete_targets = value.split(",")
     delete_targets = list(map(str.strip, delete_targets))
     for delete_target in delete_targets:
         if not (
-            _is_valid_bento_tag(delete_target) or _is_valid_bento_name(delete_target)
+            is_valid_bento_tag(delete_target) or is_valid_bento_name(delete_target)
```
sauyon, line 39: We should probably introduce `Tag.is_valid` or just use `Tag(<tag to check>)` here.

bentoml/_internal/cli/bento_management.py
sauyon: Shouldn't we just annotate the correct return type, then?

bentoml/_internal/utils/analytics/cli_events.py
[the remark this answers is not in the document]
sauyon: We should probably at least cap the time taken on this. On a machine with a spinning disk summing file sizes can take upwards of minutes if there are a lot of files. This is a run with a bunch of sizes cached already:
`du -hs .  3.16s user 13.09s system 13% cpu 2:00.60 total`

bentoml/_internal/utils/analytics/schemas.py
sauyon: This is what UUID1 was made for, though, no?

docs/source/index.rst
```diff
@@ -62,22 +62,22 @@ There are many ways to contribute to the project:
...
 ---------------
 
-BentoML by default collects anonymous usage data using `Amplitude <https://amplitude.com/>`_. 
-It only collects BentoML library's own actions and parameters, no user or model data will be collected. 
-Here is the `code <https://github.com/bentoml/BentoML/blob/main/bentoml/utils/usage_stats.py>`_ that does it.
-
-This helps the BentoML team to understand how the community is using this tool and what to build next. 
-You can easily opt-out of usage tracking by running the BentoML commands with the `--do-not-track` option.
+BentoML collects anonymous usage data that helps our team to improve the product.
+Only BentoML's internal API calls and CLI commands are being reported and we striped out any potentially
```
sauyon, line 66: Same suggestion as above.

sauyon (commented):
Two notes:
 - Has this been tested in an environment where `t.bentoml.com` isn't reachable?
 - I think we should have a more permanent way of disabling tracking. It's worth noting most of the time when enabling tracking by default there's a prompt on first run asking the user whether they want to opt-out; I think that's probably worth doing, even if it would mean less tracking data, since I think a decent proportion of users will never see the disclaimers.

aarnphm:
> Two notes:
> 
> * Has this been tested in an environment where `t.bentoml.com` isn't reachable?
We silent the log with timeout, thus I believe it should work when not connected to the internet.

> * I think we should have a more permanent way of disabling tracking. It's worth noting most of the time when enabling tracking by default there's a prompt on first run asking the user whether they want to opt-out; I think that's probably worth doing, even if it would mean less tracking data, since I think a decent proportion of users will never see the disclaimers.

If they want to permanently disable tracking, then just put `BENTOML_DO_NOT_TRACK=True` under the shell rc. Chaoyu told me to remove that from the doc.

bentoml/_internal/utils/analytics/schemas.py
aarnphm: we actually want uuid1 for this

sauyon:
> If they want to permanently disable tracking, then just put `BENTOML_DO_NOT_TRACK=True` under the shell rc. Chaoyu told me to remove that from the doc.

My issue is less with the fact that it's configured via env var and more with the fact that it's silent and doesn't present an option to disable it permanently for you.

sauyon:
> We silent the log with timeout, thus I believe it should work when not connected to the internet.

We should make sure this is tested in CI, I think.

bentoml/_internal/cli/bento_management.py
parano: yes we should, will leave that to follow up pr

bentoml/_internal/utils/analytics/cli_events.py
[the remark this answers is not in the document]
parano: we also calculate bento/model size in the `bentoml list` CLI command. Maybe we should calculate it once and save it in a file in the Bento/Model directory?

bentoml/_internal/yatai_rest_api_client/config.py
```diff
@@ -27,7 +27,9 @@ class YataiClientContext:
     name: str
     endpoint: str
     api_token: str
-    email: Optional[str] = attr.field(default=None)
+    email: Optional[str] = attr.field(
```
parano, line 30: why is this change? @aarnphm

setup.cfg
```diff
@@ -103,20 +103,20 @@ formats = gztar
...
```
parano, line 116:
```suggestion
```

bentoml/_internal/bento/bento.py
parano: changing to `BentoModelInfo`

bentoml/_internal/yatai_rest_api_client/config.py
aarnphm: it was to fix the previous issue with yatai config, now I don't think we need to do this anymore.

bentoml/_internal/cli/click_utils.py
```diff
@@ -180,19 +195,27 @@ def wrapper(func):
...
+            error_msg = str(e)
+            original_cmd_name = click.utils.make_str(args[0])
+            matches = difflib.get_close_matches(
+                original_cmd_name, self.list_commands(ctx), 3, 0.5
+            )
+            if matches:
+                fmt_matches = "\n    ".join(matches)
+                error_msg += "\n\n"
+                error_msg += f"Did you mean?\n    {fmt_matches}"
+            raise UsageError(error_msg, e.ctx)
```
parano, line 213: @aarnphm this is really nice! 

@yubozhao @jjmachan let's do something similar in bentoctl?

parano:
> > We silent the log with timeout, thus I believe it should work when not connected to the internet.
> 
> We should make sure this is tested in CI, I think.

This behavior is now covered in unit test

PR #2323, opened 2022-03-02
chore: fetch yatai version and store email and yatai version in yatai config

yubozhao:
## Motivation and Context

Retrieve and store user email & Yatai version during `bentoml yatai login` CLI command.

## Checklist:

### DIFF
--- a/bentoml/_internal/cli/yatai.py
+++ b/bentoml/_internal/cli/yatai.py
@@ -42,7 +42,7 @@ def add_login_command(cli):
 
         if user is None:
             raise CLIException("current user is not found")
-        ctx.email = user.email if user.email else None
+        ctx.email = user.email if user.email is not None else None
 
         org = yatai_rest_client.get_current_organization()
 
--- a/bentoml/_internal/cli/yatai.py
+++ b/bentoml/_internal/cli/yatai.py
@@ -5,6 +5,7 @@ import click
 from bentoml.exceptions import CLIException
 
 from ..yatai_rest_api_client.config import add_context
+from ..yatai_rest_api_client.config import update_context
 from ..yatai_rest_api_client.config import YataiClientContext
 from ..yatai_rest_api_client.config import default_context_name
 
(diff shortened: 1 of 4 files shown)


bentoml/_internal/yatai_rest_api_client/config.py
```diff
@@ -26,6 +26,8 @@ class YataiClientContext:
     name: str
     endpoint: str
     api_token: str
+    email: str
+    yatai_version: str
```
yubozhao, line 30: We don't need to track yatai version here, since we are already tracking it on the yatai side.

Also, as Yatai version continue increases, we will have to keep checking the version to the latest.  

My suggestion is not track yatai version in bentoml

aarnphm: can we exported to an envars, like `YATAI_VERSION` so that we can just retrieved it on bentoml side

bentoml/_internal/cli/yatai.py
```diff
@@ -41,12 +42,21 @@ def login(endpoint: str, api_token: str) -> None:
 
         if user is None:
             raise CLIException("current user is not found")
+        ctx.email = user.email
```
parano, line 45: verify if `user.email` filed exist in the response??

bentoml/_internal/cli/yatai.py
```diff
@@ -42,19 +42,13 @@ def login(endpoint: str, api_token: str) -> None:
 
         if user is None:
             raise CLIException("current user is not found")
-        ctx.email = user.email
+        ctx.email = user.email if user.email else None
```
aarnphm, line 45:
```suggestion
        ctx.email = user.email if user.email is not None else None
```

aarnphm (commented):
otherwise LGTM

aarnphm (approved):

aarnphm:
merging now @yubozhao ?

bentoml/_internal/cli/yatai.py
```diff
@@ -41,12 +42,15 @@ def login(endpoint: str, api_token: str) -> None:
 
         if user is None:
             raise CLIException("current user is not found")
+        ctx.email = user.email if user.email is not None else None
```
parano, line 45: Does the user attrs class already set missing email files to None? If so we don’t need this check

aarnphm: oh we already set the email to None

parano: then we can remove this line right?

aarnphm: we just need to set ctx.emaill to user.email

yubozhao: Going to revert back to original implementation

aarnphm (approved):

PR #2336, opened 2022-03-03
feat: multi-architecture support for bentoml docker images

aarnphm:

## Checklist:


aarnphm:
CI will get updated separately.

docker/tests/sanity_check.sh
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env bash
```
parano, line 1: 👍

docker/tests/00-version-check.bats
```diff
@@ -0,0 +1,30 @@
...
+
+@test "conda_version" {
+    docker_run --rm  ${image} conda --version
+    [ "$status" -eq 0 ]
+}
+
+@test "bash_version" {
+    docker_run --rm  ${image} bash --version
+    [ "$status" -eq 0 ]
+}
```
parano, line 30: should we also check the python and pip version?

aarnphm:
For future reference: alpine is limited to x86_64 due to conda not having support for arm64 installer, yet.

docker/.env.example
```diff
@@ -1,2 +1,15 @@
-HUB_PASSWORD=
-HUB_USERNAME=
\ No newline at end of file
+DOCKER_PASSWORD=world
+DOCKER_USER=hello
+DOCKER_URL=docker.io/<org_name>
+
+AWS_USER=
+AWS_PASSWORD=<...>
```
sauyon, line 6: why is this the only password env to have a placeholder value?

docker/README.md
```diff
@@ -43,110 +43,161 @@ See all available tags [here](https://hub.docker.com/repository/docker/bentoml/b
 
 Before starting:
 
+* We will utilize `docker buildx` underthe hood, so make sure to have that installed with `docker buildx install` (buildx should already installed for folks with Docker > 19.03)
+
+* we will also use [docker-pushrm](https://github.com/christian-korneck/docker-pushrm) to push the given readme, so make sure to have that installed as well.
```
sauyon, line 48: Do we?

docker/README.md
```diff
@@ -43,110 +43,161 @@ See all available tags [here](https://hub.docker.com/repository/docker/bentoml/b
 
 Before starting:
 
+* We will utilize `docker buildx` underthe hood, so make sure to have that installed with `docker buildx install` (buildx should already installed for folks with Docker > 19.03)
```
sauyon, line 46: Should we require this anywhere that's not the CI? Can we make it so that locally we can just use `docker build` and then use `buildx` to create multiarch images when running CD? I feel like that would limit the complexity that users that just want to tool around with the dockerfile a little bit are exposed to.

We can still have all the fun `manager` stuff but it doesn't seem to me like we should expose most of it to the average user.

sauyon:
@aarnphm would be good to have you run me through all this stuff at some point if you have time.

docker/README.md
aarnphm: yes we do

docker/README.md
aarnphm: users can still build the docker images per each dockerfile, but for us we will use buildx to support multiple architecture

sauyon: Sure, I mean that we don't need to expose this much complexity in the README, then. I'd prefer if `buildx` was only really necessary for building on CD, but I need to really review everything first.

aarnphm:
For amazonlinux will suport 3.7 for now.

aarnphm:
CI and linter will be addressed separately, for now its just placeholder

docker/generated/bento-server/debian10/runtime/Dockerfile
```diff
@@ -1,16 +1,28 @@
...
+FROM base-${TARGETARCH}
+
+ARG TARGETARCH
 
-ENV BENTOML_VERSION=1.0.0a5
+ARG TARGETPLATFORM
 
-RUN pip install bentoml==${BENTOML_VERSION} --no-cache-dir
\ No newline at end of file
+RUN rm -rf /var/lib/{apt,dpkg,cache,log}
```
aarnphm, line 28: I need some feedback here. For all debian images, if we do this, then users won't be able to install new packages to the container. They will only have access to python, curl, bash and bentoml

docker/manager/_internal/_gen.py
```diff
@@ -0,0 +1,309 @@
...
+        checker, templates_fs.walk.files(filter=["*.yaml"])
+    ):
+        if check in p:
+            p = p.strip("/").strip(".yaml")
+            results[filter_key(p)] = f"!include include.d/{templates_dir}/{p}.yaml"
+    return results
+
+
+def unpack_include_content(include_: IncludeMapping, indent: int = 0) -> str:
+    mem_fs = fs.open_fs("mem://")
```
sauyon, line 78: Seems like you really want to just use `io.StringIO` or `io.BytesIO` depending on whether pyyaml supports string i/o here.

docker/manager/_internal/_gen.py
```diff
@@ -0,0 +1,309 @@
...
+    tmp_file = partial(mem_fs.open, f".{uuid1()}.yml", encoding="utf-8")
+    include_string = "\n".join([f"{k}: {v}\n" for k, v in include_.items()])
+
+    with tmp_file("w") as inf:
+        yaml.dump(
+            yaml.load(include_string.encode("utf-8"), Loader=yaml.FullLoader), inf
```
sauyon, line 84: It seems like `yaml` supports just string input?

docker/manager/_internal/_gen.py
```diff
@@ -0,0 +1,309 @@
...
+            for k, v in rltx.release_tags.items()
+            if "base" not in k
+        ]
+
+        (k_sort := [t[0] for t in results]).sort(
+            key=lambda tag: get_python_version_from_tag(tag)
+        )
+        order = {k: v for v, k in enumerate(k_sort)}
+
+        tag_ref[distro] = sorted(results, key=lambda k: order[k[0]])
```
sauyon, line 259: Why sort twice here?

```suggestion
        def result_sort_key(result):
            return get_python_version_from_tag(result[0])

        tag_ref[distro] = sorted(results, key=result_sort_key)
```

docker/manager/_internal/_gen.py
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    readme_tmpl = fs.path.combine("docs", "README.md.j2")
+    tmp_fs = fs.open_fs("temp://")
+
+    def render_final_readmes(
+        out_name: str,
+        out_fs: FS,
+        *,
+        template_files: t.List[str] = ["headers.md.j2", "body.md.j2"],
```
sauyon, line 278: This doesn't seem to be used?

docker/manager/_internal/_gen.py
```diff
@@ -0,0 +1,309 @@
...
+            readme_tmpl,
+            ctx._templates_dir,
+            out_name,
+            out_fs,
+            **final_context,
+        )
+
+    render_final_readmes(ctx.docker_package, ctx._generated_dir)
+    render_final_readmes("/", ctx._fs, extends={"main_readme": True})
+    render_final_readmes("/", ctx._generated_dir, extends={"emphemeral": True})
```
sauyon, line 309: What are the different uses of these readmes?

sauyon (commented):
Halfway review so you can start responding to stuff.

docker/README.md
```diff
@@ -1,239 +1,269 @@
...
+[![join_slack](https://badgen.net/badge/Join/BentoML%20Slack/cyan?icon=slack)](https://join.slack.bentoml.org)
+
+
+## Why BentoML ##
+
+- The easiest way to turn your ML models into production-ready API endpoints.
+- High performance model serving, all in Python.
+- Standardized model packaging and ML service definition to streamline deployment.
+- Support all major machine-learning training [frameworks](https://docs.bentoml.org/en/latest/frameworks/index.html).
+- Deploy and operate ML serving workload at scale on Kubernetes via [Yatai](https://github.com/bentoml/yatai).
```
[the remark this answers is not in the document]
parano: I agree we should remove it here, also makes it hard to keep everything in sync

docker/README.md
```diff
@@ -1,239 +1,269 @@
...
 
 Example image tags:
 - `bento-server:devel-python3.7-debian`
 - `bento-server:1.0.0-python3.8-ubi8-cudnn`
 - `bento-server:1.0.0-python3.7-ami2-runtime`
 
-See all available tags [here](https://hub.docker.com/repository/docker/bentoml/bento-server/tags).
+## NOTICE: MISSING PYTHON VERSION ON UBI
 
+Python 3.7 and 3.10 is missing. The reason being RedHat doesn't provide support for these Python version.
```
parano, line 67: add a check for this in BuildConfig

docker/README.md
```diff
@@ -1,239 +1,269 @@
...
+If you need to use UBI and Python 3.7 make sure to contact the BentoML team for supports..
+
+## NOTICE: CONDA AVAILABILITY ONLY ON DEBIAN
+
+From 1.0.0a7 onwards, BentoML will only provide conda supports with debian variants only.
```
parano, line 72: add a check for this in BentoBuildConfig, that verify CondaOptions.is_empty and DockerOptions.distro are compatible

.github/workflows/ci.yml
```diff
@@ -29,19 +30,45 @@ jobs:
...
               - "tests/**"
-              - "docker/**"
             docs:
               - *related
               - requirements/docs-requirements.txt
               - "docs/**"
+            docker_readme:
+              - docker/generated/bento-server/README.md
+
+  bento_server_push_readme:
```
parano, line 51: could we add a CI task on new commits on main branch, that builds and pushes new devel images?

docker/README.md
```diff
@@ -1,239 +1,269 @@
...
+
+We ran into a lot of trouble building BentoML to supports Python from 3.6 to 3.10 with conda environment on other distros than debian. In order to reduce 
+complexity we will now only provides conda on Debian-based image. Conda will be available with all of BentoML image type, including `runtime`, `devel`, `cudnn`. 
+
+If you need to use conda on other distros contact the BentoML team for supports.
+
+Example conda tags:
+- `bento-server:1.0.0a7-python3.8-debian11-runtime-conda`
```
parano, line 80: let's remove the `-conda` suffix in Debian images - I think doing the validation logic in BentoBuildConfig is good enough

parano: do we have a Debian image that doesn't have Conda?

does that mean we need a `use_conda` field in our DockerOptions user facing config in bentofile.yaml?

docker/generated/README.md
```diff
@@ -1,5 +1,5 @@
+<!-- P/S: even this file is autogenerated :happy: -->
 ## GENERATED DIRECTORY - DO NOT EDIT
 
-The files are maintained by `manager.py`, which generates Dockerfile from `templates`.
+These files are maintained by [`manager`](../manager), which generates Dockerfile from `templates`.
```
parano, line 4: these links are not going to work when pushed to docker io?

docker/generated/bento-server/README.md
```diff
@@ -1,8 +1,10 @@
+<!-- P/S: even this file is autogenerated :happy: -->
+
 # Unified Model Serving Framework  [![Tweet](https://img.shields.io/twitter/url/http/shields.io.svg?style=social)](https://twitter.com/intent/tweet?text=BentoML:%20The%20Unified%20Model%20Serving%20Framework%20&url=https://github.com/bentoml&via=bentomlai&hashtags=mlops,bentoml)
 
 BentoML is an open platform that simplifies ML model deployment and enables you to serve your models at production scale in minutes
 
-👉 [Pop into our Slack community!](https://join.slack.bentoml.org) We're happy to help with any issue you face or even just to meet you and hear what you're working on :)
+👉 [Pop into our Slack community!](https://l.linklyhq.com/l/ktOu) We're happy to help with any issue you face or even just to meet you and hear what you're working on :)
```
parano, line 7: can you confirm with Tim about the link to use here?

docker/generated/bento-server/README.md
```diff
@@ -32,80 +34,155 @@ BentoML is an open platform that simplifies ML model deployment and enables you
...
+If you need to use UBI and Python 3.7 make sure to contact the BentoML team for supports..
+
+## NOTICE: CONDA AVAILABILITY ONLY ON DEBIAN
+
+From 1.0.0a7 onwards, BentoML will only provide conda supports with debian variants only.
+
+We ran into a lot of trouble building BentoML to supports Python from 3.6 to 3.10 with conda environment on other distros than debian. In order to reduce 
+complexity we will now only provides conda on Debian-based image. Conda will be available with all of BentoML image type, including `runtime`, `devel`, `cudnn`. 
+
+If you need to use conda on other distros contact the BentoML team for supports.
```
parano, line 76:
```suggestion
If you need to use conda on other distros, you may build and use your own custom docker base image with BentoML.
```

docker/.env.example
```diff
@@ -1,2 +1,3 @@
-HUB_PASSWORD=
-HUB_USERNAME=
\ No newline at end of file
+DOCKER_PASSWORD=world
```
parano, line 1: why do we need this?

.github/workflows/ci.yml
sauyon: Let's do a nightly deploy? Probably best not to have *too* many releases on docker hub :P

docker/README.md
aarnphm: we have debian images without conda.

docker/generated/README.md
aarnphm: we won't push this readme to docker.io This is to warn not to remove this directory

docker/.env.example
aarnphm: we don't need this anymore

docker/hack/shells/modprobe.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
aarnphm: creating the docker build image.

docker/generated/bento-server/README.md
aarnphm: @timliubentoml can you confirm that the link we use is correct?

aarnphm:
discussed offline with new design, closed for now.

PR #2339, opened 2022-03-06
feat(cli): add export/import cli for bento and model

larme:
## Checklist:

### DIFF
--- a/bentoml/_internal/cli/bento_management.py
+++ b/bentoml/_internal/cli/bento_management.py
@@ -9,6 +9,7 @@ from simple_di import Provide
 from rich.table import Table
 from rich.console import Console
 
+from bentoml.bentos import import_bento
 from bentoml.bentos import build_bentofile
 
 from ..utils import calc_dir_size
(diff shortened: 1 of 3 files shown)


bentoml/_internal/models/model.py
```diff
@@ -151,7 +151,6 @@ def save(
         self, model_store: "ModelStore" = Provide[BentoMLContainer.model_store]
     ) -> "Model":
         self._save(model_store)
-        logger.info(f"Successfully saved {self}")
```
larme, line 154: remove this line to prevent printing this message twice.

bentoml/_internal/cli/bento_management.py
```diff
@@ -170,35 +174,56 @@ def delete_target(target):
...
                 if delete_confirmed:
                     bento_store.delete(bento.tag)
-                    click.echo(f"{bento} deleted")
+                    logger.info(f"{bento} deleted")
 
         for target in delete_targets:
             delete_target(target)
 
-    @cli.command(help="Export Bento to a tar file")
+    @cli.command()
```
aarnphm, line 182: why we delete the help message here?

larme: in `_save()` we have the same message so it will display twice

larme: @aarnphm to let click use docstring to  display help message. The docstring is more detailed.

aarnphm: i see

aarnphm (approved):

bentoml/_internal/cli/bento_management.py
```diff
@@ -170,35 +174,56 @@ def delete_target(target):
...
+        BENTO_TAG: bento identifier
+        OUT_PATH: output path of exported bento.
+          If this argument is not provided, bento is exported to name-version.bento in the current directory.
+          Supported formats are tar ('tar'), tar.gz ('gz'), tar.xz ('xz'), tar.bz2 ('bz2'), and zip
 
-        bentoml export FraudDetector:latest > my_bento.tar
+        \b
+        examples:
+        bentoml export FraudDetector:20210709_DE14C9
         bentoml export FraudDetector:20210709_DE14C9 ./my_bento.tar
```
parano, line 203: @larme @sauyon do we allow user to provide a path/directory when exporting? 

e.g. `bentoml export s3://foo/bar/` which creates a file `s3://foo/bar/FraudDetector_20210709_DE14C9.bento` automatically?

bentoml/_internal/cli/bento_management.py
```diff
@@ -170,35 +174,56 @@ def delete_target(target):
...
+        bentoml export FraudDetector:latest ./my_bento.zip
```
parano, line 204: in our CLI help message examples, could we use `.bento` extension, as a way to build this convention?

bentoml/_internal/cli/model_management.py
```diff
@@ -172,37 +176,47 @@ def delete_target(target):
...
+        \b
+        MODEL_TAG: model identifier
+        OUT_PATH: output path of exported model.
+          If this argument is not provided, model is exported to name-version.bentomodel in the current directory.
+          Supported formats are tar('tar'), tar.gz ('gz'), tar.xz ('xz'), tar.bz2 ('bz2'), and zip
 
+        examples:
+
+        \b
+        bentoml models export FraudDetector:latest > my_model.tar
```
parano, line 201: does this syntax still work ? now we've changed the click argument type from `click.File` to `click.String`?  it is ok to not support this syntax, but let's make sure the documentation is correct

bentoml/_internal/cli/model_management.py
```diff
@@ -172,37 +176,47 @@ def delete_target(target):
...
         bentoml models export FraudDetector:latest > my_model.tar
         bentoml models export FraudDetector:20210709_DE14C9 ./my_model.tar
+        bentoml models export FraudDetector:20210709_DE14C9 s3://mybucket/models/my_model.gz
```
parano, line 204: similarly, let's use `.bentomodel` extension here as the default.  We can mention it also supports `.tar` or `.gz`, but always recommend users to use `.bento` or `.bentomodel`, this is because it makes it obvious what it is during `import`, user would know immediately which API to use, `bentoml models import` or `bentoml import`.

bentoml/_internal/cli/bento_management.py
sauyon: Yes, that should indeed work. Note that for the remote sources I didn't add a check that the remote was a directory, so I consciously chose not to document this because sometimes shells can be weird with including final `/`s with autocomplete and such.

PR #2343, opened 2022-03-07
Better options for `bentoml build`

sauyon:
Adds two features to `bentoml build`:
 - can now set the tag of the output bento (like `docker build`) with `--tag <bento>:<ver>` or `--tag <bento> --version <ver>`
 - `build_ctx` is now automatically guessed from the `bentofile` parameter if the `bentofile` parameter is absolute, so commands like `bentoml build -f /path/to/bento/bentofile.yaml` should work as expected.

/cc @bojiang

### DIFF
--- a/bentoml/_internal/bento/bento.py
+++ b/bentoml/_internal/bento/bento.py
@@ -137,6 +137,7 @@ class Bento(StoreItem):
         build_config: BentoBuildConfig,
         version: t.Optional[str] = None,
         build_ctx: t.Optional[str] = None,
+        tag: t.Optional[t.Union[str, Tag]] = None,
         model_store: "ModelStore" = Provide[BentoMLContainer.model_store],
     ) -> "Bento":
         from ..service.loader import import_service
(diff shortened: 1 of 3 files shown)


bentoml/_internal/bento/bento.py
```diff
@@ -137,6 +137,7 @@ def create(
         build_config: BentoBuildConfig,
         version: t.Optional[str] = None,
         build_ctx: t.Optional[str] = None,
+        tag: t.Optional[t.Union[str, Tag]] = None,
```
parano, line 140: The name in the tag should always be the bentoml.Service name specified in code, right? I think it make sense to pass version here, but tag is a bit confusing to me

sauyon: On second thought I think you're right; I misunderstood what the docker tag does.

@bojiang did you have something to add here?

parano: Maybe we should name the `bentoml containerize .. —tag` option to `—docker-image-tag`?

sauyon: Maybe we should allow for CLI labeling though?

sauyon: > Maybe we should name the `bentoml containerize .. —tag` option to `—docker-image-tag`? Or add an alias?

I think it makes sense as is. Do we expose the name tag in the documentation?

bojiang: > The name in the tag should always be the bentoml.Service name specified in code, right? I think it make sense to pass version here, but tag is a bit confusing to me
Shall make it customizable?

parano: I think it will be confusing to users. Better to just name it in one place only. Do you see a use case for it?

parano:
@bojiang `tag` should not be configurable during build, user can provide a custom version for the tag via `--version`, but the name is fixed in the `bentoml.Service` creation code. Closing now.

PR #2360, opened 2022-03-15
fix(frameworks): use bentoml.models.create instead of Model.create

sauyon:
As a prelude to some changes which will freeze `Model` info after creation.

### DIFF
--- a/bentoml/_internal/frameworks/lightgbm.py
+++ b/bentoml/_internal/frameworks/lightgbm.py
@@ -6,8 +6,8 @@ import joblib  # type: ignore[reportMissingTypeStubs]
 from simple_di import inject
 from simple_di import Provide
 
+import bentoml
 from bentoml import Tag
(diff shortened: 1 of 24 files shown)


bentoml/_internal/frameworks/lightgbm.py
```diff
@@ -185,34 +185,32 @@ def save(
...
-    ):
-        joblib.dump(model, _model.path_of(f"{SAVE_NAMESPACE}{PKL_EXT}"))
-        _model.info.options["sklearn_api"] = True
-    else:
-        model.save_model(_model.path_of(f"{SAVE_NAMESPACE}{TXT_EXT}"))  # type: ignore
-
-    _model.save(model_store)
-
-    return _model.tag
+    ) as _model:
```
parano, line 196: should we rename all the `_model` to `bento_model`, as @larme suggested today?

PR #2346, opened 2022-03-09
docs: update advanced guide on building bentos

splch:
## Description

This PR will add the Building Bentos guide for users. It covers the `bentofile.yaml`'s generation and use in BentoML.

## Motivation and Context

This guide will help explain how to use Bentos to incorporate BentoML into existing development environments.

Mentioned in MLH meetings and by #2326

## Checklist:

- [ ] My code follows the bentoml code style, both `make format` and
  `make lint` script have passed
  ([instructions](https://github.com/bentoml/BentoML/blob/master/DEVELOPMENT.md#style-check-and-auto-formatting-your-code)).
- [ ] My change reduces project test coverage and requires unit tests to be added
- [ ] I have added unit tests covering my code change
- [x] My change requires a change to the documentation
- [x] I have updated the documentation accordingly

PR #2362, opened 2022-03-17
fix: extract correct desired Python version

matheusMoreno:
## Description

This PR proposes a fix for the way that the desired Python version of the Docker image is extracted by the `init.sh`.

## Motivation and Context

Currently, the `init.sh` script used by BentoML infers the desired Python version by slicing the value of the saved Python version, retrieving the first three chars. But if the saved Python version is >3.9 (such as 3.10), the script returns 3.1 as the desired Python version, which is incorrect.

The proposed change uses parameter expansion to remove the patch version from the saved Python version, fixing the issue.

## How Has This Been Tested?

I was unable to test the entire script, but I tested the area affected, and can confirm that it works. Since parameter expansion is native to bash, this modification should not break the code.

## Checklist:

- [ ] My code follows the bentoml code style, both `make format` and `make lint` script have passed ([instructions](https://github.com/bentoml/BentoML/blob/master/DEVELOPMENT.md#style-check-and-auto-formatting-your-code)).

### DIFF
--- a/bentoml/_internal/bento/docker/init.sh
+++ b/bentoml/_internal/bento/docker/init.sh
@@ -20,9 +20,9 @@ if [ $# -eq 0 ] || [ $1 == "ensure_python" ] ; then
   if [ -f ./env/python/version.txt ]; then
     PY_VERSION_SAVED=$(cat ./env/python/version.txt)
     # remove PATCH version - since most patch version only contains backwards compatible
-    # bug fixes and the BentoML defautl docker base image will include the latest
+    # bug fixes and the BentoML default docker base image will include the latest
     # patch version of each Python minor release
-    DESIRED_PY_VERSION=${PY_VERSION_SAVED:0:3} # returns 3.7, 3.8 or 3.9
+    DESIRED_PY_VERSION=${PY_VERSION_SAVED%.*} # returns major.minor
     CURRENT_PY_VERSION=$(python -c 'import sys; print(f"{sys.version_info.major}.{sys.version_info.minor}")')
 
     if [[ "$DESIRED_PY_VERSION" == "$CURRENT_PY_VERSION" ]]; then


bentoml/_internal/bento/docker/init.sh
```diff
@@ -20,9 +20,9 @@ if [ $# -eq 0 ] || [ $1 == "ensure_python" ] ; then
   if [ -f ./env/python/version.txt ]; then
     PY_VERSION_SAVED=$(cat ./env/python/version.txt)
     # remove PATCH version - since most patch version only contains backwards compatible
-    # bug fixes and the BentoML defautl docker base image will include the latest
+    # bug fixes and the BentoML default docker base image will include the latest
     # patch version of each Python minor release
-    DESIRED_PY_VERSION=${PY_VERSION_SAVED:0:3} # returns 3.7, 3.8 or 3.9
+    DESIRED_PY_VERSION=${PY_VERSION_SAVED%.*} # returns major.minor
```
parano, line 25: This will return the `major.minor.patch` right?  e.g. if `PY_VERSION_SAVED=3.10.12`, it will return `3.10.12` here? is it possible to use `grep` or `sed` to extract only the `major.minor` here?  Since BentoML only pre-built base image for all the python minor versions supported, but not for all the `patch` versions.

aarnphm: This will only return `major.minor`. But we eventually want `major.minor.patch`

parano:
Thank you for the PR @matheusMoreno - @aarnphm is actually working on python 3.10 base image support here https://github.com/bentoml/BentoML/pull/2346 and we indeed will need this change for full 3.10 support.

bentoml/_internal/bento/docker/init.sh
parano: nvm, just realize what `%.*` does in bash

PR #2363, opened 2022-03-17
freeze model info and validate metadata entries

sauyon:


bentoml/_internal/utils/__init__.py
```diff
@@ -167,6 +174,76 @@ def resolve_user_filepath(filepath: str, ctx: t.Optional[str]) -> str:
...
+        )
+    elif LazyType("pandas", "Series").isinstance(entry):
+        entry = {entry.name: entry.to_numpy().tolist()}
+        _validate_metadata_entry(entry)
+    elif LazyType("pandas.api.extensions", "ExtensionArray").isinstance(entry):
+        entry = entry.to_numpy().tolist()  # type: ignore (LazyType)
+        _validate_metadata_entry(entry)
+    elif LazyType("pandas", "DataFrame").isinstance(entry):
+        entry = entry.to_dict("list")  # type: ignore (LazyType)
+        validate_metadata(entry)  # type: ignore (LazyType)
```
larme, line 227: same I wonder if we should just use `to_dict` instead of `to_dict("list")`

sauyon:
Just got around to rebasing this properly...

sauyon:
Oh, `to_dict` generates dicts like `{0: 1, 1: 2, 2: 4}`. Should I go through and just convert the number keys to strings? Or should we just allow integer keys?

sauyon:
Hm, does anybody know what's going on with the pytorch test?

PR #2375, opened 2022-03-25
feat(internal): implement request contexts and check inference API types

sauyon:
Very WIP, context still needs to be merged with tracing context. Mostly for @bojiang to take a quick look at inference API implementation.

### DIFF
--- a/bentoml/_internal/context.py
+++ b/bentoml/_internal/context.py
@@ -1,5 +1,157 @@
 import typing as t
 import contextvars
+from abc import ABC
+from abc import abstractmethod
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    import starlette.requests
+    import starlette.responses
+
+
+class Metadata(ABC):
+    @abstractmethod
+    def __setitem__(self, key: str, value: str) -> None:
+        """
+        Set the header ``key`` to ``value``, removing any duplicate entries.
+        Retains insertion order.
+        """
+
+    @abstractmethod
+    def __delitem__(self, key: str) -> None:
+        """
+        Remove the header ``key``.
+        """
+
+    @abstractmethod
+    def __ior__(self, other: t.Mapping[t.Any, t.Any]) -> "Metadata":
+        """
+        Updates this metadata with the contents of ``other``
+        """
+
+    @abstractmethod
+    def __or__(self, other: t.Mapping[t.Any, t.Any]) -> "Metadata":
+        """
+        Returns a new metadata object with the contents of this metadata object updated with the contents of ``other``
+        """
+
+    @abstractmethod
+    def setdefault(self, key: str, value: str) -> str:
+        """
+        If the header ``key`` does not exist, then set it to ``value``.
+        Returns the header value.
+        """
+
+    @abstractmethod
+    def update(self, other: t.Mapping[t.Any, t.Any]) -> None:
+        """
+        Sets all the headers in ``other`` in this object.
+
+        For example, if this object is ``{"my-header": "my-value", "my-other-header": "my-other-value"}``
+        and other is {"my-header": 3¸ "other-header": 4}
+        """
+
+    @abstractmethod
+    def append(self, key: str, value: str) -> None:
+        """
+        Append a header, preserving any duplicate entries.
+        """
+
+
+class InferenceApiContext:
+    request: "RequestContext"
+    response: "ResponseContext"
+
+    def __init__(self, request: "RequestContext", response: "ResponseContext"):
+        self.request = request
+        self.response = response
+
+    @staticmethod
+    def from_http(
+        request: "starlette.requests.Request", response: "starlette.responses.Response"
+    ) -> "InferenceApiContext":
+        request_ctx = InferenceApiContext.RequestContext.from_http(request)
+        response_ctx = InferenceApiContext.ResponseContext.from_http(response)
+
+        return InferenceApiContext(request_ctx, response_ctx)
+
+    class RequestContext:
+        metadata: Metadata
+        headers: Metadata
+
+        def __init__(self, metadata: Metadata):
+            self.metadata = metadata
+            self.headers = metadata
+
+        @staticmethod
+        def from_http(
+            request: "starlette.requests.Request",
+        ) -> "InferenceApiContext.RequestContext":
+            return InferenceApiContext.RequestContext(request.headers)  # type: ignore (coercing Starlette headers to Metadata)
+
+    class ResponseContext:
+        _raw_response: "starlette.responses.Response"
+        metadata: Metadata
+        headers: Metadata
+        set_cookie: t.Callable[
+            [
+                str,
+                str,
+                int,
+                int,
+                str,
+                str,
+                bool,
+                bool,
+            ],
+            None,
+        ]
+        delete_cookie: t.Callable[[str, str, str], None]
+        response_code: int = 200
+
+        def __init__(
+            self,
+            raw: "starlette.responses.Response",
+            metadata: Metadata,
+            set_cookie: t.Callable[
+                [
+                    str,
+                    str,
+                    int,
+                    int,
+                    str,
+                    str,
+                    bool,
+                    bool,
+                ],
+                None,
+            ],
+            delete_cookie: t.Callable[[str, str, str], None],
+        ):
+            self._raw_response = raw
+            self.metadata = metadata
+            self.headers = metadata
+            self.set_cookie = set_cookie
+            self.delete_cookie = delete_cookie
+
+        @property
+        def status_code(self) -> int:
+            return self._raw_response.status_code
+
+        @status_code.setter
+        def status_code(self, code: int) -> None:
+            self._raw_response.status_code = code
+
+        @staticmethod
+        def from_http(
+            response: "starlette.responses.Response",
+        ) -> "InferenceApiContext.ResponseContext":
+            return InferenceApiContext.ResponseContext(
+                response,
+                response.headers,  # type: ignore (coercing starlette Headers to Metadata)
+                response.set_cookie,
+                response.delete_cookie,
+            )
 
 
 class ServiceTraceContext:
--- a/bentoml/_internal/io_descriptors/file.py
+++ b/bentoml/_internal/io_descriptors/file.py
@@ -8,7 +8,6 @@ from starlette.responses import Response
 
 from .base import IODescriptor
 from ..types import FileLike
-from ..types import LazyType
 from ..utils.http import finalize_http_response
 from ...exceptions import BentoMLException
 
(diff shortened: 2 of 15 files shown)


bentoml/_internal/io_descriptors/multipart.py
```diff
@@ -191,10 +209,15 @@ async def from_http_request(self, request: Request) -> MultipartIO:
...
 
-    async def to_http_response(self, obj: MultipartIO) -> Response:
+    async def init_http_response(self) -> Response:
+        return Response(None)
+
+    async def finalize_http_response(self, response: Response, obj: MultipartIO):
         res_mapping: t.Dict[str, Response] = {}
         for k, io_ in self._inputs.items():
             data = obj[k]
             # TODO(aarnphm): fix with stubs
```
parano, line 219:
```suggestion
```

parano:
LGTM!

bentoml/_internal/server/service_app.py
```diff
@@ -313,17 +315,37 @@ async def api_func(
...
+                        if api.needs_ctx:
+                            ctx = Context.from_http(request, response)
+                            output = await run_in_threadpool(
+                                api.func(input_data, ctx=ctx)
+                            )
+                        else:
+                            output = await run_in_threadpool(api.func, input_data)
+
+                await api.output.finalize_http_response(response, output)
+                finalize_http_response(response)
```
parano, line 348: it looks like we are calling this function both here and in the `api.output.finalize_http_response` above, do we need both?

bentoml/_internal/service/inference_api.py
```diff
@@ -39,7 +45,77 @@ def __init__(
         InferenceAPI._validate_route(route)
 
         self.name = name
+        self.needs_ctx = False
+        self.ctx_param = None
         self.func = user_defined_callback
+        input_type = input_descriptor.input_type()
+        self.multi_input = isinstance(input_type, dict)
```
parano, line 52: should we do `isinstance(input_descriptor, MultiInput)` here just in case another new IODescriptor added that has input type dict?

bentoml/_internal/service/inference_api.py
```diff
@@ -39,7 +45,77 @@ def __init__(
...
+        sig = inspect.signature(user_defined_callback)
+        if isinstance(input_type, dict):
```
parano, line 54: same here

parano: or just use `if self.multi_input` here?

bentoml/_internal/service/inference_api.py
sauyon: This is checking that the `input_type` is a `dict` (and therefore not a `t.Type[dict]`). It's a bit weird but I decided the way that we'd represent multiple outputs is a dict of types, while an IO descriptor that returns a dict would just return `dict` itself.

bentoml/_internal/service/inference_api.py
sauyon: Pyright complains if I use `self.multi_input`, so I gave up and used this instead.

bentoml/_internal/server/service_app.py
sauyon: I think this is unnecessary since I'm calling that when I need to. Honestly it should be renamed to something like `set_content_length` now too.

bojiang:
reviewing

bentoml/_internal/utils/http.py
```diff
@@ -0,0 +1,36 @@
...
+                if item_value == b"0":
+                    found_indexes.append(idx)
+                else:
+                    update = False
+
+        if update:
+            for idx in reversed(found_indexes[1:]):
+                del response.raw_headers[idx]
+
+            new_context_len = str(len(body)).encode("latin-1")
```
aarnphm, line 27: what happens if the body is not using latin-1 encoding?

sauyon: `body` is bytes here; this is encoding the content-length.

bentoml/_internal/service/inference_api.py
```diff
@@ -39,7 +45,77 @@ def __init__(
...
+            # note: in python 3.6 kwarg order was not guaranteed to be preserved,
+            #       though it is in practice.
+            for key in input_type:
+                if key not in sig.parameters:
+                    if (
+                        key in ["context", "ctx"]
+                        or sig.parameters[key].annotation == bentoml.Context
```
aarnphm, line 61: we should import `Context` from `..context` instead of `import bentoml`

bentoml/_internal/service/inference_api.py
```diff
@@ -1,9 +1,15 @@
+from __future__ import annotations
+
 import re
 import typing as t
+import inspect
 from typing import Optional
 
 import yaml
 
+import bentoml
```
aarnphm, line 10:
```suggestion
from ..context import InferenceApiContext as Context
```

bentoml/_internal/io_descriptors/numpy.py
```diff
@@ -149,6 +151,9 @@ def _items_schema(self) -> t.Dict[str, t.Any]:
             return {"type": self._infer_types()}
         return {}
 
+    def input_type(self) -> LazyType["ext.NpNDArray"]:
+        return LazyType("numpy", "NdArray")
```
aarnphm, line 155:
```suggestion
        return LazyType("numpy", "NDArray")
```

bentoml/_internal/service/inference_api.py
parano: I think for public facing API, we should just use the public facing import path

bentoml/_internal/io_descriptors/file.py
```diff
@@ -8,6 +8,7 @@
 
 from .base import IODescriptor
 from ..types import FileLike
+from ..utils.http import finalize_http_response
```
aarnphm, line 11: should we call this function with a different name? something like `set_headers` so that we don't confuse ourself with `finalize_http_response` from each of the io descriptors?

bentoml/_internal/context.py
```diff
@@ -1,5 +1,157 @@
...
+        """
+
+    @abstractmethod
+    def append(self, key: str, value: str) -> None:
+        """
+        Append a header, preserving any duplicate entries.
+        """
+
+
+class InferenceApiContext:
```
aarnphm, line 61: `set_headers` missing?

bentoml/_internal/server/service_app.py
```diff
@@ -313,17 +315,37 @@ async def api_func(
...
+                            ctx = Context.from_http(request, response)
+                            input_data[api.ctx_param] = ctx
                         output: t.Any = await run_in_threadpool(api.func, **input_data)
                     else:
-                        output: t.Any = await run_in_threadpool(api.func, input_data)
-                response = await api.output.to_http_response(output)
+                        if api.needs_ctx:
+                            ctx = Context.from_http(request, response)
+                            output = await run_in_threadpool(
+                                api.func(input_data, ctx=ctx)
```
aarnphm, line 342: This line will fail if users have something like:
```python
@svc.api(input=NumpyNdarray(), output=NumpyNdarray())
def classify(input_series: np.ndarray, context: Context) -> np.ndarray:
```

aarnphm, line 342:
```suggestion
                            output = await run_in_threadpool(
                                api.func, input_data, ctx=ctx
```

bentoml/_internal/io_descriptors/file.py
sauyon: Yeah, I'll rename it to set_content_length; that's what it does anyway.

bentoml/_internal/service/inference_api.py
ssheng: Wouldn't `isinstance(input_descriptor, MultiInput)` be a more direct indication?

bentoml/_internal/service/inference_api.py
```diff
@@ -39,7 +45,77 @@ def __init__(
...
+        input_type = input_descriptor.input_type()
+        self.multi_input = isinstance(input_type, dict)
+        sig = inspect.signature(user_defined_callback)
+        if isinstance(input_type, dict):
+            # note: in python 3.6 kwarg order was not guaranteed to be preserved,
+            #       though it is in practice.
+            for key in input_type:
+                if key not in sig.parameters:
+                    if (
+                        key in ["context", "ctx"]
```
ssheng, line 60: Should we allow both `context` and `ctx`? Or allow only one.

bentoml/_internal/service/inference_api.py
```diff
@@ -39,7 +45,77 @@ def __init__(
         InferenceAPI._validate_route(route)
 
         self.name = name
+        self.needs_ctx = False
+        self.ctx_param = None
         self.func = user_defined_callback
+        input_type = input_descriptor.input_type()
+        self.multi_input = isinstance(input_type, dict)
+        sig = inspect.signature(user_defined_callback)
```
ssheng, line 53: Raise is arguments count is zero?

bentoml/_internal/service/inference_api.py
```diff
@@ -39,7 +45,77 @@ def __init__(
...
+                second_arg = next(param_iter)
+                annotation = sig.parameters[second_arg].annotation
+                if (
+                    isinstance(annotation, t.Type)
+                    and annotation != inspect.Signature.empty
+                ):
+                    if not annotation == bentoml.Context:
+                        raise TypeError(
+                            f"Expected type of argument '{second_arg}' to be '{input_type}', got '{sig.parameters[second_arg].annotation}'"
+                        )
```
ssheng, line 117: Seems like we are enforcing the `context` argument to be the last argument is in the singular input case but not the multi-input case. Is my understanding correct? If yes, what is the reason behind?

ssheng (commented):
Overall LGTM with minor comments. Good work!

bentoml/_internal/io_descriptors/numpy.py
bojiang: +1

bentoml/_internal/service/inference_api.py
parano: I slightly prefer to ignore type check here and use `isinstance(input_descriptor, MultiInput)` instead, as it makes the code more readable

sauyon: This was a conscious choice to allow users to write their own I/O descriptor that passes kwargs.

bentoml/_internal/context.py
sauyon: I don't think we intend to have a `set_headers` method.

bentoml/_internal/service/inference_api.py
sauyon: I think this was purely because I didn't want to introduce another function call as overhead by adding a `partial` call here since this is technically on the critical serving path.

At some point I think we could factor the if statements out of the serving loop too, but it's probably worth getting benchmarks actually setup first.

bentoml/_internal/service/inference_api.py
sauyon: I'd want to lean towards allowing any name that could reasonably imply that a user wants a context parameter; but maybe we should just use any name left over after checking the rest of the IO descriptor defined ones?

ssheng: It probably doesn't provide much more flexibility for allowing any names. A lot of times, a well defined rule and naming is simpler for understanding.

bentoml/_internal/service/inference_api.py
sauyon: The code here is not, I was talking about the nested ifs in the server code.

PR #2398, opened 2022-04-06
feat: allow serving from project dir using import str from bentofile.yaml

parano:
## Description

Allow users to run just `bentoml serve` from a project directory containing a `bentofile.yaml` build file. It will import the service using the "service" field in the `bentofile.yaml` by default. All existing load/serve behavior remain the same.

## Checklist:

### DIFF
--- a/bentoml/_internal/cli/bento_server.py
+++ b/bentoml/_internal/cli/bento_server.py
@@ -15,7 +15,7 @@ DEFAULT_RELAOD_DELAY = 0.25
 
 def add_serve_command(cli: click.Group) -> None:
     @cli.command()
-    @click.argument("bento", type=click.STRING)
+    @click.argument("bento", type=click.STRING, default=".")
     @click.option(
         "--production",
         type=click.BOOL,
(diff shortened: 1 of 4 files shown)


bentoml/_internal/cli/bento_server.py
```diff
@@ -15,7 +15,7 @@
 
 def add_serve_command(cli: click.Group) -> None:
     @cli.command()
-    @click.argument("bento", type=click.STRING)
+    @click.argument("bento", type=click.STRING, default=".")
```
timliubentoml, line 18: So this would now let us run: "bentoml serve ."?

We previously didn't allow this. But I like it. Isn't there somewhere that doesn't allow serving a particular bento name based on a regex?

timliubentoml: I think at some point we had this error: "bentoml.exceptions.InvalidArgument: . is not a valid tag: a tag's name or version must consist of alphanumeric characters, '_', '-', or '.', and must start and end with an alphanumeric character"

timliubentoml (approved):

bentoml/_internal/cli/bento_server.py
parano: @timliubentoml yes, both `bentoml serve` and `bentoml serve .` should just work now, as long as there's a `bentofile.yaml` file in the directory

PR #2359, opened 2022-03-15
docs(catboost): improve CatBoost documentation

sauyon:
Rewrote a bunch of docs for clarity. Should happen for the rest of them but I thought I'd put this up now for review.

@timliubentoml maybe to take a look?

PR #2400, opened 2022-04-07
docs(catboost): improve CatBoost documentation

kakokat:
Reworked pull request #2359 by fixing minor grammar and formatting issues.


bentoml/_internal/frameworks/catboost.py
```diff
@@ -84,37 +81,28 @@ def _load_helper(
...
-            a CatBoost model. Following parameters can be specified:
-                - model_type(:code:`str`): :obj:`classifier` (`CatBoostClassifier`) or :obj:`regressor` (`CatBoostRegressor`)
-        model_store (:mod:`~bentoml._internal.models.store.ModelStore`, default to :mod:`BentoMLContainer.model_store`):
-            BentoML modelstore, provided by DI Container.
-
+        tag (``str`` ``|`` :obj:`~bentoml.Tag`):
+            The tag of the model to get from the store.
+        model_params (:code:`Dict[str, Union[str, Any]]`, `optional`, default to :code:`None`): The parameters for loading the CatBoost model. model_type (``str``): can be one of the following:
+                * ``"classifier"`` (``CatBoostClassifier``) 
+                * ``"regressor"`` (``CatBoostRegressor``)
```
aarnphm, line 95: Hi, please make sure to follow https://github.com/bentoml/BentoML/tree/main/docs#readme

aarnphm: I prefer the Union type documentation since we are still supporting python < 3.9. I believe the way you are proposing is the 3.9 syntax?

sauyon: I think I made that change---do you think it's unclear?

aarnphm: ye

sauyon: IMO it looks nicer and I don't think it would be confusing but it can be reverted if people don't think we should use the new syntax.

sauyon: @parano @timliubentoml thoughts?

bentoml/_internal/frameworks/catboost.py
```diff
@@ -45,9 +43,8 @@
 def _get_model_info(
     tag: Tag,
     model_params: t.Optional[t.Dict[str, t.Union[str, int]]],
-    model_store: "ModelStore",
 ) -> t.Tuple["Model", str, t.Dict[str, t.Any]]:
-    model = model_store.get(tag)
+    model = bentoml.models.get(tag)
```
sauyon, line 47: This looks like a bad merge.

bentoml/_internal/frameworks/catboost.py
timliubentoml: I'm a bit bias because I'm not used to python, but I was confused by Union[x,y,z]. I initially thought it was some sort of generic declaration. pipe seems a bit nicer imo

parano:
just discussed with @sauyon , will close this PR for now due to conflicts, and will look into it again after the ongoing Runner API refactoring

PR #2410, opened 2022-04-13
feat: add parameters in flask post

jung235:
## Description

- I added `Parameters` in the POST method to use a variable URL.

Screenshot from 2022-04-14 02-10-23
Screenshot from 2022-04-14 02-18-47
Screenshot from 2022-04-14 02-13-56

## Motivation and Context

- While dealing with a massive inference job, I needed `Parameters`, especially when the job shares a dataset but depends on a few different parameters.

## How Has This Been Tested?

```python3
@env(infer_pip_packages=True)
@artifacts([SklearnModelArtifact("model")])
class BentoServiceWithParameters(BentoService):
    @api(input=DataframeInput(), route="predict/items/<string:id>", batch=True)
    def predict_dataframe(self, df: pd.DataFrame, tasks: Tuple[InferenceTask], id: str):
        print(id)  # ASDFZXCV (from the task in Description)
        ...

    @api(
        input=JsonInput(), 
        route="predict/items/<name>/categories/<int:category_id>", 
        batch=False,
    )
    def predict_json(self, parsed_json: Dict[str, str], task: InferenceTask, **kwargs):
        print(kwargs)  # {'name': 'qwer', 'category_id': 1234} (from the task in Description)
        print(type(kwargs["name"]))  # <class 'str'> (is the default if <CONVERTER>: is not assigned)
        ...
```

- After `BentoServiceWithParameters` is served, we can use `curl` command:
  ```bash
  curl -X POST \
    "http://0.0.0.0:5000/predict/items/qwer/categories/1234" \
    -H "accept: */*" \
    -H "Content-Type: application/json" \
    -d "[[5.1,3.5,1.4,0.2]]"
  ```
- I also tested the code for a different framework: `PytorchModelArtifact`.

## Checklist:

### DIFF
--- a/bentoml/server/model_app.py
+++ b/bentoml/server/model_app.py
@@ -428,12 +428,14 @@ class ModelApp:
 
             return response
 
-        def api_func_with_tracing():
+        def api_func_with_tracing(**view_kwargs):
             with self.tracer.span(
                 service_name=f"BentoService.{self.bento_service.name}",
                 span_name=f"InferenceAPI {api.name} HTTP route",
                 request_headers=request.headers,
             ):
+                if view_kwargs:
+                    api.user_func_kwargs = view_kwargs
                 return api_func()
 
         return api_func_with_tracing
--- a/bentoml/service/inference_api.py
+++ b/bentoml/service/inference_api.py
@@ -270,7 +270,9 @@ class InferenceAPI(object):
                     user_return, tasks=filtered_tasks
                 )
         else:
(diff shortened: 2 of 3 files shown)


aarnphm:
Hi @jung235 can you fix the failing CI?

parano:
@ssheng should we include this for the 0.13.3 release?

aarnphm:
I will probably close this for now due to inactivity. If one requests for such feature in 1.0 we can add it.

PR #2423, opened 2022-04-22
fix: ensure compatibility with attrs 20.1.0

sauyon:
I now have 20.1.0 in my env, so hopefully this shouldn't happen again.

We probably also want to start fixing the versions of our dependencies in the CI to the lowest ones we claim to support.

/cc @timliubentoml

### DIFF
--- a/bentoml/_internal/bento/bento.py
+++ b/bentoml/_internal/bento/bento.py
@@ -391,7 +391,7 @@ class BentoModelInfo:
         )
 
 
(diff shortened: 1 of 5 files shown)


timliubentoml:
Hooray!!

aarnphm:
@sauyon code format :(

sauyon:
Hah, how'd changing the import name from `attrs` to `attr` change the correct import order lop

aarnphm (approved):

PR #2424, opened 2022-04-22
docs: Update Development.md 

kakokat:
## Checklist:


DEVELOPMENT.md
```diff
@@ -1,72 +1,167 @@
...
+   ```bash
+   git clone https://github.com/yourusername/BentoML.git && cd BentoML
+   ```
 
-This will make `bentoml` available on your system which links to the sources of
-your local clone and pick up changes you made locally.
+4. Add the BentoML upstream remote to your local BentoML clone:
+   ```bash
+   git remote add upstream git@github.com:bentoml/BentoML.git
+   ```
```
aarnphm, line 22: I prefer not to encapsulate this under a dropdown details.

DEVELOPMENT.md
```diff
@@ -1,72 +1,167 @@
-## Install BentoML from source
+# Developer Guide
 
-Make sure to have [Git](https://git-scm.com/) and [Python3.7+](https://www.python.org/downloads/) installed.
+## Start Developing
```
aarnphm, line 3: lets revert this

DEVELOPMENT.md
```diff
@@ -1,72 +1,167 @@
...
-```
+2. Fork the BentoML project on [GitHub](https://github.com/bentoml/BentoML).
 
-Install BentoML with pip in `editable` mode:
-```bash
-pip install -e .
-```
+3. Clone the source code from your fork of BentoML's GitHub repository:
+   ```bash
+   git clone https://github.com/yourusername/BentoML.git && cd BentoML
```
aarnphm, line 16: Actually we want to keep this like before, since this sections tells users how to install bentoml from the official repository, below this section we do mention how to install from fork.

DEVELOPMENT.md
```diff
@@ -1,72 +1,167 @@
...
+5. Stage all your changes:
+    1. Enter and select 'Git: Stage All Changes...' in the command palette.
+
+6. Commit your changes:
+    1. Open the command palette and enter 'Git: Commit'.
+
+7. Push your changes:
+    1. Enter and select 'Git: Push...' in the command palette.
+
+</details>
```
aarnphm, line 164: I'm abit hesistant to add this. What is your opinion on this @parano ? I think if we provide a way to do it in the terminal then using VSCode should be pretty self-explanatory.

DEVELOPMENT.md
```diff
@@ -93,13 +188,13 @@ Flags:
   -v, --verbose         set verbose scripts
 
 
-If pytest_additional_arguments is given, this will be appended to given tests run.
+If pytest_additional_arguments is given, the additional arguments will be passed to all of the tests run by the tests script.
```
aarnphm, line 191:
```suggestion
If `pytest_additional_arguments` is given, the additional arguments will be passed to all of the tests run by the tests script.
```

DEVELOPMENT.md
```diff
@@ -295,41 +389,18 @@ BentoML also provides git hooks that developers can install with:
...
- in the main repository.
+In order for pyright to function correctly you must also run the following scripts alongside the stubs provided in the main repository.
 
-One can also clone a [copy](https://github.com/bentoml/stubs) of all dependencies stubs used by BentoML to `typings/` via:
+You can also clone a [copy](https://github.com/bentoml/stubs) of all the dependencies stubs used by BentoML to `typings/` via:
 ```bash
 # Assuming at $GIT_ROOT
 git clone git@github.com:bentoml/stubs.git
 \rm -rf typings/.git* typings/*.{sh,toml,txt,cfg,py,md}
 ```
```
aarnphm, line 401: I forgot to fix this, but I think we should remove this part. Can you help us removing this? Thank you

README.md
```diff
@@ -36,6 +36,19 @@ reporting issues, contribute to the documentation and create sample gallery proj
...
+
+1. Install custom BentoML in editable mode. e.g.:
+   * git clone your bentoml fork
+   * `pip install -e PATH_TO_THE_FORK`
+2. Set env var `export BENTOML_BUNDLE_LOCAL_BUILD=True`
+   * make sure you have the latest setuptools installed:  `pip install -U setuptools`
+3. Build a new Bento with `bentoml build` in your project directory
+4. The new Bento will include a wheel file built from the BentoML source, and 
+`bentoml containrize` will install it to override the default BentoML installation in base image
+
```
aarnphm, line 51: I don't think this section should live under the main README.md. this should stay in the DEVELOPMENT.md

DEVELOPMENT.md
```diff
@@ -295,41 +389,18 @@ BentoML also provides git hooks that developers can install with:
...
-follows the given regex : `^(feature|bugfix|improv|lib|prerelease|release|hotfix)\/[a-zA-Z0-9._-]+$`. This
-is partially based on how [Azure DevOps](https://docs.microsoft.com/en-us/azure/devops/repos/git/require-branch-folders?view=azure-devops&tabs=browser)
-manages its repos and conventional commits spec (see below for more information):
-- feature: new features/proposals users want to integrate into the library, not a new feature for a build script
-- bugfix: bugfix of a feature, not a fix to a build script
-- improv: improvements/refactor/cleanup production code, eg. reformat, pylint, etc.
-- lib: related to internal libraries, features required by the production code
-- prerelease: alpha/beta features that might should be included in the prerelease of the library. This would help testing new features/integrations for the library
-- release: included all features that is production ready
-- hotifx: patch of current bugs in production code
```
aarnphm, line 332: can you restore this part, we are putting this here since we want to follow conventional commits. Refers to [here](https://www.conventionalcommits.org/en/v1.0.0/)

aarnphm (changes requested):
Thanks for the PR. There are some changes and comments for the PR. One quick question. What is the purpose of the added vscode images under `docs/sources/_static/img`?

DEVELOPMENT.md
sauyon: Since it's a guide for developers I think the correct thing to do is to tell them to fork. I don't think it really makes sense for somebody who doesn't know what they're doing to clone the master repo, and if you do know what you're doing you probably don't need the help.

DEVELOPMENT.md
sauyon: Er, what, exactly?

README.md
sauyon: The reasoning for this is that it's not really developers who want to do this, it's probably an end user who wants to try out changes from a PR. We might want to group it with the rest of the installation instructions, though.

DEVELOPMENT.md
sauyon: My opinion on conventional commits: with our workflow they're kind of useless.

Branch hygiene is kind of meaningless because it's not our problem; we don't develop in a monolithic repo (thank goodness) so users can do branches however they like.

Because we're squash-and-merging, commits on people's branches are also kind of meaningless. Commit hygiene in general is a bit lackluster overall, and enforcing these arbitrary rules aren't actually helping. We should just make an effort to write sensible commit titles and messages when we squash-and-merge; I think that's generally the better way to handle this kind of thing anyway. We're still asking users to title their PRs in a nice way, which should get us a decent amount of the way there.

DEVELOPMENT.md
```diff
@@ -188,7 +188,7 @@ Flags:
   -v, --verbose         set verbose scripts
 
 
-If pytest_additional_arguments is given, the additional arguments will be passed to all of the tests run by the tests script.
+If 'pytest_additional_arguments' is given, the additional arguments will be passed to all of the tests run by the tests script.
```
aarnphm, line 191:
```suggestion
If `pytest_additional_arguments` is given, the additional arguments will be passed to all of the tests run by the tests script.
```

kakokat:
> Thanks for the PR. There are some changes and comments for the PR. One quick question. What is the purpose of the added vscode images under `docs/sources/_static/img`?

Oh I added some images to the Development.md just to clarify a couple steps for setting up VSCode.

DEVELOPMENT.md
parano: I like this addition, this could be very helpful for beginners

parano:
Thank for the PR @kakokat! It all look great! 

I moved around the testing section a bit here to make it easier to discover the most useful information first: https://github.com/bentoml/BentoML/pull/2424/commits/7150a928e047ac9471a705abf8cef4d8ba62e083

README.md
parano: will move this back to development.md for now

DEVELOPMENT.md
parano: that should be ok since it's for beginners only and they probably only read it once. But I can see people come back fro the rest of the docs during development

PR #2426, opened 2022-04-25
fix: consistent naming accross docker and build config

aarnphm:

## Checklist:


timliubentoml (approved):

bentoml/_internal/bento/build_config.py
```diff
@@ -27,7 +27,7 @@
 DOCKER_SUPPORTED_DISTROS: t.List[str] = [
     "debian",
     "amazonlinux2",
-    "alpine",
```
parano, line 30: should we just use alpine in both places?  and alpine 3.15 already have stable release, any reason we don't use that?

aarnphm: we can support `alpine3.15`, but we can explicitly ask user to add that? so this would be `alpine3.14` ?

aarnphm: also right now we just have alpine3.14 as tags for our docker images.

PR #2432, opened 2022-04-27
refactor: runner/runnable interface

bojiang:
## Checklist:

### DIFF
--- a/bentoml/_internal/runner/resource.py
+++ b/bentoml/_internal/runner/resource.py
@@ -1,0 +1,217 @@
+from __future__ import annotations
+
+import os
+import re
+import math
+import typing as t
+import logging
+import functools
+
+import attr
+import psutil
+
+logger = logging.getLogger(__name__)
+
+
+def cpu_converter(cpu: t.Union[int, float, str]) -> float:
+    assert isinstance(cpu, (int, float, str)), "cpu must be int, float or str"
+
+    if isinstance(cpu, (int, float)):
+        return float(cpu)
+
+    milli_match = re.match("([0-9]+)m", cpu)
+    if milli_match:
+        return float(milli_match[1]) / 1000.0
+    raise BentoMLException(f"Invalid CPU resource limit '{cpu}'. ")
+
+
+def mem_converter(mem: t.Union[int, str]) -> int:
+    assert isinstance(mem, (int, str)), "mem must be int or str"
+
+    if isinstance(mem, int):
+        return mem
+
+    unit_match = re.match("([0-9]+)([A-Za-z]{1,2})", mem)
+    mem_multipliers = {
+        "k": 1000,
+        "M": 1000 ** 2,
+        "G": 1000 ** 3,
+        "T": 1000 ** 4,
+        "P": 1000 ** 5,
+        "E": 1000 ** 6,
+        "Ki": 1024,
+        "Mi": 1024 ** 2,
+        "Gi": 1024 ** 3,
+        "Ti": 1024 ** 4,
+        "Pi": 1024 ** 5,
+        "Ei": 1024 ** 6,
+    }
+    if unit_match:
+        base = int(unit_match[1])
+        unit = unit_match[2]
+        if unit in mem_multipliers:
+            return base * mem_multipliers[unit]
+    raise ValueError(f"Invalid MEM resource limit '{mem}'")
+
+
+@functools.lru_cache(maxsize=1)
+def query_cgroup_cpu_count() -> float:
+    # Query active cpu processor count using cgroup v1 API, based on OpenJDK
+    # implementation for `active_processor_count` using cgroup v1:
+    # https://github.com/openjdk/jdk/blob/master/src/hotspot/os/linux/cgroupSubsystem_linux.cpp
+    # For cgroup v2, see:
+    # https://github.com/openjdk/jdk/blob/master/src/hotspot/os/linux/cgroupV2Subsystem_linux.cpp
+    # Possible supports: cpuset.cpus on kubernetes
+    def _read_cgroup_file(filename: str) -> float:
+        with open(filename, "r", encoding="utf-8") as f:
+            return int(f.read().strip())
+
+    cgroup_root = "/sys/fs/cgroup/"
+    cfs_quota_us_file = os.path.join(cgroup_root, "cpu", "cpu.cfs_quota_us")
+    cfs_period_us_file = os.path.join(cgroup_root, "cpu", "cpu.cfs_period_us")
+    shares_file = os.path.join(cgroup_root, "cpu", "cpu.shares")
+    cpu_max_file = os.path.join(cgroup_root, "cpu.max")
+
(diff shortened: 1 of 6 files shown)


bentoml/_internal/runner/runnable.py
```diff
@@ -0,0 +1,124 @@
...
+import typing as t
+import logging
+import functools
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    InputType = t.TypeVar("InputType")
+    OutputType = t.TypeVar("OutputType")
+
+    WrappedMethod = t.TypeVar("WrappedMethod", bound=t.Callable[..., t.Any])
```
aarnphm, line 13:
```suggestion
    P = t.ParamSpec("P")
    
    WrappedMethod = t.TypeVar("WrappedMethod", bound=t.Callable[P, t.Any])
```

bojiang: they are different

bojiang: Here I just want a methed

aarnphm: i see

bentoml/_internal/runner/runner.py
```diff
@@ -183,238 +140,179 @@ def get_runnable_methods(cls) -> t.List[RunnableMethod]:
...
-    @property
-    def resource_config(self):
-        # look up configuration
-        # else return self._resource_config
-        ...
+    def destroy_local(self):
+        if not self._runnable:
+            logger.warning("local runnable not found")
+        else:
+            del self._runnable
```
ssheng, line 186: Is `destroy_local()` necessary? Do we expect the `RunnerHandle` class to be re-init? Same for `destroy_client()` below.

bentoml/_internal/runner/runner.py
```diff
@@ -183,238 +140,179 @@ def get_runnable_methods(cls) -> t.List[RunnableMethod]:
...
 
-    def _run(self, runner_method_name, *args, **kwargs):
-        if self._runner_app_client:
-            return self._runner_app_client.run(runner_method_name, *args, **kwargs)
+    def init_client(self, runner: Runner):
```
ssheng, line 188: How do we like `init_remote()`?

bentoml/_internal/runner/runner.py
```diff
@@ -183,238 +140,179 @@ def get_runnable_methods(cls) -> t.List[RunnableMethod]:
...
+    def __init__(
+        self,
+        runnable_class: t.Type[Runnable],
+        init_params: t.Dict[str, t.Any],
+        name: str,
+        strategy: t.Type[Strategy],
+        models: t.List[Model],
+        cpu: int,
+        nvidia_gpu: int,
+        custom_resources: t.Dict[str, int | float],
```
ssheng, line 258: Would it be cleaner if we ask users to pass a `Resource` instance. The `Resource` instance can be overridden to add custom resources as properties. The custom `Strategy` will know how to consume the custom `Resource)

bentoml/_internal/runner/runner.py
```diff
@@ -183,238 +140,179 @@ def get_runnable_methods(cls) -> t.List[RunnableMethod]:
...
             "runner not initialized"
         )  # TODO: make this UninitializedRunnerException
 
+
+@attr.define(frozen=True)
+class Runner:
+    runnable_class: t.Type[Runnable] = attr.field()
+    init_params: t.Dict[str, t.Any] = attr.field()
+    name: str = attr.field()
+    strategy: t.Type[Strategy] = attr.field(default=DefaultStrategy)
```
ssheng, line 243: Maybe either `schedule_strategy` or `scheduling_strategy`. Strategy is maybe too generic.

bentoml/_internal/server/schedule_strategy.py
```diff
@@ -0,0 +1,16 @@
...
+from typing import TYPE_CHECKING
+from contextvars import ContextVar
+
+if TYPE_CHECKING:
+    from .. import external_typing as ext
+
+import abc
+
+
+class SchedulingStrategy(abc.ABC):
```
ssheng, line 12: What's this for?

bentoml/_internal/runner/runnable.py
```diff
@@ -0,0 +1,88 @@
...
+            input_spec=input_spec,
+            output_spec=output_spec,
+        ),
+    )
+    return meth
+
+
+class Runnable(abc.ABC):
+    SUPPORT_NVIDIA_GPU: bool
+    SUPPORT_MULTIPLE_CPU_THREADS: bool
```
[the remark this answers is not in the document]
parano: will address in a follow up PR

bentoml/_internal/runner/runner.py
parano: it will be used in the RunnerApp only, not meant to be invoke by users

bentoml/_internal/runner/runner.py
parano: sounds good, this will be a internal method as well

bentoml/_internal/runner/runner.py
parano: It means user always need to import an extra class for passing in resource config, I feel passing in the parameters directly is nicer from user perspective. Resource class is nice for internal use and validation. We took a similar approach in Model and BuildConfig as well

bentoml/_internal/server/schedule_strategy.py
parano: The is the base class for extending custom scheduling strategy

bentoml/_internal/runner/runner.py
bojiang: I've thinking abt that. `init_remote` sounds like will launch a remote runner server, which is not what is happening.

bentoml/_internal/runner/runner.py
ssheng: Passing `Resource` may give better extensibility in the future. Say when we add `another_type_gpu`, we won't need to update the signature here. We can also get rid of the `custom_resources` field because a custom implementation can add any additional fields that are only needed to be understand by their custom strategies.

bentoml/_internal/runner/resource.py
```diff
@@ -0,0 +1,217 @@
...
+    elif quota:
+        limit_count = quota
+    elif shares:
+        limit_count = shares
+
+    return float(min(limit_count, os_cpu_count))
+
+
+def query_os_cpu_count() -> int:
+    cpu_count = os.cpu_count()
```
rehevkor5, line 123: Shouldn't this (and above) use `len(os.sched_getaffinity(0))` instead?

aarnphm: hey If you have an suggestion please make a new issue for this instead of commenting to older PR.

sauyon: Feel free to make a PR for this!

PR #2434, opened 2022-04-27
framework template

sauyon:
Somewhat non-final framework template, committed so work can (maybe) begin on framework impl.

Somewhat subject to change.


aarnphm:
are we going to generate new frameworks file from this templates file? or this would just be there?

parano:
Template updated in PR https://github.com/bentoml/BentoML/pull/2439

parano:
> are we going to generate new frameworks file from this templates file? or this would just be there?

No, it's just an example

PR #2437, opened 2022-04-28
model.py update

sauyon:


parano:
Changes included here https://github.com/bentoml/BentoML/pull/2439/files

PR #2438, opened 2022-04-28
runnablehandle proposal

sauyon:
I think this is a little cleaner than what we had before.

### DIFF
--- a/bentoml/_internal/runner/runner.py
+++ b/bentoml/_internal/runner/runner.py
@@ -2,7 +2,8 @@ from __future__ import annotations
 
 import typing as t
 import logging
-import collections
+from abc import ABC
+from abc import abstractmethod
 from typing import TYPE_CHECKING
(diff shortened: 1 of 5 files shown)


bentoml/_internal/runner/runner.py
```diff
@@ -279,22 +275,37 @@ async def async_run_method(
...
-        self._runner_handle.init_local(self)
+        logger.warning("for debugging and testing only")  # if not called from RunnerApp
+        if self._runner_handle is not None:
+            raise BentoMLException(
+                "Runner already initialized"
+            )  # TODO: better exception type (InitializedRunnerException?)
 
-    def destroy_local(self):
-        self._runner_handle.destroy_local()
+        self._set_runner_handle(LocalRunner(self.runnable_class))
```
parano, line 297: also need the init Params here?

PR #2439, opened 2022-04-28
Runnable refactors and Model info update

sauyon:
Did a bunch of random stuff, all the changes not nicely packaged in PR are in the last commit.

This should probably supersede the template PR since I added more stuff to that...

### DIFF
--- a/bentoml/_internal/frameworks/template.py
+++ b/bentoml/_internal/frameworks/template.py
@@ -1,0 +1,139 @@
+from __future__ import annotations
+
+import typing as t
+from typing import TYPE_CHECKING
+
+import attr
+
+import bentoml
+from bentoml import Tag
+from bentoml import __version__ as BENTOML_VERSION
+from bentoml.models import ModelContext
+from bentoml.models import ModelOptions
+
+if TYPE_CHECKING:
+    from bentoml.types import ModelSignatureDict
+
+    ModelType = ...
+
+MODULE_NAME = "bentoml.MY_MODULE"
+
+
+@attr.define(frozen=True)
+class FrameworkOptions(ModelOptions):
+    pass
+
+
+def load_model(
+    bento_model: str | Tag | bentoml.Model,
+    *,
+    ...
+) -> FrameworkModelType:
+    """
+    Load the <FRAMEWORK> model with the given tag from the local BentoML model store.
+
+    Args:
+        bento_model (``str`` ``|`` :obj:`~bentoml.Tag` ``|`` :obj:`~bentoml.Model`):
+            Either the tag of the model to get from the store, or a BentoML `~bentoml.Model`
+            instance to load the model from.
+        ...
+    Returns:
+        <MODELTYPE>:
+            The <FRAMEWORK> model loaded from the model store or BentoML :obj:`~bentoml.Model`.
+    Example:
+    .. code-block:: python
+        import bentoml
+        <LOAD EXAMPLE>
+    """  # noqa
+    if not isinstance(bento_model, bentoml.Model):
+        bento_model = bentoml.models.get(bento_model)
+
+    ...
+
+
+def save_model(
+    name: str,
+    model: FrameworkModelType,
+    *,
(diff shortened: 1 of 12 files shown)


bentoml/_internal/runner/runner.py
```diff
@@ -99,7 +100,7 @@ async def async_run(self, *args: t.Any, **kwargs: t.Any) -> t.Any:
 
 
 @attr.define
-class RunnerHandle:
+class RunnerHandle(ABC):
```
bojiang, line 103: lol, so you take the `RunnerImpl` back with another name. I write the current because my previous implementation (RunnerImpl) is disliked.

bentoml/_internal/runner/runnable.py
```diff
@@ -13,60 +14,74 @@
...
-            batch_dim=batch_dim,
-            input_spec=input_spec,
-            output_spec=output_spec,
-        ),
-    )
-    return meth
+class Runnable(ABC):
+    @property
+    @abstractmethod
+    def supports_nvidia_gpu(self) -> bool:
```
bojiang, line 28: This can not be a property because we need this before init the runnable.

bojiang (changes requested):
SUPPORT_NVIDIA_GPU cannot be a property

bentoml/_internal/runner/runnable.py
sauyon: Ah, I meant for these to be static but apparently that's not possible, will revert.

bentoml/_internal/runner/runner.py
sauyon: Oh, why?

bentoml/_internal/frameworks/template.py
```diff
@@ -0,0 +1,139 @@
...
+        custom_objects=custom_objects,
+        metadata=metadata,
+        context=context,
+    ) as bentoml_model:
+        ...
+
+        return bentoml_model.tag
+
+
+def _to_runnable(
```
parano, line 106: This should be get_runnable?

sauyon: Still wasn't sure exactly what we had decided on and saw larme's PR had _to_runnable, I can change this though.

parano: `model.py` still uses `get_runnable`, l talked to Larme to change this to use `get_runnable` as well, let's also update that in the framework template

parano, line 106:
```suggestion
def get_runnable(
```

bentoml/_internal/frameworks/template.py
```diff
@@ -0,0 +1,139 @@
...
+                bento_model,
+                # bento_model.info.options.*
+            )
+
+    for method_name, options in bento_model.info.signatures.items():
+
+        def _run(input_data) -> output_data:
+            ...
+
+        FrameworkRunnable.add_method(
```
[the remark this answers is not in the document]
parano: cc @larme

bentoml/_internal/runner/runnable.py
```diff
@@ -53,48 +36,54 @@ def supports_nvidia_gpu(self) -> bool:
...
+                batch_dim=batch_dim,
+                input_spec=input_spec,
+                output_spec=output_spec,
+            ),
+        )
 
     @staticmethod
     def method(
-        batchable_or_method: bool | WrappedMethod = False,  # type: ignore (pyright bug?)
+        meth: WrappedMethod | None = None,
```
parano, line 64: user who's using `Runnable.method` as a decorator may get confused by this parameter?  I will try to refactor the to_method to get rid of this parameter here

previously with the `batchable_or_method`, it was trying to support using the decorator without any parameters, so both of the following works:

```
@Runnable.method
def __call__():
   ...

@Runnable.method()
def __call__():
   ...
```

parano, line 64: nice refactoring to get rid of the `batchable_or_method` 👍

PR #2440, opened 2022-04-29
Runners refactor: service & bento build changes

parano:
## Checklist:


bentoml/_internal/bento/bento.py
```diff
@@ -30,6 +30,7 @@
 from .build_config import BentoBuildConfig
 from ..configuration import BENTOML_VERSION
 from ..configuration.containers import BentoMLContainer
+from ..runner.resource import Resource
```
aarnphm, line 33: should we do this under `TYPE_CHECKING` and `from __future__ import annotations`?

bentoml/_internal/service/service.py
```diff
@@ -161,6 +219,7 @@ def mount_asgi_app(
     def mount_wsgi_app(
         self, app: WSGI_APP, path: str = "/", name: t.Optional[str] = None
     ) -> None:
+        # TODO: Migrate to a2wsgi
```
aarnphm, line 222: https://github.com/abersheeran/a2wsgi is this the library @parano ?

tests/unit/_internal/bento/test_bento.py
```diff
@@ -81,10 +81,9 @@ def test_bento_info(tmpdir: "Path"):
         assert bentoinfo_b_from_yaml == bentoinfo_b
 
 
-def build_test_bento(model_store: "ModelStore") -> Bento:
+def build_test_bento() -> Bento:
```
aarnphm, line 84: This `model_store` is actually a fixture under `conftest.py` to create a model_store under pytest tmp test directory

bentoml/_internal/service/service.py
bojiang: I wonder why do we do that

parano: @bojiang it's a warning from starlette that it's deprecating this API

bojiang: That lib is not using anyio. I thought we should do more investigation. or even wrote ourselves

sauyon (commented):
Very minor comments.

bentoml/_internal/service/loader.py
```diff
@@ -11,14 +13,16 @@
 
 from ..bento import Bento
 from ..models import ModelStore
-from ...exceptions import NotFound
+from ...exceptions import NotFound, ImportServiceError
```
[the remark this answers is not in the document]
parano: looks like it's because template.py failed in previous step so isort was not executed, just renamed the template file

bentoml/_internal/tag.py
```diff
@@ -40,7 +40,6 @@ def validate_tag_str(value: str):
         raise InvalidArgument(f"{value} is not a valid tag: " + ", and ".join(errors))
 
 
-@attr.define
```
parano, line 43: ops, will add this back

PR #2446, opened 2022-05-03
Fix ModelInfo cattrs serialization issue 

parano:
## Checklist:


bentoml/_internal/models/model.py
```diff
@@ -318,25 +328,30 @@ def __init__(self, base_path: "t.Union[PathType, FS]"):
...
 
     @staticmethod
-    def from_dict(data: dict[str, str | dict[str, str]]) -> ModelContext:
-        return cattrs.structure(data, ModelContext)
+    def from_dict(data: dict[str, str | dict[str, str]] | ModelContext) -> ModelContext:
+        if isinstance(data, ModelContext):
+            return data
+        return bentoml_cattr.structure(data, ModelContext)
+
+attr.resolve_types(ModelContext, globals(), locals())
```
parano, line 341: @sauyon looks like this is required for cattrs to work, when using `from __future__ import annotations`, see https://www.attrs.org/en/stable/types.html

sauyon: Ah, got it! I think according to the cattrs people `cattrs.Converter()` should also be able to structure these types.

sauyon: Though probably for our types we want to do this anyway.

parano: @sauyon cattrs.Convertor doesn't have a resolve_types method, it seems to be a attrs behavior that cattrs did not expect

parano:
Besides using `attr.resolve_types`, attrs & cattrs doesn't seem to work with type annotation `dict`, had to change all the `attrs` class fields to use `t.Dict` instead. @sauyon

sauyon:
> Besides using `attr.resolve_types`, attrs & cattrs doesn't seem to work with type annotation `dict`, had to change all the `attrs` class fields to use `t.Dict` instead. @sauyon 

Hmm, that's probably the best solution until we drop support for 3.7/3.8 😛

bentoml/_internal/bento/bento.py
```diff
@@ -463,14 +474,18 @@ def validate(self):
...
 bentoml_cattr.register_unstructure_hook(
     BentoInfo,
     # Ignore tag, tag is saved via the name and version field
-    make_dict_unstructure_fn(BentoInfo, bentoml_cattr, tag=override(omit=True)),
+    make_dict_unstructure_fn(BentoInfo, bentoml_cattr, tag=override(omit=True)),  # type: ignore
 )
 
 
 def _BentoInfo_dumper(dumper: yaml.Dumper, info: BentoInfo) -> yaml.Node:
+    print(info.to_dict())
```
parano, line 488:
```suggestion
```

PR #2459, opened 2022-05-11
feat(xgboost): 1.0 XGBoost implementation

sauyon:


bentoml/_internal/frameworks/xgboost.py
```diff
@@ -162,120 +123,104 @@ def save(
...
-        predict_fn_name=predict_fn_name,
-        booster_params=booster_params,
-        name=name,
-    )
+    Private API: use :obj:`~bentoml.Model.to_runnable` instead.
+    """
+
+    class XGBoostRunnable(bentoml.Runnable):
+        SUPPORT_NVIDIA_GPU = True
+        SUPPORT_CPU_MULTI_THREADING = True
```
sauyon, line 187: I think we should do something like

```python
SUPPORTED_RESOURCES = set(bentoml.resources.NVIDIA_GPU, bentoml.resources.MULTIPLE_THREADS)
```

or maybe

```python
SUPPORTED_RESOURCES = [bentoml.resources.NVIDIA_GPU]
SUPPORTS_CPU_MULTI_THREADING = True
```

so that this is more generic.

bentoml/_internal/frameworks/xgboost.py
```diff
@@ -162,120 +123,104 @@ def save(
...
+
+        def __init__(self):
+            super().__init__()
+            self.model = load_model(bento_model)
+
+            # check for resources
+            available_gpus = os.getenv("NVIDIA_VISIBLE_DEVICES")
+            if available_gpus is not None and available_gpus != "":
```
sauyon, line 195: I think this generally doesn't look that nice; with custom resources in the future it might be nicer to pass a `Resource` object to the constructor or via context vars instead.

bentoml/_internal/frameworks/xgboost.py
```diff
@@ -31,6 +29,7 @@
     )
 
 MODULE_NAME = "bentoml.xgboost"
+MODEL_FILENAME = "model.ubj"
```
sauyon, line 32: I think we should probably do this instead of a global `MODEL_NAMESPACE`; we should never change this and I think it improves code readability to not combine two constants from another file here.

bentoml/_internal/exportable.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import os
```
aarnphm, line 3: should we just do 
```
# type: ignore
```
at the beginning of this file and remove all of the inline type ignore?

sauyon: The rest of the type checking is still working fine, I don't see much reason to do this.

bentoml/_internal/frameworks/xgboost.py
parano: I agree, let's remove the global model namespace

bentoml/_internal/frameworks/xgboost.py
parano: sounds good, let's do it this way for now, we can refactor it with something like context vars later

bentoml/_internal/frameworks/xgboost.py
```diff
@@ -162,120 +123,108 @@ def save(
...
-    predict_fn_name: str = "predict",
-    *,
-    booster_params: t.Optional[t.Dict[str, t.Union[str, int]]] = None,
-    name: t.Optional[str] = None,
-) -> "_XgBoostRunner":
     """
-    Runner represents a unit of serving logic that can be scaled horizontally to
-    maximize throughput. `bentoml.xgboost.load_runner` implements a Runner class that
-    wrap around a Xgboost booster model, which optimize it for the BentoML runtime.
+    Save a model instance to BentoML modelstore.
```
parano, line 151: I'm a bit confused by the docstring here, this part should probably go to the `save_model` docstring above?

bentoml/_internal/frameworks/xgboost.py
```diff
@@ -162,120 +123,108 @@ def save(
...
-        "pip_dependencies": [f"xgboost=={get_pkg_version('xgboost')}"],
-    }
+    """  # noqa: LN001
+    context: ModelContext = ModelContext(
+        framework_name="xgboost",
+        framework_versions={"xgboost": get_pkg_version("xgboost")},
+    )
+
+    if signatures is None:
+        signatures = {
```
parano, line 133: could we add a `logger.info` statement here, similar to the sklearn module did?

bentoml/_internal/frameworks/xgboost.py
```diff
@@ -31,6 +30,7 @@
     )
 
 MODULE_NAME = "bentoml.xgboost"
+MODEL_FILENAME = "model.ubj"
 
 # TODO: support xgb.DMatrix runner io container
```
parano, line 35: we can remove this part now, the interface has been changed

bentoml/_internal/frameworks/xgboost.py
ssheng: Users should have the flexibility to support resources not listed under `bentoml.resources` with a custom strategy. So we might just keep them as strings.

bentoml/_internal/frameworks/xgboost.py
```diff
@@ -31,123 +31,86 @@
...
-            BentoML modelstore, provided by DI Container.
-
+        name (``str``):
+            The name to give to the model in the BentoML store. This must be a valid
+            :obj:`~bentoml.Tag` name.
+        model (:obj:`~xgboost.core.Booster`):
+            The XGBoost model to be saved.
+        signatures (``dict[str, ModelSignatureDict]``, optional):
+            Signatures of predict methods to be used. If not provided, the signatures default to 
+            ``{"predict": {"batchable": False}}``. See :obj:`~bentoml.types.ModelSignature` for more
```
parano, line 93: looks like we currently don't expose this type in bentoml.types?

sauyon: I think I fixed it in my other PR. Probably should put this one on top of it.

bentoml/_internal/frameworks/xgboost.py
```diff
@@ -162,120 +125,84 @@ def save(
...
-    }
+    """  # noqa: LN001
+    context: ModelContext = ModelContext(
+        framework_name="xgboost",
+        framework_versions={"xgboost": get_pkg_version("xgboost")},
+    )
+
+    if signatures is None:
+        logger.info(
+            'Using default model signature `{"predict": {"batchable": False}}` for sklearn model'
```
parano, line 136:
```suggestion
            'Using default model signature `{"predict": {"batchable": False}}` for xgboost model'
```

bentoml/_internal/frameworks/xgboost.py
```diff
@@ -162,120 +125,84 @@ def save(
...
+    for method_name, options in bento_model.info.signatures.items():
+
+        def _run(
+            self: XGBoostRunnable,
+            input_data: ext.NpNDArray | ext.PdDataFrame | xgb.DMatrix,
+        ) -> ext.NpNDArray:
+            if not isinstance(input_data, xgb.DMatrix):
+                input_data = xgb.DMatrix(input_data)
+
+            res = self.predict_fns[method_name](input_data)
```
parano, line 196: @larme found an issue with this implementation, where as `method_name` is always the last item in the list, due to it being bound to the loop scope variable. We can do something like `predict_fn = self.predict_fns[method_name]` to the beginning of this for loop, to avoid this issue. @sauyon could you help fix that in a follow up PR?

sauyon: Oh, python captures by reference 🤦. Well, some more stuff that should be going into my integration tests...

PR #2460, opened 2022-05-11
chore(internal): fix typing

sauyon:
Didn't bother separating out some actual changes:
 - new docs for `ModelSignature`
 - removed `flushed` from `Model` and added a manual flush call inside of `bentoml.models.create`

We should probably start documenting the methods of `Model` now that at least some of them are user-facing.

### DIFF
--- a/bentoml/_internal/models/model.py
+++ b/bentoml/_internal/models/model.py
@@ -14,10 +14,10 @@ import attr
 import yaml
 import fs.errors
 import fs.mirror
-import cloudpickle
+import cloudpickle  # type: ignore (no cloudpickle types)
 from fs.base import FS
-from cattr.gen import override
-from cattr.gen import make_dict_unstructure_fn
+from cattr.gen import override  # type: ignore (incomplete cattr types)
+from cattr.gen import make_dict_unstructure_fn  # type: ignore (incomplete cattr types)
 from simple_di import inject
 from simple_di import Provide
 
@@ -45,6 +45,8 @@ if TYPE_CHECKING:
     ]
 
 
+T = t.TypeVar("T")
+
 logger = logging.getLogger(__name__)
 
 PYTHON_VERSION: str = f"{pyver.major}.{pyver.minor}.{pyver.micro}"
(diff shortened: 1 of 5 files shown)


bentoml/_internal/models/model.py
```diff
@@ -352,6 +392,36 @@ def from_dict(data: dict[str, str | dict[str, str]] | ModelContext) -> ModelCont
 
 @attr.frozen
 class ModelSignature:
+    """
+    A model signature represents a method on a model object that can be called.
+
+    Note that anywhere a ``ModelSignature`` is used, a ``dict`` with keys corresponding to the fields
```
parano, line 398: great job documenting this 👍

bentoml/_internal/models/model.py
```diff
@@ -352,6 +392,36 @@ def from_dict(data: dict[str, str | dict[str, str]] | ModelContext) -> ModelCont
...
+    can be used instead. For example, instead of ``{"predict": ModelSignature(batchable=True)}``, one
+    can pass ``{"predict": {"batchable": True}}``.
+
+    Fields:
+        batchable:
+            Whether the method should support batching. Note that batching must also
+            be explicitly enabled at runtime in the ``Model.to_runner`` or ``bentoml.runner`` call.
+        batch_dim:
+            The dimension(s) that contain multiple data. For example, if when passing two input arrays
+            ``[1, 2]`` and ``[5, 6]``, if the input array is ``[[1, 2], [3, 4]]``, then the batch dimension
```
parano, line 408:
```suggestion
            ``[1, 2]`` and ``[3, 4]``, if the input array is ``[[1, 2], [3, 4]]``, then the batch dimension
```

bentoml/_internal/models/model.py
```diff
@@ -352,6 +392,36 @@ def from_dict(data: dict[str, str | dict[str, str]] | ModelContext) -> ModelCont
...
 @attr.frozen
 class ModelSignature:
+    """
+    A model signature represents a method on a model object that can be called.
+
+    Note that anywhere a ``ModelSignature`` is used, a ``dict`` with keys corresponding to the fields
+    can be used instead. For example, instead of ``{"predict": ModelSignature(batchable=True)}``, one
+    can pass ``{"predict": {"batchable": True}}``.
+
+    Fields:
```
parano, line 402: Should this also mention that batchable and batch_dim are a `bentoml.Runnable` concept, and this is used for converting models to Runnable/Runner in BentoML?

bentoml/_internal/models/model.py
```diff
@@ -352,6 +392,36 @@ def from_dict(data: dict[str, str | dict[str, str]] | ModelContext) -> ModelCont
...
+        batchable:
+            Whether the method should support batching. Note that batching must also
+            be explicitly enabled at runtime in the ``Model.to_runner`` or ``bentoml.runner`` call.
+        batch_dim:
+            The dimension(s) that contain multiple data. For example, if when passing two input arrays
```
parano, line 407: Maybe add a bit language that's familiar to data scientist here? something like:

`batch_dim` represents the batch dimension in a multidimensional array(numpy.ndarray or tf.Tensor). BentoML uses this information to do dynamic batching of input data.

For example, with `batch_dim=0`, two separate input array `[[1,2]]` and `[[3,4]]` will be grouped along axis 0, into a larger batch input `[[1,2], [3,4]]`.  And similarly with `batch_dim=1`, the two separate input array `[[1], [2]]` and `[[3], [4]]` will be grouped as `[[1,3], [2,4]]`.

bentoml/_internal/models/model.py
parano: oh since we are recommending the `batch-to-batch` style API now, this may not be a great example. updated it slightly in the other comment.

sauyon: I think it's more explainable if you just use two arrays that you want to do prediction on and we talk about the input to the call, but I should make more clear what's going on, I'll rewrite this.

bentoml/_internal/models/model.py
```diff
@@ -66,25 +68,56 @@ def with_options(cls, **kwargs: t.Any) -> ModelOptions:
...
 
     _runnable: Runnable | None = None
 
+    def __init__(
+        self,
+        tag: Tag,
+        fs: FS,
+        info: ModelInfo,
+        custom_objects: dict[str, t.Any] | None = None,
+        runnable: Runnable | None = None,
```
parano, line 98: Is this meant for a reserved API to customize runnable for a model?

sauyon: Ah, this is meant to be the internal way to construct a model; I wanted to be able to set all the fields with a single init call.

If the `_internal` argument is not set it will error and inform the user they should use `bentoml.models.create` or `bentoml.framework.save`.

PR #2463, opened 2022-05-13
docs(model): update ModelSignature documentation

sauyon:


bentoml/_internal/models/model.py
```diff
@@ -395,29 +397,56 @@ class ModelSignature:
...
+            pass to the predict method would be ``[[1, 3], [2, 4]]``, then the batch dimension would
+            be ``1``.
+
+            If there are multiple arguments to the predict method and there is only one batch
+            dimension supplied, all arguments will use that batch dimension.
+
+            Example:
+            .. code-block:: python
+                # create a runner from 'my_bento_model' which batches on the dimension 0
+                runner0 = my_bento_model.to_runner(signatures={"predict": {"batchable": True, "batch_dim": 0}})
```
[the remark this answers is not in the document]
sauyon: Seems like we may as well, though we probably shouldn't use it in examples, now that I think about it.

parano, line 425:
```suggestion
              bentoml.pytorch.save_model("demo", model, signatures={"predict": {"batchable": True, "batch_dim": 0}})
```

bentoml/_internal/models/model.py
```diff
@@ -395,29 +397,56 @@ class ModelSignature:
...
+
+                # if the following calls are batched, the input to the actual predict method on the
+                # runner would be [[1, 2], [3, 4], [5, 6]]
+                runner0.run(np.array([[1, 2], [3, 4]]))
+                runner0.run(np.array([[5, 6]]))
+
+                # if the following calls are batched, the input to the actual predict method on the
+                # runner would be [[1, 2, 5], [3, 4, 6]]
+                runner1.run(np.array([[1, 2], [3, 4]]))
+                runner1.run(np.array([[5], [6]]))
```
parano, line 437:
```suggestion
                # Save two models with `predict` method that supports taking input batches on the dimension 0 and the other on dimension 1:
                bentoml.pytorch.save_model("demo0", model_0, signatures={"predict": {"batchable": True, "batch_dim": 0}})
                bentoml.pytorch.save_model("demo1", model_1, signatures={"predict": {"batchable": True, "batch_dim": 1}})

                # if the following calls are batched, the input to the actual predict method on the
                # model.predict method would be [[1, 2], [3, 4], [5, 6]]
                runner0 = bentoml.pytorch.get("demo0:latest").to_runner()
                runner0.init_local()
                runner0.predict.run(np.array([[1, 2], [3, 4]]))
                runner0.predict.run(np.array([[5, 6]]))

                # if the following calls are batched, the input to the actual predict method on the
                # model.predict would be [[1, 2, 5], [3, 4, 6]]
                runner1 = bentoml.pytorch.get("demo1:latest").to_runner()
                runner1.init_local()
                runner1.predict.run(np.array([[1, 2], [3, 4]]))
                runner1.predict.run(np.array([[5], [6]]))
```

PR #2469, opened 2022-05-17
fix(runner): fix bugs in runner batching

sauyon:
Now properly deals with edge cases, I think... I want to factor this out into its own module so we can reuse it, I think that would make sense. (Also we could write unit tests for the logic...)

### DIFF
--- a/bentoml/_internal/runner/runnable.py
+++ b/bentoml/_internal/runner/runnable.py
@@ -135,7 +135,9 @@ class BatchDimMapping(BatchDimSuper):
 
         self.args = []
         self.kwargs = {}
-        for i, (name, param) in enumerate(params.items()):
+        # drop self parameter from params
+        param_items = list(params.items())[1:]
+        for i, (name, param) in enumerate(param_items):
             if param.kind in [
                 inspect.Parameter.POSITIONAL_ONLY,
                 inspect.Parameter.POSITIONAL_OR_KEYWORD,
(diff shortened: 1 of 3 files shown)


bentoml/_internal/runner/runnable.py
```diff
@@ -135,7 +135,9 @@ def get_batch_dim(idx: int) -> int:
 
         self.args = []
         self.kwargs = {}
-        for i, (name, param) in enumerate(params.items()):
+        # drop self parameter from params
+        param_items = list(params.items())[1:]
```
bojiang, line 139: not all the cases have self as the first arg

bentoml/_internal/runner/runner_handle/local.py
```diff
@@ -16,15 +17,17 @@ class LocalRunnerRef(RunnerHandle):
...
-    def run_method(self, method_name: str, *args: t.Any, **kwargs: t.Any) -> t.Any:
-        return getattr(self._runnable, method_name)(*args, **kwargs)
+    def run_method(
+        self, __bentoml_method_name: str, *args: t.Any, **kwargs: t.Any
+    ) -> t.Any:
+        return getattr(self._runnable, __bentoml_method_name)(*args, **kwargs)
 
     async def async_run_method(
-        self, method_name: str, *args: t.Any, **kwargs: t.Any
+        self, __bentoml_method_name: str, *args: t.Any, **kwargs: t.Any
```
bojiang, line 26: 👍

bentoml/_internal/runner/runnable.py
sauyon: I think it's ok to not support `staticmethod` here, though we may want to have better error messages for those cases.

PR #2476, opened 2022-05-18
test(runner): update DataContainer tests

larme:
## Checklist:

### DIFF
--- a/bentoml/_internal/runner/container.py
+++ b/bentoml/_internal/runner/container.py
@@ -14,7 +14,7 @@ from ..configuration.containers import DeploymentContainer
 
 SingleType = t.TypeVar("SingleType")
 BatchType = t.TypeVar("BatchType")
-BatchIndicesType = t.Sequence[int]
+BatchIndicesType = list[int]
 
 if TYPE_CHECKING:
     from .. import external_typing as ext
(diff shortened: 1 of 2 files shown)


larme:
@parano `DefaultContainer` is the default container we use in `AutoContainer` so I'm hesitated to rename it to `SequenceContainer`, other comments in last pr are addressed

bentoml/_internal/runner/container.py
```diff
@@ -172,7 +172,7 @@ def from_batch_payloads(  # pylint: disable=arguments-differ
         payloads: t.Sequence[Payload],
         batch_dim: int = 0,
         plasma_db: "ext.PlasmaClient" = Provide[DeploymentContainer.plasma_db],
-    ) -> t.Tuple["ext.NpNDArray", t.List[int]]:
+    ) -> t.Tuple["ext.NpNDArray", BatchIndicesType]:
```
bojiang, line 175: afaik, for params, the type annotation should be as general as possible. And for the return values, annotations should be as concrete as possible. Thus it's okay to keep this list[int] since it does return list[int]

bentoml/_internal/runner/container.py
```diff
@@ -14,7 +14,7 @@
 
 SingleType = t.TypeVar("SingleType")
 BatchType = t.TypeVar("BatchType")
-IndexType = t.Union[None, int]
+BatchIndicesType = list[int]
```
bojiang, line 17: This will cause exception on lower version of Python. Using list/tuple as type is only allowed in type annotation area, or in `if TYPE_CHECKING:` since it's always Python 3.10

bojiang: @larme

larme: @bojiang fixed

PR #2479, opened 2022-05-19
feat: Transformers framework runner implementation 1.0

ssheng:
## Checklist:

### DIFF
--- a/bentoml/_internal/frameworks/transformers.py
+++ b/bentoml/_internal/frameworks/transformers.py
@@ -1,86 +1,37 @@
-import re
+from __future__ import annotations
+
 import typing as t
 import logging
 import importlib.util
+import pydantic
+import os
 from typing import TYPE_CHECKING
-from importlib import import_module
-
-from simple_di import inject
-from simple_di import Provide
 
 import bentoml
 from bentoml import Tag
+from bentoml.models import Model, ModelContext, ModelOptions
 from bentoml.exceptions import BentoMLException
 from bentoml.exceptions import MissingDependencyException
+from bentoml.exceptions import NotFound
 
 from ..types import LazyType
 from ..utils.pkg import get_pkg_version
-from .common.model_runner import BaseModelRunner
-from ..configuration.containers import BentoMLContainer
+
+if TYPE_CHECKING:
+    from bentoml.types import ModelSignatureDict
+    from bentoml.types import ModelSignature
+    from ..external_typing import transformers as ext
+
+
+MODULE_NAME = "bentoml.transformers"
+
 
 logger = logging.getLogger(__name__)
 
-if TYPE_CHECKING:
 
-    from ..models import ModelStore
-    from ..external_typing import transformers as ext
-
-try:
-    import transformers
-    from transformers.models.auto.configuration_auto import AutoConfig
-except ImportError:  # pragma: no cover
-    raise MissingDependencyException(
-        """\
-        transformers is required in order to use module `bentoml.transformers`.
-        Instruction: Install transformers with `pip install transformers`.
-        """
-    )
-
-
-def check_model_type(identifier: t.Any) -> bool:
-    return (
-        LazyType["ext.TransformersModelType"](
-            "transformers.modeling_utils.PreTrainedModel"
-        ).isinstance(identifier)
-        or LazyType["ext.TransformersModelType"](
-            "transformers.modeling_flax_utils.FlaxPreTrainedModel"
-        ).isinstance(identifier)
-        or LazyType["ext.TransformersModelType"](
-            "transformers.modeling_tf_utils.TFPreTrainedModel"
-        ).isinstance(identifier)
-    )
-
-
-def check_tokenizer_type(tokenizer: t.Any) -> bool:
-    return LazyType["ext.TransformersTokenizerType"](
-        "transformers.tokenization_utils.PreTrainedTokenizer"
-    ).isinstance(tokenizer) or LazyType["ext.TransformersTokenizerType"](
-        "transformers.tokenization_utils_fast.PreTrainedTokenizerFast"
-    ).isinstance(
-        tokenizer
-    )
-
-
-def check_fe_type(fe: t.Any) -> bool:
-    return LazyType["ext.PreTrainedFeatureExtractor"](
-        "transformers.feature_extraction_sequence_utils.SequenceFeatureExtractor"
-    ).isinstance(fe) or LazyType["ext.PreTrainedFeatureExtractor"](
-        "transformers.feature_extraction_utils.FeatureExtractionMixin"
-    ).isinstance(
-        fe
-    )
-
-
-def clean_name(name: str) -> str:
-    return re.sub(r"\W|^(?=\d)-", "_", name)
-
-
-def check_flax_supported() -> None:  # pragma: no cover
+def _check_flax_supported() -> None:  # pragma: no cover
     _supported: bool = get_pkg_version("transformers").startswith("4")
-    _flax_available = (
-        importlib.util.find_spec("jax") is not None
-        and importlib.util.find_spec("flax") is not None
-    )
+
     if not _supported:
         logger.warning(
             "Detected transformers version: "
@@ -90,11 +41,15 @@ def check_flax_supported() -> None:  # pragma: no cover
             "above to have Flax supported."
         )
     else:
+        _flax_available = (
+            importlib.util.find_spec("jax") is not None
+            and importlib.util.find_spec("flax") is not None
+        )
         if _flax_available:
             _jax_version = get_pkg_version("jax")
             _flax_version = get_pkg_version("flax")
             logger.info(
-                f"JAX version {_jax_version}, "
+                f"Jax version {_jax_version}, "
                 f"Flax version {_flax_version} available."
             )
         else:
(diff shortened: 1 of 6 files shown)


bentoml/_internal/frameworks/transformers.py
```diff
@@ -106,427 +65,209 @@ def check_flax_supported() -> None:  # pragma: no cover
...
-    """
-    return _TransformersRunner(
-        tag=tag,
-        tasks=tasks,
-        framework=framework,
-        device=device,
-        name=name,
-        **pipeline_kwargs,
-    )
+    return TransformersRunnable
```
aarnphm, line 273: address new changes from `FRAMEWORK_TEMPLATE_PY`

bentoml/_internal/frameworks/transformers.py
```diff
@@ -1,86 +1,41 @@
-import re
+from __future__ import annotations
+
+import os
 import typing as t
 import logging
 import importlib.util
```
aarnphm, line 6: lets also do
```python
import importlib
```
above this line for robust.

bentoml/_internal/frameworks/transformers.py
```diff
@@ -1,86 +1,41 @@
...
 from typing import TYPE_CHECKING
-from importlib import import_module
 
-from simple_di import inject
-from simple_di import Provide
+import pydantic
```
aarnphm, line 9: Is it possible that we use `attrs` instead of `pydantic`. I believe that `pydantic` is a very heavy library to import, and for the cases of `_Schema` a simple attrs class as a dataclass would work completely well and fast.

bentoml/_internal/frameworks/transformers.py
```diff
@@ -106,427 +65,209 @@ def check_flax_supported() -> None:  # pragma: no cover
...
+    if not isinstance(bento_model, Model):
+        bento_model = get(bento_model)
 
-        tmodel: "ext.TransformersModelType" = getattr(import_module("transformers"), _model).from_pretrained(  # type: ignore[reportUnknownMemberType]
-            model.path,
-            config=config,
-            **model_kwargs,
+    if bento_model.info.module not in (MODULE_NAME, __name__):
+        raise BentoMLException(
+            f"Model {bento_model.tag} was saved with module {bento_model.info.module}, failed loading with {MODULE_NAME}."
```
aarnphm, line 135: no need for this double check, since we have already did this during `get`, hence remove :)

bentoml/_internal/frameworks/transformers.py
```diff
@@ -106,427 +65,209 @@ def check_flax_supported() -> None:  # pragma: no cover
...
         )
 
-        if return_config:
-            return config, tmodel, tfe  # type: ignore
-        return None, tmodel, tfe  # type: ignore
+    return transformers.pipeline(
+        bento_model.info.options["task"], bento_model.path, **kwargs
+    )
```
aarnphm, line 140: lets also do 
```python
            try:
                transformers.pipelines.check_task(task)  # type: ignore
            except KeyError as e:
                raise BentoMLException(
                    f"{e}, as `{task}` is not recognized by transformers."
```

bentoml/_internal/frameworks/transformers.py
```diff
@@ -106,427 +65,209 @@ def check_flax_supported() -> None:  # pragma: no cover
...
-        config_kwargs = kwargs.pop("config_kwargs", {})
-        config: "ext.PretrainedConfig" = AutoConfig.from_pretrained(
-            model.path, **config_kwargs
-        )
 
-        model_kwargs = kwargs.pop("model_kwargs", {})
+def load_model(
+    bento_model: str | Tag | Model,
+    **kwargs: t.Any,
+) -> "ext.TransformersPipeline":
```
aarnphm, line 110: actually, since you import `from __future__ import annotations`, no need to quote this `ext.TransformersPipeline` anymore

bentoml/_internal/frameworks/transformers.py
```diff
@@ -1,86 +1,41 @@
...
-from ..configuration.containers import BentoMLContainer
-
-logger = logging.getLogger(__name__)
 
 if TYPE_CHECKING:
+    from bentoml.types import ModelSignature
+    from bentoml.types import ModelSignatureDict
 
-    from ..models import ModelStore
     from ..external_typing import transformers as ext
```
aarnphm, line 27: Since we are simplifying a lot of concepts, probably we don't need the transformers type under `external_typing` anymore. For the pipeline type I think we should just import directly from transformers and put it under `TYPE_CHECKING`.

bentoml/_internal/frameworks/transformers.py
```diff
@@ -106,427 +65,209 @@ def check_flax_supported() -> None:  # pragma: no cover
...
+    return transformers.pipeline(
+        bento_model.info.options["task"], bento_model.path, **kwargs
+    )
 
 
-def save(
+def save_model(
     name: str,
-    obj: t.Union["ext.TransformersModelType", "ext.TransformersPipeline"],  # type: ignore
+    pipeline: "ext.TransformersPipeline",
```
aarnphm, line 145: same comment as above

bentoml/_internal/frameworks/transformers.py
```diff
@@ -106,427 +65,209 @@ def check_flax_supported() -> None:  # pragma: no cover
...
+        tag = bentoml.transformers.save_model("text-generation-pipeline", generator)
+
+        # load the model back:
+        loaded = bentoml.transformers.load_model("text-generation-pipeline:latest")
+        # or:
+        loaded = bentoml.transformers.load_model(tag)
+    """  # noqa
+    if not LazyType["ext.TransformersPipeline"](
+        "transformers.pipelines.base.Pipeline"
+    ).isinstance(pipeline):
```
aarnphm, line 189: when you remove everything under `external_typing.transformers`
```suggestion
    if not LazyType["transformers.pipelines.base.Pipeline"](
        "transformers.pipelines.base.Pipeline"
    ).isinstance(pipeline):
```

bentoml/_internal/frameworks/transformers.py
```diff
@@ -106,427 +65,209 @@ def check_flax_supported() -> None:  # pragma: no cover
...
+        framework_name="transformers",
+        framework_versions={"transformers": get_pkg_version("transformers")},
+    )
+    options = TransformersOptions(task=pipeline.task)
+
+    if signatures is None:
+        signatures = {
+            "__call__": {"batchable": True},
+        }
+        logger.info(
```
aarnphm, line 204: lets make this a debug log instead

bentoml/_internal/frameworks/transformers.py
```diff
@@ -106,427 +65,209 @@ def check_flax_supported() -> None:  # pragma: no cover
...
 
-    If you want to returns an config object:
+class TransformersOptions(ModelOptions):
+    """Options for the Transformers model."""
 
-    .. code-block:: python
+    def __init__(self, **kwargs):
+        class _Schema(pydantic.BaseModel):
+            task: str
+            pipeline: bool = True
```
aarnphm, line 87: It would also be nice to know about the Model, Tokenizer, Feature Extractor if any that is used in this Pipeline

bentoml/_internal/frameworks/transformers.py
```diff
@@ -106,427 +65,209 @@ def check_flax_supported() -> None:  # pragma: no cover
...
-            return_config=True,
-            revision=self._revision,
-            use_fast=self._use_fast,
-            use_auth_token=self._use_auth_token,
-            model_store=self.model_store,
-            **self._kwargs,
-        )
+    class TransformersRunnable(bentoml.Runnable):
+        SUPPORT_NVIDIA_GPU = True  # type: ignore
+        SUPPORT_CPU_MULTI_THREADING = True  # type: ignore
```
aarnphm, line 231: we don't want to support multi-threading on huggingface. What do you think about this?

bentoml/_internal/frameworks/transformers.py
```diff
@@ -106,427 +65,209 @@ def check_flax_supported() -> None:  # pragma: no cover
...
-                device=self._device,
-                **self._kwargs,
-            )
-
-    def _run_batch(self, *args: t.Any, **kwargs: t.Any) -> t.Any:
-        return self._pipeline(*args, **kwargs)  # type: ignore
+                # assign CPU resources
+                kwargs = {}
 
+            self.pipeline = load_model(bento_model, **kwargs)
```
aarnphm, line 253: this kwargs would only have one option `device`, so if users want to pass in other transformers kwargs this would not work. I think we should make a 1-to-1 map from transformers kwargs to our load kwargs and save it under `ModelOptions`

bentoml/_internal/frameworks/transformers.py
```diff
@@ -106,427 +65,209 @@ def check_flax_supported() -> None:  # pragma: no cover
...
+            return getattr(self.pipeline, method_name)(*args, **kwargs)
 
-    .. code-block:: python
+        TransformersRunnable.add_method(
+            _run,
+            name=method_name,
+            batchable=options.batchable,
+            batch_dim=options.batch_dim,
+            input_spec=options.input_spec,
+            output_spec=options.output_spec,
```
aarnphm, line 270: refers to https://github.com/bentoml/BentoML/blob/runners-1.0/bentoml/_internal/frameworks/FRAMEWORK_TEMPLATE_PY for bug fix

aarnphm (commented):
overall first run everything looks good to me. Needs some changes here and there. Will discuss more offline

bentoml/_internal/frameworks/transformers.py
larme, line 87: I wonder if we can move this part to `bentoml._internal.models.model`, by using attrs' `make_class` to dynamically generate the `_Schema` class for each framework module? So for all frameworks we have single `ModelOptions` class but with different fields. Or do we think it's better to subclass the `ModelOptions` for each framework?

bentoml/_internal/frameworks/transformers.py
larme: or refers to the sklearn one: https://github.com/bentoml/BentoML/blob/runners-1.0/bentoml/_internal/frameworks/sklearn.py#L185

bentoml/_internal/frameworks/transformers.py
parano: currently pydantic is not a dependency of BentoML, would be great to replace it with attrs

bentoml/_internal/frameworks/transformers.py
aarnphm: I like to use `attr.make_class`, however, we need to create a type hint for this dynamic class

larme: I think we can use `attr.ib` to make a field with type info?

bentoml/_internal/frameworks/transformers.py
bojiang: Why not? We should always considering multi-threading first

bentoml/_internal/frameworks/transformers.py
bojiang: or shall we put them under with_args?

bentoml/_internal/frameworks/transformers.py
sauyon: `make_class` is the same thing as defining a new class, just uglier, isn't it? I thought it's for creating classes programmatically.

bentoml/_internal/frameworks/transformers.py
sauyon: Actually, we might want to do this if the type was already `Model` because we wouldn't be sure it's from `get`.

bentoml/_internal/frameworks/transformers.py
sauyon: I think this should be info level, it should show up by default.

bentoml/_internal/models/model.py
```diff
@@ -64,22 +64,22 @@
...
-        if len(kwargs) != 0:
-            for k in kwargs:
-                raise ValueError(f"Option {k} is not supported")
-        return cls()
+        return cls(**kwargs)
 
     @staticmethod
     def to_dict(options: ModelOptions) -> dict[str, t.Any]:
-        return dict(options)
+        return options.data
```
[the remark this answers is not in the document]
sauyon: This should probably remain as is and subclasses should just redefine this as necessary?

bentoml/_internal/frameworks/transformers.py
ssheng: Added.

bentoml/_internal/frameworks/transformers.py
ssheng: Good catch. I missed this check.

bentoml/_internal/frameworks/transformers.py
```diff
@@ -106,427 +65,209 @@ def check_flax_supported() -> None:  # pragma: no cover
...
-            raise BentoMLException(
-                "`model` is neither a Transformers pipeline nor a Transformers model."
-            )
+        generator = pipeline(task="text-generation")
+        tag = bentoml.transformers.save_model("text-generation-pipeline", generator)
+
+        # load the model back:
+        loaded = bentoml.transformers.load_model("text-generation-pipeline:latest")
+        # or:
+        loaded = bentoml.transformers.load_model(tag)
```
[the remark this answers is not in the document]
ssheng: Do you mean when we construct the pipeline, we also pass in the model and tokenizer like below?
```
generator = pipeline(task="text-generation", model=..., tokenizer=...)
```

bentoml/_internal/frameworks/transformers.py
ssheng: Huggingface models are essential TF, PT, or JAX models. All of them support multi-threading, no?

bentoml/_internal/frameworks/transformers.py
ssheng: Thanks for the suggestion. Updated to the following.

Transformers types provided other than Pipeline are not supported in BentoML. 
To save models, tokenizers, configs, feature extractors, construct a pipeline 
with the model, tokenizer, config, or feature extractor specified as arguments, 
then call save_model with the pipeline.

```python
import bentoml
from transformers import pipeline, AutoTokenizer, AutoModelForCausalLM

tokenizer = AutoTokenizer.from_pretrained("distilgpt2")
model = AutoModelForCausalLM.from_pretrained("distilgpt2")
generator = pipeline(task="text-generation", model=model, tokenizer=tokenizer)

bentoml.transformers.save_model("text-generation-pipeline", generator)
```

Refers to https://huggingface.co/docs/transformers/main_classes/pipelines for 
more information on pipelines.

bentoml/_internal/frameworks/transformers.py
ssheng: I also think `info` is maybe the right logging level. Please let me know if you feel strong about this.

bentoml/_internal/frameworks/transformers.py
ssheng: What do you suggest that we can do with that information?

bentoml/_internal/frameworks/transformers.py
aarnphm: I don't have a strong opinion on this. we can just keep it as it 👍🏼

bentoml/_internal/frameworks/transformers.py
ssheng: How do we get an exhaustive list of Transformers pipeline argument? I was also not clear what you meant by saving it under `ModelOptions` under the `load_model()` path.

bentoml/_internal/frameworks/transformers.py
ssheng: Added as a part of the `attrs` class field validator.

bentoml/_internal/frameworks/transformers.py
```diff
@@ -1,86 +1,43 @@
...
 import logging
+import importlib
 import importlib.util
 from typing import TYPE_CHECKING
-from importlib import import_module
 
-from simple_di import inject
-from simple_di import Provide
+import attr
+import attrs
```
aarnphm, line 11: you only need to import `attr` here. `attrs` is a newer api that does route everything back to `attr`.
```suggestion

```

bentoml/_internal/frameworks/transformers.py
```diff
@@ -106,427 +67,228 @@ def check_flax_supported() -> None:  # pragma: no cover
...
-        import bentoml
-        config, model, tokenizer = bentoml.transformers.load('custom_gpt2', return_config=True, tokenizer_kwargs={"use_fast":True})
+    @classmethod
+    def with_options(cls, **kwargs: t.Any) -> ModelOptions:
+        return cls(**kwargs)
 
-    If the pipeline is saved with :code:`bentoml.transformers.save()`, then :code:`load()` will return pipeline objects:
+    @staticmethod
+    def to_dict(options: ModelOptions) -> dict[str, t.Any]:
+        return attrs.asdict(options)
```
aarnphm, line 103:
```suggestion
        return attr.asdict(options)
```

bentoml/_internal/frameworks/transformers.py
aarnphm: got it

bentoml/_internal/frameworks/transformers.py
```diff
@@ -106,427 +67,228 @@ def check_flax_supported() -> None:  # pragma: no cover
...
+            """
+            ```python
+            import bentoml
+            from transformers import pipeline, AutoTokenizer, AutoModelForCausalLM
+
+            tokenizer = AutoTokenizer.from_pretrained("distilgpt2")
+            model = AutoModelForCausalLM.from_pretrained("distilgpt2")
+            generator = pipeline(task="text-generation", model=model, tokenizer=tokenizer)
+
+            bentoml.transformers.save_model("text-generation-pipeline", generator)
```
aarnphm, line 215: even more verbose suggestion. 
```
If transformers doesn't provide a task you need, refers to the custom pipeline section to create your own pipeline...
```

bentoml/_internal/frameworks/transformers.py
aarnphm: I think it is impossible to get an exhaustive list of all of the pipeline arguments. I think one way is that we accept `**kwargs` for `get_runnable`, and then append our default to that kwargs if user is not set. Then that `kwargs` can be passed to `load_model`

bentoml/_internal/frameworks/transformers.py
aarnphm: noted

bentoml/_internal/frameworks/transformers.py
```diff
@@ -89,10 +89,13 @@ class TransformersOptions(ModelOptions):
             lambda instance, attribute, value: transformers.pipelines.check_task(value),  # type: ignore
         ]
     )
+
     pipeline: bool = attr.field(
         default=True, validator=attr.validators.instance_of(bool)
     )
 
+    kwargs: t.Dict[str, t.Any] = attr.field(default={})
```
aarnphm, line 97:
```suggestion
    kwargs: t.Dict[str, t.Any] = attr.field(factory=dict)
```

bentoml/_internal/frameworks/transformers.py
```diff
@@ -106,427 +66,240 @@ def check_flax_supported() -> None:  # pragma: no cover
...
+    bento_model.info.parse_options(TransformersOptions)
+
+    pipeline_task: str = bento_model.info.options.task  # type: ignore
+    pipeline_kwargs: t.Dict[str, t.Any] = bento_model.info.options.kwargs  # type: ignore
+    pipeline_kwargs.update(kwargs)
+    if len(pipeline_kwargs) > 0:
+        logger.info(
+            f"Loading '{pipeline_task}' pipeline '{bento_model.tag}' with kwargs {pipeline_kwargs}."
+        )
+    return transformers.pipeline(pipeline_task, bento_model.path, **pipeline_kwargs)  # type: ignore
```
aarnphm, line 157:
```suggestion
    return transformers.pipeline(pipeline_task, model=bento_model.path, **pipeline_kwargs)  # type: ignore
```

bentoml/_internal/frameworks/transformers.py
```diff
@@ -106,427 +66,240 @@ def check_flax_supported() -> None:  # pragma: no cover
...
-            accordingly to your use case provided by :obj:`transformers`. Refers to
-            `Feature Extractor API <https://huggingface.co/docs/transformers/main_classes/feature_extractor>`_
-            for more information
+        pipeline (:code:`Pipeline`):
+            Instance of the Transformers pipeline to be saved.
+        signatures (:code: `Dict[str, bool | BatchDimType | AnyType | tuple[AnyType]]`)
+            Methods to expose for running inference on the target model. Signatures are
+             used for creating Runner instances when serving model with bentoml.Service
         labels (:code:`Dict[str, str]`, `optional`, default to :code:`None`):
             user-defined labels for managing models, e.g. team=nlp, stage=dev
```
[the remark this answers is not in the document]
aarnphm: because with tokenizer they should use `Tokenizer` class instead.

bentoml/_internal/models/model.py
```diff
@@ -539,6 +535,11 @@ def with_options(self, **kwargs: t.Any) -> ModelInfo:
     def to_dict(self) -> t.Dict[str, t.Any]:
         return bentoml_cattr.unstructure(self)  # type: ignore (incomplete cattr types)
 
+    def parse_options(self, options_class: type[ModelOptions]) -> None:
+        object.__setattr__(
+            self, "options", options_class.with_options(**self.options.data)
```
aarnphm, line 540: I don't think that `options` have attrributes `data`. This only applies to pydantic model only

ssheng: The `data` attribute came from `UserDict` actually, https://docs.python.org/2/library/userdict.html

bentoml/_internal/frameworks/transformers.py
ssheng: Let's add `task=` as well.

bentoml/_internal/frameworks/transformers.py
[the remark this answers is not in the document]
ssheng: It's a case of a dilemma for Transformers. While it is nice to keep the consistency across frameworks, the concept of Transformers is to build everything into the pipeline. Arguably, `custom_objects` can be used in the preprocessing logic. I will add it to the next iteration and see how we feel.

bentoml/_internal/models/model.py
aarnphm: cool cool

bentoml/_internal/frameworks/transformers.py
```diff
@@ -106,427 +66,248 @@ def check_flax_supported() -> None:  # pragma: no cover
...
-        model, tokenizer = bentoml.transformers.load('custom_gpt2')
+    task: str = attr.field(
+        validator=[
+            attr.validators.instance_of(str),
+            lambda instance, attribute, value: transformers.pipelines.check_task(value),  # type: ignore
+        ]
+    )
 
-    If you want to returns an config object:
+    pipeline: bool = attr.field(
```
aarnphm, line 93: where do we use this again? @ssheng

ssheng: We don't use it today. This is more of an informative metadata which is useful if we add support for models, tokenizers, config, and feature extractors. I'm also OK if we want to exclude this field for now.

aarnphm: lets leave it out for now, since it doesn't provide any value yet. But agree with adding this when we have more metadata about model, tokenizer.

PR #2486, opened 2022-05-23
fix(server): correctly use starlette APIs

sauyon:


pyproject.toml
```diff
@@ -119,63 +119,7 @@ skip_glob = ["typings/*", "docs/*"]
 
 [tool.pyright]
 pythonVersion = "3.10"
-include = ["bentoml", "typings"]
+include = ["bentoml"]
 exclude = ['bentoml/_version.py','bentoml/__main__.py']
-strictListInference = true
-strictDictionaryInference = true
-strictSetInference = true
```
bojiang, line 126: why do we remove these

bentoml/_internal/types.py
```diff
@@ -217,7 +217,7 @@ def is_compatible_type(t1: AnyType, t2: AnyType) -> bool:
 
 @json_serializer(fields=["uri", "name"], compat=True)
 @dataclass(frozen=False)
-class FileLike(t.Generic[t.AnyStr]):
+class FileLike(t.Generic[t.AnyStr], io.IOBase, t.IO[t.AnyStr]):  # type: ignore (python IO types)
```
bojiang, line 220: Is this necessary for runtime or just for typing?

sauyon: Just typing. I actually realized I can use `t.cast` later so I don't think this is necessary at all anymore.

pyproject.toml
sauyon: I wanted to go back to full strict mode and just `# type: ignore` all the unknownmembertype and missingstubs issues.

bojiang: Is full strict mode stricter or looser than the current?

sauyon: stricter

bojiang: I prefer not to do too much style changes till a stable code base.

PR #2487, opened 2022-05-24
fix(tests): update model unit tests for new batch_dim type

sauyon:
We no longer support non-tuple batch_dim.

### DIFF
--- a/tests/unit/_internal/models/test_model.py
+++ b/tests/unit/_internal/models/test_model.py
@@ -59,7 +59,9 @@ signatures:
     batchable: false
   classify:
     batchable: true
-    batch_dim: 0
+    batch_dim:
+    - 0
+    - 0
   predict_ii:
     batchable: true
     batch_dim:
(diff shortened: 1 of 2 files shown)


tests/unit/_internal/models/test_model.py
```diff
@@ -59,7 +59,9 @@
     batchable: false
   classify:
     batchable: true
-    batch_dim: 0
+    batch_dim:
```
bojiang, line 62: shall we have
```
input_batch_dim: 0
output_batch_dim: 0
```
here? @parano

parano: yes, I thought you had some refactoring work for batch_dim that will do this?

bojiang: @parano yeah but I need an extra approval since it's simplified a lot.

bojiang: will do that in a small refactoring PR

bojiang (approved):

PR #2488, opened 2022-05-24
fix(server): fix development_mode in the config

bojiang:
## Description

> bentoml serve --reload

## Checklist:


aarnphm (approved):

sauyon:
Uh, this doesn't work, does it? If on reload it's defaulting to `development_mode = False` it'll try to init a remote runner client but there aren't any remote runners.

The easy fix would be to just set this to `True`, but that might be a bit strange too.

bojiang:
Do you mean it should not be a config item?

bojiang:
discussed in the IM. will propose another PR

PR #2499, opened 2022-05-26
feat(framework): pytorch

bojiang:
## Checklist:


bentoml/_internal/frameworks/sklearn.py
```diff
@@ -149,7 +149,14 @@ def save_model(
     )
 
     if signatures is None:
-        signatures = {"predict": {"batchable": False}}
+        signatures = {
```
parano, line 152: revert this change? I think it's nicer to print a short info log that shows the default

bentoml/_internal/frameworks/sklearn.py
```diff
@@ -11,6 +11,7 @@
 from bentoml.exceptions import NotFound
 from bentoml.exceptions import BentoMLException
 from bentoml.exceptions import MissingDependencyException
+from bentoml._internal.models.model import ModelSignaturesType
```
parano, line 14: @sauyon  @larme let's use this type for `signature` parameter in all `save_model` methods?

bentoml/_internal/frameworks/pytorch_lightning.py
```diff
@@ -63,32 +72,35 @@ def load(
...
         )
     weight_file = bentoml_model.path_of(f"{SAVE_NAMESPACE}{PT_EXT}")
-    model_format = bentoml_model.info.context.get("model_format")
-    # backward compatibility
-    if not model_format:
-        model_format = "pytorch_lightning:v1"
+    # model_format = bentoml_model.info.context.get("model_format")
+    # # backward compatibility
+    # if not model_format:
+    # model_format = "pytorch_lightning:v1"
```
parano, line 86: remove this part?

bentoml/_internal/frameworks/pytorch.py
```diff
@@ -100,8 +115,6 @@ def save(
             e.g. a tokenizer instance, preprocessor function, model configuration json
         metadata (:code:`Dict[str, Any]`, `optional`,  default to :code:`None`):
             Custom metadata for given model.
-        model_store (:mod:`~bentoml._internal.models.store.ModelStore`, default to :mod:`BentoMLContainer.model_store`):
-            BentoML modelstore, provided by DI Container.
 
     Returns:
         :obj:`~bentoml.Tag`: A :obj:`tag` with a format `name:version` where `name` is the user-defined model's name, and a generated `version` by BentoML.
```
parano, line 120: update the docstring below?

bentoml/_internal/frameworks/torchscript.py
```diff
@@ -103,104 +114,98 @@ def save(
...
+            def configure_optimizers(self):
+                return torch.optim.Adam(self.parameters(), lr=self.hparams.learning_rate)
 
-        tag = bentoml.torchscript.save("resnet50", resnet50)
-    """  # noqa
-    context: t.Dict[str, t.Any] = {
-        "framework_name": "torch",
-        "pip_dependencies": [f"torch=={get_pkg_version('torch')}"],
-    }
+        tag = bentoml.pytorch_lightning.save("lit_classifier", LitClassifier())
```
parano, line 157:
```suggestion
        tag = bentoml.pytorch_lightning.save_model("lit_classifier", LitClassifier())
```

parano: also, `signature` is required for this example

bentoml/_internal/frameworks/torchscript.py
```diff
@@ -1,97 +1,108 @@
...
+    _include_pytorch_lightning_version: bool = False,
 ) -> Tag:
     """
     Save a model instance to BentoML modelstore.
 
     Args:
         name (:code:`str`):
             Name for given model instance. This should pass Python identifier check.
-        model (:code:`torch.jit.ScriptModule`):
+        model (`pl.LightningModule`):
```
parano, line 98: this should be `torch.jit.ScriptModule` right?

parano: also make sure to use .rst format, e.g.  
```
model (:code:`torch.jit.ScriptModule`)
```

bentoml/_internal/frameworks/torchscript.py
```diff
@@ -103,104 +114,98 @@ def save(
...
+        }
+    else:
+        framework_versions = {"torch": get_pkg_version("torch")}
+
+    context: ModelContext = ModelContext(
+        framework_name="torchscript",
+        framework_versions=framework_versions,
+    )
+
+    if signatures is None:
```
parano, line 173: does it make sense also to apply a default signature for all PyTorch models, with { "__call__": { batchable: True }}` ?

bentoml/_internal/frameworks/common/pytorch.py
```diff
@@ -41,141 +41,156 @@
...
 class PyTorchTensorContainer(DataContainer[torch.Tensor, torch.Tensor]):
     @classmethod
-    def singles_to_batch(cls, singles, batch_axis=0):
-        return torch.stack(singles, dim=batch_axis)
+    def batches_to_batch(
+        cls,
+        batches: tuple[torch.Tensor] | list[torch.Tensor],
+        batch_dim: int = 0,
+    ) -> t.Tuple[torch.Tensor, list[int]]:
+        # numpy.concatenate may consume lots of memory, need optimization later
```
sauyon, line 107: Does this still apply to tensors?

bentoml/_internal/frameworks/common/pytorch.py
```diff
@@ -41,141 +41,156 @@
...
+        indices: t.Sequence[int],
+        batch_dim: int = 0,
+    ) -> t.List[torch.Tensor]:
+        sizes = [indices[i] - indices[i - 1] for i in range(1, len(indices))]
+        return torch.split(batch, sizes, axis=batch_dim)
 
     @classmethod
     @inject
-    def single_to_payload(
+    def to_payload(  # pylint: disable=arguments-differ
```
sauyon, line 130: `to_payload` should take a `batch_dim` now and set payload length for validation on the service app side.

bentoml/_internal/frameworks/torchscript.py
```diff
@@ -1,97 +1,108 @@
...
+
+
+try:
+    import torch
+except ImportError:  # pragma: no cover
+    raise MissingDependencyException(_PL_IMPORT_ERROR)
 
-MODULE_NAME = "bentoml.torchscript"
+MODULE_NAME = "bentoml.pytorch_lightning"
+MODEL_FILENAME = "savd_model.pt"
```
sauyon, line 33:
```suggestion
MODEL_FILENAME = "saved_model.pt"
```

bentoml/_internal/frameworks/torchscript.py
sauyon: Should we be enabling `batchable` by default?

bentoml/_internal/models/model.py
```diff
@@ -42,8 +42,8 @@
     from ..types import PathType
 
     class ModelSignatureDict(t.TypedDict, total=False):
-        batch_dim: tuple[int, int]
-        batchable: bool
+        batch_dim: tuple[int, int] | int | None
+        batchable: bool | None
         input_spec: tuple[AnyType] | AnyType | None
         output_spec: AnyType | None
```
sauyon, line 48: Does `total=False` not do the optional bit by default?

sauyon (commented):
LGTM minus minor comments.

bentoml/_internal/frameworks/torchscript.py
parano: @sauyon I think it make sense for frameworks like PyTorch and TF, although I understand it might be better to keep the default behavior consistent and ask users to set this explicitly.

Btw we currently have batchable=True as default in the Keras module

bentoml/_internal/frameworks/pytorch.py
```diff
@@ -1,48 +1,53 @@
...
-from .common.pytorch import PyTorchTensorContainer  # pylint: disable=unused-import
-from ..configuration.containers import BentoMLContainer
+from ..models.model import ModelContext
+from ..models.model import ModelSignaturesType
+from .common.pytorch import torch  # type: ignore
+from .common.pytorch import PyTorchTensorContainer
 
-if TYPE_CHECKING:
-    from ..models import ModelStore
+__all__ = ["load_model", "get_runnable", "get", "PyTorchTensorContainer"]
```
aarnphm, line 22: why do we need this here? we should only set `__all__` at the top level

bentoml/_internal/frameworks/torchscript.py
bojiang: @parano @sauyon Pytorch modules are not using `__call__` as the default inference method, nor in its official guide, thus it doesn't make sense, I believe.

bentoml/_internal/frameworks/pytorch.py
```diff
@@ -53,37 +58,29 @@ def load(
...
-            model: "torch.nn.Module" = cloudpickle.load(file).to(device_id)
-    elif model_format == "torch.save:v1":
-        with Path(weight_file).open("rb") as file:
-            model: "torch.nn.Module" = torch.load(file, map_location=device_id)
-    else:
-        raise BentoMLException(f"Unknown model format {model_format}")
 
+    weight_file = bentoml_model.path_of(MODEL_FILENAME)
+    with Path(weight_file).open("rb") as file:
+        model: "torch.nn.Module" = torch.load(file, map_location=device_id)
```
aarnphm, line 72: should we keep `model_format` behaviour like before?

bentoml/_internal/frameworks/pytorch_lightning.py
```diff
@@ -156,72 +103,24 @@ def configure_optimizers(self):
...
-        name=name,
-        partial_kwargs=partial_kwargs,
+    return script_save_model(
+        name,
+        script_module,
+        signatures=signatures,
+        labels=labels,
+        custom_objects=custom_objects,
+        metadata=metadata,
+        _include_pytorch_lightning_version=True,
```
aarnphm, line 125: hmm this seems to be like an internal arg from torchscript. 🤔

bentoml/_internal/frameworks/torchscript.py
aarnphm, line 157: Since this is the torchscript model, We shouldn't show the pytorch lightning examples. PyTorch lightning example should be under pytorch lightning instead. Even though pytorch lightning use torchscript under the hood, I think it is still better to have the separation and let user know about that.

bentoml/_internal/models/model.py
```diff
@@ -490,12 +488,19 @@ def __init__(
         api_version: str | None = None,
         creation_time: datetime | None = None,
     ):
-        ...
-
-    def __attrs_post_init__(self):
-        object.__setattr__(self, "name", self.tag.name)
-        object.__setattr__(self, "version", self.tag.version)
-
```
aarnphm, line 498: should we use `field_transformer` instead here?

bentoml/_internal/frameworks/pytorch.py
parano: What is that for?

bentoml/_internal/models/model.py
parano: The current approach is ok

bentoml/_internal/frameworks/torchscript.py
bojiang: Let me check about the original usage or __call__ in its doc

bentoml/_internal/frameworks/pytorch.py
aarnphm: for compatibility i believe right @larme ?

larme: I think we can drop model format backward compatibility before 1.0 release to simplify the code. But keep backward compatibility after 1.0 release

bentoml/_internal/frameworks/torchscript.py
bojiang: Just updated. will use __call__ as the default method. But I don't think we should have batchable=True for models, which may cause unwanted batching behavior

bentoml/_internal/frameworks/pytorch.py
bojiang: This is the standard way to get linter to stop complaining about A `imported but not used`

bentoml/_internal/frameworks/pytorch_lightning.py
bojiang: We don't want to expose this arg to users, I think.

aarnphm: agree.

bentoml/_internal/models/model.py
bojiang: the "__attrs_init__" is the recommended way in case the initialization parameter type of the class is not the same as the property type. Naturally friendly to typing tools.

bentoml/_internal/frameworks/pytorch.py
bojiang: done

bentoml/_internal/frameworks/common/pytorch.py
bojiang: done

bentoml/_internal/models/model.py
bojiang: TIL

PR #2500, opened 2022-05-26
Testing framework and runner app update

sauyon:


bentoml/_internal/runner/container.py
```diff
@@ -178,6 +178,70 @@ def from_batch_payloads(
         return cls.batches_to_batch(batches, batch_dim)
 
 
+class DMatrixContainer(
```
parano, line 181: when we implement this container, could we move it to the xgboost module? we are doing that for the PyTorch and TF tensor, so that those containers are only registered when those modules got imported by user code

bentoml/_internal/runner/runnable.py
```diff
@@ -25,42 +26,69 @@ class Runnable(ABC):
     SUPPORT_NVIDIA_GPU: bool
     SUPPORT_CPU_MULTI_THREADING: bool
 
+    methods: dict[str, RunnableMethod[t.Any, t.Any, t.Any]] = {}
```
parano, line 29: change default value to None here and create it in `add_method` if it's None?

bentoml/_internal/external_typing/__init__.py
```diff
@@ -20,6 +20,8 @@
 
     NpNDArray = _NDArray[t.Any]
 
+    from xgboost import DMatrix
```
parano, line 23: does this mean developers have to install xgboost, in order to run type checking?

sauyon: If they work on any files that use it, yeah.

pyproject.toml
```diff
@@ -121,6 +121,7 @@ skip_glob = ["typings/*", "docs/*"]
 pythonVersion = "3.10"
 include = ["bentoml"]
 exclude = ['bentoml/_version.py','bentoml/__main__.py']
+useLibraryCodeForTypes = true
```
parano, line 124: note from: https://github.com/microsoft/pyright/blob/main/docs/settings.md

```
pyright.useLibraryCodeForTypes [boolean]: This setting is deprecated in favor of python.analysis.useLibraryCodeForTypes. It will be removed at a future time.
```

bentoml/_internal/models/model.py
```diff
@@ -372,47 +371,42 @@ class ModelSignature:
         batch_dim:
             The dimension(s) that contain multiple data when passing to this prediction method.
 
-            For example, if you have two inputs you want to run prediction on, ``[1, 2]`` and
-            ``[3, 4]``, if the array you would pass to the predict method would be
-            ``[[1, 2], [3, 4]]``, then the batch dimension would be ``0``. If the array you would
-            pass to the predict method would be ``[[1, 3], [2, 4]]``, then the batch dimension would
-            be ``1``.
+            For example, if you have two inputs you want to run prediction on, ``[1, 2]`` and ``[3,
```
parano, line 374: 👍 also adding this to the documentation

bentoml/_internal/runner/container.py
sauyon: Actually I only included this because it was a pain to separate out properly but I ended up making a commit for it anyway so we can just drop it.

DMatrix has no native way to concatenate, so if we wanted to do that we'd either have to hack memory or save to binary -> use their internal format to hack together the files -> load.

bentoml/_internal/runner/utils.py
```diff
@@ -65,42 +68,6 @@ def from_dict(cls, data: dict[str | int, T]) -> Params[T]:
             **{k: v for k, v in data.items() if isinstance(k, str)},
         )
 
-    @classmethod
-    def agg(
-        cls,
-        params_list: t.Sequence[Params[T]],
-        agg_func: t.Callable[[t.Sequence[T]], To] = pass_through,
```
parano, line 72: remove `pass_through` above?

bentoml/_internal/server/runner_app.py
```diff
@@ -181,23 +180,26 @@ async def _run(requests: t.Iterable[Request]) -> list[Response]:
...
-                params_list=params_list,
-                agg_func=lambda i: AutoContainer.from_batch_payloads(
-                    i, batch_dim=batch_dim[0]
-                ),
-            )
-            batched_params, indices_params = mix_params.vsplit(lambda i: (i[0], i[1]))
+            batch_dim = runner_method.config.batch_dim
+
+            try:
+                batched_params, indices = batch_params(params_list, batch_dim[0])
```
bojiang, line 186: we actually do need that batch_params, I think. How about just checking by iterating on the indices_params?

bojiang: This refactoring makes general container utils on Pramas circumstance specified, with more cupling.

PR #2507, opened 2022-05-27
fix(framework): adapt changes for Tensorflow DataContainer

larme:
## Checklist:

### DIFF
--- a/bentoml/_internal/frameworks/tensorflow_v2.py
+++ b/bentoml/_internal/frameworks/tensorflow_v2.py
@@ -229,9 +229,6 @@ def get_runnable(
                 self.device_name = "/device:GPU:0"
             else:
                 self.device_name = "/device:CPU:0"
-                num_threads = int(os.environ["BENTOML_NUM_THREAD"])
-                tf.config.threading.set_inter_op_parallelism_threads(num_threads)
-                tf.config.threading.set_intra_op_parallelism_threads(num_threads)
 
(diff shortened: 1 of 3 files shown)


bentoml/_internal/frameworks/tensorflow_v2.py
```diff
@@ -229,9 +229,6 @@ def __init__(self):
                 self.device_name = "/device:GPU:0"
             else:
                 self.device_name = "/device:CPU:0"
-                num_threads = int(os.environ["BENTOML_NUM_THREAD"])
-                tf.config.threading.set_inter_op_parallelism_threads(num_threads)
-                tf.config.threading.set_intra_op_parallelism_threads(num_threads)
```
larme, line 234: @parano Because resource strategy already setup `TF_NUM_INTEROP_THREADS` etc., we don't need these lines at all

bentoml/_internal/frameworks/tensorflow_v2.py
```diff
@@ -250,7 +247,9 @@ def _run_method(
             with tf.device(runnable_self.device_name):
 
                 def _mapping(item: "TFArgType") -> "tf_ext.TensorLike":
```
bojiang, line 249: shall we move this function definition out of the inference path? Or it will create the function for each request

parano: @larme could you do that in a follow-up PR?

PR #2519, opened 2022-05-30
fix(bug): Make `model.with_options()` additive

ssheng:
Make `model.with_options()` additive as opposed to overriding. Changed the `ModelOptions` from UserDict to an `attrs` class. Specific framework implementation can still provide a customized `FrameworkOptions(ModelOptions)` to type check and validate the options attributes.

## Checklist:

### DIFF
--- a/bentoml/_internal/frameworks/tensorflow_v2.py
+++ b/bentoml/_internal/frameworks/tensorflow_v2.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import os
 import pickle
 import typing as t
 import logging
@@ -9,10 +8,13 @@ import itertools
 import contextlib
 from typing import TYPE_CHECKING
 
+import attr
+
 import bentoml
 from bentoml import Tag
 from bentoml import Runnable
 from bentoml.models import ModelContext
+from bentoml.models import ModelOptions
 from bentoml.exceptions import NotFound
 from bentoml.exceptions import MissingDependencyException
 
@@ -47,6 +49,13 @@ if TYPE_CHECKING:
 
 MODULE_NAME = "bentoml.tensorflow"
 API_VERSION = "v1"
+
+
+@attr.define
+class KerasOptions(ModelOptions):
+    """Options for the Keras model."""
+
+    partial_kwargs: t.Dict[str, t.Any] = attr.field(factory=dict)
 
 
 def get(tag_like: str | Tag) -> bentoml.Model:
@@ -219,9 +228,7 @@ def get_runnable(
     Private API: use :obj:`~bentoml.Model.to_runnable` instead.
     """
 
-    partial_kwargs: t.Dict[str, t.Any] = bento_model.info.options.get(
-        "partial_kwargs", dict()
-    )
+    partial_kwargs: t.Dict[str, t.Any] = bento_model.info.options.partial_kwargs
 
     class TensorflowRunnable(Runnable):
         SUPPORT_NVIDIA_GPU = True
--- a/bentoml/_internal/frameworks/tensorflow_v2.py
+++ b/bentoml/_internal/frameworks/tensorflow_v2.py
@@ -52,7 +52,7 @@ API_VERSION = "v1"
 
 
 @attr.define
-class KerasOptions(ModelOptions):
+class TensorflowOptions(ModelOptions):
     """Options for the Keras model."""
 
     partial_kwargs: t.Dict[str, t.Any] = attr.field(factory=dict)
--- a/bentoml/_internal/models/model.py
+++ b/bentoml/_internal/models/model.py
@@ -57,8 +57,8 @@ CUSTOM_OBJECTS_FILENAME = "custom_objects.pkl"
 
 
 @attr.define
-class ModelOptions():
-    def with_options(self, **kwargs: t.Any) -> 'ModelOptions':
+class ModelOptions:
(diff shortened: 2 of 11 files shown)


bentoml/_internal/models/model.py
```diff
@@ -57,24 +56,25 @@ class ModelSignatureDict(t.TypedDict, total=False):
...
-
-
-class ModelOptions(ModelOptionsSuper):
-    @classmethod
-    def with_options(cls, **kwargs: t.Any) -> ModelOptions:
-        return cls(**kwargs)
+@attr.define
+class ModelOptions:
+    def with_options(self, **kwargs: t.Any) -> "ModelOptions":
+        return attr.evolve(self, **kwargs)
```
sauyon, line 62: I think it might be better to have these be fully abstract, or maybe have an ABC that people can subclass that's a supertype of this `attrs` base class.

parano:
Looks like the transformers integration test is failing here: https://github.com/bentoml/BentoML/runs/6744037769?check_suite_focus=true

bentoml/_internal/frameworks/tensorflow_v2.py
```diff
@@ -49,6 +51,13 @@
 API_VERSION = "v1"
 
 
+@attr.define
+class KerasOptions(ModelOptions):
+    """Options for the Keras model."""
+
+    partial_kwargs: t.Dict[str, t.Any] = attr.field(factory=dict)
+
+
```
aarnphm, line 60: this should be called `TensorflowOptions`?

bentoml/_internal/models/model.py
aarnphm: @ssheng are we still need to do something like this?

bentoml/_internal/models/model.py
```diff
@@ -585,6 +573,20 @@ def from_yaml_file(stream: t.IO[t.Any]):
             del yaml_content["context"]["pip_dependencies"]
             yaml_content["context"]["framework_versions"] = {}
 
+        # register hook for model options
+        module_name: str = yaml_content["module"]
+        try:
+            module = importlib.import_module(module_name)
+        except ValueError or ModuleNotFoundError as e:
```
aarnphm, line 580:
```suggestion
        except (ValueError, ModuleNotFoundError) as e:
```

bentoml/keras.py
```diff
@@ -4,6 +4,10 @@
 from ._internal.frameworks.keras import load_model
 from ._internal.frameworks.keras import save_model
 from ._internal.frameworks.keras import get_runnable
+from ._internal.frameworks.keras import KerasOptions
+
+ModelOptions = KerasOptions
+
```
aarnphm, line 10:
```suggestion
from ._internal.frameworks.keras import KerasOptions as ModelOptions
```

aarnphm: Screenshot 2022-06-09 at 23 53 42

refactor(style): If we go to the above change, `ModelOptions` is under the module scope.

bentoml/tensorflow.py
```diff
@@ -4,6 +4,10 @@
 from ._internal.frameworks.tensorflow_v2 import load_model
 from ._internal.frameworks.tensorflow_v2 import save_model
 from ._internal.frameworks.tensorflow_v2 import get_runnable
+from ._internal.frameworks.tensorflow_v2 import TensorflowOptions
+
+ModelOptions = TensorflowOptions
+
```
aarnphm, line 10: same as above

bentoml/transformers.py
```diff
@@ -4,6 +4,10 @@
 from ._internal.frameworks.transformers import load_model
 from ._internal.frameworks.transformers import save_model
 from ._internal.frameworks.transformers import get_runnable
+from ._internal.frameworks.transformers import TransformersOptions
+
+ModelOptions = TransformersOptions
+
```
aarnphm, line 10: ^

bentoml/_internal/frameworks/tensorflow_v2.py
ssheng: Yes. Change was uncommitted. Fixed now.

bentoml/_internal/models/model.py
ssheng: Not all frameworks will have options. For those that don't have options, we can still have it fallback to using ModelOptions. So we probably don't want to change it to an abstract class.

bentoml/keras.py
ssheng: Will update as suggested.

bentoml/transformers.py
```diff
@@ -4,6 +4,7 @@
 from ._internal.frameworks.transformers import load_model
 from ._internal.frameworks.transformers import save_model
 from ._internal.frameworks.transformers import get_runnable
+from ._internal.frameworks.transformers import TransformersOptions as ModelOptions  # type: ignore
```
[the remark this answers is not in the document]
ssheng: Fixed in the next commit.

aarnphm (approved):
Tested locally e2e. Kudos @ssheng

bentoml/_internal/models/model.py
sauyon: Some part of me likes the purity of having an ABC define the trait and then implementing it (with a default class) but it's probably realistically it's not important enough to add the code complexity for.

PR #2526, opened 2022-05-31
chore: more robust prometheus directory ensuring

bojiang:
## Description

P2

We could not launch multiple server instances by the same project/bento, because there would be a contest about Prometheus db files. But for now it gets fixed

## Checklist:

### DIFF
--- a/bentoml/_internal/server/__init__.py
+++ b/bentoml/_internal/server/__init__.py
@@ -71,7 +71,7 @@ def ensure_prometheus_dir(
         f"Failed to ensure the prometheus multiproc directory {directory}, "
         f"using alternative: {alternative}",
     )
-    DeploymentContainer.prometheus_multiproc_dir.set(alternative)
+    BentoMLContainer.prometheus_multiproc_dir.set(alternative)
     return alternative
 
 
--- a/bentoml/_internal/server/__init__.py
+++ b/bentoml/_internal/server/__init__.py
@@ -9,6 +9,7 @@ import typing as t
 import logging
 import tempfile
 import contextlib
+from pathlib import Path
 
 import psutil
 from simple_di import inject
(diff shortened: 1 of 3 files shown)


bentoml/_internal/server/cli/dev_api_server.py
```diff
@@ -9,23 +9,33 @@
...
 
 @click.command()
 @click.argument("bento_identifier", type=click.STRING, required=False, default=".")
 @click.option("--bind", type=click.STRING, required=True)
-@click.option("--backlog", type=click.INT, default=2048)
 @click.option("--working-dir", required=False, type=click.Path(), default=None)
+@click.option("--backlog", type=click.INT, default=2048)
+@click.option(
+    "--prometheus-dir",
+    type=click.Path(exists=True),
```
sauyon, line 22: Don't we ensure the directory exists and create it now?

bojiang: Here could be multiple workers, if we do the init here, there would be a potential contest. Then we will have to introduce a file lock, and remember to clean it if an exception happens.

bentoml/_internal/server/cli/api_server.py
```diff
@@ -42,6 +42,11 @@
     type=click.Path(exists=True),
     help="Working directory for the API server",
 )
+@click.option(
```
parano, line 45: this will require changes on Yatai deployment operator side as well, right? @yetone @bojiang

parano: looks like Yatai does not run api_server workers from here?

I guess when we add metrics to runner_app, we will need to update the runner_app CLI command to include a prometheus directory?

bentoml/_internal/server/__init__.py
```diff
@@ -31,17 +32,45 @@
 
 @inject
 def ensure_prometheus_dir(
-    prometheus_multiproc_dir: str = Provide[
-        DeploymentContainer.prometheus_multiproc_dir
-    ],
+    directory: str = Provide[DeploymentContainer.prometheus_multiproc_dir],
```
parano, line 35: it seems to me the root cause of this issue might be we are using a fixed directory for all BentoServer on the same host: 

```
    prometheus_multiproc_dir = providers.Factory[str](
        os.path.join,
        bentoml_container.bentoml_home,
        "prometheus_multiproc_dir",
    )
```

Should we change it to have a `serve_id` or something similar in the file path? similar to the serve_id in the ServeInfo tracking event?

bentoml/_internal/server/cli/api_server.py
bojiang: No. This option is not required. On the other hand, the e2e test of simulated yatai passed so no worry. In Yatai the default directory will always work, so no need to specify it.

bentoml/_internal/server/__init__.py
bojiang: yeah, I believe this is the best practice here

bojiang: Yeah it can be a solution. But it's a strange coupling that seems good, I think. The `prometheus_multiproc_dir` is just similar to a common pid file in Linux daemons. I think totally randomly generated in the `/var` or `/tmp` and passed to children is the best practice.

aarnphm:
@bojiang if this is crucial for 1.0, maybe update this?

ssheng (approved):

aarnphm (dismissed):

aarnphm (approved):

aarnphm (approved):

PR #2533, opened 2022-06-01
feat(server): provide different backend for `--reload`

aarnphm:
depends on #2657 for `Bento.includes`

This PR provides different backend for `--reload`: `default` (python implementation), `watchfiles`
- refactor our circus watchers for development server to different backend
- file changes are respected via `bentofile.yaml` include and exclude as well as `.bentoignore`

to use `--reload` user should install `pip install bentoml[watchfiles]`.

```bash
bentoml serve service.py:svc --reload
```
https://github.com/samuelcolvin/watchfiles


bentoml/_internal/utils/circus.py
```diff
@@ -21,10 +33,10 @@ def create_circus_socket_from_uri(
     *args: t.Any,
     name: str = "",
     **kwargs: t.Any,
-) -> CircusSocket:
-    from circus.sockets import CircusSocket  # type: ignore[reportMissingTypeStubs]
+) -> CircusSocket | t.NoReturn:  # pragma: no cover (test coverred by circus)
```
sauyon, line 36: Isn't `NoReturn` for when a function will *never* return?

bentoml/_internal/utils/circus.py
```diff
@@ -61,94 +75,160 @@ def create_standalone_arbiter(watchers: list[Watcher], **kwargs: t.Any) -> Arbit
...
+            # We just need include and exclude fields to create spec.
+            # Hence, service can be an empty string.
+            build_config = BentoBuildConfig(service="").with_defaults()
+
+        include = build_config.include
+        exclude = build_config.exclude
+
+        # bentoml/_internal/bento/bento.py#L191
+        self.include_spec = PathSpec.from_lines("gitwildmatch", include)
+        self.exclude_spec = PathSpec.from_lines("gitwildmatch", exclude)
```
sauyon, line 203: Hm, should we factor this out into its own method? Or maybe attach it to the `BentoBuildConfig`?

sauyon: Something like `BentoBuildConfig.includes_file(filename)` maybe would be nice.

setup.cfg
```diff
@@ -71,6 +71,7 @@ install_requires =
     simple-di>=0.1.4
     starlette
     typing-extensions>=4.0
+    watchfiles>=0.14.1
```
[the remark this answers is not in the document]
sauyon: What does this do better than the current implementation? Maybe we could have this as a feature that can be enabled by installing watchfiles if it's much better?

bentoml/_internal/utils/circus.py
aarnphm: sounds good

setup.cfg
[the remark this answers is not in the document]
aarnphm: since they listen file changes from os file notification rather than polling from Python, which makes it more performant

bentoml/_internal/utils/circus.py
aarnphm: `NoReturn` is for function when it raises an exception

bojiang: @aarnphm actually not. Every function may raise exceptions. `NoRetuen` means a function never returns. Constantly raising an exception is just one of the cases. The case hers should not use `NoReturn`

bentoml/_internal/utils/__init__.py
```diff
@@ -201,8 +201,10 @@ def reserve_free_port(
     else:
         sock.bind((host, 0))
     port = sock.getsockname()[1]
-    yield port
-    sock.close()
+    try:
+        yield port
```
bojiang, line 205: Why do we have this line? yield will always success right?

bentoml/_internal/utils/circus.py
```diff
@@ -61,94 +75,160 @@ def create_standalone_arbiter(watchers: list[Watcher], **kwargs: t.Any) -> Arbit
...
 
     Args:
         working_dir: The directory of the bento service.
-        reload_delay: The delay in seconds between checking for changes.
     """
 
     name = "bento_change_reloader"
     config: dict[str, t.Any]
 
+    @inject
```
bojiang, line 90: It seems an unnecessary inject here

bentoml/_internal/utils/circus.py
```diff
@@ -61,94 +75,160 @@ def create_standalone_arbiter(watchers: list[Watcher], **kwargs: t.Any) -> Arbit
...
+
+        self.watchers = watchfiles.watch(
+            *self.watch_dirs,
+            watch_filter=None,
+            yield_on_timeout=True,  # to stop hanging on our tests, doesn't affect the behaviour
+            rust_timeout=self.watcher_timeout,  # no need to setup timeout
+            stop_event=self.exit_event,
+        )
+
+    def filtered(
```
bojiang, line 131: Is it possible to make use of the internal filter of watchfiles?

bentoml/_internal/utils/circus.py
aarnphm: Got it let me update accordingly

bentoml/_internal/utils/circus.py
aarnphm: we want to format the output correctly, hense we can't use the internal `watch_filter` here

bentoml/_internal/cli/bento_server.py
```diff
@@ -92,28 +92,38 @@ def serve(
...
         working_dir: str,
         run_with_ngrok: bool,
     ) -> None:
-        """Start BentoServer from BENTO
+        """Start a :code:`BentoServer` from a given ``BENTO`` 🍱
+
+        ``BENTO`` is the serving target, it can be the import as:
+            - a path of a :code:`bentoml.Service` instance
+            - a tag to a Bento in local Bento store
+            - a file path to a Bento directory
```
parano, line 104:
```suggestion
            - the import path of a :code:`bentoml.Service` instance
            - a tag to a Bento in local Bento store
            - a folder containing a valid `bentofile.yaml` build file with a `service` field, which provides the import path of a :code:`bentoml.Service` instance
            - a path to a built Bento
```

bentoml/_internal/cli/bento_server.py
```diff
@@ -92,28 +92,38 @@ def serve(
...
+            :code:`bentoml serve fraud_detector:4tht2icroji6zput3suqi5nl2`
+            :code:`bentoml serve fraud_detector:latest`
 
         \b
         Serve from a Bento directory:
-            bentoml serve ./fraud_detector_bento
+            :code:`bentoml serve ./fraud_detector_bento`
+
+        \b
+        If :code:`--reload` is parsed, BentoML will detect file changes to restart the service.
```
parano, line 122:
```suggestion
        If :code:`--reload` is provided, BentoML will detect code and model store changes during development, and restart the service automatically.
```

bentoml/_internal/cli/bento_server.py
```diff
@@ -92,28 +92,38 @@ def serve(
...
+
+            The behaviour is as follows:
+            - BentoML will respect :obj:`include` and :obj:`exclude` specfied under :obj:`bentofile.yaml`.
+            - If :obj:`.bentoignore` is present, then it will be respected after filtering from :obj:`bentofile.yaml`.
```
parano, line 126:
```suggestion
            The `--reload` flag will:
            - be default, all file changes under `--working-dir` (default to current directory) will trigger a restart
            - when specified, respect :obj:`include` and :obj:`exclude` under :obj:`bentofile.yaml` as well as the :obj:`.bentoignore` file in `--working-dir`, for code and file changes
            - all model store changes will also trigger a restart (new model saved or existing model removed)
```

setup.cfg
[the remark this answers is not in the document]
sauyon: FWIW I don't think this is particularly important for the development server.

bentoml/_internal/utils/circus/watchfilesplugin.py
```diff
@@ -0,0 +1,138 @@
...
+        # a thread to restart circus
+        self.exit_event = Event()
+
+        self.file_changes = watch(
+            *self.watch_dirs,
+            watch_filter=None,
+            yield_on_timeout=True,  # NOTE: stop hanging on our tests, doesn't affect the behaviour
+            stop_event=self.exit_event,
+            rust_timeout=0,  # NOTE: we should set reload_delay to 0 to avoid unnecessary logs
+        )
```
ssheng, line 73: What is the performance issues of watchfiles if the watch set is very large?

bentoml/_internal/bento/build_config.py
```diff
@@ -702,6 +703,54 @@ def to_yaml(self, stream: t.TextIO) -> None:
...
+    _exclude: PathSpec = attr.field(
+        converter=lambda x: PathSpec.from_lines("gitwildmatch", x)
+    )
+    # we want to ignore .git folder in cases the .git folder is very large.
+    git: PathSpec = attr.field(
+        default=PathSpec.from_lines("gitwildmatch", [".git"]), init=False
+    )
+
+    def __init__(self, build_config: "FilledBentoBuildConfig") -> None:
+        return self.__attrs_init__(build_config.include, build_config.exclude)  # type: ignore (unfinished __attr_init__ type)
```
ssheng, line 720: Let's pass in just what we need, the include and exclude fields, to be specific.

bentoml/_internal/bento/build_config.py
```diff
@@ -702,6 +703,54 @@ def to_yaml(self, stream: t.TextIO) -> None:
         raise NotImplementedError
 
 
+@attr.define(frozen=True, init=False)
+class BentoPatternSpec:
```
ssheng, line 707: BentoPathSpec?

requirements/tests-requirements.txt
```diff
@@ -12,5 +12,5 @@ pytest-asyncio
 pandas
 scikit-learn
 imageio>=2.5.0
-docker
```
ssheng, line 15: Do we want to add `>=0.15.0`

ssheng (commented):
Great usability improvement.

bentoml/_internal/utils/circus/watchfilesplugin.py
aarnphm: i haven't explored that part just yet.

aarnphm: There are none I believe.

ssheng (approved):

PR #2539, opened 2022-06-02
test: build config tests

aarnphm:

## Description

- added unit tests
- wait for #2581


bentoml/_internal/bento/docker/entrypoint.sh
```diff
@@ -18,6 +18,7 @@ _main() {
 	# Overide the BENTOML_PORT if PORT env var is present. Used for Heroku
 	if [[ -v PORT ]]; then
 		echo "\$PORT is set! Overiding \$BENTOML_PORT with \$PORT ($PORT)"
+		echo "NOTE: Yatai also set \$PORT, hence \$BENTOML_PORT will be overridden."
```
parano, line 21: move this to the inline comments above ? No need to log it

aarnphm: done

bentoml/_internal/bento/docker/entrypoint.sh
```diff
@@ -11,6 +11,7 @@ _is_sourced() {
 
 _main() {
 	# if first arg looks like a flag
+	echo "${1:0:1}"
```
parano, line 14: Remove?

bentoml/_internal/bento/docker/entrypoint.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -Eeuo pipefail
+set -Eeuox pipefail
```
parano, line 2: Why adding x here?

aarnphm: I was debugging, removed now

bentoml/_internal/bento/docker/entrypoint.sh
```diff
@@ -17,6 +17,7 @@ _main() {
 
 	# Overide the BENTOML_PORT if PORT env var is present. Used for Heroku
 	if [[ -v PORT ]]; then
+		# NOTE: Yatai also set \$PORT, hence \$BENTOML_PORT will be overridden.
```
parano, line 20: Simply update the comment above: 

	# Overide the BENTOML_PORT if PORT env var is present. Used for Heroku **and Yatai**

bentoml/_internal/bento/gen.py
```diff
@@ -21,11 +6,13 @@
 import logging
 from typing import TYPE_CHECKING
 
+import fs
 import attr
 from jinja2 import Environment
 from jinja2.loaders import FileSystemLoader
 
-from ..utils import bentoml_cattr
+from ..utils import bentoml_cattr as cattr
```
parano, line 14: let's not do this, it's very confusing - we should make it very clear it is our cattr instance

parano: note that there are still places where we use cattr directly.

bentoml/_internal/bento/gen.py
```diff
@@ -84,28 +83,89 @@ class CustomizableEnv:
...
-def get_template_env(
-    docker_options: DockerOptions, spec: DistroSpec
+def validate_user_template(template: Template, loader: BaseLoader) -> None | t.NoReturn:
+    """
+    Validate all user-defined templates in the given environment.
+    """
+    ctx = template.new_context()
+    if "bento__dockerfile" not in ctx.keys():
+        exc_info = (
+            f"User-defined template `{template}` does not contain `bento__dockerfile`. "
```
parano, line 113: why is that a problem?  user should be able to generate everything from scratch without our base templates right?

parano: the user could also be just extending one block? right?

requirements/tests-requirements.txt
```diff
@@ -12,5 +12,5 @@ pytest-asyncio
 pandas
 scikit-learn
 imageio>=2.5.0
-docker
```
parano, line 15: why removing docker?

parano: fs is already a bentoml dependency, no need to add it here in tests dependency

bentoml/_internal/utils/__init__.py
```diff
@@ -211,9 +211,20 @@ def copy_file_to_fs_folder(
     dst_folder_path: str = ".",
     dst_filename: t.Optional[str] = None,
 ):
-    """Copy the given file at src_path to dst_fs filesystem, under its dst_folder_path
-    folder with dst_filename as file name. When dst_filename is None, keep the original
+    """
```
parano, line 214: probably no need to change this at all - i'd prefer not using `:code:` for internal docs and Args section doesn't really add any value

bentoml/_internal/utils/__init__.py
```diff
@@ -223,9 +234,22 @@ def copy_file_to_fs_folder(
     fs.copy.copy_file(src_fs, file_name, dst_fs, dst_path)
 
 
-def resolve_user_filepath(filepath: str, ctx: t.Optional[str]) -> str:
-    """Resolve the abspath of a filepath provided by user, which may contain "~" or may
+def resolve_user_filepath(filepath: str, ctx: t.Optional[str]) -> str | t.NoReturn:
+    """
```
parano, line 238: same here

bentoml/_internal/bento/gen.py
aarnphm: actually we need they to have our base template to make sure that bento still works

requirements/tests-requirements.txt
aarnphm: we don't use docker anymore

.gitattributes
```diff
@@ -1,2 +1,3 @@
 typings/**/*.pyi linguist-generated=true
+tests/unit/_internal/bento/testdata/**/*.yaml linguist-generated=true
```
parano: these YAML files should probably not be linguist-generated=true

bentoml/_internal/utils/__init__.py
aarnphm: it actually helps with lsp hover

bentoml/_internal/bento/build_config.py
```diff
@@ -74,11 +65,13 @@ def _convert_python_version(py_version: t.Optional[str]) -> t.Optional[str]:
     return f"{major}.{minor}"
 
 
-def _convert_cuda_version(cuda_version: t.Optional[str]) -> t.Optional[str]:
+def _convert_cuda_version(
+    cuda_version: t.Optional[t.Union[str, float, int]]
```
parano, line 69: float?
```suggestion
    cuda_version: t.Optional[t.Union[str, int]]
```

bentoml/_internal/bento/build_config.py
```diff
@@ -451,10 +529,10 @@ def write_to_bento(self, bento_fs: "FS", build_ctx: str) -> None:
             # Note: "--generate-hashes" is purposefully not used here because it will
             # break if user includes PyPI package from version control system
             pip_compile_in = bento_fs.getsyspath(
-                fs.path.join(py_folder, "requirements.txt")
+                fs.path.combine(py_folder, "requirements.txt")
```
parano, line 532: Would be nice to document this type of changes in PR in a comment like this:
---

Combine: Join two paths together.

This is faster than [join()](https://docs.pyfilesystem.org/en/latest/reference/path.html#fs.path.join), but only works when the second path is relative, and there are no back references in either path.

https://docs.pyfilesystem.org/en/latest/reference/path.html#fs.path.combine

tests/conftest.py
```diff
@@ -10,16 +14,20 @@
     framework_name="testing", framework_versions={"testing": "v1"}
 )
 
+if TYPE_CHECKING:
+    from _pytest.tmpdir import TempPathFactory
```
parano, line 18: what's the difference between this and `from pytest import TempPathFactory`?

bentoml/_internal/bento/build_config.py
aarnphm: for example `11.6` should works as well

tests/conftest.py
aarnphm: oh the autocompletion imports from _pytest

aarnphm: Screenshot 2022-06-05 at 17 45 00

aarnphm:
changes for bentoctl is https://github.com/bentoml/BentoML/pull/2547. the same files are kept here for testing purposes. 2547 should get merged first

.gitattributes
aarnphm: fixed

aarnphm:
TODO @aarnphm:
- add command args validation for pip-tools

aarnphm:
close this for now, since we have done many manual QA. I will take a look into https://github.com/HypothesisWorks/hypothesis for possible property-based test. Will revisit this after 1.0

cc @ssheng.

PR #2549, opened 2022-06-06
style(runner): revert b14919db(factor out batching)

bojiang:
## Description

Discussed with sauyon before. It's unnecessary to factor that batching logic out while the Params utils already provided the function. I didn't do that because I'm concerned about the checking will increase the overhead of the runner inference callback. Compared to type safety, lighter inference callback usually takes precedence.

Still do the checking now, but we need to consider disabling it if there's too much CPU overhead.

## Checklist:


bentoml/_internal/runner/batching.py
```diff
@@ -1,71 +0,0 @@
...
-
-    if len(paramss[0].args) > 0:
-        arg_idxs = iter(range(len(paramss[0].args)))
-
-        idx = next(arg_idxs)
-        batched, first_indices = AutoContainer.from_batch_payloads(
-            [params.args[idx] for params in paramss],
-            batch_dim=batch_dim,
-        )
-        args.append(batched)
```
bojiang, line 34: We should prefer
* generators & iterators
* then list comprehension
* then `append` or `expand`
in performance sensitive scenarios

PR #2556, opened 2022-06-07
chore(ci): cover all the characters of a tag in e2e

bojiang:
## Checklist:

### DIFF
--- a/bentoml/testing/server.py
+++ b/bentoml/testing/server.py
@@ -112,7 +112,7 @@ def bentoml_build(project_path: str) -> t.Generator["Tag", None, None]:
     logger.info(f"Building bento: {project_path}")
     output = subprocess.check_output(["bentoml", "build", project_path])
     match = re.search(
-        r'Bento\(tag="(\w+:[a-z0-9]+)"\)',
+        r'Bento\(tag="([A-Za-z0-9\-_\.]+:[a-z0-9]+)"\)',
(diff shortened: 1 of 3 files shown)


bentoml/_internal/server/instruments.py
```diff
@@ -38,6 +38,10 @@ def _setup(
     ):
         self.metrics_client = metrics_client
         service_name = self.bento_service.name
+        # a valid tag name may includes invalid characters, so we need to escape them
+        # ref: https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels
+        service_name = service_name.replace("-", ":").replace(".", "::")
```
parano, line 43: instead of ":", could we replace them with just underscore? "_"

it seems very uncommon for metric names to contain ":" https://github.com/prometheus/docs/blob/main/content/docs/practices/naming.md

parano, line 43: instead of ":", could we replace them with just underscore? "_"

it seems very uncommon for metric names to contain ":" https://github.com/prometheus/docs/blob/main/content/docs/practices/naming.md

bojiang: If so, there would be a collision between service "service-1" and "service_1", since "_" is valid for bentoml tag. But ":" is not. Thus it is the best replacement.

PR #2568, opened 2022-06-08
feat: improve logging

sauyon:
@yetone requesting a review just in case changing the server CLI arguments affects yatai.


bentoml/_internal/cli/click_utils.py
```diff
@@ -88,15 +90,14 @@ def wrapper(
...
                 if verbose:
-                    logger.warning(
-                        "The bentoml command option `--verbose/--debug` is ignored when"
-                        "the `--quiet` flag is also in use"
-                    )
+                    logger.warning("'--quiet' passed; ignoring '--verbose/--debug'")
             elif verbose:
                 set_debug_mode(True)
 
+            configure_logging()
```
aarnphm, line 99: does this mean that when user use our API from python, logging won't get configured?

sauyon: Indeed.

aarnphm: should we also configure logs for python API separately?

sauyon: I believe the decision was made because logging configuration is mainly an application's job and not a library's; https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library.

sauyon:
Removed places where paths were being printed to console output, and fixed the CLI handler to actually print stuff.

bentoml/_internal/configuration/__init__.py
```diff
@@ -99,6 +95,18 @@ def get_debug_mode() -> bool:
     return False
 
 
+def set_quiet_mode(enabled: bool) -> None:
+    os.environ[DEBUG_ENV_VAR] = str(enabled)
+
+    # do not log setting quiet mode
```
parano, line 101: should it also call `configure_logging` here?  If the user calls this after a `configure_logging`, the env var won't be applied to the logging module?

sauyon: I removed the `configure_logging` calls from both of these since when users use this as a library `configure_logging` should never be called.

bentoml/_internal/cli/model_management.py
```diff
@@ -116,7 +116,6 @@ def list_models(model_name: str, output: str, no_trunc: bool) -> None:
             {
                 "tag": str(model.tag),
                 "module": model.info.module,
-                "path": display_path_under_home(model.path),
```
parano, line 119: also need to remove the `path` usage below, it will raises a KeyError

bentoml/_internal/configuration/__init__.py
sauyon: We could have a flag that checks whether `configure_logging` has been called and call this again if the user has called that, but I think this is ok; `configure_logging` is an internal API right now anyway.

parano: Sounds good, if users need to customize the logging, they can probably do it via the standard logging library as well.

bojiang (approved):

PR #2573, opened 2022-06-09
fix(build): use subprocess to call pip-compile

sauyon:
This fixes the issue where importing pip-tools was causing the root logger to be configured.

### DIFF
--- a/bentoml/_internal/bento/build_config.py
+++ b/bentoml/_internal/bento/build_config.py
@@ -1,7 +1,9 @@
 import os
 import re
+import sys
 import typing as t
 import logging
+import subprocess
 from sys import version_info as pyver
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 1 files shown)


bentoml/_internal/bento/build_config.py
```diff
@@ -472,10 +473,11 @@ def write_to_bento(self, bento_fs: "FS", build_ctx: str) -> None:
                 ]
             )
             logger.info("Locking PyPI package versions..")
-            click_ctx = pip_compile_cli.make_context("pip-compile", pip_compile_args)
+            cmd = [sys.executable, "-m", "piptools", "compile"]
+            cmd.extend(pip_compile_args)
             try:
-                pip_compile_cli.invoke(click_ctx)
-            except Exception as e:
+                subprocess.check_call(cmd)
```
aarnphm, line 479: I think check_call doesn't work on windows? Maybe use check_output instead here?

aarnphm (approved):
LGTM if this needs to get in fast. I will update the BuildConfig PR accordingly

bentoml/_internal/bento/build_config.py
sauyon: Oh, what's the issue with `check_call` on windows?

aarnphm: oh nvm it was to related to this.

PR #2576, opened 2022-06-09
chore(runner): add shcedule shortcuts to runners

bojiang:
* reuse code
* should be useful for writing streaming udf from a runner


aarnphm (approved):
LGTM

PR #2579, opened 2022-06-10
fix(cli): cli encoding error on Windows

bojiang:
## Description

We are using `subprocess.check_output().encode()` to evoke other commands. But for Windows cmd.exe, the default encoding is not utf-8. So all features calling other commands like that will fail by default on Windows now.

## Checklist:

### DIFF
--- a/bentoml/_internal/cli/__init__.py
+++ b/bentoml/_internal/cli/__init__.py
@@ -1,4 +1,5 @@
 import click
+import psutil
 
 from bentoml import __version__ as BENTOML_VERSION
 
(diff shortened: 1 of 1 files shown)


bentoml/_internal/utils/buildx.py
```diff
@@ -35,8 +36,10 @@ def health() -> None:
     """
     cmds = DOCKER_BUILDX_CMD + ["--help"]
     try:
-        output = subprocess.check_output(cmds, stderr=subprocess.STDOUT)
-        assert "--builder string" in output.decode("utf-8")
+        output = subprocess.check_output(cmds, stderr=subprocess.STDOUT).decode(
+            sys.stdout.encoding
+        )
+        assert "--builder string" in output
```
parano, line 42: @aarnphm just curious, in what case the output may not contain this? and should we provide a better error message here?

aarnphm: and we should provide better error message here for sure

aarnphm (approved):

PR #2580, opened 2022-06-10
fix: update conda env yaml file name and default channel

parano:
## Description

* Using conda "defaults" channel may require user to have a commercial license, this PR changes the default channel to the community maintained "conda-forge" channel
* fix the generated conda yaml file path
* update related docs

## Checklist:

- [ ] My code follows the bentoml code style, both `make format` and
  `make lint` script have passed
  ([instructions](https://github.com/bentoml/BentoML/blob/master/DEVELOPMENT.md#style-check-and-auto-formatting-your-code)).
- [ ] My change reduces project test coverage and requires unit tests to be added
- [ ] I have added unit tests covering my code change
- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly

PR #2581, opened 2022-06-10
qa: build config

aarnphm:

## Description

refactor from #2539, easier for review (probably need to merge this first then
I will try to run test again)

some notes about this PR:
- fixes `convert_python_version` regex to support python version that is
  installed from beta (e.g: 3.11.0b1)
- environment variables uses regex to validate instead of loosely check for
  capitalization

`DockerOptions`:
- `system_packages` should not return and empty list, otherwise generated
  dockerfile will contains an empty line that tries to install system_packages

`CondaOptions`:
- add validator for `dependencies`:
  dependencies can be used:
    ```yaml
    dependencies:
      - numpy
      pip: 
        - spacy
    ```
- update structure hooks to handle dependencies correctly.
- defaults value should always be None

NOTE: should update default channels changes after #2580 got merged

`PythonOptions`:
- set default value to `None` if not specified

- updates hooks accordingly
- changelog: move from `fs.path.join` to `fs.path.combine`, refers to [here]( https://docs.pyfilesystem.org/en/latest/reference/path.html#fs.path.combine)


bentoml/_internal/bento/build_config.py
```diff
@@ -450,22 +525,28 @@ def write_to_bento(self, bento_fs: "FS", build_ctx: str) -> None:
...
         # write pip install args to a text file if applicable
         if pip_args:
-            with bento_fs.open(fs.path.join(py_folder, "pip_args.txt"), "w") as f:
+            with bento_fs.open(fs.path.combine(py_folder, "pip_args.txt"), "w") as f:
                 f.write(" ".join(pip_args))
 
-        if self.lock_packages:
+        if self.lock_packages:  # pragma: no cover
+            # This section is not covered by tests and we relies
+            # on jazzband/pip-tools test coverage
```
parano, line 533: we can still test this section by verifying the right pip-tools commands/APIs are invoked, also possible to cover it in integration or e2e tests.

bentoml/_internal/bento/build_config.py
```diff
@@ -450,22 +525,28 @@ def write_to_bento(self, bento_fs: "FS", build_ctx: str) -> None:
...
             # Note: "--allow-unsafe" is required for including setuptools in the
             # generated requirements.lock.txt file, and setuptool is required by
             # pyfilesystem2. Once pyfilesystem2 drop setuptools as dependency, we can
             # remove the "--allow-unsafe" flag here.
 
             # Note: "--generate-hashes" is purposefully not used here because it will
             # break if user includes PyPI package from version control system
+
+            # changelog: move from fs.path.join to fs.path.combine
+            # refers to https://docs.pyfilesystem.org/en/latest/reference/path.html#fs.path.combine
```
parano, line 544: remove this? we can document this in PR description

bentoml/_internal/bento/build_config.py
```diff
@@ -554,9 +635,11 @@ class BentoBuildConfig:
...
 
     def __attrs_post_init__(self) -> None:
-        use_conda = any(
-            val is not None for _, val in bentoml_cattr.unstructure(self.conda).items()
-        )
+        use_conda = False
+        for attr_ in [a.name for a in attr.fields(self.conda.__class__)]:
+            if getattr(self.conda, attr_) is not None:
+                use_conda = True
+                break
```
parano, line 642: move this to conda options:

```
   def is_used(self):
          return not self == CondaOptions()
```

`is not None` may not be the right check here, it should verify if it's the default value. it could break if a new option being added has a default value of an empty list.

bentoml/_internal/bento/build_config.py
```diff
@@ -570,7 +653,8 @@ def __attrs_post_init__(self) -> None:
                     f"{self.docker.distro} does not supports conda. BentoML will only support conda with the following distros: {get_supported_spec('miniconda')}."
                 )
             if use_cuda and self.docker.distro not in get_supported_spec("cuda"):
-                raise BentoMLException(
+                # This is helpful when we support more CUDA version
```
parano, line 656: what does this comment mean?

bentoml/_internal/bento/build_dev_bentoml_whl.py
```diff
@@ -62,6 +62,6 @@ def build_bentoml_editable_wheel(target_path: str) -> None:
             # copy the built wheel file to target directory
             shutil.copytree(dist_dir, target_path)  # type: ignore
     else:
-        logger.info(
+        logger.info(  # pragma: no cover
```
parano, line 65: why "no cover" here? We can come back and add tests for this, no reason to exclude it

bentoml/_internal/bento/build_config.py
```diff
@@ -110,28 +107,29 @@ def _envars_validator(
...
-
     if isinstance(value, list):
-        for envar in value:
-            match = envars_rgx.match(envar)
-            if match is None:
-                raise InvalidArgument(f"{envar} doesn't follow format ENVAR=value")
-
-            k, _ = envar.split("=")
-            if not k.isupper():
+        for envar in value:  # pragma: no cover
```
parano, line 111: remove `pragma: no cover`?

bentoml/_internal/bento/build_config.py
```diff
@@ -110,28 +107,29 @@ def _envars_validator(
...
+            #  we have tests for this cases, but pytest refuse to cover
```
parano, line 112: it's ok if pytest fail to report the test coverage here, let's just report the issue upstream. no need to comment about it here.

bentoml/_internal/bento/build_config.py
```diff
@@ -110,28 +107,29 @@ def _envars_validator(
...
                 )
     elif isinstance(value, dict):
-        if not all(k.isupper() for k in value):
-            raise InvalidArgument("All envars should be in UPPERCASE.")
+        uppercase_rgx = re.compile(r"(^[A-Z_]+\w*)")
+        filter_non_envar: list[str] = []
+        for k, v in value.items():
+            if not uppercase_rgx.match(k):
+                raise InvalidArgument(
+                    "All keys in `env` dict must be UPPERCASE and start with a letter, e.g: `ENV_123: value`"
```
parano, line 124: env var written in uppercase is just a convention, no need to enforce it BentoML.

bentoml/_internal/bento/build_config.py
```diff
@@ -296,12 +272,52 @@ def _docker_options_structure_hook(d: t.Any, _: t.Type[DockerOptions]) -> Docker
...
+
+def conda_dependencies_validator(
+    _: t.Any, __: "Attribute[DependencyType]", value: DependencyType
+) -> None:
+    if not isinstance(value, list):
+        raise InvalidArgument(
+            f"Expected `conda.dependencies` to be type `list`, got type: `{type(value)}` instead"
+        )
+    else:
+        conda_pip: "t.List[CondaPipType]" = [x for x in value if isinstance(x, dict)]
```
parano, line 290: isinstance list?

bentoml/_internal/bento/build_config.py
```diff
@@ -296,12 +272,52 @@ def _docker_options_structure_hook(d: t.Any, _: t.Type[DockerOptions]) -> Docker
...
+        if conda_pip:
+            try:  # need to test this since conda didn't cover this :(
+                if len(conda_pip) > 1:
+                    raise InvalidArgument(
+                        "Expected dictionary under `conda.dependencies` to ONLY have key `pip`"
+                    )
+                pip_list: t.List[str] = conda_pip[0]["pip"]
```
parano, line 297: verify `"pip" in conda_pip[0]` before this line

bentoml/_internal/bento/build_config.py
```diff
@@ -296,12 +272,52 @@ def _docker_options_structure_hook(d: t.Any, _: t.Type[DockerOptions]) -> Docker
...
+    _: t.Any, __: "Attribute[DependencyType]", value: DependencyType
+) -> None:
+    if not isinstance(value, list):
+        raise InvalidArgument(
+            f"Expected `conda.dependencies` to be type `list`, got type: `{type(value)}` instead"
+        )
+    else:
+        conda_pip: "t.List[CondaPipType]" = [x for x in value if isinstance(x, dict)]
+        if conda_pip:
+            try:  # need to test this since conda didn't cover this :(
```
parano, line 292:
```suggestion
            try:
```

bentoml/_internal/bento/build_config.py
```diff
@@ -296,12 +272,52 @@ def _docker_options_structure_hook(d: t.Any, _: t.Type[DockerOptions]) -> Docker
...
+                    raise InvalidArgument(
+                        "Expected dictionary under `conda.dependencies` to ONLY have key `pip`"
+                    )
+                pip_list: t.List[str] = conda_pip[0]["pip"]
+                if not all(isinstance(x, str) for x in pip_list):
+                    raise InvalidArgument("Expected `pip` values to be type `str`")
+            except KeyError:
+                raise InvalidArgument(
+                    "Conda dependencies can ONLY include `pip` value as a dictionary."
+                )
```
parano, line 303: use `"pip" in conda_pip[0]` check instead of KeyError

bentoml/_internal/bento/build_config.py
```diff
@@ -369,29 +385,87 @@ def with_defaults(self) -> "CondaOptions":
...
+                pip_from_deps = dep.pop("pip")
+                pip_list.extend(pip_from_deps)
+        deps = list(filter(lambda x: not isinstance(x, dict), deps))
+        d["dependencies"] = deps
+    d["pip"] = None if not pip_list else pip_list
+
+    return CondaOptions(**d)
+
+
+bentoml_cattr.register_structure_hook(CondaOptions, _conda_options_structure_hook)
```
parano, line 412: this is probably not necessary, I think we can just store the CondaOptions the way the user specified it.

bentoml/_internal/bento/build_config.py
```diff
@@ -592,7 +676,7 @@ def __attrs_post_init__(self) -> None:
                     if self.docker.cuda_version != "default" and (
                         self.docker.cuda_version not in _spec.supported_cuda_versions
                     ):
-                        raise BentoMLException(
+                        raise BentoMLException(  # pragma: no cover
```
parano, line 679: why no cover?

bentoml/_internal/bento/build_config.py
aarnphm: seems like we can do `__eq__` for all of the options class?

bentoml/_internal/bento/build_config.py
aarnphm: let me remove that this is not related.

bentoml/_internal/bento/build_config.py
aarnphm: I have tests for this, but it doesn't seem to add it into coverage.

aarnphm: could mean that this statement is never reached?

sauyon: If this is using branch coverage it might be that value never has length 0.

bentoml/_internal/bento/build_config.py
sauyon: I concur; we probably shouldn't worry about pragma no cover until we're at 100% coverage.

bentoml/_internal/bento/build_config.py
```diff
@@ -110,28 +107,29 @@ def _envars_validator(
...
+            envars_rgx = re.compile(r"(^[A-Z_]+\w*)(\=)(.*)")
+            if not envars_rgx.match(envar):
                 raise InvalidArgument(
-                    f'Invalid env var defined, key "{k}" should be in uppercase.'
+                    "All value in `env` list must follow format ENVAR=value"
                 )
     elif isinstance(value, dict):
-        if not all(k.isupper() for k in value):
-            raise InvalidArgument("All envars should be in UPPERCASE.")
+        uppercase_rgx = re.compile(r"(^[A-Z_]+\w*)")
```
sauyon, line 119: This doesn't ensure uppercase right now, only that the name starts with an uppercase character. But I agree we probably shouldn't enforce this.

bentoml/_internal/bento/build_config.py
sauyon: This one doesn't seem resolved?

bentoml/_internal/bento/build_config.py
sauyon: Unresolved this one too, are there commits pending?

sauyon (commented):
Bunch of comments I thought I already posted :sweat_smile:

bentoml/_internal/bento/build_config.py
aarnphm: noted.

bentoml/_internal/bento/build_config.py
```diff
@@ -26,24 +26,22 @@
 from ...exceptions import BentoMLException
 from .build_dev_bentoml_whl import build_bentoml_editable_wheel
 
-if pyver >= (3, 8):
+if version_info >= (3, 8):
     from typing import Literal
-else:
+else:  # pragma: no cover
```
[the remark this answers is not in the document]
aarnphm: I will cover it

bentoml/_internal/bento/build_config.py
```diff
@@ -52,21 +50,14 @@
...
-
 def _convert_python_version(py_version: t.Optional[str]) -> t.Optional[str]:
     if py_version is None:
         return None
 
     if not isinstance(py_version, str):
         py_version = str(py_version)
 
-    match = re.match(r"^(\d+).(\d+)", py_version)
+    match = re.match(r"^(\d+)\.(\d+)(?:|.(?:\d+|\w+))$", py_version)
```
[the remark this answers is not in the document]
aarnphm: I think with your recommendation it would match loosely right? So `3.11.0.a` would work, tho there wouldn't be a python version for that?

bentoml/_internal/bento/build_config.py
```diff
@@ -284,24 +191,76 @@ def write_to_bento(
...
+else:
+    DependencyType = list
+
+
+def conda_dependencies_validator(
+    _: t.Any, __: "Attribute[DependencyType]", value: DependencyType
+) -> None:
+    if not isinstance(value, list):
+        raise InvalidArgument(
+            f"Expected `conda.dependencies` to be type `list`, got type: `{type(value)}` instead"
```
sauyon, line 235:
```suggestion
            f"Expected 'conda.dependencies' to be a list of dependencies, got a '{type(value)}' instead."
```

bentoml/_internal/bento/build_config.py
```diff
@@ -450,22 +525,28 @@ def write_to_bento(self, bento_fs: "FS", build_ctx: str) -> None:
 
         # write pip install args to a text file if applicable
         if pip_args:
-            with bento_fs.open(fs.path.join(py_folder, "pip_args.txt"), "w") as f:
+            with bento_fs.open(fs.path.combine(py_folder, "pip_args.txt"), "w") as f:
                 f.write(" ".join(pip_args))
 
-        if self.lock_packages:
+        if self.lock_packages:  # pragma: no cover
```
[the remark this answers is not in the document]
sauyon: ^

bentoml/_internal/bento/build_config.py
sauyon: Though I think this is an understandable no cover if we were aiming for 100% test coverage (obvious error cases probably don't need tests) it should probably be justified.

bentoml/_internal/bento/build_config.py
```diff
@@ -52,35 +50,30 @@
...
 
-    match = re.match(r"^(\d+).(\d+)", py_version)
+    match = re.match(r"^(\d+)\.(\d+)(?:\.\w+)?", py_version)
     if match is None:
         raise InvalidArgument(
             f'Invalid build option: docker.python_version="{py_version}", python '
             f"version must follow standard python semver format, e.g. 3.7.10 ",
         )
-    major, minor = match.groups()
+    major, minor = match.groups()[:2]
```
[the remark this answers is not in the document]
aarnphm: ugh this would match everything after 3.11 as well, so match would be `[3, 11, ""]`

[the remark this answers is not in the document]
sauyon: Why? There's no third capture group in the regex.

bentoml/_internal/bento/build_config.py
```diff
@@ -450,22 +448,23 @@ def write_to_bento(self, bento_fs: "FS", build_ctx: str) -> None:
 
         # write pip install args to a text file if applicable
         if pip_args:
-            with bento_fs.open(fs.path.join(py_folder, "pip_args.txt"), "w") as f:
+            with bento_fs.open(fs.path.combine(py_folder, "pip_args.txt"), "w") as f:
                 f.write(" ".join(pip_args))
 
-        if self.lock_packages:
+        if self.lock_packages:  # pragma: no cover
```
aarnphm, line 454:
```suggestion
        if self.lock_packages:
```

bentoml/_internal/bento/build_config.py
```diff
@@ -238,26 +155,16 @@ def with_defaults(self) -> "DockerOptions":
...
-        if self.dockerfile_template is None:
-            defaults["dockerfile_template"] = None
+                python_version = f"{version_info.major}.{version_info.minor}"
+                defaults["python_version"] = python_version
 
         return attr.evolve(self, **defaults)
 
     def write_to_bento(
-        self, bento_fs: "FS", build_ctx: str, conda_options: "CondaOptions"
+        self, bento_fs: "FS", build_ctx: str, conda_cls: "CondaOptions"
```
parano, line 164: this is an instance right? `_cls` usually mean it's a Class

bentoml/_internal/bento/build_config.py
```diff
@@ -285,24 +192,84 @@ def write_to_bento(
 
 def _docker_options_structure_hook(d: t.Any, _: t.Type[DockerOptions]) -> DockerOptions:
```
parano, line 193: move this to a convertor, remove the `bentoml_cattr.register_structure_hook` below

bentoml/_internal/bento/build_config.py
```diff
@@ -285,24 +192,84 @@ def write_to_bento(
...
+    if "env" in d and d["env"] is not None:
+        if isinstance(d["env"], str):
+            if not os.path.isfile(d["env"]):
+                raise BentoMLException(f"Invalid env file path: {d['env']}")
             d["env"] = dotenv_values(d["env"])
-        except FileNotFoundError:
-            raise BentoMLException(f"Invalid env file path: {d['env']}")
+        elif isinstance(d["env"], list):
+            for envar in d["env"]:
+                if not envar.contains("="):
```
parano, line 202:
```
AttributeError: 'str' object has no attribute 'contains'
```

let's add a unit test for it

bentoml/_internal/bento/build_config.py
```diff
@@ -369,30 +340,64 @@ def with_defaults(self) -> "CondaOptions":
...
+    extra_index_url: t.Optional[t.List[str]] = attr.field(
+        default=None,
+        validator=attr.validators.optional(attr.validators.instance_of(list)),
+    )
+    pip_args: t.Optional[str] = attr.field(
+        default=None,
+        validator=attr.validators.optional(
+            lambda _, __, value: isinstance(value, str)
+            and re.match(r"^(?:--)\w*", value)
+        ),
```
parano, line 386: this doesn't seem to work

bentoml/_internal/bento/build_config.py
```diff
@@ -147,52 +95,24 @@ class DockerOptions:
...
-    # A python or sh script that executes during docker build time
     setup_script: t.Optional[str] = None
-
-    # A user-provided custom docker image
     base_image: t.Optional[str] = None
-
-    # A user-provided dockerfile jinja2 template
     dockerfile_template: str = attr.field(
         default=None,
         validator=attr.validators.optional(
```
parano, line 118: this will fail if Dockerfiles is provided as a relative path and user provided a build_ctx

bentoml/_internal/bento/build_config.py
parano: a validator doesn't return bool, it should raise value error

bentoml/_internal/bento/docker.py
```diff
@@ -18,7 +18,13 @@
 # Python supported versions
 SUPPORTED_PYTHON_VERSIONS = ["3.7", "3.8", "3.9", "3.10"]
 # CUDA supported versions
-SUPPORTED_CUDA_VERSIONS = {"11": "11.6.2"}
+SUPPORTED_CUDA_VERSIONS = ["11.6.2"]
```
parano, line 21: made this a list of CUDA versions that can actually be installed on distro. The ALLOWED_CUDA_VERSION_ARGS provides a map from user provided version string to the actual version to be installed

bentoml/_internal/bento/build_config.py
```diff
@@ -293,6 +302,9 @@ def __attrs_post_init__(self):
                 )
 
     def write_to_bento(self, bento_fs: "FS", build_ctx: str) -> None:
+        if self.is_empty():
```
parano, line 305: we no longer create an empty `conda` folder if conda options are not provided

bentoml/_internal/bento/build_config.py
```diff
@@ -58,33 +59,74 @@ def _convert_python_version(py_version: t.Optional[str]) -> t.Optional[str]:
...
+
+def _convert_env(
+    env: t.Optional[t.Union[str, t.List[str], t.Dict[str, str]]]
+) -> t.Optional[t.Dict[str, str]]:
+    if env is None:
+        return None
+
+    if isinstance(env, str):
+        logger.debug("Reading dot env file '%s' specified in config", env)
+        return dict(dotenv_values(env))
```
parano, line 109: dotenv_values returns an OrderedDict and it causes issues with pyyaml - thus converting it to `dict`

bentoml/_internal/bento/build_config.py
```diff
@@ -58,33 +59,74 @@ def _convert_python_version(py_version: t.Optional[str]) -> t.Optional[str]:
...
+            py_version,
+        )
+    return target_python_version
 
 
 def _convert_cuda_version(
     cuda_version: t.Optional[t.Union[str, int]]
 ) -> t.Optional[str]:
-    if cuda_version is None or cuda_version == "":
+    if cuda_version is None or cuda_version == "" or cuda_version == "None":
```
parano, line 82: cuda_version will be str "None" if user accidentally has `cuda_version: null` in the bentofile.yaml

bentoml/_internal/bento/build_config.py
```diff
@@ -380,10 +397,7 @@ class PythonOptions:
     )
     pip_args: t.Optional[str] = attr.field(
         default=None,
-        validator=attr.validators.optional(
-            lambda _, __, value: isinstance(value, str)
-            and re.match(r"^(?:--)\w*", value)
-        ),
+        validator=attr.validators.optional(attr.validators.instance_of(str)),
```
parano, line 400: I removed the regex check here since it doesn't add too much value and tricky to get it exactly right.

for future reference, we could be using something like this:

```python
        validator=attr.validators.optional([
            attr.validators.instance_of(str),
            attr.validators.matches_re(r"\s*(-{1,2}[^ =]+(=[^ =]+){0,1}[\s\n]*)+"),
        ]),
    )
```

bentoml/_internal/bento/build_config.py
```diff
@@ -409,17 +423,16 @@ def write_to_bento(self, bento_fs: "FS", build_ctx: str) -> None:
         with bento_fs.open(fs.path.join(py_folder, "version.txt"), "w") as f:
             f.write(f"{version_info.major}.{version_info.minor}.{version_info.micro}")
 
+        # Build BentoML whl from local source if BENTOML_BUNDLE_LOCAL_BUILD=True
+        build_bentoml_editable_wheel(bento_fs.getsyspath(wheels_folder))
```
parano, line 427: build_bentoml_editable_wheel uses shutil.copytree underneath which requires the target directory to be empty. thus moving it to above the `bento_fs.makedirs(wheels_folder, recreate=True)` line following

bentoml/_internal/bento/build_config.py
```diff
@@ -58,33 +59,74 @@ def _convert_python_version(py_version: t.Optional[str]) -> t.Optional[str]:
...
         raise InvalidArgument(
             f'Invalid build option: docker.python_version="{py_version}", python '
             f"version must follow standard python semver format, e.g. 3.7.10 ",
         )
     major, minor = match.groups()
-    return f"{major}.{minor}"
+    target_python_version = f"{major}.{minor}"
+    if target_python_version != py_version:
+        logger.warning(
+            "BentoML will install the latest python%s instead of the specified version %s. To use the exact python version, use a custom docker base image. See https://docs.bentoml.org/en/latest/concepts/bento.html#custom-base-image-advanced",
```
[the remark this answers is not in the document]
aarnphm: 👍🏼

bentoml/_internal/bento/docker/templates/miniconda_debian.j2
```diff
@@ -14,7 +14,7 @@ if [ -f {{ __environment_yml__ }} ]; then
   # set pip_interop_enabled to improve conda-pip interoperability. Conda can use
   # pip-installed packages to satisfy dependencies.
   echo "Updating conda base environment with environment.yml"
-  conda config --set pip_interop_enabled True || true
+  conda config --set pip_interop_enabled True
```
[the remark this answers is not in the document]
aarnphm: the `|| true` is when the case where we can't set the config, it won't fail the docker build process

bentoml/_internal/bento/build_config.py
```diff
@@ -595,12 +599,12 @@ class BentoBuildConfig:
...
 
     def __attrs_post_init__(self) -> None:
-        use_conda = not self.conda.is_default()
+        use_conda = not self.conda.is_empty()
         use_cuda = self.docker.cuda_version is not None
 
         if use_cuda and use_conda:
-            logger.warning(
-                f'Conda will be ignored and will not be available inside bento container when "cuda={self.docker.cuda_version}" is set.'
+            raise BentoMLException(
```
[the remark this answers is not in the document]
aarnphm: sg

bentoml/_internal/bento/build_config.py
```diff
@@ -491,26 +501,32 @@ def write_to_bento(self, bento_fs: "FS", build_ctx: str) -> None:
...
+        if self.pip_args:
+            pip_args.extend(self.pip_args.split())
+
+        with bento_fs.open(fs.path.combine(py_folder, "install.sh"), "w") as f:
             args = " ".join(map(quote, pip_args)) if pip_args else ""
-            install_python_packages = (
+            install_script_content = (
                 """\
 #!/usr/bin/env bash
+set -ex
```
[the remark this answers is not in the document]
aarnphm: noted

bentoml/_internal/bento/build_config.py
aarnphm: maybe we can revisit this and have a better null handling @aarnphm

bentoml/_internal/bento/build_config.py
```diff
@@ -283,26 +231,55 @@ def write_to_bento(
...
+    if not isinstance(value, list):
+        raise InvalidArgument(
+            f"Expected 'conda.dependencies' to be a list of dependencies, got a '{type(value)}' instead."
+        )
+    else:
+        conda_pip: "t.List[CondaPipType]" = [x for x in value if isinstance(x, dict)]
+        if len(conda_pip) > 0:
+            if len(conda_pip) > 1 or "pip" not in conda_pip[0]:
+                raise InvalidArgument(
+                    "Expected dictionary under `conda.dependencies` to ONLY have key `pip`"
```
aarnphm, line 253:
```suggestion
            if "pip" not in conda_pip[0]:
                invalid_keys=[k for k in conda_pip[0] if k != "pip"]
                raise InvalidArgument(
                    f"Expected ONLY dependency names or 'pip' keys in 'conda.dependencies', found {', '.join(invalid_keys)} instead."
            if len(conda_pip) > 1:
                raise InvalidArgument(
                    "Expected ONLY one 'pip' key inside 'conda.dependencies'"
```

bentoml/_internal/bento/docker/templates/miniconda_debian.j2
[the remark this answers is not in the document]
parano: why would it fail to set the config?

bentoml/_internal/bento/docker/templates/miniconda_alpine.j2
```diff
@@ -14,7 +14,7 @@ if [ -f {{ __environment_yml__ }} ]; then
   # set pip_interop_enabled to improve conda-pip interoperability. Conda can use
   # pip-installed packages to satisfy dependencies.
   echo "Updating conda base environment with environment.yml"
-  conda config --set pip_interop_enabled True || true
```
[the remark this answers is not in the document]
parano: could you explain why this command may fail?

bentoml/_internal/bento/docker/templates/base.j2
```diff
@@ -76,21 +74,11 @@ WORKDIR $BENTO_PATH
 COPY --chown={{ bento__user }}:{{ bento__user }} . ./
 
 {% endblock %}
-# Install bentoml=={{ bento__pypi_version }}
-ENV BENTOML_VERSION={{ bento__pypi_version }}
-RUN --mount=type=cache,mode=0777,target=/root/.cache/pip \
-    {{ bento__pip_cmd }} install bentoml==${BENTOML_VERSION} --no-cache-dir
```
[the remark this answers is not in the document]
parano: it has been refactored to the install.sh

bentoml/_internal/bento/docker.py
aarnphm: sg

bentoml/_internal/bento/docker/templates/miniconda_debian.j2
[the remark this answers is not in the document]
aarnphm: I guess we can just have what you are having for now, we will see if user reports fail. I believe we have `|| true` before. So I'm not sure if we should remove it.

bentoml/_internal/bento/build_config.py
```diff
@@ -484,26 +500,37 @@ def write_to_bento(self, bento_fs: "FS", build_ctx: str) -> None:
...
+            install_script_content = (
                 """\
 #!/usr/bin/env bash
+set -ex
 
-# https://stackoverflow.com/a/246128/8643197
+# Parent directory https://stackoverflow.com/a/246128/8643197
 BASEDIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]:-$0}"; )" &> /dev/null && pwd 2> /dev/null; )"
 
-PIP_ARGS=(-U --force-reinstall --no-warn-script-location """
```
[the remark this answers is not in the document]
aarnphm: -U is for upgrading repeated wheels dependencies, --no-warn-script-location is needed to silent unnecessary warning from pip install

bentoml/_internal/bento/build_config.py
```diff
@@ -588,25 +618,19 @@ class BentoBuildConfig:
...
-            val is not None for _, val in bentoml_cattr.unstructure(self.conda).items()
-        )
+        use_conda = not self.conda.is_empty()
         use_cuda = self.docker.cuda_version is not None
 
         if use_cuda and use_conda:
-            logger.warning(
-                f'Conda will be ignored and will not be available inside bento container when "cuda={self.docker.cuda_version}" is set.'
+            raise BentoMLException(
+                f"BentoML does not support using both conda dependencies and setting a CUDA version for GPU. If you need both conda and CUDA, use a custom base image or create a docker_template, see https://docs.bentoml.org/en/latest/concepts/bento.html#custom-base-image-advanced"
```
aarnphm, line 626:
```suggestion
                "BentoML does not support using both conda dependencies and setting a CUDA version for GPU. If you need both conda and CUDA, use a custom base image or create a docker_template, see https://docs.bentoml.org/en/latest/concepts/bento.html#custom-base-image-advanced"
```

Empty f-string :)

PR #2588, opened 2022-06-14
fix(runner): validate runner names

### LINKED ISSUE #2583: ValueError: invalid literal for int() with base 10: '$(circus.sockets.{runner_name}) when runner name has a space character in production mode
**Describe the bug**
A Runner object can be instantiated with a `name` argument. It seem as if this name contains a space seperator mentioned ValueError is raised when serving in production mode. As you can see in the screenshot, name of the runner is *Document Klassifikation*. I added some print statements starting with `>>>`:
image
Removing the space, `name="DocumentKlassifikation"` fixed the problem. 
I tracked the problem down to `/bentoml/_internal/server/cli/runner.py`. Seems the `runner_name` needs to be sanitized before using it for the bind I guess.

**To Reproduce**

Steps to reproduce the issue:
1. Create a Runner object 
2. set argument `name="Document Classification"
3. run `bentoml serve service.py --production`
4. See ValueError

**Expected behavior**
No ValueError is raised.

**Environment:**
 - OS: SLES 12 SP5
 - Python Version 3.10.4
 - BentoML Version bentoml, version 1.0.0rc0

sauyon:
Ensure that runner names are valid Python identifiers.

Resolves #2583.


sauyon:
@yetone is this sufficient for Yatai? Should we also require names to be < 63 bytes in length (punycode encoded?) so they can be used in a DNS subdomain?

yetone:
@sauyon yatai cannot use the runner name as the k8s resource name because, as you said, in addition to the runner name meeting the DNS naming rules, the resource name needs to be less than 63 in length, but the k8s resource name is a string that splices the bento deployment name and the runner name together, because the length of the bento deployment name is of unknown length, so it cannot be followed by a runner name of unknown length

sauyon:
Should we just require them both to be 31 chars or less?

yetone:
@sauyon If we want to include the runner name as part of the k8s resource name, it needs to follow the DNS 1035 standard

```go
const dns1035LabelFmt string = "[a-z]([-a-z0-9]*[a-z0-9])?"
```

sauyon:
Do you think that would make sense? I think our tags already follow that standard.

yetone:
@sauyon I don't think the Python identifier names follow the [DNS 1035](https://datatracker.ietf.org/doc/html/rfc1035). For example: `Abcd_` is a valid Python identifier, but it's not a valid k8s resource name

sauyon:
@yetone er, yeah, I meant the `Tag` validator; we could use that again here if you think that makes sense (and make the length limit smaller).

yetone:
@sauyon I think it make sense

yetone:
@sauyon We need to make sure that the bento deployment name length plus the model name length plus the model tag length cannot exceed a total of 63

bojiang:
Would this solution break users' workflow?

If a user just saved a model with an invalid runner name but a valid tag name, they would find their "to_runner" raising that exception, and didn't know what's happening.

sauyon:
> If a user just saved a model with an invalid runner name but a valid tag name, they would find their "to_runner" raising that exception, and didn't know what's happening.

I'm going to change this to Tag validation too, I think. We should have a quick chat about it when the two of you are available.

sauyon:
As per discussion, I've limited runner names to the same restrictions as `Tag`s.

Thus, currently tag name, version, and runner name are required to abide by the RFC 1123 naming restrictions, but the max length of each is 63, meaning yatai will have to hash combinations.

I think now that we're clear of Kubernetes restrictions we may want to start allowing user names to, say, include unicode or have arbitrary length, but let's leave the validation strict for now.

@yetone @bojiang @parano

bojiang:
@sauyon As long as they have the same restrictions, I'm okay with it.

sauyon:
@yetone If this looks good to you should we get this one merged?

bentoml/_internal/runner/runner.py
```diff
@@ -99,7 +100,13 @@ def __init__(
             method_configs: per method configs
         """
 
-        name = runnable_class.__name__ if name is None else name
+        if name is None:
+            name = runnable_class.__name__
+        elif not validate_tag_str(name):
```
parano, line 105:
```suggestion
        if name is None:
            name = runnable_class.__name__
        
        if not validate_tag_str(name):
```

parano: The runner name is used as a Kubernetes labels value, and here are the requirements: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set validate_tag_str is perfect for this check, but a Python class name may not meet the requirment

sauyon:
Needs a reformat, am out for dinner but can do when I come back.

bentoml/_internal/runner/runner.py
```diff
@@ -99,7 +100,14 @@ def __init__(
...
 
-        name = runnable_class.__name__ if name is None else name
+        if name is None:
+            name = runnable_class.__name__
+
+        if not validate_tag_str(name):
+            # TODO: link to tag validation documentation
+            raise ValueError(
+                f"Runner name '{name}' is not valid; it must be a valid BentoML Tag name."
+            )
```
ssheng, line 110:
```suggestion
        try:
            validate_tag_str(name)
        except ValueError as e:
            # TODO: link to tag validation documentation
            raise ValueError(
                f"Runner name '{name}' is not valid; it must be a valid BentoML Tag name."
            ) from e
```

ssheng (approved):

PR #2589, opened 2022-06-14
feat: Bring LightGBM back

mqk:
## Description
Implemented the LightGBM framework for the new 1.0.0 design.

## Motivation and Context
LightGBM support disappeared with rc0.
```
NotImplementedError: Support for "bentoml.lightgbm" is temporarily unavailable as BentoML transition to the new design in version 1.0.0 release.
Before this module is officially implemented in BentoML, users may use Custom Runner as a workaround. Learn more at http://docs.bentoml.org/
```

## How Has This Been Tested?
The test I ran are is the new one I introduced (tests/integration/frameworks/models/lightgbm.py).


.github/workflows/frameworks.yml
```diff
@@ -538,8 +538,7 @@ jobs:
 
   lightgbm_integration_tests:
     needs: diff
-    #if: ${{ (github.event_name == 'pull_request' && needs.diff.outputs.lightgbm == 'true') || github.event_name == 'push' }}
-    if: ${{ false }}
+    if: ${{ (github.event_name == 'pull_request' && needs.diff.outputs.lightgbm == 'true') || github.event_name == 'push' }}
```
mqk, line 541: I wasn't sure if this change is good or not, but the xgboost entry has it and I've been following xgboost as an example.

sauyon: This is fine! We just disabled all the tests for old frameworks by setting `if false`.

bentoml/_internal/frameworks/lightgbm.py
```diff
@@ -176,121 +166,96 @@ def save(
...
+            model = model.booster_
+        except AttributeError as e:
+            logger.error(
+                "Unable to obtain a lightgbm.basic.Booster from the specified model."
+            )
+            raise e
+
+    assert isinstance(
+        model, lgb.basic.Booster
+    ), "Unable to obtain a lightgbm.basic.Booster from the specified model."
```
mqk, line 189: In the previous implementation of LightGBM support we accepted either one of these scikit-learn wrapper objects or the lightgbm Booster directly. In the former case, we'd serialize the object with `joblib.dump` and in the latter by directly calling `.save_model`. I'm now following the XGBoost implementation and am _only_ supporting directly saving the model with `model.save_model`. However, I still want to allow the user to provide an sklearn wrapper object, so this code simply extracts the Booster object from it.

bentoml/_internal/frameworks/lightgbm.py
```diff
@@ -176,121 +166,96 @@ def save(
...
-
-    Returns:
-        :obj:`~bentoml._internal.runner.Runner`: Runner instances for :mod:`bentoml.lightgbm` model
-
-    Examples:
 
-    .. code-block:: python
+    class LightGBMRunnable(bentoml.Runnable):
+        SUPPORT_NVIDIA_GPU = False  # afaik LightGBM only supports GPU during training, not for inference.
+        SUPPORT_CPU_MULTI_THREADING = True
```
mqk, line 226: I don't believe LightGBM supports using a GPU for `.predict`, but please correct me if I'm mistaken!

Regarding cpu-multithreading support, I know LightGBM supports it, and I usually just have `n_jobs = -1` set in the model itself, which I believe utilizes all available cores. We could follow the XGBoost implementation and set `n_jobs` to the value of the envvar `OMP_NUM_THREADS`, but I don't know if that's really a common use case so I didn't implement it.

.github/workflows/frameworks.yml
```diff
@@ -113,7 +113,7 @@ jobs:
               - *runner
               - bentoml/lightgbm.py
               - bentoml/_internal/frameworks/lightgbm.py
-              - tests/integration/frameworks/test_lightgbm_impl.py
+              - tests/integration/frameworks/test_frameworks.py
```
aarnphm, line 116: we should add this under `related` keys so that all frameworks can have this by default.

bentoml/_internal/frameworks/lightgbm.py
```diff
@@ -28,119 +31,106 @@
...
-@inject
-def load(
-    tag: t.Union[str, Tag],
-    booster_params: t.Optional[t.Dict[str, t.Union[str, int]]] = None,
-    model_store: "ModelStore" = Provide[BentoMLContainer.model_store],
-) -> t.Union["lgb.basic.Booster", "_LightGBMModelType"]:
+def load_model(bento_model: str | Tag | bentoml.Model) -> lgb.basic.Booster:
     """
-    Load a model from BentoML local modelstore with given name.
+    Load the LightGNM model with the given tag from the local BentoML model store.
```
parano, line 68:
```suggestion
    Load the LightGBM model with the given tag from the local BentoML model store.
```

bentoml/_internal/frameworks/lightgbm.py
```diff
@@ -28,119 +31,106 @@
...
-    tag: t.Union[str, Tag],
-    booster_params: t.Optional[t.Dict[str, t.Union[str, int]]],
-    model_store: "ModelStore",
-) -> t.Tuple["Model", str, t.Dict[str, t.Any]]:
-    model = model_store.get(tag)
+    .. code-block:: python
+
+        import bentoml
+        # target model must be from the BentoML model store
+        model = bentoml.lightgbm.get("my_lightgbm_model")
```
parano, line 56:
```suggestion
        model = bentoml.lightgbm.get("my_lightgbm_model:latest")
```

bentoml/_internal/frameworks/lightgbm.py
```diff
@@ -176,121 +166,96 @@ def save(
             params, lgb_train, num_boost_round=20, valid_sets=lgb_eval
         )
 
+        # `save` the booster to BentoML modelstore:
         tag = bentoml.lightgbm.save("my_lightgbm_model", gbm, booster_params=params)
```
parano, line 170:
```suggestion
        tag = bentoml.lightgbm.save_model("my_lightgbm_model", gbm)
```

bentoml/_internal/frameworks/lightgbm.py
aarnphm: You are correct, LightGBM doesn't have support for GPU, hence no need to support GPU here.

Boosting library such as LightGBM or XGBoost uses openmp to implements multi-threading, and hence we can control the behaviour via `OMP_NUM_THREADS`. I believe that we can mimic the behaviour from xgboost here for lightgbm, since these two are mostly similar.

.github/workflows/frameworks.yml
sauyon: Probably after we convert all the framework tests over, though.

bentoml/_internal/frameworks/lightgbm.py
parano: BentoML will set OMP_NUM_THREADS for each Runner process based on 1. the CPU cores available, 2. the runner resource configuration and 3. the cgroup settings in the context of Docker and Kubernetes. So in this case, I think you are right, we don't need to set the n_jobs or num_threads parameter, BentoML's Runner scheduling_strategy will take care of this by setting the `OMP_NUM_THREADS` env var (because SUPPORT_CPU_MULTI_THREADING is set to True), and it will be the optimal default configuration for lightGBM.

bentoml/_internal/frameworks/lightgbm.py
```diff
@@ -176,121 +166,96 @@ def save(
...
-        "framework_name": "lightgbm",
-        "pip_dependencies": [f"lightgbm=={get_pkg_version('lightgbm')}"],
-    }
+    """  # noqa: LN001
+
+    # Ensure that `model` is actually the Booster object, and not for example one of the scikit-learn wrapper objects.
+    if not isinstance(model, lgb.basic.Booster):
+        # Deal with a LightGBM issue (https://github.com/microsoft/LightGBM/issues/3014)
+        if not hasattr(model, "fitted_"):
+            model.fitted_ = True
```
sauyon, line 177: Is this necessary if we're extracting the booster from the sklearn wrapper?

sauyon (dismissed):
Thanks so much for this! It looks great, sans all the (very) minor niggles.

We'd be happy to take over and apply all the little fix if you'd like us to.

bentoml/_internal/frameworks/lightgbm.py
mqk: So I have a LGBMClassifier model (sklearn-wrapper) that was trained and saved with LightGBM 3.0.0. When I unpickle it and try to directly access the booster, I get an exception:
```
In [2]: clf.booster_
---------------------------------------------------------------------------
NotFittedError                            Traceback (most recent call last)
Input In [2], in <cell line: 1>()
----> 1 clf.booster_

File ~/.virtualenvs/bentoml-lightgbm/lib/python3.8/site-packages/lightgbm/sklearn.py:854, in LGBMModel.booster_(self)
    852 """Booster: The underlying Booster of this model."""
    853 if not self.__sklearn_is_fitted__():
--> 854     raise LGBMNotFittedError('No booster found. Need to call fit beforehand.')
    855 return self._Booster

NotFittedError: No booster found. Need to call fit beforehand.
```

But the model is fitted of course. According to that issue linked to in the comment, this is due to some misconfiguration in the attributes, which was fixed in a release post 3.0.0.

mqk:
> We'd be happy to take over and apply all the little fix if you'd like us to.

That sounds good to me. Thanks!

bentoml/_internal/frameworks/lightgbm.py
aarnphm: Then should we set the support for lightgbm to be >=3.0.0? I feel like we don't want to coupling lightgbm logics here.

mqk: It would have to be >3.0.0 (not >=), because the issue is still present in 3.0.0. But that's fine, I can understand not wanting to couple to lightgbm internals. I guess we'll just have to ask people (like myself) that have a <=3.0.0 model to deal with this issue themselves. Not a big deal.

sauyon: I think it's fine to set fitted ourselves, I didn't realize `booster_` would actually check that it's fitted.

sauyon (approved):
Thanks again for this!

PR #2590, opened 2022-06-14
test(frameworks): minor test improvements

sauyon:
Minor update to the XGBoost tests to cleanup and skip GPU tests depending on the `SUPPORT_NVIDIA_GPU` flag.

Note that batching tests are enabled when there are multiple inputs, because we're setting `batchable` to false in most frameworks so I'm not going to skip that test when it's false.


parano:
LGTM!

bojiang (approved):

PR #2608, opened 2022-06-17
fix(model): load ModelOptions lazily

sauyon:
Also add a less opinionated base `ModelOptions`; that's not really necessary and I'm not actually sure I want it to exist. @ssheng thoughts?

### DIFF
--- a/bentoml/_internal/models/model.py
+++ b/bentoml/_internal/models/model.py
@@ -5,6 +5,7 @@ import typing as t
 import logging
 import importlib
 from sys import version_info as pyver
+from types import ModuleType
 from typing import overload
 from typing import TYPE_CHECKING
 from datetime import datetime
@@ -63,10 +64,6 @@ class ModelOptions:
 
     def to_dict(self: ModelOptions) -> dict[str, t.Any]:
         return attr.asdict(self)
-
-
-bentoml_cattr.register_structure_hook(ModelOptions, lambda d, cls: cls(**d))
-bentoml_cattr.register_unstructure_hook(ModelOptions, lambda v: v.to_dict())  # type: ignore  # pylint: disable=unnecessary-lambda # lambda required
 
 
 @attr.define(repr=False, eq=False, init=False)
@@ -307,8 +304,7 @@ class Model(StoreItem):
 
     def to_runnable(self) -> t.Type[Runnable]:
         if self._runnable is None:
-            module = importlib.import_module(self.info.module)
-            self._runnable = module.get_runnable(self)
+            self._runnable = self.info.imported_module.get_runnable(self)
         return self._runnable
 
     def with_options(self, **kwargs: t.Any) -> Model:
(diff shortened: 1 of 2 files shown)


bentoml/_internal/models/model.py
```diff
@@ -56,19 +58,25 @@ class ModelSignatureDict(t.TypedDict, total=False):
 CUSTOM_OBJECTS_FILENAME = "custom_objects.pkl"
 
 
-@attr.define
 class ModelOptions:
+    def with_options(self, **kwargs: t.Any) -> ModelOptions:
+        if len(kwargs) > 0:
+            raise InvalidArgument("This model does not support any options.")
```
ssheng, line 64: More accurately this "framework" does not support any options. Can we pass the`info.module` field into `ModelOptions` and include the module name in the exception message?

bentoml/_internal/models/model.py
```diff
@@ -532,6 +546,36 @@ def with_options(self, **kwargs: t.Any) -> ModelInfo:
             creation_time=self.creation_time,
         )
 
+    # cached_property doesn't support __slots__ classes
+    @property
+    def imported_module(self) -> ModuleType:
```
ssheng, line 551: May be we can get rid of `self._cached_module` if we use a `@lru_cache(maxsize=1)` decorator here.

ssheng (commented):
Thanks, Sauyon. I think this is a good and clean solution.

bentoml/_internal/models/model.py
```diff
@@ -56,19 +58,25 @@ class ModelSignatureDict(t.TypedDict, total=False):
...
+        if len(kwargs) > 0:
+            raise InvalidArgument("This model does not support any options.")
+        return self
+
+    def to_dict(self) -> dict[str, t.Any]:
+        return {}
+
+
+@attr.define
+class AttrsModelOptions(ModelOptions):
```
parano, line 72: add this to `bentoml.models`?

bentoml/_internal/models/model.py
aarnphm: i don't think that would work?

sauyon: I would think `lru_cache` also uses the dict, but will test this one out.

bentoml/_internal/frameworks/FRAMEWORK_TEMPLATE_PY
```diff
@@ -46,9 +46,12 @@ def get(tag_like: str | Tag) -> bentoml.Model:
...
 
 
-@attr.define(frozen=True)
 class FrameworkOptions(ModelOptions):
-    pass
+    def with_options(self, **kwargs: t.Any) -> FrameworkOptions:
+        ...
+
+    def to_dict(self) -> dict[str, t.Any]:
+        ...
```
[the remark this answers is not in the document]
aarnphm: Exporting `AttrsModelOptions` to `bentoml.models` would solve this issue

bentoml/_internal/models/model.py
sauyon: We probably could, let's do that.

bentoml/_internal/models/model.py
sauyon: Hm, after reading up on this `lru_cache` might not do exactly what we want; it'll cache externally over the `self` parameter without modifying the actual object so we'd have to either let it be unbounded, which would be a leak, or only cache a certain number.

aarnphm (approved):

Issue #2609, opened 2022-06-17
'ModelInfo' object has no attribute '__attrs_init__'


ebbunnim:
**Describe the bug**

`AttributeError` occurs when executing the sample code below. (1.0.0-rc1 version)
Please tell me how to solve it.

```
import bentoml
from sklearn import svm
from sklearn import datasets

iris = datasets.load_iris()
X, y = iris.data, iris.target

clf = svm.SVC(gamma='scale')
clf.fit(X, y)
bentoml.sklearn.save_model("iris_clf", clf)
```

**To Reproduce**

**Expected behavior**

save model normally

**Screenshots/Logs**

```
2022/06/17 11시 20분 00초 INFO     [cli] Using the default model signature for sklearn ({'predict': {'batchable': False}}) for model iris_clf.                                                                 
Traceback (most recent call last):
  File "pipeline.py", line 57, in <module>
    bentoml.sklearn.save_model("iris_clf", clf)
  File "/Users/jaylene/opt/anaconda3/envs/py38_pre/lib/python3.8/site-packages/bentoml/_internal/frameworks/sklearn.py", line 156, in save_model
    with bentoml.models.create(
  File "/Users/jaylene/opt/anaconda3/envs/py38_pre/lib/python3.8/contextlib.py", line 113, in __enter__
    return next(self.gen)
  File "/Users/jaylene/opt/anaconda3/envs/py38_pre/lib/python3.8/site-packages/bentoml/models.py", line 239, in create
    res = Model.create(
  File "/Users/jaylene/opt/anaconda3/envs/py38_pre/lib/python3.8/site-packages/bentoml/_internal/models/model.py", line 191, in create
    ModelInfo(
  File "/Users/jaylene/opt/anaconda3/envs/py38_pre/lib/python3.8/site-packages/bentoml/_internal/models/model.py", line 497, in __init__
    self.__attrs_init__(  # type: ignore
AttributeError: 'ModelInfo' object has no attribute '__attrs_init__'
```

**Environment:**
- OS: [MacOS 11.3.1]
- Python Version [Python 3.8.4]
- BentoML Version [BentoML-1.0.0-rc1]

**Additional context**

bojiang:
@ebbunnim 
Hi. I can not reproduce it. Could you please run the command `pip freeze` and paste the result here?

ebbunnim:
@bojiang 
here is the result
```
aiofiles==0.8.0
aiohttp==3.8.1
[...]

PR #2610, opened 2022-06-17
fix(model): lock attrs to >=21.1.0

bojiang:
## Description

fix: #2609 

## Checklist:

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,7 +38,7 @@ install_requires =
     aiofiles
     aiohttp<=3.8.1  # prevent it from installing preview release
     attrs>=21.1.0
-    cattrs
+    cattrs>=1.9.0
     chardet
     circus
     click>=7.0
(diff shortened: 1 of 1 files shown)


aarnphm:
We might need to check if this break cattrs.

bojiang:
I didn't find the `cattrs` version compatibility in its docs. We can only try our best. I'd lock it to >= 1.9.0 because of python 3.10 support

aarnphm:
If we locking attrs to 21.1.0, it might conflict with others library that is older attr.

aarnphm:
> @aarnphm the current bentoml rc1 already requires `attrs>= 21.1.0` in fact. This PR doesn't change that. If users complained about that, we could open a larger PR to refactor to lower version APIs of attrs.

Our current API uses `attr` which is the underlying api, so compatibility issue on that front is not worrying. 

Seems that newer attrs version will have some features that are incompatible. @sauyon any inputs on this?

sauyon:
It seems like `__attrs_init__` doesn't exist for older versions of attrs, this makes sense for me until somebody complains.

aarnphm (approved):

PR #2618, opened 2022-06-20
feat(wip): CLI autocompletion

aarnphm:

added zsh completion.

Ideally we want to extend click-completion, but right now click-completion is very slow and wouldn't understand how to autocomplete bento and models


aarnphm:
I will close this for now.

PR #2619, opened 2022-06-21
fix: create wheels directory when BENTOML_BUNDLE_LOCAL_BUILD=True

aarnphm:
## Description

- Wheels folder should be created when BENTOML_BUNDLE_LOCAL_BUILD=True
- set `dirs_exist_ok` since wheels folder will be created via `write_to_bento`
- address #2618

## Checklist:

### DIFF
--- a/bentoml/_internal/bento/build_config.py
+++ b/bentoml/_internal/bento/build_config.py
@@ -398,7 +398,7 @@ class PythonOptions:
     def write_to_bento(self, bento_fs: "FS", build_ctx: str) -> None:
         py_folder = fs.path.join("env", "python")
         wheels_folder = fs.path.join(py_folder, "wheels")
-        bento_fs.makedirs(py_folder, recreate=True)
+        bento_fs.makedirs(wheels_folder, recreate=True)
 
         # Save the python version of current build environment
         with bento_fs.open(fs.path.join(py_folder, "version.txt"), "w") as f:
--- a/bentoml/_internal/bento/build_dev_bentoml_whl.py
+++ b/bentoml/_internal/bento/build_dev_bentoml_whl.py
@@ -60,7 +60,7 @@ def build_bentoml_editable_wheel(target_path: str) -> None:
                 ],
             )
             # copy the built wheel file to target directory
-            shutil.copytree(dist_dir, target_path)  # type: ignore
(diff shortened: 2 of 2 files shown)


bentoml/_internal/bento/build_dev_bentoml_whl.py
```diff
@@ -1,9 +1,14 @@
...
-import shutil
 import logging
 import importlib
 import importlib.util
+from sys import version_info
+
+if version_info >= (3, 8):
+    from shutil import copytree
+else:
+    from backports.shutil_copytree import copytree
```
parano, line 11: why do we need this?

bentoml/_internal/bento/build_dev_bentoml_whl.py
```diff
@@ -60,7 +65,7 @@ def build_bentoml_editable_wheel(target_path: str) -> None:
                 ],
             )
             # copy the built wheel file to target directory
-            shutil.copytree(dist_dir, target_path)  # type: ignore
+            copytree(dist_dir, target_path, dirs_exist_ok=True)  # type: ignore
```
parano, line 68: could we change to use `shut.copy`?

bentoml/_internal/bento/build_dev_bentoml_whl.py
aarnphm: for 3.7, `dir_exists_ok` does not exist.

bentoml/_internal/bento/build_dev_bentoml_whl.py
aarnphm: copy will throw error here since the directory is already created. What we can do is just to move the wheels file into the target path?

aarnphm:
@parano

bentoml/_internal/bento/build_config.py
```diff
@@ -418,7 +418,10 @@ def __attrs_post_init__(self):
     def write_to_bento(self, bento_fs: "FS", build_ctx: str) -> None:
         py_folder = fs.path.join("env", "python")
         wheels_folder = fs.path.join(py_folder, "wheels")
-        bento_fs.makedirs(py_folder, recreate=True)
+
+        build_bentoml_editable_wheel(bento_fs.getsyspath(wheels_folder))
+
```
aarnphm, line 423:
```suggestion
```

Issue #2620, opened 2022-06-21
bentoml serve with transformer pipeline fails to set GPU device


thechaos16:
**Describe the bug**

`bentoml serve` with transformers pipeline fails running with GPU on the assigning device step.

The available GPU device collected by env variable (https://github.com/bentoml/BentoML/blob/main/bentoml/_internal/frameworks/transformers.py#L265-L273) is string, and it is not compatible with transformer's code (https://github.com/huggingface/transformers/blob/main/src/transformers/pipelines/base.py#L762-L765) because string is not comparable with integer.

**To Reproduce**

1. Save any transformer pipeline with bentoml
2. run `bentoml build` with gpu (exactly same as https://docs.bentoml.org/en/latest/guides/gpu.html#docker-images-options)
3. `bentoml containerize` and deploy the image
4. See error while starting `bentoml serve --production`

**Expected behavior**

Correctly assigns visible GPU and run inference server

**Screenshots/Logs**

Error logs
```
Traceback (most recent call last):             
File "/usr/local/lib/python3.8/dist-packages/starlette/routing.py", line 645, in lifespan            
 async with self.lifespan_context(app):           
File "/usr/local/lib/python3.8/dist-packages/starlette/routing.py", line 540, in __aenter__             
 await self._router.startup()                     
File "/usr/local/lib/python3.8/dist-packages/starlette/routing.py", line 624, in startup                
 handler()                                        
File "/usr/local/lib/python3.8/dist-packages/bentoml/_internal/runner/runner.py", line 204, in init_local                                           
 self._init_local()                               
File "/usr/local/lib/python3.8/dist-packages/bentoml/_internal/runner/runner.py", line 195, in _init_local                                          
 self._init(LocalRunnerRef)                       
File "/usr/local/lib/python3.8/dist-packages/bentoml/_internal/runner/runner.py", line 189, in _init    
 runner_handle = handle_class(self)               
File "/usr/local/lib/python3.8/dist-packages/bentoml/_internal/runner/runner_handle/local.py", line 23, in __init__                                          
 self._runnable = runner.runnable_class()         
File "/usr/local/lib/python3.8/dist-packages/bentoml/_internal/frameworks/transformers.py", line 293, in __init__                                             
 self.pipeline = load_model(bento_model, **kwargs)
File "/usr/local/lib/python3.8/dist-packages/bentoml/_internal/frameworks/transformers.py", line 158, in load_model                                           
[...]

PR #2624, opened 2022-06-22
fix: Transformers NVIDIA_VISIBLE_DEVICES value type casting

ssheng:
As reported in https://github.com/bentoml/BentoML/issues/2620, Transformers accepts an integer device ID. The current implementation pass the value of NVIDIA_VISIBLE_DEVICES environment variable as device without converting to an integer.

## Checklist:

### DIFF
--- a/bentoml/_internal/frameworks/FRAMEWORK_TEMPLATE_PY
+++ b/bentoml/_internal/frameworks/FRAMEWORK_TEMPLATE_PY
@@ -179,7 +179,7 @@ def get_runnable(bento_model: bentoml.Model) -> t.Type[bentoml.Runnable]:
                 # bento_model.info.options.*
             )
 
-            available_gpus = os.getenv("NVIDIA_VISIBLE_DEVICES")
+            available_gpus = os.getenv("CUDA_VISIBLE_DEVICES")
             if available_gpus is not None and available_gpus != "":
                 # assign GPU resources
                 ...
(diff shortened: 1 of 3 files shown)


aarnphm (approved):
We need to add test for this. Maybe unit test under `tests/unit/_internal/frameworks`

bentoml/_internal/frameworks/transformers.py
```diff
@@ -262,12 +262,17 @@ class TransformersRunnable(bentoml.Runnable):
         def __init__(self):
             super().__init__()
 
-            available_gpus = os.getenv("NVIDIA_VISIBLE_DEVICES")
+            available_gpus: str = os.getenv("NVIDIA_VISIBLE_DEVICES")
```
sauyon, line 265: Should we support `NVIDIA_VISIBLE_DEVICES=all` or `NVIDIA_VISIBLE_DEVICES=1,2`?

ssheng: My understanding is that breaking down `all` and `1,2` to individual devices is the job of the scheduling strategy.

bentoml/_internal/frameworks/FRAMEWORK_TEMPLATE_PY
```diff
@@ -179,7 +179,7 @@ def get_runnable(bento_model: bentoml.Model) -> t.Type[bentoml.Runnable]:
                 # bento_model.info.options.*
             )
 
-            available_gpus = os.getenv("NVIDIA_VISIBLE_DEVICES")
```
aarnphm, line 182: https://gitlab.com/nvidia/container-images/cuda/-/blob/master/templates/ubuntu/base-dockerfile.j2#L96

aarnphm (approved):

PR #2627, opened 2022-06-22
chore(pylint): update pylint configuration

sauyon:
New pylint versions no longer have a bunch of these configuration options.


sauyon:
We should probably pin pylint like we pin black, yeah.

ssheng (approved):

PR #2629, opened 2022-06-22
feat(framework): onnx support

larme:
## Checklist:


larme:
will add tests and docs soon

bentoml/onnx.py
```diff
@@ -1,7 +1,38 @@
...
+def load_runner(tag, *args, **kwargs):
+    if len(args) != 0 or len(kwargs) != 0:
+        logger.error(
+            f'The "{__name__}.load_runner" method is being deprecated. "load_runner" arguments will be ignored. Use `{__name__}.get("{tag}").to_runner()` instead'
+        )
+    else:
+        logger.warning(
+            f'The "{__name__}.load_runner" method is being deprecated. Use `{__name__}.get("{tag}").to_runner()` instead'
+        )
+    return get(tag).to_runner()
```
aarnphm, line 35: Since we don't have 1.0 support for ONNX yet, no need for this.
```suggestion
```

bentoml/_internal/frameworks/onnx.py
```diff
@@ -0,0 +1,392 @@
...
+            )
+
+            self.predict_fns: dict[str, t.Callable[..., t.Any]] = {}
+            for method_name in bento_model.info.signatures:
+                self.predict_fns[method_name] = getattr(self.model, method_name)
+
+
+    def _mapping(
+        item: ONNXArgType
+    ) -> "ext.NpNDArray":
```
aarnphm, line 347:
```suggestion
    def _mapping(item: ONNXArgType) -> "ext.NpNDArray":
```

tests/integration/frameworks/models/__init__.py
```diff
@@ -14,6 +14,15 @@ class FrameworkTestModel:
...
     save_kwargs: dict[str, t.Any] = attr.Factory(dict)
+    # when raw model method call is not simply `method(*args,
+    # **kwargs)` format or raw model method call does not simply
+    # return the outputs, then use this to override default behavior
+    # when testing raw model inputs with expected outputs
+    model_method_caller: t.Callable[
+        [FrameworkTestModel, str, list[t.Any], dict[str, t.Any]], t.Any
+    ] | None = attr.field(default=None)
+    # when framework has some special signatures requirements
+    model_signatures: dict[str, t.Any] | None = attr.field(default=None)
```
larme, line 25: @sauyon some changes for `FrameworkTest` for onnx or other more flexible frameworks.

tests/integration/frameworks/test_frameworks.py
```diff
@@ -54,9 +54,13 @@ def test_generic_arguments(framework: types.ModuleType, test_model: FrameworkTes
...
-    kwargs["signatures"] = {
-        "pytest-signature-rjM5": {"batchable": True, "batch_dim": (1, 0)}
-    }
+    default_meth = "pytest-signature-rjM5"
+    if test_model.model_signatures:
+        kwargs["signatures"] = test_model.model_signatures
+        meths = list(test_model.model_signatures.keys())
+    else:
+        kwargs["signatures"] = {default_meth: {"batchable": True, "batch_dim": (1, 0)}}
+        meths = [default_meth]
```
larme, line 63: cc @sauyon some changes for `FrameworkTest` for onnx or other more flexible frameworks.

tests/integration/frameworks/test_frameworks.py
```diff
@@ -139,7 +144,12 @@ def test_load(
                     key: inp.preprocess(kwarg)
                     for key, kwarg in inp.input_kwargs.items()
                 }
-                out = getattr(model, method)(*args, **kwargs)
+                if test_model.model_method_caller:
+                    out = test_model.model_method_caller(
+                        test_model, method, args, kwargs
+                    )
+                else:
+                    out = getattr(model, method)(*args, **kwargs)
```
larme, line 152: cc @sauyon some changes for `FrameworkTest` for onnx or other more flexible frameworks.

bentoml/_internal/frameworks/onnx.py
```diff
@@ -0,0 +1,397 @@
...
+        elif LazyType["ext.PdDataFrame"]("pandas.DataFrame").isinstance(item):
+            item = item.to_numpy(dtype=np.float32)
+        elif LazyType["tf.Tensor"]("tensorflow.Tensor").isinstance(item):
+            item = np.array(memoryview(item)).astype(np.float32, copy=False)  # type: ignore
+        elif LazyType["torch.Tensor"]("torch.Tensor").isinstance(item):
+            item = item.numpy().astype(np.float32, copy=False)
+        else:
+            raise TypeError(
+                f"`run` of {self.__class__.__name__} only takes "
+                "`numpy.ndarray` or `pd.DataFrame` as input parameters"
```
aarnphm, line 370:
```suggestion
                "`numpy.ndarray` or `pd.DataFrame`, `tf.Tensor`, or `torch.Tensor` as input parameters"
```

aarnphm (commented):
some comments I never sent :)

bentoml/_internal/frameworks/onnx.py
```diff
@@ -0,0 +1,392 @@
...
+def flatten_list(lst: t.List[t.Any]) -> t.List[str]:  # pragma: no cover
+    return [k for i in lst for k in _yield_providers(i)]
+
+
+
+@attr.define
+class ONNXOptions(ModelOptions):
+    """Options for the ONNX model"""
+    providers: list[str] | None = attr.field(default=None)
+    session_options: "ort.SessionOptions" | None = attr.field(default=None)
```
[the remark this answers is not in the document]
larme: I think here using an empty list or None is the same thing. We still need to change the value to `[CPUExecutionProvider]` in our code logic. Then I prefer to just using `None` so it's more clear that nothing is set for this field.

aarnphm:
Lets also address all the PEP8 bot warning. maybe we should look into pylint and bandit after 1.0 for code quality instead of pep8 bot :(

sauyon (commented):
Minor comments.

bentoml/_internal/frameworks/onnx.py
```diff
@@ -0,0 +1,403 @@
...
+    signatures: dict[str, ModelSignatureDict | ModelSignature] | None = None,
+    labels: dict[str, str] | None = None,
+    custom_objects: dict[str, t.Any] | None = None,
+    metadata: dict[str, t.Any] | None = None,
+) -> bentoml.Model:
+    """
+    Save a onnx model instance to the BentoML model store.
+
+    Args:
+        name (``str``):
```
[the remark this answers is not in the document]
sauyon: Whoops, forgot to remove this style comment.

bentoml/_internal/frameworks/onnx.py
```diff
@@ -0,0 +1,403 @@
...
+        raise NotFound(
+            f"Model {model.tag} was saved with module {model.info.module}, not loading with {MODULE_NAME}."
+        )
+    return model
+
+
+def load_model(
+    bento_model: str | Tag | bentoml.Model,
+    providers: ProvidersType | None = None,
+    session_options: "ort.SessionOptions" | None = None,
```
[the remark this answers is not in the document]
parano: yes I'd recommend use kwargs for all additional options in save_model

bentoml/_internal/frameworks/onnx.py
```diff
@@ -0,0 +1,403 @@
...
+        framework_versions={
+            "onnx": get_pkg_version("onnx"),
+            _onnxruntime_pkg: _onnxruntime_version,
+        },
+    )
+
+    if signatures is None:
+        signatures = {
+            "run": {"batchable": False},
+        }
```
ssheng, line 253: Is `run()` the only allowed signature? With run, calling it from a runner will need the following syntax.
```
runner.run.run()
```

bentoml/_internal/frameworks/onnx.py
```diff
@@ -0,0 +1,403 @@
...
+            raise TypeError(
+                f"`run` of {self.__class__.__name__} only takes "
+                "`numpy.ndarray` or `pd.DataFrame`, `tf.Tensor`, or `torch.Tensor` as input parameters"
+            )
+        return item
+
+    def add_runnable_method(method_name: str, options: ModelSignature):
+        def _run(self, *args: ONNXArgType) -> t.Any:
+            params = Params["ONNXArgType"](*args)
+            params = params.map(_mapping)
```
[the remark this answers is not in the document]
aarnphm: isn't that the same thing?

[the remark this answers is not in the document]
parano: they are a bit different, see  https://github.com/larme/BentoML/blob/cd124763d079dc73c7b83af52625f622d25f8239/bentoml/_internal/runner/utils.py#L63-L70

bentoml/_internal/frameworks/onnx.py
```diff
@@ -0,0 +1,403 @@
...
+                f"allowed method name is `run` instead of provided {provided_methods}"
+            )
+
+    if not isinstance(model, onnx.ModelProto):
+        raise TypeError(f"Given model ({model}) is not a onnx.ModelProto.")
+
+    if len(model.graph.output) > 1:
+        logger.warning(
+            "The onnx model you want to save has more than one "
+            "output. bentoml.onnx runner will only return the first output."
```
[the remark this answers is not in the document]
larme: I think return multiple output may break our batching behavior

bentoml/_internal/frameworks/onnx.py
```diff
@@ -0,0 +1,403 @@
...
+    import onnxruntime as ort
+except ImportError:  # pragma: no cover (trivial error checking)
+    raise MissingDependencyException(
+        """onnx is required in order to use module `bentoml.onnx`, install
+        onnx with `pip install onnx`. For more information, refer to
+        https://onnx.ai/get-started.html
+        `onnxruntime` or `onnxruntime-gpu` is also required by `bentoml.onnx`.
+        Refer to https://onnxruntime.ai/ for more information.
+        """
+    )
```
aarnphm, line 39: Ideally we should have two separate `ImportError` for two dependencies requirements like this. Any thoughts on this @ssheng?

bentoml/_internal/frameworks/onnx.py
```diff
@@ -0,0 +1,403 @@
...
+try:
+    import importlib.metadata as importlib_metadata
+except ImportError:
+    import importlib_metadata
+
+# prefer "onnxruntime-gpu" over "onnxruntime" for framework versioning
+_PACKAGE = ["onnxruntime-gpu", "onnxruntime"]
+for p in _PACKAGE:
+    try:
+        _onnxruntime_version = importlib_metadata.version(p)
```
aarnphm, line 69:
```suggestion
        _onnxruntime_version = get_pkg_version(p)
```

bentoml/_internal/frameworks/onnx.py
```diff
@@ -0,0 +1,403 @@
...
+    ProvidersType = list[str | t.Tuple[str, t.Dict[str, t.Any]]]
+    ONNXArgType = t.Union[
+        "ext.NpNDArray", "ext.PdDataFrame", "torch.Tensor", "tf_ext.Tensor"
+    ]
+
+
+try:
+    import importlib.metadata as importlib_metadata
+except ImportError:
+    import importlib_metadata
```
aarnphm, line 63:
```suggestion
```

bentoml/_internal/frameworks/onnx.py
```diff
@@ -0,0 +1,403 @@
...
+
+# prefer "onnxruntime-gpu" over "onnxruntime" for framework versioning
+_PACKAGE = ["onnxruntime-gpu", "onnxruntime"]
+for p in _PACKAGE:
+    try:
+        _onnxruntime_version = importlib_metadata.version(p)
+        _onnxruntime_pkg = p
+        break
+    except importlib_metadata.PackageNotFoundError:
+        pass
```
aarnphm, line 73: Since this whole components are being used only under `save_model()`, we should lazily put this check under `save_model` instead of here.

bentoml/_internal/frameworks/onnx.py
```diff
@@ -0,0 +1,403 @@
...
+    if model.info.module not in (MODULE_NAME, __name__):
+        raise NotFound(
+            f"Model {model.tag} was saved with module {model.info.module}, not loading with {MODULE_NAME}."
+        )
+    return model
+
+
+def load_model(
+    bento_model: str | Tag | bentoml.Model,
+    providers: ProvidersType | None = None,
```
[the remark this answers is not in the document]
aarnphm: resolved this during meetings yesterday.

bentoml/_internal/frameworks/onnx.py
```diff
@@ -0,0 +1,403 @@
...
+    session_options: "ort.SessionOptions" | None = None,
+) -> "ort.InferenceSession":
```
aarnphm, line 113:
```suggestion
) -> ort.InferenceSession:
```
House keeping.

bentoml/_internal/frameworks/onnx.py
```diff
@@ -0,0 +1,403 @@
...
+    if not isinstance(bento_model, bentoml.Model):
+        bento_model = get(bento_model)
+
+    if bento_model.info.module not in (MODULE_NAME, __name__):
+        raise NotFound(
+            f"Model {bento_model.tag} was saved with module {bento_model.info.module}, not loading with {MODULE_NAME}."
+        )
+
+    if providers:
+        if not all(i in ort.get_all_providers() for i in flatten_list(providers)):
```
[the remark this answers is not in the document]
aarnphm: ONNX sadly does not have a check for this. It is ok for now that we do this. We can make a PR upstream to ONNX later if needed.

bentoml/_internal/frameworks/onnx.py
aarnphm: `run` is the entrypoint of a `InferenceSession`.
To avoid name collision, we should make this an alias for signature
```python
runner.infer.run()
```
We can frozen the signature `infer` and forbid user to add `run` as a signature keys. If users provide a signature keys then we add this special signature to their signature dictionary.

With this we just need to mention how to use it in our docs with .
Thoughts @ssheng @larme ?

tests/integration/frameworks/test_frameworks.py
```diff
@@ -81,9 +81,13 @@ def test_generic_arguments(framework: types.ModuleType, test_model: FrameworkTes
...
     kwargs = test_model.save_kwargs.copy()
-    kwargs["signatures"] = {
-        "pytest-signature-rjM5": {"batchable": True, "batch_dim": (1, 0)}
-    }
+    default_meth = "pytest-signature-rjM5"
+    if test_model.model_signatures:
+        kwargs["signatures"] = test_model.model_signatures
+        meths = list(test_model.model_signatures.keys())
+    else:
+        kwargs["signatures"] = {default_meth: {"batchable": True, "batch_dim": (1, 0)}}
```
aarnphm, line 89:
```suggestion
    if test_model.model_signatures:
        kwargs["signatures"] = test_model.model_signatures
        meths = list(test_model.model_signatures.keys())
    else:
        default_meth = "pytest-signature-rjM5"
        kwargs["signatures"] = {default_meth: {"batchable": True, "batch_dim": (1, 0)}}
```

address sauyon comments

aarnphm (dismissed):
Final house keeping. Otherwise GTG

bentoml/_internal/frameworks/onnx.py
larme: @aarnphm  currently I just use `runner.run.run()` and explain it in doc. Do you want to add `runner.infer.run()` because `runner.run.run()` sounds a little strange?

bentoml/_internal/frameworks/onnx.py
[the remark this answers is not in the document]
larme: For `args` only it should be the same. But I want to keep the `Params` usage so onnx codes looks like pytorch and tensorflow. I think it's better to make our codes across similar frameworks look similar. However I have no strong opinion on this one

bentoml/_internal/frameworks/onnx.py
sauyon: Does `run` not get set as the default, meaning users can just use `runner.run()`?

tests/integration/frameworks/test_frameworks.py
larme: addressed

bentoml/_internal/frameworks/onnx.py
larme: I think we still need `importlib_metadata.PackageNotFoundError`

bentoml/_internal/frameworks/onnx.py
larme: @sauyon no because the original `runner.run` is overridden by runner method `run`

bentoml/_internal/frameworks/onnx.py
larme: @aarnphm moved the entire section (including `importlib_metadata` import part) to `save_model`

bentoml/_internal/frameworks/onnx.py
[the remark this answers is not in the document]
larme: Most ml model return multiple outputs inside a tuple or list, I think our DataContainer may not correctly handle these cases?

bentoml/_internal/frameworks/onnx.py
aarnphm: I guess `runner.run.run()` is confusing, since the first `run` is actually the signature of InferenceSession, where as the second `run` is the Runner's run

bentoml/_internal/frameworks/onnx.py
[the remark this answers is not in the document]
aarnphm: i see. @bojiang

bentoml/_internal/frameworks/onnx.py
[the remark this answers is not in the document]
aarnphm: understood. Let's keep it consistent among frameworks.

bentoml/_internal/frameworks/onnx.py
larme: will keep it this way now and allow `runner.run()` in later stage

bentoml/_internal/frameworks/onnx.py
ssheng: Even if we write them outside of a try block, it will throw on the first import error, right, not both? Maybe this is OK.

aarnphm (approved):

PR #2632, opened 2022-06-23
feat(runner): implement resource configuration

sauyon:


sauyon:
Undrafting for e2e tests.

sauyon:
This should be ready for review now.

tests/e2e/bento_server_general_features/tests/test_io.py
```diff
@@ -9,29 +8,10 @@
...
-
-
-@pytest.fixture()
-def bin_file(tmpdir):
-    bin_file_ = tmpdir.join("bin_file.bin")
-    with open(bin_file_, "wb") as of:
-        of.write("â".encode("gb18030"))
-    return str(bin_file_)
-
-
```
aarnphm, line 34: doesn't seem to be related to this PR? We can cleanup function later.

aarnphm, line 34: doesn't seem to be related to this PR? We can cleanup function later.

docs/source/concepts/runner.rst
```diff
@@ -138,10 +140,11 @@ and used in the same service. For example:
...
     import torch
 
-    class MyModelRunnable(bentoml.Runnable):
-        SUPPORT_NVIDIA_GPU = True
-        SUPPORT_CPU_MULTI_THREADING = True
-
+    class MyModelRunnable(
+        bentoml.Runnable
+        supported_resources={"nvidia.com/gpu"},
+        supports_multi_threading=True
```
aarnphm, line 146:
```suggestion
        bentoml.Runnable,
        supported_resources={"nvidia.com/gpu"},
        supports_multi_threading=True),
```

docs/source/concepts/runner.rst
```diff
@@ -114,11 +114,13 @@ Send a test request:
...
-threading. Since NLTK library doesn't support utilizing GPU nor multi-cpu-core natively,
-they are both set to False. This information is used by the BentoServer scheduler
-to determine the worker pool size of this runner.
+The :code:`supported_resources` class parameter indicates which resources this Runnable class
+implementation supports. The only currently pre-defined resource is ``"nvidia.com/gpu"``.
+:code:`supports_multi_threading` indicates whether or not the runner supports CPU multi-threading.
+Since the NLTK library doesn't support utilizing GPU or multiple CPU cores natively, no resources
+are specified, and ``supports_multi_threading`` is set to False. This is the default configuration.
+This information is used by the BentoServer scheduler to determine the worker pool size of this
+runner.
```
aarnphm, line 123:
```suggestion
The :code:`supported_resources` class parameter indicates which resources this Runnable class
implementation supports. The only currently pre-defined resource is ``"nvidia.com/gpu"``.

The :code:`supports_multi_threading` class parameter indicates whether or not the runner supports CPU multi-threading.

Since the NLTK library doesn't support utilizing GPU or multiple CPU cores natively, no resources
are specified, and ``supports_multi_threading`` is set to False. This is the default configuration.
This information is used by the BentoServer scheduler to determine the worker pool size of this
runner.
```

bentoml/_internal/resource.py
```diff
@@ -0,0 +1,279 @@
...
+class ResourceMeta(ABCMeta):
+    def __new__(  # pylint: disable=arguments-differ # we don't want to be generic
+        cls,
+        name: str,
+        bases: tuple[type],
+        clsdict: dict[t.Any, t.Any],
+        *,
+        resource_id: str,
+    ):
+        res = super().__new__(cls, name, bases, clsdict)
```
aarnphm, line 54: very very nit thing
```suggestion
        dct: dict[t.Any, t.Any],
        *,
        resource_id: str,
    ):
        res = super().__new__(cls, name, bases, dct)
```

bentoml/_internal/runner/runnable.py
```diff
@@ -24,19 +24,101 @@
...
+        cls,
+        name: str,
+        bases: t.Tuple[type, ...],
+        instDict: dict[t.Any, t.Any],
+        *,
+        supported_resources: Set[str] | None = None,
+        supports_multi_threading: bool | None = None,
+        **_kwargs: t.Any,
+    ) -> RunnableMeta:
+        res = super().__new__(cls, name, bases, instDict)
```
aarnphm, line 43: same as above
```suggestion
        dct: dict[t.Any, t.Any],
        *,
        supported_resources: Set[str] | None = None,
        supports_multi_threading: bool | None = None,
        **_kwargs: t.Any,
    ) -> RunnableMeta:
        res = super().__new__(cls, name, bases, dct)
```

aarnphm: then also maybe change all `instDict` to `dct` ?

aarnphm (commented):
First runthrough, some comments here and there.

tests/e2e/bento_server_general_features/tests/test_io.py
sauyon: Yeah, this commit should probably live in another PR.

bentoml/_internal/resource.py
sauyon: Thought about this one a bit and I think `attrs` is probably better than both.

tests/e2e/bento_server_general_features/tests/test_io.py
sauyon: Done.

bentoml/_internal/resource.py
sauyon: Named it `attr_dict` instead; I think that's decent?

bentoml/_internal/frameworks/FRAMEWORK_TEMPLATE_PY
```diff
@@ -167,10 +167,11 @@ def get_runnable(bento_model: bentoml.Model) -> t.Type[bentoml.Runnable]:
...
     """
 
-    class FrameworkRunnable(bentoml.Runnable):
-        SUPPORT_NVIDIA_GPU = True  # type: ignore
-        SUPPORT_CPU_MULTI_THREADING = True  # type: ignore
-
+    class FrameworkRunnable(
+        bentoml.Runnable,
+        supported_resources={"nvidia.com/gpu"},
+        supports_multi_threading=True
```
bojiang, line 173: as mentioned before, even GPUs have a multi_threading concept, too.
still suggest using "support_cpu_multi_threading" to be clear

bojiang (changes requested):
Meta classes introduce extra limitations for custom runnable classes. I think it doesn't worth the benefit.

bentoml/_internal/resource.py
aarnphm: I think it is a convention for writting __new__ to use dct right?

bentoml/_internal/runner/runnable.py
```diff
@@ -24,19 +24,101 @@
 RUNNABLE_METHOD_MARK: str = "_bentoml_runnable_method"
 
 
-class Runnable(ABC):
-    SUPPORT_NVIDIA_GPU: bool
-    SUPPORT_CPU_MULTI_THREADING: bool
+class RunnableMeta(ABCMeta, t.Type[t.Any]):
```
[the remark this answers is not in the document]
sauyon: Oh, how does ABCMeta normally resolve metaclass conflicts?

bentoml/_internal/resource.py
```diff
@@ -0,0 +1,279 @@
...
+        res = super().__new__(cls, name, bases, attr_dict)
+        if resource_id != "":
+            # register the class with RESOURCE_REGISTRY
+            assert issubclass(res, Resource)
+            _RESOURCE_REGISTRY[resource_id] = res
+
+        return res  # pylint: disable=useless-super-delegation (???)
+
+
+class Resource(t.Generic[T], metaclass=ResourceMeta, resource_id=""):
```
aarnphm, line 63: I feel like attrs class seems to be a nicer solution than having a metaclass. I guess the metaclass solution seems to be more magical (which i don't really have a strong opinion on) as this is more about how we want to maintain the code.

bentoml/_internal/resource.py
```diff
@@ -0,0 +1,279 @@
...
+    @abstractmethod
+    def from_spec(cls, spec: t.Any) -> T:
+        """
+        Get an instance of this resource from user input. For example, a CPU resource
+        might parse "10m" and return a CPU resource with 0.01 CPUs.
+        """
+
+    @classmethod
+    @abstractmethod
+    def from_system(cls) -> T:
```
aarnphm, line 74:
```suggestion
    def from_system(cls) -> "t.TypeGuard[T]":
```

bentoml/_internal/resource.py
```diff
@@ -0,0 +1,279 @@
...
+            assert issubclass(res, Resource)
+            _RESOURCE_REGISTRY[resource_id] = res
+
+        return res  # pylint: disable=useless-super-delegation (???)
+
+
+class Resource(t.Generic[T], metaclass=ResourceMeta, resource_id=""):
+    @classmethod
+    @abstractmethod
+    def from_spec(cls, spec: t.Any) -> T:
```
aarnphm, line 66:
```suggestion
    def from_spec(cls, spec: t.Any) -> "t.TypeGuard[T]":
```

bentoml/_internal/resource.py
```diff
@@ -0,0 +1,279 @@
...
+@functools.lru_cache(maxsize=1)
+def query_os_cpu_count() -> int:
+    cpu_count = os.cpu_count()
+    if cpu_count is not None:
+        return cpu_count
+    logger.warning("os.cpu_count() is NoneType")
+    return 1
+
+
+# class MemResource(Resource, resource_id="mem"):
```
aarnphm, line 186: Is this part not working yet?

bentoml/_internal/resource.py
sauyon: Hm, is this a type guard? Aren't type guards supposed to return booleans?

bentoml/_internal/resource.py
sauyon: We've never supported memory limitations.

bentoml/_internal/resource.py
sauyon: This is to allow users to add custom resources; it functions effectively the same as the deploymentcontainer registry. An attrs class wouldn't work for that I think.

bentoml/_internal/resource.py
aarnphm: 😄

bentoml/_internal/resource.py
```diff
@@ -0,0 +1,279 @@
...
+        *,
+        resource_id: str,
+    ):
+        res = super().__new__(cls, name, bases, attr_dict)
+        if resource_id != "":
+            # register the class with RESOURCE_REGISTRY
+            assert issubclass(res, Resource)
+            _RESOURCE_REGISTRY[resource_id] = res
+
+        return res  # pylint: disable=useless-super-delegation (???)
```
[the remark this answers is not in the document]
sauyon: Hm, I'll try that. I think this check is only supposed to catch useless calls to super, though.

bentoml/_internal/resource.py
aarnphm: type guards ensure the return type is the correct type, not a boolean. Essentially it does type narrowing for us without having to implement `if isinstance`

bentoml/_internal/resource.py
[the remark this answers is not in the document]
aarnphm: pylint changes a bunch of error message recently so we might have to check it.

bentoml/_internal/resource.py
aarnphm: I was thinking to use `attr.make_class` 🤔, but honestly it doesn't good typing.

Again, I don't have a strong opinion, but I'm curious to learn why meta classes will introduce limitation for custom runnable class @bojiang

aarnphm: Correct me if im wrong, but having metaclass can introduce conflicts if user custom runnable class also implements/extends a metaclass, hence resolving metaclass in the case of multiple inheritance is not recommended.

 However, we should discourage using [multiple inheritance](https://www.youtube.com/watch?v=3MNVP9-hglc&ab_channel=NextDayVideo) anyways.

bentoml/_internal/frameworks/FRAMEWORK_TEMPLATE_PY
aarnphm: I think imo `supports_multi_threading` is nicer UI, where we this is use in conjunction with `support_resources`

bentoml/_internal/resource.py
aarnphm: https://peps.python.org/pep-0647/

sauyon: "...Return statements within a type guard function should return bool values, and type checkers should verify that all return paths return a bool."

I'm pretty sure that `TypeGuard` is mostly for implementing custom `isinstance`.

aarnphm: noted.

bentoml/_internal/resource.py
```diff
@@ -0,0 +1,279 @@
...
+            raise TypeError("cpu must be int, float or str")
+
+        if isinstance(spec, (int, float)):
+            return float(spec)
+
+        milli_match = re.match("([0-9]+)m", spec)
+        if milli_match:
+            return float(milli_match[1]) / 1000.0
+
+        raise BentoMLConfigException(f"Invalid CPU resource limit '{spec}'. ")
```
ssheng, line 102: I think the regex does not current handle the case with a non-milli string value.
```
- "1" -> 1.0
```

bentoml/_internal/resource.py
```diff
@@ -0,0 +1,279 @@
...
+        except FileNotFoundError as err:
+            logger.warning(f"Caught exception while calculating CPU quota: {err}")
+    if quota is not None and quota < 0:
+        quota = None
+    elif quota == 0:
+        quota = 1
+
+    if os.path.exists(shares_file):
+        try:
+            shares = _read_cgroup_file(shares_file) / float(1024)
```
ssheng, line 159: I know you are just copying the logic over here but assuming dividing `cpu.shares` by 1024 will give us CPU core allocation is probably incorrect. CGroup `cpu.shares` is a relative value as opposed to absolute.

bentoml/_internal/resource.py
```diff
@@ -0,0 +1,279 @@
...
+        except (pynvml.nvml.NVMLError, OSError):
+            return 0
+        finally:
+            try:
+                pynvml.nvmlShutdown()
+            except Exception:  # pylint: disable=broad-except
+                pass
+
+
+def get_gpu_memory(dev: int) -> t.Tuple[float, float]:
```
ssheng, line 247: Is this being used?

bentoml/_internal/runner/runnable.py
```diff
@@ -24,19 +24,101 @@
...
+                    supported_resources = set()
+                logger.warning(
+                    f"{name} is using deprecated 'SUPPORT_NVIDIA_GPU'. Please convert to using 'supported_resources':\n"
+                    f"class {name}(Runnable, supported_resources={supported_resources}):\n"
+                    "    ..."
+                )
+            else:
+                logger.warning(
+                    f"Deprecated 'SUPPORT_NVIDIA_GPU' is being ignored in favor of 'supported_resources' for {name}."
+                )
```
ssheng, line 59: We probably want to raise a warning or error if the legacy `SUPPORT_NVIDIA_GPU` and `supported_resources`  do not agree like in the multi-threading case.

bentoml/_internal/runner/runnable.py
```diff
@@ -24,19 +24,101 @@
...
+
+        if supported_resources is None:
+            # attempt to get supported_resources from a superclass
+            found_base = None
+            for base in bases:
```
ssheng, line 64: For simplicity, should we just assume `supported_resources` must be provided in the leaf runnable class, so we do not have to check for base classes?

bentoml/_internal/resource.py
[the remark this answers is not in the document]
sauyon: Hm, the warning went away by itself, that's odd...

ssheng (commented):
Great refactoring to make the resource implementation cleaner.

bentoml/_internal/resource.py
sauyon: AFAICT there is no convention here. `ABCMeta`'s `__new__` uses `namespace` (which is confusing, I think), `type`'s stubs use `__dict` (which is pretty bad). I've only seen `dct` used in one tutorial, I think.

bentoml/_internal/configuration/containers.py
```diff
@@ -59,30 +69,47 @@ def _is_ip_address(addr: str) -> bool:
         return False
 
 
+RUNNER_CFG_SCHEMA = {
+    "batching": {
+        # for review: should we allow configuring batching per-method as well as per-runner?
```
parano, line 74: ideally it should allow per-method batching options but practically very few use cases will need that. this is something we can add in the future right?

bentoml/_internal/configuration/containers.py
```diff
@@ -68,21 +100,16 @@ def _is_ip_address(addr: str) -> bool:
...
             "logging": {
                 # TODO add logging level configuration
                 "access": {
-                    "enabled": bool,
+                    Optional("enabled", default=True): bool,
                     "request_content_length": Or(bool, None),
                     "request_content_type": Or(bool, None),
                     "response_content_length": Or(bool, None),
-                    "response_content_type": Or(bool, None),
+                    Optional("response_content_type", default=None): bool,
```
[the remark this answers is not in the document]
parano: is `Or(bool, None)` different than `Optional`?  I assume the config keys can be missing in the final configuration for `Optional`. although we do a deep merge of user config with the default config before applying the schema validation, we probably don't need to validate the key exist?

bentoml/_internal/resource.py
sauyon: As per discussion with @ssheng I'm removing shares for now.

bentoml/_internal/resource.py
sauyon: No, I don't believe so.

bentoml/_internal/runner/runnable.py
sauyon: We actually do this ourselves in [`pytorch`](https://github.com/bentoml/BentoML/blob/21267df1fc4239d4a9ae3d0c682f4cbd6dbf94da/bentoml/_internal/frameworks/common/pytorch.py#L46), which is why I added this. We could just pass the parameters to the base class but it's a little ugly on the framework impl side.

bentoml/_internal/configuration/containers.py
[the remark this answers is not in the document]
sauyon: I think I started changing them all then we decided we shouldn't change anything; Will revert this.

Why are we allowing `None` as a value for these anyway?

bentoml/_internal/configuration/containers.py
sauyon: Yes, that should be backwards-compatible.

bentoml/_internal/resource.py
yetone: How about use `__init_subclass__`?

```python
class Resource(t.Generic[T]):
    def __init_subclass__(cls, resource_id="", **kwargs):
        super().__init_subclass__(**kwargs)
        if resource_id == "":
            return
        # register the class with RESOURCE_REGISTRY
        _RESOURCE_REGISTRY[resource_id] = cls
```

aarnphm: I believe we want to abstract away from initialization process. But from the breakout chat we decided to just use the old behaviour.

sauyon:
As per discussion, removed metaclasses for maintainability.

Thanks to @yetone for finding `__init_subclass__`, which is a much nicer way of doing resource registration.

bentoml/_internal/resource.py
sauyon: That was for `Runnable`, not `Resource`; I've switched to `__init_subclass__`.

bentoml/_internal/frameworks/FRAMEWORK_TEMPLATE_PY
sauyon: We're using `SUPPORTS_CPU_MULTI_THREADING`.

bojiang:
@sauyon mark as merge-hold for you. feel free to remove that~

bentoml/_internal/resource.py
```diff
@@ -0,0 +1,281 @@
...
+        except ValueError:
+            raise BentoMLConfigException(f"Invalid CPU resource limit '{spec}'. ")
+
+    @classmethod
+    @functools.lru_cache(maxsize=1)
+    def from_system(cls) -> float:
+        if psutil.POSIX:
+            return query_cgroup_cpu_count()
+        else:
+            return float(query_os_cpu_count())
```
sauyon, line 107: Pyright is complaining about this and I currently have no idea why.

bojiang: I think it's okay to remove lru_cache here?

sauyon: Yeah, I think that's fine. Just confused.

bojiang: I guess the lru_cache is not well type annotated to `_lru_cache_wrapper[float]`

sauyon:
> @sauyon mark as merge-hold for you. feel free to remove that~

Should be ready for (re-)review now.

sauyon:
Added tests. This PR is slowly becoming a monster :P

bentoml/_internal/configuration/containers.py
sauyon: Going to delete this line; we can do this later.

bentoml/_internal/frameworks/common/pytorch.py
```diff
@@ -54,8 +54,8 @@ def __init__(self, *inner_args: t.Any, **inner_kwargs: t.Any) -> None:
 
 
 class PytorchModelRunnable(bentoml.Runnable):
-    SUPPORT_NVIDIA_GPU = True
-    SUPPORT_CPU_MULTI_THREADING = True
+    SUPPORTED_RESOURCES = ("nvidia.com/gpu",)
```
bojiang, line 57: shall we have "cpu" in it by default?

sauyon: Hm, that's a good point. We're kind of special casing the strategies around CPU right now but it definitely wouldn't hurt to include `"cpu"` everywhere.

bojiang: Yeah it make users even easier to understand

bentoml/_internal/resource.py
aarnphm: This is useful for our GPU strategy. We should just put it here for now.

bentoml/_internal/runner/runnable.py
```diff
@@ -24,15 +23,31 @@
 RUNNABLE_METHOD_MARK: str = "_bentoml_runnable_method"
 
 
-class Runnable(ABC):
-    SUPPORT_NVIDIA_GPU: bool
+class Runnable:
+    SUPPORTED_RESOURCES: set[str]
     SUPPORT_CPU_MULTI_THREADING: bool
```
aarnphm, line 28:
```suggestion
    SUPPORTS_CPU_MULTI_THREADING: bool
```

bentoml/_internal/runner/runnable.py
```diff
@@ -24,15 +23,31 @@
...
 
-    methods: dict[str, RunnableMethod[t.Any, t.Any, t.Any]] | None = None
+    bentoml_runnable_methods__: dict[
```
aarnphm, line 30: can we make it a special attributes
```suggestion
    __bentoml_runnable_methods__: dict[
```

aarnphm: seems like a good idea to have a special `__` that is unique to runners and more pythonic

tests/unit/_internal/test_resource.py
```diff
@@ -0,0 +1,88 @@
...
+    assert math.isclose(CpuResource.from_spec("100m"), 0.1)
+    assert math.isclose(CpuResource.from_spec(0.3), 0.3)
+    assert math.isclose(CpuResource.from_spec(1), 1.0)
+    assert math.isclose(CpuResource.from_spec("5"), 5.0)
+
+    with pytest.raises(TypeError):
+        CpuResource.from_spec((1, 2, 3))
+
+
+def test_NvidiaGpuResource():
```
aarnphm, line 66: NOTE: revisit when we have GPU test

tests/unit/_internal/test_resource.py
```diff
@@ -0,0 +1,88 @@
...
+    assert NvidiaGpuResource.from_system() >= 0  # TODO: real from_system tests
+
+    with pytest.raises(BentoMLConfigException):
+        NvidiaGpuResource.validate(NvidiaGpuResource.from_system() + 1)
+    with pytest.raises(BentoMLConfigException):
+        NvidiaGpuResource.validate(-2)
+
+    NvidiaGpuResource.validate(-1)
+    NvidiaGpuResource.validate(0)
+    # NvidiaGpuResource.validate(1)  # TODO: work out how to skip this on systems with no GPU
```
aarnphm, line 76: either a pytest mark or monkeypatch the attribute

aarnphm (commented):
You might to also include fastai and catboost changes to here as well.

Some second round review

bojiang (dismissed):

bentoml/_internal/runner/runner.py
```diff
@@ -51,18 +50,13 @@ async def async_run(self, *args: t.Any, **kwargs: t.Any) -> t.Any:
...
-
-
 @attr.define(slots=False, frozen=True, eq=False)
 class Runner:
     runnable_class: t.Type[Runnable]
     runnable_init_params: dict[str, t.Any]
     name: str
     models: list[Model]
-    resource_config: Resource
+    resource_config: dict[str, t.Any]
```
ssheng, line 59: nit: Config can be injected as default value through `Provide[BentoMLContainer.key_name]`

docs/source/concepts/runner.rst
```diff
@@ -114,11 +113,20 @@ Send a test request:
     {"is_positive":true}%
 
 
-The :code:`SUPPORT_NVIDIA_GPU` and :code:`SUPPORT_CPU_MULTI_THREADING` class attribute indicates
-whether or not this Runnable class implementation supports using GPU or utilizes multi
-threading. Since NLTK library doesn't support utilizing GPU nor multi-cpu-core natively,
-they are both set to False. This information is used by the BentoServer scheduler
-to determine the worker pool size of this runner.
+The constant attribute ``SUPPORTED_RESOURCES`` indicates which resources this Runnable class
+implementation supports. The only currently pre-defined resource is ``"nvidia.com/gpu"``.
```
ssheng, line 117: We may need to cover the "cpu" case.

bentoml/_internal/runner/strategy.py
```diff
@@ -44,39 +47,55 @@ class DefaultStrategy(Strategy):
...
 
         # use CPU
-        if resource_request.cpu is not None and resource_request.cpu > 0:
-            if runnable_class.SUPPORT_CPU_MULTI_THREADING:
+        cpus = get_resource(resource_request, "cpu")
+        if cpus is not None and cpus > 0:
+            if "cpu" not in runnable_class.SUPPORTED_RESOURCES:
+                logger.warning(
+                    f"No known supported resource available for {runnable_class}, falling back to using CPU."
+                )
```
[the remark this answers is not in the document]
sauyon: We could define it to have CPU in the base class, but I like requiring users to specify. Runners are an advanced API anyway.

Automatic validation at definition time would require a descriptor or metaclass ;)

But really I think it's not so bad for the strategy to just warn.

tests/unit/_internal/test_resource.py
sauyon: let's worry about this later, I think.

bentoml/_internal/runner/runnable.py
sauyon: I don't think this name needs to be internal to `BentoML`. It's something that can be public, which is why I originally named it `methods`.

Probably we don't need the underscores if we named it `bentoml_`... if you don't think that's nice.

bentoml/_internal/runner/runner.py
sauyon: Nevermind, I thought this was the method arguments. We can just default to the global config here, but we should just use `get`; I don't think simple_di supports `Provide` on class attributes? @bojiang

ssheng (approved):

tests/integration/frameworks/models/onnx.py
```diff
@@ -11,7 +11,7 @@
 import onnxruntime as ort
 
 import bentoml
-from bentoml._internal.runner.resource import Resource
+from bentoml._internal.runner.runner import Runner
```
aarnphm, line 14:
```suggestion
```

aarnphm (dismissed):
remove the imports so pep8 bot won't complain, otherwise gtg

aarnphm:
@sauyon seems like the frameworks CI is failing.

sauyon:
>  ValueError: Runner name 'CatBoostRunnable' is not valid; it must be a valid BentoML Tag name.

Ruh-roh. We should add `runner.py` to all the framework tests too.

bentoml/_internal/runner/runnable.py
aarnphm: it just seems nonsymmetrical. but again this is more of a codestyle thing.

bentoml/_internal/runner/strategy.py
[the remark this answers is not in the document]
aarnphm: can't we add `cpu` to SUPPORTED_RESOURCES if it is not defined via `__setattr__`?

aarnphm:
@sauyon

aarnphm (approved):

sauyon:
`test_runner` shouldn't exist, let me clean this up.

ssheng (approved):

bentoml/_internal/resource.py
```diff
@@ -0,0 +1,292 @@
...
+
+    if quota:
+        limit_count = quota
+
+    return float(min(limit_count, os_cpu_count))
+
+
+@functools.lru_cache(maxsize=1)
+def query_os_cpu_count() -> int:
+    cpu_count = os.cpu_count()
```
rehevkor5, line 176: Shouldn't this (and above) use `os.sched_getaffinity(0)` instead?

PR #2635, opened 2022-06-24
fix: Server silently crash without logging exceptions

ssheng:
## Description

Server (dev_api_server, api_server, and runner) logs are currently omitted. If a runner raises an error during initialization, the runner will silently enter a crash loop without logging the exceptions. This is due to Python's arcane implementation of `disable_existing_loggers`, see https://docs.python.org/3/library/logging.config.html.

Setting `https://docs.python.org/3/library/logging.config.html` to `False` in the `SERVER_LOGGING_CONFIG` under log.py fixes the problem but the solution is unclean. The solution in this PR sets `log_config` in Uvicorn options as `None` to prevent Uvicorn from configuring logging again unnecessarily and fallback to using BentoML's `SERVER_LOGGING_CONFIG`.

## Checklist:

### DIFF
--- a/bentoml/_internal/server/cli/api_server.py
+++ b/bentoml/_internal/server/cli/api_server.py
@@ -11,7 +11,6 @@ import psutil
 
 import bentoml
 
-from ...log import SERVER_LOGGING_CONFIG
 from ...context import component_context
 
 if TYPE_CHECKING:
@@ -119,9 +118,8 @@ def main(
 
     parsed = urlparse(bind)
     uvicorn_options: dict[str, t.Any] = {
-        "log_level": SERVER_LOGGING_CONFIG["root"]["level"],
         "backlog": backlog,
-        "log_config": SERVER_LOGGING_CONFIG,
+        "log_config": None,
         "workers": 1,
     }
     if psutil.WINDOWS:
--- a/bentoml/_internal/server/cli/dev_api_server.py
+++ b/bentoml/_internal/server/cli/dev_api_server.py
@@ -7,7 +7,6 @@ import psutil
 
 from bentoml import load
 
-from ...log import SERVER_LOGGING_CONFIG
 from ...context import component_context
 
 
@@ -44,9 +43,8 @@ def main(
             component_context.bento_version = svc.tag.version
 
         uvicorn_options = {
-            "log_level": SERVER_LOGGING_CONFIG["root"]["level"],
             "backlog": backlog,
-            "log_config": SERVER_LOGGING_CONFIG,
+            "log_config": None,
             "workers": 1,
             "lifespan": "on",
         }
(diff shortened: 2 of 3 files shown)


aarnphm (approved):
Screenshot 2022-06-24 at 02 52 36

Works well with surfacing issues from service.py

aarnphm:
Screenshot 2022-06-24 at 03 04 50
We are double logging tho

ssheng:
> Screenshot 2022-06-24 at 03 04 50
> 
> We are double logging tho

That's strange. We should be logging `INFO` logs from Uvicorn server anymore. Could you please share your repro steps @aarnphm?

aarnphm:
> > Screenshot 2022-06-24 at 03 04 50
> > We are double logging tho
> 
> That's strange. We shouldn't be logging `INFO` logs from Uvicorn server anymore. Could you please share your repro steps @aarnphm?

To reproduce:
1. import yolov5 from torch hub:
```python
import torch
import bentoml

_ = bentoml.pytorch.save_model("yolov5", torch.hub.load("ultralytics/yolov5", "yolov5s", force_reload=True)) 
2. import the correct modules for torch hub:
```python
git clone --bare git@github.com:ultralytics/yolov5.git && git --git-dir=./yolov5.git --work-tree=. checkout a2a1ed201d150343a4f9912d644be2b210206984 -- data models utils LICENSE README.md detect.py export.py hubconf.py train.py val.py requirements.txt && rm -rf yolov5.git
```

3. create a `service.py`:
```python
from typing import List
  
import bentoml
import numpy as np
import torch
from bentoml.io import JSON

runner = bentoml.pytorch.get("yolov5:latest").to_runner()
service = bentoml.Service("yolov5", runners=[runner])


@service.api(input=JSON(), output=JSON())
def predict(input_dict):
    filename = input_dict['filename']
    img = input_dict['image_content']
    input_img = torch.from_numpy(np.asarray(img[0]))
    predictions = runner.run_batch(input_img)
    return predictions
    
```
4. Run `serve`:
```python
bentoml serve service.py:service --reload
```

ssheng:
> I believe the issue stems from here https://github.com/ultralytics/yolov5/blob/a2a1ed201d150343a4f9912d644be2b210206984/utils/general.py#L82
> 
> By this the module add another streamhandler, hence double logging. I don't know how we can do this ;)

Thanks for the repro steps. It is unfortunate when libraries don't play nice with logging. I don't think there is much we can do actually.

aarnphm:
> > I believe the issue stems from here [ultralytics/yolov5@`a2a1ed2`/utils/general.py#L82](https://github.com/ultralytics/yolov5/blob/a2a1ed201d150343a4f9912d644be2b210206984/utils/general.py#L82)
> > By this the module add another streamhandler, hence double logging. I don't know how we can do this ;)
> 
> Thanks for the repro steps. It is unfortunate when libraries don't play nice with logging. I don't think there is much we can do actually.

otherwise this PR i gtg

sauyon:
> Thanks for the repro steps. It is unfortunate when libraries don't play nice with logging. I don't think there is much we can do actually.

We should probably open issues eventually.

PR #2640, opened 2022-06-27
feat: Support Transformers custom pipeline

ssheng:
## Description

This PR adds support for Huggingface Transformers custom pipelines as requested in https://github.com/bentoml/BentoML/issues/2534.

The Huggingface documentation instructed creation of custom pipelines by modifying the `SUPPORTED_TASKS`, see https://huggingface.co/docs/transformers/add_new_pipeline. To support saving custom pipelines, the framework code will first check the presence of the task name of the custom pipeline and save all relevant task definition information (task name, PyTorch and Tensorflow auto class names, default models, and pipeline type) to the model options. When loading pipeline, the framework code will add the custom task to the `SUPPORTED_TASKS` with the metadata saved in the model options and construct the pipeline.

## How Has This Been Tested?

Added framework test for saving and loading custom Transformers pipelines. This PR does not cover the following two tasks which I want to cover in the future.
- Refactor the framework test to use the new template.
- Refactor the framework test to use simpler mocks of PyTorch and Tensorflow models instead of using the actual models from the model hub.

Current working on adding framework tests for custom pipeline. Encountered issue with compatibility issues with Tensorflow  and Apple Silicon.

## Checklist:

- [ ] My code follows the bentoml code style, both `make format` and
  `make lint` script have passed
  ([instructions](https://github.com/bentoml/BentoML/blob/master/DEVELOPMENT.md#style-check-and-auto-formatting-your-code)).
- [ ] My change reduces project test coverage and requires unit tests to be added
- [ ] I have added unit tests covering my code change
- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly

PR #2642, opened 2022-06-28
chore(internal): fix LazyType typing

sauyon:
Noticed some type ignores in #2640, I think this should help.

/cc @ssheng

### DIFF
--- a/bentoml/_internal/types.py
+++ b/bentoml/_internal/types.py
@@ -76,7 +76,13 @@ JSONSerializable = t.NewType("JSONSerializable", object)
 T = t.TypeVar("T")
 
 
-class LazyType(t.Generic[T]):
+if TYPE_CHECKING:
+    LazyTypeSuper = t.Type
+else:
+    LazyTypeSuper = t.Generic
+
+
+class LazyType(LazyTypeSuper[T]):
     """
     LazyType provides solutions for several conflicts when applying lazy dependencies,
         type annotations and runtime class checking.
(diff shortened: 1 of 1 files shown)


bentoml/_internal/types.py
```diff
@@ -150,7 +156,7 @@ def __init__(
             else:
                 self.qualname: str = getattr(module_or_cls, "__name__")
 
-    def __instancecheck__(self, obj) -> "t.TypeGuard[T]":
```
aarnphm, line 153: can you explain more about this change?

sauyon: Assuming you figured this one out because you marked it as resolved, but for others: pyright complains about the type not matching the `type` superclass, and I think it's a type guard by default since we're now inheriting `type[T]`.

bojiang: So this change is invoked because of the previous usage change from `lazy_type.isinstance(obj)` to `isinstance(obj, lazy_type)`?

bojiang: @sauyon On my side this change breaks type guard. It can not narrow the unions down anymore.
How about just keeping using the original solution?

bojiang (changes requested):
it breaks the typeguard

sauyon:
I think the `isinstance()` check never actually worked in the first place, I was looking at code where the type was already determined...

We can hack around the type checker a little bit by returning `type[T]`, but I'm not sure whether we really want to do this in the end.

sauyon:
Going to draft this until post-1.0.

bojiang:
Discussed with Sauyon. We may just keep using `lazy_type.isinstance()` since it works well.

bojiang:
And the built-in `isinstance` in Python doesn't fit the current typing system well since it mixed the runtime and typing. We don't need to implement it forcibly

PR #2644, opened 2022-06-28
docs: Add model in bentoml.pytorch.save_model() pytorch integration example

AlexandreNap:
In the pytorch section of framework-specific guides, model was not specified in the save_model() call.


aarnphm (approved):
nice catch

PR #2645, opened 2022-06-28
feat: add `from_grpc_request` and `to_grpc_response` for numpy_ndarray

Proto007:
## Description
Add `from_grpc_request` and `to_grpc_response` method in io_descriptor numpy to handle grpc requests and response with numpy_ndarrays

## Motivation and Context
This is a necessary to convert protobuf request to an array format which can be used with the service API functions. It is required to handle grpc requests with protobuf array data. Similarly, the response needs to be in protobuf format which requires conversion from numpy_ndarray to protobuf.

## How Has This Been Tested?
The methods are tested using arrays of different data and dimensions to convert from protobuf to array and array to protobuf.

## Checklist:


bentoml/_internal/io_descriptors/numpy.py
```diff
@@ -237,6 +237,67 @@ async def to_http_response(self, obj: ext.NpNDArray, ctx: Context | None = None)
...
+        """
+        import io_descriptors_pb2
+        from google.protobuf.timestamp_pb2 import Timestamp
+        from google.protobuf.duration_pb2 import Duration
+
+        tuple_arr = []
+        [tuple_arr.append(i) for i in getattr(proto_tuple, "value_")]
+
+        if(not tuple_arr):
+            raise ValueError("Provided tuple is either empty or invalid.")
```
aarnphm, line 252:
```suggestion
        tuple_arr = [i for i in getattr(proto_tuple, "value_")]

        if len(tuple_arr) == 0:
            raise ValueError("Provided tuple is either empty or invalid.")
```

bentoml/_internal/io_descriptors/numpy.py
```diff
@@ -237,6 +237,67 @@ async def to_http_response(self, obj: ext.NpNDArray, ctx: Context | None = None)
...
+    def proto_to_arr(self, proto_arr):
+        """
+        Convert given protobuf array to python list
+        """
+        import io_descriptors_pb2
+        from google.protobuf.timestamp_pb2 import Timestamp
+        from google.protobuf.duration_pb2 import Duration
+
+        return_arr = []
+        [return_arr.append(i) for i in getattr(proto_arr, proto_arr.dtype)]
```
aarnphm, line 281:
```suggestion
        return_arr = [i for i in getattr(proto_arr, proto_arr.dtype)]
```

bentoml/_internal/io_descriptors/numpy.py
```diff
@@ -237,6 +237,67 @@ async def to_http_response(self, obj: ext.NpNDArray, ctx: Context | None = None)
...
+            val = getattr(tuple_arr[i], tuple_arr[i].WhichOneof("dtype"))
+
+            if(tuple_arr[i].WhichOneof("dtype") == "timestamp_"):
+                val = Timestamp.ToDatetime(val)
+            elif(tuple_arr[i].WhichOneof("dtype") == "duration_"):
+                val = Duration.ToTimedelta(val)
+
+            if(type(val) == io_descriptors_pb2.NumpyNdarray):
+                val = self.proto_to_arr(val)
+            elif(type(val) == io_descriptors_pb2.Tuple):
```
aarnphm, line 266: usually isinstance is a more pythonic way:
```suggestion
            if isinstance(val, io_descriptors_pb2.NumpyNdarray):
                val = self.proto_to_arr(val)
            elif isinstance(val, io_descriptors_pb2.Tuple):
```

bentoml/_internal/io_descriptors/numpy.py
```diff
@@ -237,6 +237,67 @@ async def to_http_response(self, obj: ext.NpNDArray, ctx: Context | None = None)
...
+                val = self.handle_tuple(val)
```
aarnphm, line 267: you should also handle exception here. What happens if the val doesn't belong to the previous two types?
```python
                val = self.handle_tuple(val)
            else:
                ...
```

bentoml/_internal/io_descriptors/numpy.py
```diff
@@ -237,6 +237,67 @@ async def to_http_response(self, obj: ext.NpNDArray, ctx: Context | None = None)
...
+            return_arr = [Duration.ToTimedelta(td) for td in return_arr]
+
+        if(not return_arr):
+            raise ValueError("Provided array is either empty or invalid")
+
+        for i in range(len(return_arr)):
+            if(type(return_arr[i]) == io_descriptors_pb2.NumpyNdarray):
+                return_arr[i] = self.proto_to_arr(return_arr[i])
+            elif(type(return_arr[i]) == io_descriptors_pb2.Tuple):
+                return_arr[i] = self.handle_tuple(return_arr[i])
```
aarnphm, line 295: same comment as above

aarnphm (commented):
quick initial thoughts.

bentoml/_internal/io_descriptors/numpy.py
Proto007: Done.

bentoml/_internal/io_descriptors/numpy.py
Proto007: Done.

bentoml/_internal/io_descriptors/numpy.py
Proto007: The `else` case would not apply here because if it is not a tuple or an array, it is one of the other types given `val` is not empty or invalid. I added an exception check in my latest commit for `val` to verify that it is valid before doing the type checks.

aarnphm: noted.

aarnphm (approved):
LGTM

aarnphm (approved):
LGTM.

aarnphm:
Lets proceed with the next item on the list.

PR #2646, opened 2022-06-29
chore(ci): always enable actions on PR

sauyon:
Noticed there's no CI on #2645; let's just have actions run on every PR?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,10 +4,7 @@ on:
   push:
     branches:
       - main
-  pull_request:
-    branches:
-      - main
-      - runners-1.0
+  pull_request
   schedule:
     - cron: '0 0 * * 1/2'
 


aarnphm (approved):

.github/workflows/ci.yml
```diff
@@ -4,10 +4,7 @@ on:
   push:
     branches:
       - main
-  pull_request:
-    branches:
-      - main
-      - runners-1.0
+  pull_request
```
aarnphm, line 7:
```suggestion
  pull_request:
```

sauyon: I think we want it to just be a string if we're not configuring it.

aarnphm: can you link it to the action documentation, i can't seem to find it

aarnphm: iirc 
```yaml
on: [push, pull_request]
``` would do the trick
so what you are having is probably right.

PR #2655, opened 2022-06-30
fix(tensorflow): register data container when eager execution is enabled

aarnphm:
tensorflow data container should not be of type `_EagerTensorBase`

By default if eager_execution is enabled, then `Tensor` is of type `EagerTensor`

context: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/framework/ops.py#L1045

### DIFF
--- a/.github/workflows/frameworks.yml
+++ b/.github/workflows/frameworks.yml
@@ -1239,7 +1239,7 @@ jobs:
           fi
 
       - name: Upload test coverage to Codecov
-        if: ${{ needs.diff.outputs.tensorflow_v2 == 'true' }}
+        if: ${{ needs.diff.outputs.tf2 == 'true' }}
         uses: codecov/codecov-action@v2
         with:
           directory: ./
--- a/.github/workflows/frameworks.yml
+++ b/.github/workflows/frameworks.yml
@@ -1231,7 +1231,7 @@ jobs:
       - name: Run tensorflow_v2 without eager execution
         shell: bash
         run: |
-          OPTS=(--cov=bentoml --cov-config=./setup.cfg --cov-report=xml:"tf2.unit.no_eager.xml" --cov-report term-missing:skip-covered --disable-eager-execution)
+          OPTS=(--cov=bentoml --cov-config=./setup.cfg --cov-report=xml:"tf2.unit.no_eager.xml" --cov-report term-missing:skip-covered --disable-tf-eager-execution)
           python -m pytest ./tests/integration/frameworks/test_tensorflow_v2_unit.py "${OPTS[@]}" || ERR=1
           if [ $ERR -eq 1 ]; then
             echo "unit tests for tensorflow_v2 failed!"
--- a/.github/workflows/frameworks.yml
+++ b/.github/workflows/frameworks.yml
@@ -199,7 +199,8 @@ jobs:
               - *runner
               - bentoml/tensorflow.py
               - bentoml/_internal/frameworks/tensorflow_v2.py
-              - tests/integration/frameworks/test_tensorflow_v2_impl.py
+              - tests/integration/frameworks/models/tensorflow_v2.py
+              - tests/integration/frameworks/tensorflow_v2_unit.py
             keras_tf2:
               - *runner
               - *tf2
--- a/.github/workflows/frameworks.yml
+++ b/.github/workflows/frameworks.yml
@@ -1218,12 +1218,32 @@ jobs:
         run: make tests-tensorflow_v2
         shell: bash
 
+      - name: Run unit test for tensorflow_v2
+        shell: bash
+        run: |
+          OPTS=(--cov=bentoml --cov-config=./setup.cfg --cov-report=xml:"tf2.unit.xml" --cov-report term-missing:skip-covered)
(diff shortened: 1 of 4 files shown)


bojiang:
I agree but still need larme's approval

larme:
When eager_execution is enabled, then current code is the same before and after modification

When eager_execution is disabled, then current code will fail, that's what i want because I don't think we should support payloading a tesnor in lazy execution mode?

aarnphm:
> When eager_execution is enabled, then current code is the same before and after modification
> 
> When eager_execution is disabled, then current code will fail, that's what i want because I don't think we should support payloading a tesnor in lazy execution mode?

If this is the case then we should discuss more about data container. Lets just revisit this after 1.0 as this is related to data container API.

aarnphm:
> How should we test if the eager_execution is enabled? I think we should not test every time a payload is created or restored. We may test when bentoml.tensorflow is imported and if the eager_execution is disabled then we don't register the tensorflow tensor DataContainer. In these situations user can still use bentoml.tensorflow but should not use tensorflow tensor as data transferred between runner and api server

seems like a reasonable solution. Lets do that. I will investigate a bit more and add it in this PR.

bentoml/_internal/frameworks/tensorflow_v2.py
```diff
@@ -377,8 +377,10 @@ def from_batch_payloads(
         return cls.batches_to_batch(batches, batch_dim)
 
 
-DataContainerRegistry.register_container(
-    LazyType("tensorflow.python.framework.ops", "_EagerTensorBase"),
-    LazyType("tensorflow.python.framework.ops", "_EagerTensorBase"),
-    TensorflowTensorContainer,
-)
+# NOTE: we only register the TensorflowTensorContainer as if eager execution is enabled.
+if tf.executing_eagerly():
```
aarnphm, line 381: https://github.com/tensorflow/tensorflow/blob/d8ce9f9c301d021a69953134185ab728c1c248d3/tensorflow/python/eager/context.py#L2162 @larme

larme: @aarnphm  then should we just change `_EagerTensorBase` to `Tensor` here?

aarnphm: i had that in mind as well

ssheng:
Will include this PR in the `1.0.0` release.

larme (dismissed):
LGTM now

.github/workflows/frameworks.yml
```diff
@@ -1218,12 +1218,32 @@ jobs:
...
+
+      - name: Run tensorflow_v2 without eager execution
+        shell: bash
+        run: |
+          OPTS=(--cov=bentoml --cov-config=./setup.cfg --cov-report=xml:"tf2.unit.no_eager.xml" --cov-report term-missing:skip-covered --disable-eager-execution)
+          python -m pytest ./tests/integration/frameworks/test_tensorflow_v2_unit.py "${OPTS[@]}" || ERR=1
+          if [ $ERR -eq 1 ]; then
+            echo "unit tests for tensorflow_v2 failed!"
+            exit 1
+          fi
```
aarnphm, line 1239: We will revisit CI after releasing 1.0. But for now, we will do something like this per framework.

cc @parano @ssheng

tests/integration/conftest.py
```diff
@@ -19,11 +19,22 @@ def pytest_addoption(parser: "Parser") -> None:
     parser.addoption(
         "--gpus", action="store_true", default=False, help="run gpus related tests"
     )
+    parser.addoption(
+        "--disable-eager-execution",
+        action="store_true",
+        default=False,
+        help="Disable TF eager execution",
+    )
```
aarnphm, line 27: I put this here for now since we don't have a specific sets of arguments and parser. Again, this is QA and testing and will have more discussion offline.

tests/integration/frameworks/test_tensorflow_v2_unit.py
```diff
@@ -0,0 +1,93 @@
+from __future__ import annotations
```
aarnphm, line 1: @larme  I added this to test out tensorflow container, would love to hear feedback.

aarnphm (commented):
Added testcase for tensorflow data container and ensure the behaviour that we only register data container if eager execution is enabled, per previous discussion.

tests/integration/conftest.py
```diff
@@ -19,15 +19,32 @@ def pytest_addoption(parser: "Parser") -> None:
...
+    elif config.getoption("--gpus"):
         return
-    skip_gpus = pytest.mark.skip(reason="need --gpus option to run")
+
+    skip_gpus = pytest.mark.skip(reason="Skip gpus tests")
+    requires_eager_execution = pytest.mark.skip(reason="Requires eager execution")
     for item in items:
         if "gpus" in item.keywords:
             item.add_marker(skip_gpus)
+        elif "requires_eager_execution" in item.keywords:
```
bojiang, line 46:
```suggestion
        if "requires_eager_execution" in item.keywords:
```

larme (approved):
LGTM

aarnphm:
codecov is flaky. Also have chatted offline with Bojiang. GTG.

PR #2657, opened 2022-07-01
refactor(build_config): pathspec to a separate class

aarnphm:
refactor include, exclude pathspec to `BentoPatternSpec`. `BentoPatternSpec` has a `includes` which returns a boolean whether a file is included in spec or not:
```python
BentoPatternSpec(build_config).includes(path)
```
by default `includes` will only respect `bentofile.yaml`.

To get specs from .bentoignore recursively use `BentoPatternSpec.from_path(path)`.

Finally to use both `bentofile.yaml` and `.bentoignore` add a `recurse_exclude_spec` iterator to `BentoPatternSpec.includes`

Basically the behaviour is the same as before, but this refactors help with internal features that want to interact with `bentofile.yaml` include and exclude programmatically.

is used in #2533 

Will discuss more about implementation.

### DIFF
--- a/bentoml/_internal/bento/bento.py
+++ b/bentoml/_internal/bento/bento.py
@@ -11,7 +11,6 @@ import fs
 import attr
 import yaml
 import fs.osfs
-import pathspec
 import fs.errors
 import fs.mirror
 from fs.copy import copy_file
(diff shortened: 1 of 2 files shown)


aarnphm:
include in watchfiles PR.

PR #2661, opened 2022-07-01
fix: catboost predict params special case to 'predict' only

aarnphm:

`task_type` is only available for `predict()` callback

@yetone might be good to note that for now we only support `task_type` to `CPU`
inside docs

### DIFF
--- a/bentoml/_internal/frameworks/catboost.py
+++ b/bentoml/_internal/frameworks/catboost.py
@@ -245,7 +245,7 @@ def get_runnable(bento_model: bentoml.Model) -> t.Type[bentoml.Runnable]:
 
             self.predict_fns: dict[str, t.Callable[..., t.Any]] = {}
             for method_name in bento_model.info.signatures:
-
+                # `task_type` argument is only supported for the `predict` method
                 if method_name == "predict":
                     self.predict_params = {"task_type": "CPU"}
 
--- a/bentoml/_internal/frameworks/catboost.py
+++ b/bentoml/_internal/frameworks/catboost.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import os
+import types
 import typing as t
 import logging
 from typing import TYPE_CHECKING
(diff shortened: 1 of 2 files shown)


bentoml/_internal/frameworks/catboost.py
```diff
@@ -249,6 +245,10 @@ def __init__(self):
 
             self.predict_fns: dict[str, t.Callable[..., t.Any]] = {}
             for method_name in bento_model.info.signatures:
+
+                if method_name == "predict":
+                    self.predict_params = {"task_type": "CPU"}
```
ssheng, line 250:
```suggestion
                # `task_type` argument is only supported for the `predict` method
                if method_name == "predict":
                    self.predict_params = {"task_type": "CPU"}
```

tests/integration/frameworks/models/catboost.py
```diff
@@ -32,7 +32,7 @@ def generator_accurate_to(
     expected, accuracy: float
 ) -> t.Callable[[ext.NpNdArray], bool]:
     def check(out):
-        score = metrics.accuracy_score(expected, next(out))
```
ssheng, line 35: This is equivalent to `accurate_to`, correct?

bentoml/_internal/frameworks/catboost.py
```diff
@@ -261,9 +258,21 @@ def _run(
             self: CatBoostRunnable,
             input_data: ext.NpNDArray | ext.CbPool | ext.PdDataFrame,
         ) -> ext.NpNDArray:
+            params_ = self.predict_params.copy()
```
ssheng, line 261: Can we avoid this copy by creating two `predict_params`, one for `predict` and one for `staged_predict`.

bentoml/_internal/frameworks/catboost.py
```diff
@@ -261,9 +258,21 @@ def _run(
...
+            if method_name == "predict":
+                if self.available_gpus not in ("", "-1"):
+                    params_["task_type"] = "GPU"
+                else:
+                    params_["task_type"] = "CPU"
+
+            res = self.predict_fns[method_name](input_data, **params_)
+            if isinstance(res, types.GeneratorType):
+                # staged_predict cases
+                return np.asarray(next(res))
```
ssheng, line 275: What types other than numpy can be returned here?

PR #2664, opened 2022-07-01
fix(tensorflow): casting error on kwargs

bojiang:
## Checklist:

### DIFF
--- a/bentoml/_internal/utils/tensorflow.py
+++ b/bentoml/_internal/utils/tensorflow.py
@@ -166,7 +166,7 @@ def normalize_spec(value: t.Any) -> "tf_ext.TypeSpec":
 
 
 def cast_py_args_to_tf_function_args(
-    signature: list["tf_ext.TensorSpec"],
+    signature: list[tf_ext.TensorSpec],
     *args: t.Any,
     **kwargs: t.Any,
 ) -> tuple[t.Any, ...]:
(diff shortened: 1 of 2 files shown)


bojiang:
requires https://github.com/bentoml/BentoML/pull/2665 to pass the ci

bentoml/_internal/utils/tensorflow.py
```diff
@@ -163,9 +165,66 @@ def normalize_spec(value: t.Any) -> "tf_ext.TypeSpec":
...
+
+    Args:
+        signature (:code:`list[tf.TensorSpec]`):
+            signature of the tensorflow function.
+        *args (:code:`t.Any`):
+            positional arguments of the Python function.
+        **kwargs (:code:`t.Any`):
+            keyword arguments of the Python function.
+    """
+    import inspect
```
larme, line 184: do we need to put inspect import here? It's a builtin module so we don't need to lazy import it?

bojiang: I think lazy importing works too because it's builtin

aarnphm: inspect is fast to import and no need to lazily import it.

bojiang: @aarnphm 
Then are there any cons of lazily importing?

bojiang: utils functions rarely have dependencies overlap by might be imported everywhere. Thus lazy import by default should be good for them.

aarnphm: just for readability. Usually it is not as needed to do unless the library is very heavy to import.

bojiang: What readability?

aarnphm: don't worry. It is preference only.

larme (approved):
LGTM

bentoml/_internal/utils/tensorflow.py
sauyon: We should probably be more structured about lazy importing for sure.

PR #2668, opened 2022-07-01
fix(conda): python installation order

aarnphm:
it should install python first when using conda


sauyon:
Can we add a quick test for this?

sauyon:
Or would that be complex?

aarnphm:
we need to set up e2e and smoke tests for this.

sauyon (approved):
Ok, if it'd take a while to do that let's just get this in; I assume you've tested a conda build?

PR #2673, opened 2022-07-02
refactor: pypi check logic

aarnphm:
we can rely on the behaviour of `importlib.util.find_spec().submodule_search_locations`
where PyPI will be saved under `../site-packages/..`

this removes the dependencies on `setuptools_scm`

additionally this will also cover cases where users install bentoml from
a source directory where the latest release tag matches with bentoml PyPI latest
releases.


bentoml/_internal/configuration/__init__.py
```diff
@@ -52,39 +41,20 @@ def clean_bentoml_version(bentoml_version: str) -> str:
...
-    # In a git repo with no tag, setuptools_scm generated version starts with "0.1."
-    is_tagged = not BENTOML_VERSION.startswith("0.1.")
-    is_clean = not str(version_mod.version_tuple[-1]).split(".")[-1].startswith("d")
-    not_been_modified = BENTOML_VERSION == BENTOML_VERSION.split("+")[0]
-    return is_tagged and is_clean and not_been_modified
+    # We can use importlib.util.find_spec("bentoml").submodule_search_locations
+    # to find the source code of bentoml. If bentoml is installed with pypi, then it should
+    # be installed in 'site-packages', otherwise, it will return path to source git.
+    bentoml_path = source_locations("bentoml")
+    if bentoml_path and "site-packages" in bentoml_path:
```
parano, line 55: this doesn't mean BentoML is installed from pypi index, it can be installed from source - this logic here might be used for checking if it's in edit mode tho

aarnphm: if it is installed in edit mode, then `submodule_search_locations` returns the path to the directory.

aarnphm: This will come with a bug where if users install from source with the latest release tags, it would match with the same tag of the one released on pypi

parano: > This will come with a bug where if users install from source with the latest release tags, it would match with the same tag of the one released on pypi

That is expected behavior - our release process ensures the tags are matching PyPI releases. why is that a problem?

PR #2680, opened 2022-07-04
chore: use pypa/build as PEP517 backend

aarnphm:
We want to use pypa/build to simplify build process and PEP517 compliant.


sauyon (approved):

bentoml/_internal/utils/tempdir.py
```diff
@@ -1,62 +0,0 @@
-import shutil
```
sauyon, line 1: File removed in favor of using `tempfile.TemporaryDirectory`.

Issue #2666, opened 2022-07-01
containerize with conda fails, missing install.sh


smidm:
**Describe the bug**

`bentoml containerize` with conda options in `bentofile.yaml` fails with `chmod: cannot access '/home/bentoml/bento/env/python/install.sh': No such file or directory`.

`bentofile.yaml` with:
```
python:
      packages: 
      - scikit-learn
```
works without issues.

**To Reproduce**

`sample_bentofile.yaml`:
```
service: "sample_service:svc"
include:
    - "*.py"  
conda:
    dependencies:
    - python=3.8.13
    - pip    
    pip:
    - scikit-learn
```

`sample_service.py`:
```
import numpy as np
import bentoml
from bentoml.io import NumpyNdarray

iris_clf_runner = bentoml.sklearn.get("iris_clf:latest").to_runner()

svc = bentoml.Service("iris_classifier", runners=[iris_clf_runner])

@svc.api(input=NumpyNdarray(), output=NumpyNdarray())
def classify(input_series: np.ndarray) -> np.ndarray:
    result = iris_clf_runner.predict.run(input_series)
    return result
```

1. `bentoml build -f sample_bentofile.yaml`
2. `bentoml containerize iris_classifier:latest --debug --no-cache`
3. fails with:

```
Building BentoML service "iris_classifier:d4lyn4xzho2n2atv" from build context "/home/ubuntu/experiments/mmdetection/bentoml"
Packing model "iris_clf:7o4gi2hvzshtkatv"

██████╗░███████╗███╗░░██╗████████╗░█████╗░███╗░░░███╗██╗░░░░░
██╔══██╗██╔════╝████╗░██║╚══██╔══╝██╔══██╗████╗░████║██║░░░░░
██████╦╝█████╗░░██╔██╗██║░░░██║░░░██║░░██║██╔████╔██║██║░░░░░
██╔══██╗██╔══╝░░██║╚████║░░░██║░░░██║░░██║██║╚██╔╝██║██║░░░░░
██████╦╝███████╗██║░╚███║░░░██║░░░╚█████╔╝██║░╚═╝░██║███████╗
╚═════╝░╚══════╝╚═╝░░╚══╝░░░╚═╝░░░░╚════╝░╚═╝░░░░░╚═╝╚══════╝

Successfully built Bento(tag="iris_classifier:d4lyn4xzho2n2atv")
[...]

PR #2681, opened 2022-07-04
fix: install.sh should be optional if python is not specified

aarnphm:
Address #2666

also address one bug where if only wheels arguments is specified, it should
still respect content of wheels folder and install with given args

Also address a bug where if conda is specified without python field, install.sh should still be available.

Will update #2539 after this got merged. We should have unittest for 1.0

### DIFF
--- a/bentoml/_internal/bento/build_config.py
+++ b/bentoml/_internal/bento/build_config.py
@@ -29,11 +29,6 @@ from ...exceptions import BentoMLException
 from ..configuration import CLEAN_BENTOML_VERSION
 from .build_dev_bentoml_whl import build_bentoml_editable_wheel
 
-if version_info >= (3, 8):
-    from typing import Literal
-else:
-    from typing_extensions import Literal
-
 if TYPE_CHECKING:
     from attr import Attribute
     from fs.base import FS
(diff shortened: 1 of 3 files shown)


bentoml/_internal/bento/build_config.py
```diff
@@ -434,73 +432,22 @@ def write_to_bento(self, bento_fs: "FS", build_ctx: str) -> None:
...
-            )
-        elif self.packages is not None:
-            with bento_fs.open(fs.path.join(py_folder, "requirements.txt"), "w") as f:
-                f.write("\n".join(self.packages))
-        else:
-            # Return early if no python packages were specified
-            return
-
+        pip_compile_compat: t.List[str] = []
         pip_args: t.List[str] = []
```
parano, line 436: move this to above L450?

bentoml/_internal/bento/docker/templates/_macros.j2
```diff
@@ -22,3 +22,16 @@ fi
 EOF
 
 {%- endmacro -%}
+
+{%- macro run_script(path, echo_msg, additional_run_arg="") -%}
```
parano, line 26:
```suggestion
{%- macro run_script(path, echo_msg, run_arg="") -%}
```

bentoml/_internal/bento/docker/templates/_macros.j2
```diff
@@ -22,3 +22,16 @@ fi
 EOF
 
 {%- endmacro -%}
+
+{%- macro run_script(path, echo_msg, run_args="") -%}
+# Running {{ echo_msg }} if exists
```
parano, line 27:
```suggestion
# Running {{ echo_msg }}
```

parano:
LGTM!

bojiang (dismissed):

bojiang (approved):

PR #2683, opened 2022-07-05
feat: moving from BaseConverter to GenConverter

aarnphm:
future proof cattrs migration for default Converter (which is now GenConverter)

### DIFF
--- a/bentoml/_internal/bento/bento.py
+++ b/bentoml/_internal/bento/bento.py
@@ -16,6 +16,7 @@ import fs.errors
 import fs.mirror
 from fs.copy import copy_file
 from cattr.gen import override  # type: ignore (incomplete cattr types)
+from cattr.gen import make_dict_structure_fn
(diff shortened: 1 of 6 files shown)


bentoml/_internal/bento/bento.py
```diff
@@ -340,81 +341,69 @@ def __init__(self, base_path: t.Union[PathType, "FS"]):
...
-
-
-@attr.define(repr=False, frozen=True, on_setattr=None)
+@attr.frozen(repr=False)
 class BentoInfo:
     tag: Tag
     service: str = attr.field(
         converter=lambda svc: svc if isinstance(svc, str) else svc._import_str
-    )  # type: ignore[reportPrivateUsage]
-    name: str = attr.field(init=False)  # converted from tag in __attrs_post_init__
```
parano, line 409: let's keep the inline comment here?

PR #2685, opened 2022-07-05
chore: lock cattrs <= 21.1.0

aarnphm:
we will look into migrating to GenConverter after 1.0

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,7 +38,7 @@ install_requires =
     aiofiles
     aiohttp<=3.8.1  # prevent it from installing preview release
     attrs>=21.1.0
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -38,7 +38,7 @@ install_requires =
     aiofiles
     aiohttp<=3.8.1  # prevent it from installing preview release
     attrs>=21.1.0
-    cattrs>=1.10.0
```
parano, line 41: can we make this a range? >=1.10.0 and <=21.4.0

ssheng:
Why do we do this?

ssheng (approved):

Issue #2684, opened 2022-07-05
cattr: migration to GenConverter


aarnphm:
From 22.2 onwards, the default `Converter` is `22.1` and below `GenConverter` and the `Converter` for 22.1 and below will be renamed to `BaseConverter`.

https://github.com/bentoml/BentoML/pull/2683

aarnphm:
Finished.

PR #2688, opened 2022-07-05
refactor: move from Converter to GenConverter

aarnphm:
This should be compatible with current GenConverter.

address #2684

### DIFF
--- a/bentoml/_internal/bento/bento.py
+++ b/bentoml/_internal/bento/bento.py
@@ -15,9 +15,9 @@ import pathspec
 import fs.errors
 import fs.mirror
 from fs.copy import copy_file
-from cattr.gen import override  # type: ignore (incomplete cattr types)
+from cattr.gen import override
 from cattr.gen import make_dict_structure_fn
-from cattr.gen import make_dict_unstructure_fn  # type: ignore (incomplete cattr types)
+from cattr.gen import make_dict_unstructure_fn
 from simple_di import inject
 from simple_di import Provide
 
@@ -345,8 +345,8 @@ class BentoStore(Store[Bento]):
 class BentoRunnerInfo:
     name: str
     runnable_type: str
-    models: list[str] = attr.field(factory=list)
-    resource_config: dict[str, t.Any] | None = attr.field(default=None)
+    models: t.List[str] = attr.field(factory=list)
+    resource_config: t.Optional[t.Dict[str, t.Any]] = attr.field(default=None)
 
     @classmethod
     def from_runner(cls, r: Runner) -> BentoRunnerInfo:
@@ -400,10 +400,10 @@ class BentoInfo:
     bentoml_version: str = attr.field(default=BENTOML_VERSION)
     creation_time: datetime = attr.field(factory=lambda: datetime.now(timezone.utc))
 
-    labels: dict[str, t.Any] = attr.field(factory=dict)
-    models: list[BentoModelInfo] = attr.field(factory=list)
-    runners: list[BentoRunnerInfo] = attr.field(factory=list)
-    apis: list[BentoApiInfo] = attr.field(factory=list)
+    labels: t.Dict[str, t.Any] = attr.field(factory=dict)
+    models: t.List[BentoModelInfo] = attr.field(factory=list)
+    runners: t.List[BentoRunnerInfo] = attr.field(factory=list)
+    apis: t.List[BentoApiInfo] = attr.field(factory=list)
     docker: DockerOptions = attr.field(factory=lambda: DockerOptions().with_defaults())
     python: PythonOptions = attr.field(factory=lambda: PythonOptions().with_defaults())
     conda: CondaOptions = attr.field(factory=lambda: CondaOptions().with_defaults())
--- a/bentoml/_internal/bento/bento.py
+++ b/bentoml/_internal/bento/bento.py
@@ -350,7 +350,6 @@ class BentoRunnerInfo:
 
     @classmethod
     def from_runner(cls, r: Runner) -> BentoRunnerInfo:
-        # TODO: Add runner default resource quota and batching config here
         return cls(
             name=r.name,
             runnable_type=r.runnable_class.__name__,
--- a/bentoml/_internal/bento/bento.py
+++ b/bentoml/_internal/bento/bento.py
@@ -390,13 +390,20 @@ class BentoModelInfo:
 
 @attr.frozen(repr=False)
 class BentoInfo:
+
+    # for backward compatibility in case new fields are added to BentoInfo.
+    __forbid_extra_keys__ = False
+    # omit field in yaml file if it is not provided by the user.
+    __omit_if_default__ = True
+
     tag: Tag
     service: str = attr.field(
         converter=lambda svc: svc if isinstance(svc, str) else svc._import_str
     )
     name: str = attr.field(init=False)
     version: str = attr.field(init=False)
-    bentoml_version: str = attr.field(default=BENTOML_VERSION)
+    # using factory explicitly instead of default because omit_if_default is enabled for BentoInfo
+    bentoml_version: str = attr.field(factory=lambda: BENTOML_VERSION)
     creation_time: datetime = attr.field(factory=lambda: datetime.now(timezone.utc))
 
     labels: t.Dict[str, t.Any] = attr.field(factory=dict)
--- a/bentoml/_internal/bento/bento.py
+++ b/bentoml/_internal/bento/bento.py
@@ -16,6 +16,7 @@ import fs.errors
 import fs.mirror
 from fs.copy import copy_file
 from cattr.gen import override  # type: ignore (incomplete cattr types)
+from cattr.gen import make_dict_structure_fn
 from cattr.gen import make_dict_unstructure_fn  # type: ignore (incomplete cattr types)
 from simple_di import inject
 from simple_di import Provide
@@ -92,9 +93,9 @@ python:
 @attr.define(repr=False, auto_attribs=False)
 class Bento(StoreItem):
     _tag: Tag = attr.field()
-    __fs: "FS" = attr.field()
+    __fs: FS = attr.field()
 
-    _info: "BentoInfo"
+    _info: BentoInfo
 
     _model_store: ModelStore
     _doc: t.Optional[str] = None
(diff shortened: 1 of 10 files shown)


bentoml/_internal/bento/bento.py
```diff
@@ -340,74 +341,62 @@ def __init__(self, base_path: t.Union[PathType, "FS"]):
...
     name: str
     runnable_type: str
     models: t.List[str] = attr.field(factory=list)
     resource_config: t.Optional[t.Dict[str, t.Any]] = attr.field(default=None)
 
     @classmethod
-    def from_runner(cls, r: Runner) -> "BentoRunnerInfo":
-        # Add runner default resource quota and batching config here
+    def from_runner(cls, r: Runner) -> BentoRunnerInfo:
+        # TODO: Add runner default resource quota and batching config here
```
parano, line 353:
```suggestion
```

requirements/dev-requirements.txt
```diff
@@ -1,5 +1,6 @@
 # Dev dependencies
 -r tests-requirements.txt
+-r docs-requirements.txt
```
parano, line 3:
```suggestion
```

bentoml/_internal/utils/cattr.py
```diff
@@ -3,12 +3,18 @@
...
-import cattr
 from attr import fields
-from cattr import override  # type: ignore
+from cattr.gen import override
 from cattr.gen import AttributeOverride
 
-bentoml_cattr = cattr.Converter()
+from .pkg import pkg_version_info
+
+if pkg_version_info("cattrs")[:2] >= (22, 2):
```
parano, line 12: use packaging.version's comparison implementation

bentoml/_internal/utils/cattr.py
```diff
@@ -3,12 +3,18 @@
...
+    from cattr import Converter
+else:
+    from cattr import GenConverter as Converter
```
parano, line 15: This doesn't work at all for version < 22.2

setup.cfg
```diff
@@ -72,6 +72,7 @@ install_requires =
     starlette
     typing-extensions>=4.0
     backports.cached-property;python_version<"3.8"
+    exceptiongroup;python_version <= "3.11"
```
aarnphm, line 74: This will be the default package for python 3.11. This would actually provide better exception traceback than rich traceback.
Screenshot 2022-07-05 at 00 35 59

aarnphm: rich traceback haven't supported this yet. We could allow user to enable rich.traceback if then want to. But I think for now exceptiongroup is more verbose.

bentoml/_internal/types.py
```diff
@@ -48,27 +48,26 @@
...
+ValueType = t.Union[
+    str, bytes, bool, int, float, complex, datetime, date, time, timedelta
+]
+MetadataType = t.Union[
+    ValueType,
+    t.List[ValueType],
+    t.Dict[str, ValueType],
+    t.Dict[t.Union[str, float, int], ValueType],
+    t.Tuple[ValueType],
 ]
```
parano, line 64: Is this change necessary? it's not exactly correct.

 I think only the l66-70 is what's required for cattrs to work.

bentoml/_internal/cli/__init__.py
```diff
@@ -12,11 +12,6 @@
 
 
 def create_bentoml_cli():
-    # exclude traceback from the click library
-    from rich.traceback import install
-
-    install(suppress=[click])
-
```
aarnphm, line 19: using exceptiongroup returns better exception message

setup.cfg
aarnphm: We should move back to rich.traceback once https://github.com/Textualize/rich/issues/1859 supports it :)

bentoml/_internal/types.py
aarnphm: oh no, let me revert this back. 

You are correct. Only L66-70 is needed.

bentoml/_internal/utils/cattr.py
```diff
@@ -3,12 +3,18 @@
...
 
-bentoml_cattr = cattr.Converter()
+from .pkg import pkg_version_info
+
+if pkg_version_info("cattrs") >= (22, 2):
+    from cattr import Converter
+else:
+    from cattr import GenConverter as Converter
+
+bentoml_cattr = Converter(forbid_extra_keys=True)
```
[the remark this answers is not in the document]
parano: that was the whole point ;)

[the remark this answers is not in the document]
aarnphm: should work as expected. on my end it works with both python 3.7 and python 3.10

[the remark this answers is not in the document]
parano: > wait wdym? we want to forbid_extra_keys. This works with 22.1.0 =='

forbid_extra_keys=True is the default behavior in BaseConvertor, we need forbid_extra_keys=False for validating user provided yaml config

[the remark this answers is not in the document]
ssheng: If we want different `forbid_extra_keys` behaviors based on the `cattrs` version, should we put the converter initialization inside the if statement as well?

ssheng:
@aarnphm do you think this is needed for the 1.0 release? On a separate topic, to help focus the PR, I'd recommend that we separate the type refactoring and the functional change. 😃

aarnphm:
> @aarnphm do you think this is needed for the 1.0 release? On a separate topic, to help focus on the actual issue, I'd recommend that we separate the type refactoring and the functional change. 😃

This is needed for 1.0

setup.cfg
```diff
@@ -39,6 +39,7 @@ install_requires =
     aiohttp<=3.8.1  # prevent it from installing preview release
     attrs>=21.1.0
     cattrs>=1.10.0,<=21.1.0
+>>>>>>> 0ffe72d5 (chore: updates types for lower python version)
```
aarnphm, line 42:
```suggestion
```

bentoml/_internal/bento/bento.py
```diff
@@ -340,81 +341,69 @@ def __init__(self, base_path: t.Union[PathType, "FS"]):
         super().__init__(base_path, Bento)
 
 
-@attr.define(frozen=True, on_setattr=None)
```
sauyon, line 343: Why did you remove this? This was for compatibility with some (probably older) `attrs` version, I think.

We should definitely add a comment as explanation here, though.

aarnphm: define is a new 22.1 feature anyway right?

sauyon: Sure, but there was at least one version where not doing this did something we didn't want it to do. Don't remember which one, though...

sauyon: Seems like it's in from v 21.

aarnphm: I mean don't we already do `@attr.frozen` anyway?

bentoml/_internal/bento/bento.py
```diff
@@ -340,75 +341,69 @@ def __init__(self, base_path: t.Union[PathType, "FS"]):
...
     service: str = attr.field(
         converter=lambda svc: svc if isinstance(svc, str) else svc._import_str
-    )  # type: ignore[reportPrivateUsage]
-    name: str = attr.field(init=False)  # converted from tag in __attrs_post_init__
-    version: str = attr.field(init=False)  # converted from tag in __attrs_post_init__
-    bentoml_version: str = attr.field(default=BENTOML_VERSION)
+    )
+    name: str = attr.field(init=False)
+    version: str = attr.field(init=False)
+    # NOTE: using factory explicitly instead of default because omit_if_default is enabled globally in bentoml_cattr
```
aarnphm, line 405:
```suggestion
    # using factory explicitly instead of default because omit_if_default is enabled for BentoInfo
```

bentoml/_internal/models/model.py
```diff
@@ -318,8 +318,10 @@ def __init__(self, base_path: "t.Union[PathType, FS]"):
 class ModelContext:
     framework_name: str
     framework_versions: t.Dict[str, str]
-    bentoml_version: str = attr.field(default=BENTOML_VERSION)
-    python_version: str = attr.field(default=PYTHON_VERSION)
+
+    # NOTE: using factory explicitly instead of default because omit_if_default is enabled globally in bentoml_cattr
```
aarnphm, line 322:
```suggestion
    # using factory explicitly instead of default because omit_if_default is enabled in ModelInfo

File: bentoml/_internal/bento/bento.py (hunk 10 above)
Comment (parano) [reply bentoml/_internal/bento/bento.py:343]: @sauyon we have `attrs>=21.1.0` in our dependency now, should be ok to just use `@attr.frozen` here now?

File: bentoml/_internal/bento/build_config.py [hunk 13]
```diff
@@ -132,6 +129,12 @@ def _convert_env(
 
 @attr.frozen
 class DockerOptions:
+
+    # User shouldn't add new fields under yaml file.
```
parano, line 133:
```suggestion
    # For validating user defined bentofile.yaml.
```
```suggestion
    # User shouldn't add new fields under yaml file.
```
```suggestion
    # User shouldn't add new fields under yaml file.
```

bentoml/_internal/bento/build_config.py
```diff
@@ -132,6 +129,12 @@ def _convert_env(
...
+    __forbid_extra_keys__ = True
+    # no need to omit since BentoML has already handled the default values.
```
parano, line 135:
```suggestion
    # always omit config values in case of default values got changed in future BentoML releases
```
```suggestion
    # no need to omit since BentoML has already handled the default values.
```
```suggestion
    # no need to omit since BentoML has already handled the default values.
```

bentoml/_internal/utils/cattr.py
```diff
@@ -28,5 +26,23 @@ def datetime_decoder(dt_like: str | datetime, _: t.Any) -> datetime:
...
 
 
-bentoml_cattr.register_unstructure_hook(datetime, lambda dt: dt.isoformat())  # type: ignore
-bentoml_cattr.register_structure_hook(datetime, datetime_decoder)
+bentoml_cattr.register_structure_hook_factory(
+    attr.has,
+    lambda cls: make_dict_structure_fn(
+        cls,
+        bentoml_cattr,
+        _cattrs_forbid_extra_keys=getattr(cls, "__forbid_extra_keys__", False),
```
ssheng, line 34: I read that dunders (double leading and trailing underscore) should be reserved for python internals. Unlikely we will get a collision with this name but should we by consistent with naming and maybe use ALL_UPPERCASE or _single_leading_underscore?

ssheng (approved):
One minor comment.

bentoml/_internal/utils/cattr.py
aarnphm: I think dunder usage in this case is fine, since this is only specific to a bentoml internal. The best way is to use metaclass, but then there is no need for that complexity.

PR #2705, opened 2022-07-09
fix: logging consistency

aarnphm:
- refactor(cli): fixing types and restore previous state for environment set
- fix: logging level consistency

This PR:
Screenshot 2022-07-08 at 22 21 04

on `main`:
Screenshot 2022-07-08 at 22 22 00

Notice the last two lines from this PR that `main` is missing. Those are bentoml logs.

### DIFF
--- a/bentoml/_internal/cli/click_utils.py
+++ b/bentoml/_internal/cli/click_utils.py
@@ -6,19 +6,18 @@ import time
 import typing as t
 import difflib
 import logging
+import functools
 from typing import TYPE_CHECKING
-from functools import wraps
 
-import attr
 import click
 from click import ClickException
 from click.exceptions import UsageError
 
 from ..log import configure_logging
 from ...exceptions import BentoMLException
+from ..configuration import DEBUG_ENV_VAR
+from ..configuration import QUIET_ENV_VAR
 from ..configuration import CONFIG_ENV_VAR
-from ..configuration import get_debug_mode
-from ..configuration import get_quiet_mode
 from ..configuration import set_debug_mode
 from ..configuration import set_quiet_mode
 from ..configuration import load_global_config
@@ -26,7 +25,6 @@ from ..utils.analytics import track
 from ..utils.analytics import CliEvent
 from ..utils.analytics import cli_events_map
 from ..utils.analytics import BENTOML_DO_NOT_TRACK
-from ..utils.analytics.usage_stats import do_not_track
 
 if TYPE_CHECKING:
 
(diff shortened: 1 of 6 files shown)


bentoml/_internal/cli/click_utils.py
```diff
@@ -104,36 +112,31 @@ def wrapper(
...
+                if (
+                    cmd_group.name in cli_events_map
+                    and command_name in cli_events_map[cmd_group.name]
+                ):
+                    return cli_events_map[cmd_group.name][command_name](
                         cmd_group=cmd_group.name,
                         cmd_name=command_name,
+                        return_value=return_value,
                     )
+                return CliEvent(cmd_group=cmd_group.name, cmd_name=command_name)
```
aarnphm: To fix pyright type error.

bentoml/_internal/cli/click_utils.py
```diff
@@ -32,37 +37,39 @@ class ClickFunctionWrapper(t.Protocol[P]):
         def __call__(*args: P.args, **kwargs: P.kwargs) -> t.Callable[P, t.Any]:
             ...
 
-    WrappedCLI = t.Callable[P, ClickFunctionWrapper[t.Any]]
+    WrappedCLI = t.Callable[P, ClickFunctionWrapper[P]]
```
aarnphm: fix some type hint.

bentoml/_internal/log.py
```diff
@@ -88,7 +88,7 @@ def configure_logging():
         CLI_LOGGING_CONFIG["root"]["level"] = logging.DEBUG
     else:
         CLI_LOGGING_CONFIG["loggers"]["bentoml"]["level"] = logging.INFO
-        CLI_LOGGING_CONFIG["root"]["level"] = logging.WARNING
+        CLI_LOGGING_CONFIG["root"]["level"] = logging.INFO
```
aarnphm, line 91: This should be INFO so that it doesn't quiet server script logs

bentoml/_internal/log.py
```diff
@@ -115,7 +115,7 @@ def configure_server_logging():
         SERVER_LOGGING_CONFIG["root"]["level"] = logging.DEBUG
     else:
         SERVER_LOGGING_CONFIG["loggers"]["bentoml"]["level"] = logging.INFO
-        SERVER_LOGGING_CONFIG["root"]["level"] = logging.WARNING
+        SERVER_LOGGING_CONFIG["root"]["level"] = logging.INFO
```
aarnphm, line 118: same here

aarnphm (commented):
cc @sauyon @ssheng

bentoml/_internal/log.py
sauyon: Why aren't the server script logs logging from the bentoml namespace?

sauyon: We should configure those logs manually; we don't want to allow info logs from any other libraries through by default, I think.

ssheng:
Seems like circus is configuring logs (notice the format difference) when it shouldn't. Is there an option for skipping the circus log configuration, similar to skipping log configuration in Starlette.

aarnphm:
i believe circus is fine to show in this case, since it is a development mode. imo the config should be at least consistent. We can decide what to show or what not to.

Issue #2693, opened 2022-07-05
feat: ONNXMlir first class support


aarnphm:
Provide supports for ONNXMlir as a first-class citizen

TODO:
- [ ] use the docker container as an entrypoint for `onnx-mlir`
- [ ] provides envar `ONNX_MLIR_DIR` if users build ONNX_MLIR locally
- [ ] need a way to safely add PyRuntime to `sys.path`
- [ ] support `multi_entrypoint` compiled model.

Maybe provides support for older `ExecutionSession` for older onnx-mlir API. Since this library is pretty obscure I assume everyone are probably living on the edge and install the latest changes anyway.

This will be looked after 1.0 releases, as ONNXMlir is not a top list item.

messerb5467:
@andrewsi-z for your awareness

andrewsi-z:
Sounds good! 
@aarnphm with 'use the docker container as an entrypoint for onnx-mlir' I assume this would be to take an ONNX model as input and handle compile as part of the abstraction layer rather than require the compile as a prerequisite step (which the current approach relies on)?

aarnphm:
that is correct. We will allow any .onnx file format and we will use the docker container to compile the model.

aarnphm:
I'm thinking to also provide a helper layer where users can put any model file (let say tensorflow model format, or torch model format) and then convert it to onnx. But imo that is outside the scope abit. We could do something like that for ONNX as well in the future.

larme:
@aarnphm I wonder if we want to convert tf/pt model to onnx for users, there're a lot of fine tuning can happen in this process. Better give user some references in documentations

PR #2706, opened 2022-07-09
feat: onnxmlir

aarnphm:
- feat: skafolding onnxmlir support
- feat: onnxmlir api

work in progress, put here as a draft for trying out #2693

Test will follow accordingly.

### DIFF
--- a/bentoml/_internal/frameworks/onnxmlir.py
+++ b/bentoml/_internal/frameworks/onnxmlir.py
@@ -1,0 +1,470 @@
+from __future__ import annotations
+
+import os
+import sys
+import shutil
+import typing as t
+import logging
+import subprocess
+from copy import deepcopy
+from uuid import uuid4
+from typing import Iterable
+from typing import TYPE_CHECKING
+from pathlib import Path
+from contextlib import contextmanager
+
+import attr
+
+import bentoml
+from bentoml import Tag
+from bentoml.models import ModelContext
+from bentoml.models import ModelOptions
+from bentoml.exceptions import NotFound
+from bentoml.exceptions import InvalidArgument
+from bentoml.exceptions import BentoMLException
+from bentoml.exceptions import MissingDependencyException
+
+from ..utils import cached_contextmanager
+from ..utils import resolve_user_filepath
+from ..utils.pkg import find_spec
+from ..utils.buildx import run_docker_cmd
+from ..utils.buildx import requires_docker
+from ..utils.buildx import UNAME_M_TO_PLATFORM_MAPPING
+
+logger = logging.getLogger(__name__)
+
+MODULE_NAME = "bentoml.onnxmlir"
+MODEL_FILENAME = "saved_model.so"
+API_VERSION = "v1"
+
+# TODO: mount session to docker container and use prebuilt?
+DEV_MLIR_IMAGE = "onnxmlirczar/onnx-mlir-dev:latest"
+
+# release docker images to compile the binary
+RELEASE_MLIR_IMAGE = "onnxmlirczar/onnx-mlir:latest"
+DOCKER_WORKDIR = "/workdir"
+DOCKER_OUTDIR = "/output"
+COMPILE_ARG = ["--O3", "--EmitLib"]
+
+# onnx-mlir constant
+if sys.platform == "win32":
+    _onnx_mlir_binary = "onnx-mlir.exe"
+else:
+    _onnx_mlir_binary = "onnx-mlir"
+
+# There are loosely a few defined environment variables
+# from onnx/onnx-mlir that are available throughout its codebase.
+# here are for documentation
+ACCEPTED_ONNX_ENVVAR = ["ONNX_MLIR_HOME", "ONNX_MLIR_BIN"]
+USE_DOCKER_ENVVAR = "USE_ONNXMLIR_DOCKER"
+use_docker = os.environ.get(USE_DOCKER_ENVVAR, str(False)).lower() == "true"
+
+warning = """\
+The compiled Python runtime 'PyRuntime' is required in order to use module `bentoml.onnxmlir`. Please set either ONNX_MLIR_HOME or ONNX_MLIR_BIN so that BentoML can find onnx-mlir binary and runtime library correctly:
+
+1. Set ONNX_MLIR_HOME to the path of 'build' directory of onnx-mlir. The 'build' directory refers to parent folder containing 'bin', 'lib' in which ONNX-MLIR executables and libraries can be found, typically '/path/to/onnx-mlir/build/Debug'.
+2. Set ONNX_MLIR_BIN to 'onnx-mlir/build/Debug/bin' directory. After, add $ONNX_MLIR_BIN to PATH environment variable: "export PATH=$ONNX_MLIR_BIN:$PATH" and reload current shell.
+"""
+
+if TYPE_CHECKING:
+    from .. import external_typing as ext
+    from ..types import PathType
+    from ...types import ModelSignature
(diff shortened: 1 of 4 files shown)


bentoml/onnx.py
```diff
@@ -1,12 +1,7 @@
-import logging
```
aarnphm, line 1: Notice this when running pylint

bentoml/_internal/utils/buildx.py
```diff
@@ -15,7 +16,14 @@
 
 logger = logging.getLogger(__name__)
 
-DOCKER_BUILDX_CMD = ["docker", "buildx"]
+docker_cmd = which("docker")
+
+if not docker_cmd:
+    raise BentoMLException(
+        "docker is required. See https://www.docker.com/get-started/ for instructions on how to install Docker."
+    )
```
aarnphm, line 24: This exception imo is ok since docker is a dependency for containerize anyway

bentoml/_internal/frameworks/onnxmlir.py
```diff
@@ -0,0 +1,396 @@
...
+            self.model = load_model(bento_model)
+
+            self.entrypoints: dict[str, t.Callable[..., t.Any]] = {}
+            for method_name in bento_model.info.signatures:
+                try:
+                    self.entrypoints[method_name] = getattr(self.model, method_name)
+                except AttributeError:
+                    raise InvalidArgument(
+                        f"No method with name {method_name} found for compiled model {self.model}"
+                    )
```
aarnphm, line 374: this should work with `use_multiple_entry_point` as bentoml runner also have a signature feature.

bentoml/_internal/frameworks/onnxmlir.py
```diff
@@ -0,0 +1,396 @@
...
+                    model_path,
+                    "-o",
+                    bento_model.path_of(MODEL_FILENAME.split(".")[0]),
+                ]
+            )
+        else:
+            # docker run --rm -it -v $(pwd):/workdir onnxmlirczar/onnx-mlir:latest -O3 --EmitLib /workdir/docs/mnist_example/mnist.onnx -o torch_mnist
+            logger.info(
+                f"Using {RELEASE_MLIR_IMAGE} to compile given model {model_path}. It might take a while if this is the first time you download {RELEASE_MLIR_IMAGE}."
+            )
```
aarnphm, line 323: In this case, we can only run docker images on x86, ppc64le, and s390x. I can't seem to make the compiled model to run on arm ;(

bentoml/_internal/frameworks/onnxmlir.py
```diff
@@ -0,0 +1,396 @@
...
+def _get_onnxmlir_version() -> str:  # pragma: no cover (trivial)
+    import requests
+
+    version_url = "https://raw.githubusercontent.com/onnx/onnx-mlir/main/VERSION_NUMBER"
+    return requests.get(version_url).text.strip()
+
+
+def save_model(
+    name: str,
+    model_path: Path | PathType,
```
aarnphm, line 209: This means that we accept both `.so` and `.onnx`. For saving `.so` `trust_compiled_model` has to set to `True`.

bentoml/_internal/frameworks/onnxmlir.py
```diff
@@ -0,0 +1,396 @@
...
+ACCEPTED_ONNX_ENVVAR = ["ONNX_MLIR_HOME", "ONNX_MLIR_BIN"]
+
+
+@attr.define
+class ONNXMLirOptions(ModelOptions):
+    use_default_entry_point: bool = attr.field(default=True)
+    use_docker: bool = attr.field(default=False)
+
+
+def check_environment() -> tuple[str | None, str | None]:
```
aarnphm, line 90: This function provides best effort to find out `onnx-mlir` binary and `lib` which contains `PyRuntime`.

bentoml/_internal/frameworks/onnxmlir.py
```diff
@@ -0,0 +1,396 @@
...
+
+1. Set ONNX_MLIR_HOME to the path of 'build' directory of onnx-mlir. The 'build' directory refers to parent folder containing 'bin', 'lib' in which ONNX-MLIR executables and libraries can be found, typically '/path/to/onnx-mlir/build/Debug'.
+2. Set ONNX_MLIR_BIN to 'onnx-mlir/build/Debug/bin' directory. After, add $ONNX_MLIR_BIN to PATH environment variable: "export PATH=$ONNX_MLIR_BIN:$PATH" and reload current shell.
+"""
+
+ONNX_MLIR, ONNX_MLIR_RUNTIME = check_environment()
+
+
+@contextmanager
+def setup_execution_session() -> t.Generator[t.Type[ExecutionSession], None, None]:
```
aarnphm, line 128: Since we are creating a context manager here, I feel that we can definitely mount the docker process and use the prebuilt pyruntime. Note that this will only work for python 3.8 since the compiled PyRuntime on the docker image is for python 3.8.

seems it didn't work with other python version other than 3.8.

bentoml/_internal/frameworks/onnxmlir.py
```diff
@@ -0,0 +1,396 @@
...
+API_VERSION = "v1"
+
+# TODO: mount session to docker container and use prebuilt?
+DEV_MLIR_IMAGE = "onnxmlirczar/onnx-mlir-dev:latest"
+
+# release docker images to compile the binary
+RELEASE_MLIR_IMAGE = "onnxmlirczar/onnx-mlir:latest"
+DOCKER_WORKDIR = "/workdir"
+DOCKER_OUTDIR = "/output"
+COMPILE_ARG = ["--O3", "--EmitLib"]
```
aarnphm, line 70: opportunity for user to customize this. This will be P1 item where we can allow end user to use other onnx-mlir options.

bentoml/onnxmlir.py
```diff
@@ -1,7 +1,14 @@
...
+
+from typing import TYPE_CHECKING
+
+from ._internal.frameworks.onnxmlir import get
+from ._internal.frameworks.onnxmlir import load_model
+from ._internal.frameworks.onnxmlir import save_model
+from ._internal.frameworks.onnxmlir import get_runnable
+
+if TYPE_CHECKING:
+    from ._internal.frameworks.onnxmlir import ExecutionSession
```
aarnphm, line 11: Provides a type stub when developing using pyright or mypy

aarnphm (commented):
cc @andrewsi-z

bentoml/_internal/frameworks/onnxmlir.py
andrewsi-z: As observed PyRuntime has a hard compatibility requirement on the python level it was built for. We may be able build the onnx-mlir with multiple levels to help testing in the future.

aarnphm: what do you mean by this? Is it possible to support multiple python version upstream?

It seems to me that if bentoml provided a docker images that contains onnxmlir binary for all of the support python this would be the best.

aarnphm: but again this is still explorable components I'm hacking around.

andrewsi-z: I think the onnx-mlir image would have to include the .so pyruntime needs for each desired python level. Let me take a look at that and see if there is something that can be explored

aarnphm: sg.

aarnphm:
I will revisit this once we have an environment manager.

PR #2707, opened 2022-07-10
docs: add exception for --reload on windows

aarnphm:
add an exception for --reload on windows.

### DIFF
--- a/bentoml/_internal/cli/bento_server.py
+++ b/bentoml/_internal/cli/bento_server.py
@@ -5,6 +5,8 @@ import logging
 
 import click
 
+from bentoml.exceptions import BentoMLException
+
 from ..log import configure_server_logging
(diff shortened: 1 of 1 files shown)


bentoml/_internal/cli/bento_server.py
```diff
@@ -133,6 +135,11 @@ def serve(
         else:
             from ..server import serve_development
 
+            if reload and sys.platform == "win32":
+                raise BentoMLException(
+                    "--reload is not supported on Windows due to circus limitation."
```
sauyon, line 140:
```suggestion
                    "--reload is not supported on Windows due to circus limitations."
```

sauyon:
Can we make circus spawn the plugin with an FD instead of a stream?

PR #2708, opened 2022-07-10
docs(xgboost): XGBoost guide

sauyon:


docs/source/frameworks/xgboost.rst
```diff
@@ -2,49 +2,174 @@
...
+      # resources can be configured at the top level
+      resources:
+         nvidia.com/gpu: 0
+      # or per runner
+      my_runner_name:
+         resources:
+             nvidia.com/gpu: 0
+
+
+Adaptive Batching
```
yubozhao, line 162: Should we link to the adaptive batching docs in the following section? Otherwise looks good

docs/source/frameworks/xgboost.rst
```diff
@@ -2,49 +2,174 @@
...
+   param = {"max_depth": 3, "eta": 0.3, "objective": "multi:softprob", "num_class": 2}
+   bst = xgb.train(param, dt)
 
-      dt = xgb.DMatrix(X, label=y)
 
-      # specify parameters via map
-      param = {"max_depth": 3, "eta": 0.3, "objective": "multi:softprob", "num_class": 2}
-      bst = xgb.train(param, dt)
+After training, use save_model to save the Booster instance to BentoML model store. XGBoost has no
+framework-specific save options.
```
yubozhao, line 37: Could we bold/italic "no framework-specific" for better visibility?

sauyon: I think this is ok because I don't think a user would necessarily think there would be.

Maybe we should even remove that sentence? I think I was mostly trying to pad out the section because it felt so bare...

aarnphm (approved):

docs/source/frameworks/xgboost.rst
```diff
@@ -2,49 +2,178 @@
 XGBoost
 =======
 
-Users can now use XGBoost with BentoML with the following API: :code:`save_model`, :code:`load_model`, and :code:`get` as follows:
+This guide provides an overview of using `XGBoost <https://xgboost.readthedocs.io/en/stable/>`_ with BentoML.
```
aarnphm, line 5:
```suggestion
XGBoost is an optimized distributed gradient boosting library designed to be highly efficient, flexible and portable.. This guide provides an overview of using `XGBoost <https://xgboost.readthedocs.io/en/stable/>`_ with BentoML.
```

docs/source/frameworks/xgboost.rst
aarnphm: lets remove the sentence, I don't think this provides any value here.

aarnphm (approved):

docs/source/frameworks/xgboost.rst
sauyon: FWIW I don't think we need to describe each framework, but it's probably better to be consistent for now.

docs/source/frameworks/xgboost.rst
```diff
@@ -2,49 +2,178 @@
...
+
+.. code-block:: python
+
+   bentoml.xgboost.get("model_name:model_version").to_runner()
+
+``runner.predict.run`` is generally a drop-in replacement for ``booster.predict``. However, while it
+is possible to pass a ``DMatrix`` as input, BentoML does not support adaptive batching in that case.
+It is therefore recommended to use a NumPy ``ndarray`` or Pandas ``DataFrame`` as input instead.
+
+There are no special options for loading XGBoost.
```
sauyon, line 135: We should probably remove this too?

PR #2710, opened 2022-07-11
fix(loader): Bento's modelstore should be injected when loading from a bento

aarnphm:
address https://bentoml.slack.com/archives/CKRANBHPH/p1657508401830029

Since a bento contains the models, when loading from bento the global model store should be set to the modelstore inside bento, not the model store from BENTOML_HOME.


bojiang (changes requested):
The problem is not in the `bentoml containerize` I think.

PR #2713, opened 2022-07-11
fix(service): the bento not works without the default model store

bojiang:
## Checklist:

### DIFF
--- a/bentoml/_internal/runner/runner.py
+++ b/bentoml/_internal/runner/runner.py
@@ -20,6 +20,7 @@ from ..configuration.containers import BentoMLContainer
 if TYPE_CHECKING:
     from ..models import Model
     from .runnable import RunnableMethodConfig
+    from ..service.service import Service
 
 T = t.TypeVar("T", bound=Runnable)
 P = ParamSpec("P")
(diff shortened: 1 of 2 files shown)


bentoml/_internal/runner/runner.py
```diff
@@ -209,6 +211,13 @@ def init_local(self, quiet: bool = False) -> None:
         if not quiet:
             logger.warning("'Runner.init_local' is for debugging and testing only")
 
+        if self.service_ref:
+            service = self.service_ref()
+            if service and service.bento:
+                with BentoMLContainer.model_store.patch(service.bento._model_store):  # type: ignore
+                    self._init_local()
+                return
```
ssheng, line 219: We dot not expect an `else` case here, right? Should we raise?

aarnphm: if user run this in a separate Runner or distributedly, the behaviour is the same as before.

we only want to change the model store if it runs inside the bento.

aarnphm: hense, raise is not desirable here.

ssheng: Understood the distributed case, what do you mean by running this in a separate Runner? Do you mean a separate Service?

aarnphm: separate runner.

aarnphm (approved):

bojiang:
@sauyon let's discuss about that later~

PR #2714, opened 2022-07-11
fix(docker): sharing dependency cache

aarnphm:
https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md#run---mounttypecache

Currently, dependency cache will be invalid for any subsequent `bentoml build` and `bentoml containerize`.
This PR address this by assigning the checksum of given bentofile.yaml as a cache key.

This improves build time for subsequent iteration by 40%

### DIFF
--- a/bentoml/_internal/bento/bento.py
+++ b/bentoml/_internal/bento/bento.py
@@ -138,9 +138,8 @@ class Bento(StoreItem):
     def create(
         cls,
         build_config: BentoBuildConfig,
-        version: str | None = None,
-        build_ctx: str | None = None,
-        hash_key: str = "",
+        version: t.Optional[str] = None,
+        build_ctx: t.Optional[str] = None,
     ) -> Bento:
         from ..service.loader import import_service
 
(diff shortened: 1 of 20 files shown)


bentoml/_internal/bento/bento.py
```diff
@@ -138,8 +138,9 @@ def info(self) -> BentoInfo:
     def create(
         cls,
         build_config: BentoBuildConfig,
-        version: t.Optional[str] = None,
-        build_ctx: t.Optional[str] = None,
+        version: str | None = None,
+        build_ctx: str | None = None,
+        hash_key: str = "",
```
sauyon, line 143: Can we just hash the relevant parts of `build_config` here instead?

bentoml/_internal/bento/docker/templates/_macros.j2
```diff
@@ -39,3 +43,4 @@ if [ -f {{ path }} ]; then
 fi
 EOF
 {%- endmacro -%}
+
```
sauyon, line 46: extraneous newline?

sauyon (approved):
LGTM minus one tiny niggle

parano:
> This PR address this by assigning the checksum of given bentofile.yaml as a cache key.

does it matter if a referenced requirements.txt file change?

PR #2725, opened 2022-07-12
fix: setup_script should run with `USER bentoml`

parano:
## Description

It's common for user's setup script to download files in the "user"'s home directory, here's an example with the NLTK custom runner example: https://github.com/bentoml/gallery/pull/136/files#diff-2ab6e592645153e012e4d282fdf21b410441cb78081cdfbbb1e56730bab71bf3R8

Prior to this change, user need to modify code and figure out where to write or load the data files, which can be very complicated process from user perspective. But everything just works if we run the setup_script with `USER bentoml`, where user service code will be executed with

## Checklist:


aarnphm (approved):
I'm ok with this for now.

We probably want to limit changes such as this due to permission after 1.0 . But since now it is before the releases it is fine.

PR #2728, opened 2022-07-12
fix: revert #2725 and fix setup script permission

parano:
## Description

This fixes this issue @aarnphm brought up: https://github.com/bentoml/BentoML/pull/2725#discussion_r918904973

Setup scripts are executed as root, same as before #2724, so that users can run commands like `apt-get install`. By setting `HOME` env var to `/home/bentoml`, libraries like NLTK will respect that and download data to the specified HOME directory.

## Checklist:


bentoml/_internal/bento/docker/templates/base.j2
```diff
@@ -44,6 +44,7 @@ ENV {{ key }}={{ value }}
 ARG BENTO_PATH={{ bento__path }}
 ENV BENTO_PATH=$BENTO_PATH
 ENV BENTOML_HOME={{ bento__home }}
+ENV HOME={{ bento__home }}
```
aarnphm, line 47: I don't think this would work either, since this will mess up with install other library that use $HOME.

Currently $HOME is setting to /root, not /home/bentoml

aarnphm (approved):
I'm a bit hesitant with this approach as well, but it is ok I guess. If user ran into any issues we will circle back to this implementation.

PR #2746, opened 2022-07-14
style: gRPC

aarnphm:
Using [buf](https://docs.buf.build/introduction) for linter and formatter.
follow Google styling reference for gRPC.

Apparently it is practice to include protos directory into package itself.
@ssheng

We will ignore generate stubs for now, and use `generate.py` to generate the
stubs to bentoml.protos

Setup a versioning for proto file. This will solve imports issue and we don't have to use `sh` scripts.

This would help setting up imports and style correctly. cc @Proto007


bentoml/_internal/io_descriptors/numpy.py
```diff
@@ -516,7 +510,7 @@ def from_sample(
                 :code:`enforce_shape=False`.
 
         Returns:
-            :obj:`~bentoml._internal.io_descriptors.NumpyNdarray`: :code:`NumpyNdarray` IODescriptor from given users inputs.
+            :obj:`~bentoml._internal.payload.NumpyNdarray`: :code:`NumpyNdarray` IODescriptor from given users inputs.
```
aarnphm, line 513:
```suggestion
            :obj:`~bentoml._internal.io_descriptors.NumpyNdarray`: :code:`NumpyNdarray` IODescriptor from given users inputs.
```

bentoml/_internal/io_descriptors/numpy.py
```diff
@@ -115,7 +115,7 @@ def predict(input_array: np.ndarray) -> np.ndarray:
             must be specified.
 
     Returns:
-        :obj:`~bentoml._internal.io_descriptors.IODescriptor`: IO Descriptor that :code:`np.ndarray`.
+        :obj:`~bentoml._internal.payload.IODescriptor`: IO Descriptor that :code:`np.ndarray`.
```
aarnphm, line 118:
```suggestion
        :obj:`~bentoml._internal.io_descriptors.IODescriptor`: IO Descriptor that :code:`np.ndarray`.
```

PR #2753, opened 2022-07-15
chore: lock imageio<=2.9.0 to address #2416

### LINKED ISSUE #2416: Image adapters break with pillow==9.1.0 for grayscale for bentoml==0.13.1
**Describe the bug**

If you use an `ImageInput` or `AnnotatedImageInput` when `pillow==9.1.0` is installed and try to submit a grayscale image it will error.

The issue is caused by `imageio.imread()` if we pin `imageio==2.9.0`. Unfortunately upgrading `imageio` is not possible since it breaks  bentoml in a different way.

This is *not* an issue for bentoml 1.0 since it uses `Pillow` directly and you'll just get an image with two dimensions that you convert. But since it's not out yet, this might be helpful for others. 

**To Reproduce**

Create a simple bentoml app that uses Image input such as:

```
import bentoml
from bentoml.adapters import JsonOutput, ImageInput

class ImageTest(bentoml.BentoService):

    @bentoml.api(input=ImageInput(), output=JsonOutput())
    def predict(self, image):
        return {
            "size" : str(image.shape)
        }
```

Submit a grayscale image and it will break if `pillow==9.1.0`

**Solution**

For any services that use `bentoml<1.0` you must pin `pillow<9.1.0` and `imageio<=2.9.0`

aarnphm:

## What does this PR address?

Fixes #2416

If i run `make format` it will just reformat a lot of files :)

## Before submitting:

## Who can help review?

cc @ssheng @parano

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -78,7 +78,7 @@ test_requires = [
     "codecov",
     "coverage>=4.4",
     "flake8>=3.8.2",
-    "imageio>=2.5.0",
+    "imageio>=2.5.0,<=2.9.0",
     "mock>=2.0.0",
     "moto==1.3.14",
     "pandas",
--- a/bentoml/adapters/annotated_image_input.py
+++ b/bentoml/adapters/annotated_image_input.py
@@ -25,8 +25,8 @@ from bentoml.types import InferenceTask, JsonSerializable, Optional
 from bentoml.utils.lazy_loader import LazyLoader
 
 # BentoML optional dependencies, using lazy load to avoid ImportError
-imageio = LazyLoader('imageio', globals(), 'imageio')
-numpy = LazyLoader('numpy', globals(), 'numpy')
+imageio = LazyLoader("imageio", globals(), "imageio")
+numpy = LazyLoader("numpy", globals(), "numpy")
 
 
 ApiFuncArgs = Tuple[Sequence[numpy.ndarray], Sequence[Optional[JsonSerializable]]]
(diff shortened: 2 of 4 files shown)


setup.py
```diff
@@ -78,7 +78,7 @@
     "codecov",
     "coverage>=4.4",
     "flake8>=3.8.2",
-    "imageio>=2.5.0",
+    "imageio>=2.5.0,<=2.9.0",
```
ssheng, line 81: `imageio` is currently on 2.19 so adding this restriction will make our dependency a bit limiting. Why do we not support 2.9+?

aarnphm: https://github.com/bentoml/BentoML/issues/2416

ssheng: > Unfortunately upgrading imageio is not possible since it breaks bentoml in a different way.

Do you know what breakages were referred to here? I think locking a non-major version upper bound is not ideal since it may introduce conflicts to user environments. If the failure is very niche, we can answer this question in our Github discussion. Wdyt?

aarnphm: it seems that the 2.9.0 is released at 10 Jul 2020, which is around the time we are already working on migrating to 1.0. I think we then didn't follow up with updated changes from imageio. 

Sure we can answer it in GitHub actions, but I think a warning would deliver this faster.

aarnphm:
I believe this is not an issue for bentoml 1.0.

aarnphm:
I will take a look at CI for this sometime later, but I think I will put this in for now.

PR #2769, opened 2022-07-19
chore: clean up dependencies

sauyon:
Cleaned up unused and barely-used dependencies.

### DIFF
--- a/bentoml/_internal/yatai_rest_api_client/schemas.py
+++ b/bentoml/_internal/yatai_rest_api_client/schemas.py
@@ -21,7 +21,7 @@ def datetime_encoder(time_obj: Optional[datetime]) -> Optional[str]:
     return time_obj.strftime(time_format)
 
 
-def datetime_decoder(datetime_str: Optional[str], _) -> Optional[datetime]:
+def datetime_decoder(datetime_str: Optional[str], _: Any) -> Optional[datetime]:
     if not datetime_str:
         return None
     return parse(datetime_str)
@@ -41,7 +41,7 @@ def schema_from_json(json_content: str, cls: Type[T]) -> T:
     return converter.structure(dct, cls)
 
 
-def schema_to_json(obj: T) -> str:
+def schema_to_json(obj: Any) -> str:
     res = converter.unstructure(obj, obj.__class__)
     return json.dumps(res)
 
--- a/requirements/dev-requirements.txt
+++ b/requirements/dev-requirements.txt
@@ -1,10 +1,5 @@
 # Dev dependencies
 -r tests-requirements.txt
-attrs>=21.1
-GitPython>=2.0
 setup-cfg-fmt
-yamllint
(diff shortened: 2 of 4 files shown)


bentoml/_internal/yatai_rest_api_client/schemas.py
```diff
@@ -24,7 +23,7 @@ def datetime_encoder(time_obj: Optional[datetime]) -> Optional[str]:
 def datetime_decoder(datetime_str: Optional[str], _) -> Optional[datetime]:
     if not datetime_str:
         return None
-    return parse(datetime_str)
+    return datetime.strptime(datetime_str, time_format)
```
sauyon, line 26: @parano @yetone was there a reason you were using `dateutil` for this?

aarnphm (dismissed):

bentoml/_internal/yatai_rest_api_client/schemas.py
yetone: @sauyon Because the default JSON time format in Go is RFC 3339

https://pkg.go.dev/time#Time.MarshalJSON

image

requirements/dev-requirements.txt
```diff
@@ -1,10 +1,5 @@
 # Dev dependencies
 -r tests-requirements.txt
-attrs>=21.1
-GitPython>=2.0
 setup-cfg-fmt
-yamllint
-setuptools>=62.0.0
 twine
 wheel
-backports.cached_property
```
bojiang, line 10: Without this dependency, the linter will complain about it with Python 3.7

sauyon: Don't we have it in `setup.cfg` already though?

bentoml/_internal/yatai_rest_api_client/schemas.py
sauyon: Should we use that format when saving too then?

requirements/dev-requirements.txt
bojiang: oh you are right

bojiang (dismissed):

bentoml/_internal/yatai_rest_api_client/schemas.py
aarnphm: `datetime.strptime` should be RFC3339 compatible right?

sauyon: Maybe I'll just revert this change for now anyway; we can worry about this later.

sauyon: Maybe I'll just revert this change for now anyway; we can worry about this later.

aarnphm: Let's just revert this :)

aarnphm (approved):

PR #2772, opened 2022-07-20
fix: add missing init_http_response to io descriptor for multipart to work

aarnphm:
Signed-off-by: Aaron Pham <29749331+aarnphm@users.noreply.github.com>

## What does this PR address?

I notice that some IO descriptors miss `init_http_response`, which is required
to use with Multipart IO descriptors.

I guess we never have test for using these IO descriptor with multipart, hence
we never spotted this.

## Before submitting:

- [ ] Does the Pull Request follow [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/#summary) naming? Here are [GitHub's
guide](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request) on how to create a pull request.
- [ ] Does the code follow BentoML's code style, both `make format` and `make lint` script have passed ([instructions](https://github.com/bentoml/BentoML/blob/main/DEVELOPMENT.md#style-check-auto-formatting-type-checking))?
- [ ] Did you read through [contribution guidelines](https://github.com/bentoml/BentoML/blob/main/CONTRIBUTING.md#ways-to-contribute) and follow [development guidelines](https://github.com/bentoml/BentoML/blob/main/DEVELOPMENT.md#start-developing)?
- [ ] Did your changes require updates to the documentation? Have you updated
  those accordingly? Here are [documentation guidelines](https://github.com/bentoml/BentoML/tree/main/docs) and [tips on writting docs](https://github.com/bentoml/BentoML/tree/main/docs#writing-documentation).
- [ ] Did you write tests to cover your changes?

PR #2774, opened 2022-07-20
fix(io descriptors): finish removing init_http_response

sauyon:
This should fix the multipart issue discovered with #2772.


bentoml/protos/payload.proto
```diff
@@ -0,0 +1,55 @@
+syntax = "proto3";
```
aarnphm, line 1: oo need to remove this file :)

aarnphm (dismissed):

aarnphm (approved):

PR #2775, opened 2022-07-20
feat(model): allow custom model versions

### LINKED ISSUE #2750: feature: api for custom model and bento tags
### Feature request

Hello! Is it possible to change the version assigned to a model before saving it? I’d like to be able to assign a semantic version instead of the random generated version in the Tag, This used to be possible with artifacts in 0.13, but I can’t find a way to do it now with 1.0.0. A possible way to do this might be to allow an argument tag in the save_model method that takes a bentoml.Tag and that supersedes the argument name .Thanks :) 

### Motivation

_No response_

### Other

_No response_

sauyon:
Should resolve #2750.

### DIFF
--- a/bentoml/_internal/models/model.py
+++ b/bentoml/_internal/models/model.py
@@ -179,7 +179,7 @@ class Model(StoreItem):
         metadata = {} if metadata is None else metadata
         options = ModelOptions() if options is None else options
 
-        model_fs = fs.open_fs(f"temp://bentoml_model_{name}")
+        model_fs = fs.open_fs(f"temp://bentoml_model_{tag.name}")
 
         return cls(
             tag,
--- a/bentoml/_internal/models/model.py
+++ b/bentoml/_internal/models/model.py
@@ -18,9 +18,9 @@ import fs.errors
 import fs.mirror
 import cloudpickle  # type: ignore (no cloudpickle types)
 from fs.base import FS
-from cattr.gen import override
-from cattr.gen import make_dict_structure_fn
-from cattr.gen import make_dict_unstructure_fn
+from cattr.gen import override  # type: ignore (incomplete cattr types)
+from cattr.gen import make_dict_structure_fn  # type: ignore (incomplete cattr types)
+from cattr.gen import make_dict_unstructure_fn  # type: ignore (incomplete cattr types)
 from simple_di import inject
 from simple_di import Provide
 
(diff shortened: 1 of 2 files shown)


bentoml/_internal/models/model.py
```diff
@@ -18,9 +18,9 @@
 import fs.mirror
 import cloudpickle  # type: ignore (no cloudpickle types)
 from fs.base import FS
-from cattr.gen import override
-from cattr.gen import make_dict_structure_fn
-from cattr.gen import make_dict_unstructure_fn
+from cattr.gen import override  # type: ignore (incomplete cattr types)
+from cattr.gen import make_dict_structure_fn  # type: ignore (incomplete cattr types)
+from cattr.gen import make_dict_unstructure_fn  # type: ignore (incomplete cattr types)
```
aarnphm, line 23: no need for type ignore here. Since this will be addressed in newer version of cattrs.

aarnphm (approved):

aarnphm:
tests are failing https://github.com/bentoml/BentoML/runs/7421909493?check_suite_focus=true @sauyon

bentoml/_internal/models/model.py
```diff
@@ -172,7 +172,9 @@ def create(
         Returns:
             object: Model instance created in temporary filesystem
         """
-        tag = Tag(name).make_new_version()
```
parano, line 175: let's update the inline docstring above to say it also accepts "tag"?

should we rename all the param `name` in save_models, to `tag`? or should we just keep it a hidden feature this way so we don't encourage users to use it?

bentoml/_internal/models/model.py
sauyon: Sure, but it hasn't been addressed yet.

bentoml/_internal/models/model.py
sauyon: I don't know that we want to expose this behavior. Maybe we should add a kwarg that we can warn people about?

sauyon:
Ah, Windows.

tests/unit/_internal/models/test_model.py
```diff
@@ -173,6 +173,18 @@ def test_model_creationtime():
     assert repr(model_a) == f'Model(tag="{model_a.tag}", path="{model_a.path}")'
 
 
+def test_model_version():
```
parano, line 176: 👍

bentoml/_internal/models/model.py
parano: I agree we don't want to expose this behavior, I think the way it is should be good

PR #2803, opened 2022-07-24
bug: fix JSON io_descriptor validate_json option

parano:
## What does this PR address?

Currently, when using a JSON io descriptor with a pydantic class, the input data type also depends on the validate_json option with is very confusing. Essentially when a user provides a pydantic class, it is expected to receive an instance of that pydantic class as the inference API callback function's input parameter. However, currently, this is only the case when validate_json=True. If the user sets a pydantic model while setting validate_json=False, it will still pass in a regular dict to user-defined API function and completely ignore the pydantic option.

This PR proposes to remove the `validate_json` option and make sure the input value type to user-defined API function is consistent:

* if pydantic is None, the input value will be type object
* if pydantic is not None, the input value will be an instance of target pydantic class

As for validation logic, the user should be able to define via the pydantic class itself. E.g. user can set the pydantic class to allow extra fields or set some of the fields as optional. See the example here: https://github.com/bentoml/gallery/pull/146

This will be a breaking change since the validate_json option is being removed

## Before submitting:

## Who can help review?

### DIFF
--- a/bentoml/_internal/io_descriptors/json.py
+++ b/bentoml/_internal/io_descriptors/json.py
@@ -170,7 +170,10 @@ class JSON(IODescriptor[JSONType]):
         else:
             return json_obj
 
-    async def to_http_response(self, obj: JSONType, ctx: Context | None = None):
+    async def to_http_response(self, obj: JSONType | pydantic.BaseModel, ctx: Context | None = None):
+        if isinstance(obj, pydantic.BaseModel):
+            obj = obj.dict()
+
         json_str = json.dumps(
             obj,
             cls=self._json_encoder,
--- a/bentoml/_internal/io_descriptors/json.py
+++ b/bentoml/_internal/io_descriptors/json.py
@@ -14,6 +14,11 @@ from ..utils.http import set_cookies
 from ...exceptions import BadInput
 from ...exceptions import MissingDependencyException
 
+try:
+    import pydantic
+except ImportError:
+    pydantic = None
+
 if TYPE_CHECKING:
     from types import UnionType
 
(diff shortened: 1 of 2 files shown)


bentoml/_internal/io_descriptors/json.py
```diff
@@ -165,25 +156,26 @@ def openapi_responses_schema(self) -> t.Dict[str, t.Any]:
...
-                return json.loads(json_str)
-            except json.JSONDecodeError as e:
-                raise BadInput(f"Json validation error: {e}") from None
+            return json_obj
+
+    async def to_http_response(
+        self, obj: JSONType | pydantic.BaseModel, ctx: Context | None = None
+    ):
+        if isinstance(obj, pydantic.BaseModel):
+            obj = obj.dict()
```
parano, line 177: this may not be used so commonly, but it allows user to use a pydantic instance as API return value

bentoml/_internal/io_descriptors/json.py
```diff
@@ -14,6 +14,11 @@
 from ...exceptions import BadInput
 from ...exceptions import MissingDependencyException
 
+try:
+    import pydantic
+except ImportError:
+    pydantic = None
+
 if TYPE_CHECKING:
     from types import UnionType
```
aarnphm, line 23: Small thing
Lets use LazyLoader here instead. If user don't want to use pydantic but happens to have pydantic install, bentoml will try to load this.

aarnphm (dismissed):
I agree with this change. This would make generating openAPI schema a bit easier if user provides their pydantic model. Will address that in a separate PR wrt openapi generation.

bentoml/_internal/io_descriptors/json.py
```diff
@@ -129,13 +137,10 @@ def predict(input_arr: pd.DataFrame):
     def __init__(
         self,
         pydantic_model: t.Type[pydantic.BaseModel] | None = None,
-        validate_json: bool = True,
```
ssheng, line 132: Maybe we should leave this in for backward compatibility. We can document the deprecation and raise a warning if the user explicitly set it as `False`.

ssheng: We can remove these minor breakages in the next minor release. Within a minor release, I think we should aim for zero breaking changes.

bentoml/_internal/io_descriptors/json.py
```diff
@@ -67,58 +72,61 @@ class JSON(IODescriptor[JSONType]):
...
+            http://127.0.0.1:3000/classify
 
-            {"res":"[{\"0\":1}]"}%
+        [2]%
 
     Args:
         pydantic_model (:code:`pydantic.BaseModel`, `optional`, default to :code:`None`):
-            Pydantic model schema.
-        validate_json (:code:`bool`, `optional`, default to :code:`True`): If True, then use
-            Pydantic model specified above to validate given JSON.
```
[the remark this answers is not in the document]
parano: done

bentoml/_internal/io_descriptors/json.py
parano: @ssheng updated accordingly, there should be zero breaking change besides a minor type change. Any existing user code should work, and a warning will be printed if user have set `validate_json` in their code.

ssheng (approved):

PR #2830, opened 2022-07-28
fix: remove chmod install scripts

aarnphm:

## What does this PR address?

`install.sh` and all related scripts generated by BentoML should use bash to run
instead of `chmod +x`.

Usually `chmod +x` are not recommended for container application.

## Before submitting:

## Who can help review?

cc @sauyon

### DIFF
--- a/bentoml/_internal/bento/build_config.py
+++ b/bentoml/_internal/bento/build_config.py
@@ -225,6 +225,10 @@ class DockerOptions:
                 setup_script = resolve_user_filepath(self.setup_script, build_ctx)
             except FileNotFoundError as e:
                 raise InvalidArgument(f"Invalid setup_script file: {e}")
+            if not os.access(setup_script, os.X_OK):
+                raise InvalidArgument(
+                    f"{setup_script} is not executable. Make it executable with 'chmod +x {setup_script}' if you are on Unix."
+                )
             copy_file_to_fs_folder(
                 setup_script, bento_fs, docker_folder, "setup_script"
             )
--- a/bentoml/_internal/bento/build_config.py
+++ b/bentoml/_internal/bento/build_config.py
@@ -13,6 +13,7 @@ from typing import TYPE_CHECKING
 import fs
 import attr
 import yaml
+import psutil
 import fs.copy
 from dotenv import dotenv_values  # type: ignore
 from pathspec import PathSpec
@@ -226,9 +227,12 @@ class DockerOptions:
             except FileNotFoundError as e:
                 raise InvalidArgument(f"Invalid setup_script file: {e}")
             if not os.access(setup_script, os.X_OK):
-                raise InvalidArgument(
-                    f"{setup_script} is not executable. Make it executable with 'chmod +x {setup_script}' if you are on Unix."
-                )
+                message = f"{setup_script} is not executable."
+                if not psutil.WINDOWS:
+                    raise InvalidArgument(
+                        f"{message} Make sure to 'chmod +x {setup_script}'."
+                    )
+                raise InvalidArgument(message)
             copy_file_to_fs_folder(
                 setup_script, bento_fs, docker_folder, "setup_script"
             )
--- a/bentoml/_internal/bento/build_config.py
+++ b/bentoml/_internal/bento/build_config.py
@@ -230,7 +230,7 @@ class DockerOptions:
                 message = f"{setup_script} is not executable."
                 if not psutil.WINDOWS:
                     raise InvalidArgument(
-                        f"{message} Make sure to 'chmod +x {setup_script}'."
+                        f"{message} Ensure the script has a shebang line, then run 'chmod +x {setup_script}'."
                     )
                 raise InvalidArgument(message)
             copy_file_to_fs_folder(
--- a/bentoml/_internal/bento/docker/templates/_macros.j2
+++ b/bentoml/_internal/bento/docker/templates/_macros.j2
@@ -38,7 +38,7 @@ set -euxo pipefail
 
 if [ -f {{ path }} ]; then
   echo "{{ echo_msg }}..."
-  chmod +x {{ path }}
+  bash {{ path }}
   {{ path }}
 fi
 EOF
--- a/bentoml/_internal/bento/docker/templates/_macros.j2
+++ b/bentoml/_internal/bento/docker/templates/_macros.j2
@@ -30,16 +30,3 @@ fi
 EOF
 
 {%- endmacro -%}
-
-{%- macro run_script(path, echo_msg, run_args="", runexec=bash) -%}
-# Running {{ echo_msg }}
-RUN {{ run_args }} bash <<EOF
-set -euxo pipefail
-
-if [ -f {{ path }} ]; then
-  echo "{{ echo_msg }}..."
-  {{ runexec }} {{ path }}
-fi
-EOF
-{%- endmacro -%}
-
--- a/bentoml/_internal/bento/docker/templates/_macros.j2
+++ b/bentoml/_internal/bento/docker/templates/_macros.j2
@@ -31,14 +31,14 @@ EOF
 
 {%- endmacro -%}
 
-{%- macro run_script(path, echo_msg, run_args="") -%}
+{%- macro run_script(path, echo_msg, run_args="", runexec=bash) -%}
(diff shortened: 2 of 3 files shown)


bentoml/_internal/bento/docker/templates/_macros.j2
```diff
@@ -38,7 +38,7 @@ set -euxo pipefail
 
 if [ -f {{ path }} ]; then
   echo "{{ echo_msg }}..."
-  chmod +x {{ path }}
+  bash {{ path }}
   {{ path }}
```
sauyon, line 42: Seems to me like this line should be removed, and I vaguely recall seeing another chmod in this template somewhere.

aarnphm: the chmod inside Docker are for entrypoint, not related.

sauyon (dismissed):
LGTM; I assume you've tested this?

bentoml/_internal/bento/docker/templates/_macros.j2
```diff
@@ -38,8 +38,7 @@ set -euxo pipefail
 
 if [ -f {{ path }} ]; then
   echo "{{ echo_msg }}..."
-  chmod +x {{ path }}
-  {{ path }}
+  bash {{ path }}
```
parano, line 41: does this work for __setup_script__, where user can provide a python script?

parano: could you test if this works for this example project? https://github.com/bentoml/gallery/blob/main/custom_runner/nltk_pretrained_model/bentofile.yaml#L8

parano: See https://docs.bentoml.org/en/latest/concepts/bento.html#setup-script

"[Shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)) .. are required in a setup script provided to BentoML."

parano: however, the current implementation for `setup_script` just copies over the user-provided script and does not guarantee the provided script has an execute permission.  we can probably add an `os.chmod`  call to ensure the file permission in Bento directory.  @sauyon what do you think?

aarnphm: Maybe for setup scripts we will just use `. setup_script` instead of using bash.

bentoml/_internal/bento/docker/templates/base.j2
```diff
@@ -54,12 +54,39 @@ COPY --chown={{ bento__user }}:{{ bento__user }} . ./
...
+    lang = code[0].replace("#!", "").strip().split()[-1]
+    return subprocess.check_call([lang, script, *args[1:]])
+
+
+if __name__ == "__main__":
+    import sys
+
+    raise SystemExit(main(sys.argv))
+EOT
+RUN chmod +x {{ bento__home }}runscript
```
aarnphm, line 81: This is the solution that I came up with.

bentoml/_internal/bento/docker/templates/_macros.j2
sauyon: Yeah, we should probably do that. Bash scripts / our own python scripts run like this, and user scripts must be executable.

Can the user specify `bash script.sh` as a custom script?

aarnphm: I think they can

parano: No user can not do that, the setup script must be path to a file

aarnphm:
cc @sauyon

bentoml/_internal/bento/docker/templates/base.j2
```diff
@@ -54,12 +54,21 @@ COPY --chown={{ bento__user }}:{{ bento__user }} . ./
 # Block SETUP_BENTO_COMPONENTS
 {% block SETUP_BENTO_COMPONENTS %}
 
+{% set __runexec__ = [bento__home, "runscript"] | join("") %}
```
aarnphm, line 57:
```suggestion
```

bentoml/_internal/bento/docker/templates/base.j2
```diff
@@ -56,10 +56,18 @@ COPY --chown={{ bento__user }}:{{ bento__user }} . ./
...
 {% set __setup_script__ = expands_bento_path("env", "docker", "setup_script", bento_path=bento__path) %}
-{{ common.run_script(__setup_script__, "user setup scripts") }}
+RUN bash <<EOF
+set -euxo pipefail
+
+if [ -f {{ __setup_script__ }} ]; then
+    echo "Installing custom setup script..."
+    bash {{ __setup_script__ }}
+fi
+EOF
```
sauyon, line 70: Didn't we discuss this needed to be just 
```suggestion
RUN {{ __setup_script__ }}
```

We should probably do validation of the setup script on bento build; maybe we can chmod there.

sauyon, line 70: Didn't we discuss this needed to be just 
```suggestion
RUN {{ __setup_script__ }}
```

We should probably do validation of the setup script on bento build; maybe we can chmod there.

aarnphm: If the file arent there ur suggestion will fail to build

bentoml/_internal/bento/docker/templates/base.j2
```diff
@@ -56,10 +56,18 @@ COPY --chown={{ bento__user }}:{{ bento__user }} . ./
...
+RUN {{ __pip_cache__ }} bash -euxo pipefail {{ __install_python_scripts__ }}
 
 {% set __setup_script__ = expands_bento_path("env", "docker", "setup_script", bento_path=bento__path) %}
-{{ common.run_script(__setup_script__, "user setup scripts") }}
+RUN bash <<EOF
+set -euxo pipefail
+
+if [ -f {{ __setup_script__ }} ]; then
+    echo "Installing custom setup script..."
+    bash {{ __setup_script__ }}
```
aarnphm, line 68:
```suggestion
    {{ __setup_script__ }}
```

bentoml/_internal/bento/build_config.py
```diff
@@ -225,6 +226,13 @@ def write_to_bento(
                 setup_script = resolve_user_filepath(self.setup_script, build_ctx)
             except FileNotFoundError as e:
                 raise InvalidArgument(f"Invalid setup_script file: {e}")
+            if not os.access(setup_script, os.X_OK):
+                message = f"{setup_script} is not executable."
+                if not psutil.WINDOWS:
+                    raise InvalidArgument(
+                        f"{message} Make sure to 'chmod +x {setup_script}'."
```
sauyon, line 233:
```suggestion
                        f"{message} Ensure the script has a shebang line, then run 'chmod +x {setup_script}'."
```

sauyon (approved):
👍

bentoml/_internal/bento/build_config.py
```diff
@@ -225,6 +226,13 @@ def write_to_bento(
                 setup_script = resolve_user_filepath(self.setup_script, build_ctx)
             except FileNotFoundError as e:
                 raise InvalidArgument(f"Invalid setup_script file: {e}")
+            if not os.access(setup_script, os.X_OK):
```
parano, line 229: is this required?  don't we do `chmod +x` in the dockerfile with the changes below? @aarnphm @sauyon

aarnphm: The `chmod` in the dockerfile are just docker related, we want to have a executable check here so that users make sure the script is runnable.

parano: @aarnphm it's not necessary, right? since we only run that script in docker?

aarnphm: I think the intention for this is to check whether a script is executable or not. But that evolve using a library like [python-magic](https://github.com/ahupp/python-magic). I think we end up just doing this as a first pass.

sauyon: I'm not entirely sure what we want to do here. But the check is indeed not required.

aarnphm: 🤔

PR #2832, opened 2022-07-29
feat: use yatai proxy to upload/download bentos/models

yetone:


parano:
@yetone Do we want to make s3 presigned URL an option for user to enable?  we may need that for backwards compatible as well

yetone:
@parano Forward compatibility is complete, please review and merge

bentoml/_internal/yatai_client/__init__.py
```diff
@@ -245,9 +246,9 @@ def push_model(model: "Model"):
                 runnable_type=r.runnable_type,
                 models=r.models,
                 resource_config=BentoRunnerResourceSchema(
-                    cpu=r.resource_config.cpu,
-                    nvidia_gpu=r.resource_config.nvidia_gpu,
-                    custom_resources=r.resource_config.custom_resources,
+                    cpu=r.resource_config.get("cpu"),
```
yetone, line 249: I hope this type-level refactoring needs to be synchronized to all the places where the type is used, and it feels like we need to add all the type checks to our ci, otherwise, it's a nullity

bentoml/_internal/yatai_client/__init__.py
```diff
@@ -25,6 +25,7 @@
 from rich.progress import TimeElapsedColumn
 from rich.progress import TimeRemainingColumn
 from rich.progress import TransferSpeedColumn
+from typing_extensions import Literal
```
aarnphm, line 28: We remove typing_extensions from out dependency, lets make this inside TYPE_CHECKING

with `from __future__ import annotations`

bojiang (approved):

PR #2843, opened 2022-07-30
feat: add support for excluding urls from tracing

### LINKED ISSUE #2829: feature: allow user to specify URLs to exclude from tracing
### Feature request

Currently, if tracing is enabled, traces are collected / sampled from all routes. There should be a way to disable tracing on certain routes, especially when a deployment on Kubernetes may have liveness and readiness probes pinging the deployment frequently. Since we cannot disable tracing on the liveness/readiness probe routes, many redundant traces get collected.

For example, you can see below that a lot of the traces collected are of the liveness and readiness probe, and only one of them has a useful trace for the predict route.
image

For other opentelemetry instrumentations, we can use an environment variable like `OTEL_PYTHON_FASTAPI_EXCLUDED_URLS` to specify routes to exclude from tracing, we should be able to do something similar.

### Motivation

_No response_

### Other

_No response_

jiewpeng:
## What does this PR address?

Fixes #2829 

## Before submitting:
No: requires external dependency (Jaeger)

## Who can help review?

Tested manually by

1. Take an existing working Bento (e.g. from using the basic iris-clf example)
2. Package the new bentoml wheel file into the Bento
3. Use docker-compose to set up the Bento and a Jaeger
```
version: '3.9'
services:
  iris_clf:
    build:
      context: <Bento folder>
      dockerfile: env/docker/Dockerfile
    ports:
      - "3000:3000"
    environment:
      - OTEL_SERVICE_NAME=iris-clf
      - BENTOML_CONFIG=bentoml_config.yaml
    volumes:
      - ./bentoml_config.yaml:/home/bentoml/bento/bentoml_config.yaml
  jaeger:
    image: jaegertracing/all-in-one:1.35
    ports:
      - "6831:6831"
      - "16686:16686"
```
4. Use this bentoml_config.yaml - uncomment or comment the `excluded_urls` line to disable / enable the routes
```
tracing:
  type: jaeger
  sample_rate: 1.0
  excluded_urls: readyz,livez,healthz,static_content,docs
  jaeger:
    address: jaeger
    port: 6831
```
5. Visit `http://localhost:3000` (model's API server) on the browser to trigger the tracing. Visit `http://localhost:16686` (Jaeger) to check the traces.

When the `excluded_urls` line is commented out (i.e. trace all routes):

before-change

When the `excluded_urls` line is left uncommented (i.e. exclude the routes):

after-change

### DIFF
--- a/bentoml/_internal/configuration/__init__.py
+++ b/bentoml/_internal/configuration/__init__.py
@@ -45,7 +45,7 @@ def expand_env_var(env_var: str) -> str:
 
 def clean_bentoml_version(bentoml_version: str) -> str:
     post_version = bentoml_version.split("+")[0]
-    match = re.match(r"^(\d+)\.(\d+)\.(post)?(\d+)(?:(a|rc)\d)*", post_version)
+    match = re.match(r"^(\d+)\.(\d+)\.(\d+)(?:(a|rc)\d)*", post_version)
     if match is None:
         raise BentoMLException("Errors while parsing BentoML version.")
     return match.group()
--- a/bentoml/_internal/configuration/__init__.py
+++ b/bentoml/_internal/configuration/__init__.py
@@ -45,7 +45,7 @@ def expand_env_var(env_var: str) -> str:
 
 def clean_bentoml_version(bentoml_version: str) -> str:
     post_version = bentoml_version.split("+")[0]
-    match = re.match(r"^(\d+)\.(\d+)\.(\d+)(?:(a|rc)\d)*", post_version)
+    match = re.match(r"^(\d+)\.(\d+)\.(post)?(\d+)(?:(a|rc)\d)*", post_version)
     if match is None:
         raise BentoMLException("Errors while parsing BentoML version.")
     return match.group()
--- a/bentoml/_internal/configuration/containers.py
+++ b/bentoml/_internal/configuration/containers.py
@@ -116,6 +116,7 @@ SCHEMA = Schema(
         "tracing": {
             "type": Or(And(str, Use(str.lower), _check_tracing_type), None),
             "sample_rate": Or(And(float, lambda i: i >= 0 and i <= 1), None),
+            "excluded_urls": Or(str, None),
             Optional("zipkin"): {"url": Or(str, None)},
             Optional("jaeger"): {"address": Or(str, None), "port": Or(int, None)},
         },
@@ -407,6 +408,15 @@ class _BentoMLContainerClass:
             return provider
         else:
             return provider
+
+    @providers.SingletonFactory
+    @staticmethod
+    def tracing_excluded_urls(
+        excluded_urls: t.Optional[str] = Provide[config.tracing.excluded_urls],
+    ):
+        from opentelemetry.util.http import parse_excluded_urls
+
+        return parse_excluded_urls(excluded_urls)
 
     # Mapping from runner name to RunnerApp file descriptor
     remote_runner_mapping = providers.Static[t.Dict[str, str]]({})
--- a/bentoml/_internal/configuration/default_configuration.yaml
+++ b/bentoml/_internal/configuration/default_configuration.yaml
@@ -41,6 +41,7 @@ runners:
 tracing:
   type: zipkin
   sample_rate: 0.0
+  excluded_urls: Null
   zipkin:
     url: Null
   jaeger:
--- a/bentoml/_internal/server/runner_app.py
+++ b/bentoml/_internal/server/runner_app.py
@@ -134,7 +134,7 @@ class RunnerAppFactory(BaseAppFactory):
         middlewares.append(
             Middleware(
                 otel_asgi.OpenTelemetryMiddleware,
-                excluded_urls=None,
+                excluded_urls=BentoMLContainer.tracing_excluded_urls.get(),
                 default_span_details=None,
                 server_request_hook=None,
                 client_request_hook=client_request_hook,
--- a/bentoml/_internal/server/service_app.py
+++ b/bentoml/_internal/server/service_app.py
@@ -218,7 +218,7 @@ class ServiceAppFactory(BaseAppFactory):
         middlewares.append(
             Middleware(
                 otel_asgi.OpenTelemetryMiddleware,
-                excluded_urls=None,
+                excluded_urls=BentoMLContainer.tracing_excluded_urls.get(),
                 default_span_details=None,
                 server_request_hook=None,
                 client_request_hook=client_request_hook,


bentoml/_internal/configuration/__init__.py
```diff
@@ -45,7 +45,7 @@ def expand_env_var(env_var: str) -> str:
 
 def clean_bentoml_version(bentoml_version: str) -> str:
     post_version = bentoml_version.split("+")[0]
-    match = re.match(r"^(\d+)\.(\d+)\.(\d+)(?:(a|rc)\d)*", post_version)
+    match = re.match(r"^(\d+)\.(\d+)\.(post)?(\d+)(?:(a|rc)\d)*", post_version)
```
aarnphm, line 48: What is the reason behind this?
You might mean 
```
r"^(\d+)\.(\d+)\.(\d+)(?:(a|rc|post)\d)*
```
BentoML dirty version can't be `1.0.post3`, should always be `1.0.0post3`?

bentoml/_internal/configuration/containers.py
```diff
@@ -408,6 +409,15 @@ def tracer_provider(
...
             return provider
 
+    @providers.SingletonFactory
+    @staticmethod
+    def tracing_excluded_urls(
+        excluded_urls: t.Optional[str] = Provide[config.tracing.excluded_urls],
+    ):
+        from opentelemetry.util.http import parse_excluded_urls
+
+        return parse_excluded_urls(excluded_urls)
```
aarnphm, line 419: Does this function accept `None` type?

bentoml/_internal/configuration/default_configuration.yaml
```diff
@@ -41,6 +41,7 @@ runners:
 tracing:
   type: zipkin
   sample_rate: 0.0
+  excluded_urls: Null
```
aarnphm, line 44: I wonder whether there are others field we also want to exclude?

But I think it is good for 1st pass.

aarnphm (commented):
Got a few question and one kind request.

Is it possible that we can also have a unit test or setup some sort of test for this?

bentoml/_internal/configuration/__init__.py
jiewpeng: When I forked the repo and pulled my fork and installed it using `pip install -e .`, it seems like the version is `0.0.post1856+g1185809.d20220730`. This caused issues when I tried to import bentoml (or do anything with bentoml really).

bentoml/_internal/configuration/containers.py
jiewpeng: Yes it does, based on what I understand from the [source code](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/fbac4245159e51877efa86581724995bd1415a18/util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py#L69), it should return empty list if you pass in `None`.

bentoml/_internal/configuration/default_configuration.yaml
jiewpeng: Sorry I'm not quite sure what you mean others field we want to exclude - can you explain more?

aarnphm: So we can make `excludes` as a dictionary. 
```
exclude:
  - urls: ...
  - .... <???>
```

I'm not to familiar with tracing, hence I'm wondering if there are any other options people also want to exclude in addition to urls.

bentoml/_internal/configuration/__init__.py
aarnphm: You probably also need to fetch the tags as well.

bentoml/_internal/configuration/containers.py
aarnphm: noted.

Just take a quick peek at the source code. Seems like 

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/fbac4245159e51877efa86581724995bd1415a18/util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py#L58 also exclude additional URLs for instrumentation.

We do use some instrumentation library, and do you think it is necessary that BentoML also filter that out?

jiewpeng: Based on what I understand, BentoML uses `opentelemetry.instrumentation.asgi`, which does not call `get_excluded_urls` ([source](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/fbac4245159e51877efa86581724995bd1415a18/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py)). This is unlike `opentelemetry.instrumentation.fastapi`, which does call `get_excluded_urls` ([source](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/fbac4245159e51877efa86581724995bd1415a18/instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py)), and hence the user can just inject the environment variable and it will work.

So unfortunately, there is no environment variable that a user can inject to make use of the builtin exclude urls logic in the instrumentation.

bentoml/_internal/configuration/default_configuration.yaml
jiewpeng: I am also not too familiar with tracing myself, but it looks like at least for now, the only thing you can exclude in opentelemetry instrumentation is urls...

bentoml/_internal/configuration/__init__.py
jiewpeng: reverted the change, as my bentoml version is as expected after fetching the tags (1.0.2.post2+g1185809f.d20220730)

bentoml/_internal/configuration/default_configuration.yaml
aarnphm: sg

bentoml/_internal/configuration/containers.py
aarnphm: I think configuration-driven solution from you is fantastic. I do think there are also value for injecting an environment variables.

I'm wondering since BentoML does allow users to mount FastAPI apps, should we also respect the environment variable set by `opentelemetry.instrumentation.fastapi` 🤔 ?

Nonetheless, I think for first pass this is more than enough.

aarnphm (approved):
A first pass from me.
I will also ask team members to take a look at this as well.

bentoml/_internal/configuration/containers.py
jiewpeng: Good point - but in the event that both the environment variable and the bentoml config provide excluded urls, how should we handle it? Do we:

- Merge the 2 lists, or
- Have the bentoml config override the environment variable, or
- Have the environment variable override the bentoml config?

aarnphm: I think the order will always be respect the environment variable first, then config.

sauyon (approved):
Thanks!

We should also support a YAML list, but that can be done separately.

Let us know if you want to do that, or one of us can take this over.

jiewpeng:
Yup I can work on supporting the YAML list; should be fairly straightforward

PR #2862, opened 2022-08-04
feat: explicit GPU runner mappings

### LINKED ISSUE #2770: feat: Runner to GPUs mapping
### Feature request

The default scheduling strategy implementation schedules the same number of runner (`nvidia.com/gpu` supported) instances as the number of available GPUs. If multiple types of runners are present in the service, they will each be scheduled with the same number of instances as the available GPUs. Ideally, we should enable device isolation and allow users to configure GPU device affinity to the runner types.

### Motivation

Allowing runner to GPUs mapping can reduce the number of runner instances required while maximizing GPU resource utilization.

### Other

_No response_

jjmachan:
## What does this PR address?

Fixes #2770 
With this change, the `NvidiaGpuResource` will accept and return a list that specifies the exact GPUs to use. This enables the user to configure exactly which GPU to map to each runner worker.

```yaml
runners:
  resources: 
    nvidia.com/gpu: 4 => still valid but will be automatically converted to use GPUs [0, 1, 2, 3] internally.
  resources: 
    nvidia.com/gpu: [2, 4] => only use GPU [2, 4] for runner worker 1 and 2
```

## Before submitting:

## Who can help review?


bentoml/_internal/resource.py
```diff
@@ -211,25 +214,29 @@ def query_os_cpu_count() -> int:
...
-class NvidiaGpuResource(Resource[int], resource_id="nvidia.com/gpu"):
+class NvidiaGpuResource(Resource[t.List[int]], resource_id="nvidia.com/gpu"):
     @classmethod
-    def from_spec(cls, spec: t.Any) -> int:
-        if not isinstance(spec, (int, str)):
-            raise TypeError("NVidia GPU resource limit must be int or str")
-
-        if isinstance(spec, int):
-            return spec
+    def from_spec(cls, spec: t.Union[int, str, t.List[int | str]]) -> t.List[int]:
```
aarnphm, line 219:
```suggestion
    def from_spec(cls, spec: int | str | list[int | str]) -> list[int]:
```

bentoml/_internal/resource.py
```diff
@@ -211,25 +214,29 @@ def query_os_cpu_count() -> int:
...
+        if not isinstance(spec, (int, str, t.List)):
```
aarnphm, line 220:
```suggestion
        if not isinstance(spec, (int, str, list)):
```

bentoml/_internal/resource.py
```diff
@@ -211,25 +214,29 @@ def query_os_cpu_count() -> int:
...
+                return [int(x) for x in spec]
         except ValueError:
             raise BentoMLConfigException(
                 f"Invalid NVidia GPU resource limit '{spec}'. "
             )
 
     @classmethod
     @functools.lru_cache(maxsize=1)
-    def from_system(cls) -> int:
+    def from_system(cls) -> t.List[int]:
```
aarnphm, line 239:
```suggestion
    def from_system(cls) -> list[int]:
```

jjmachan: thanks for the review @aarnphm 🙂! can you tell me the difference between the two types signatures here?

aarnphm: See [PEP604](https://peps.python.org/pep-0604/)

bentoml/_internal/resource.py
```diff
@@ -211,25 +214,29 @@ def query_os_cpu_count() -> int:
...
-    def from_spec(cls, spec: t.Any) -> int:
-        if not isinstance(spec, (int, str)):
-            raise TypeError("NVidia GPU resource limit must be int or str")
-
-        if isinstance(spec, int):
-            return spec
+    def from_spec(cls, spec: t.Union[int, str, t.List[int | str]]) -> t.List[int]:
+        if not isinstance(spec, (int, str, t.List)):
+            raise TypeError(
+                "NVidia GPU resource limit must be int, str or a list specifing the exact GPUs to use."
```
ssheng, line 222: maybe "GPU device IDs".

bentoml/_internal/resource.py
```diff
@@ -238,22 +245,24 @@ def from_system(cls) -> int:
         try:
             pynvml.nvmlInit()
             device_count = pynvml.nvmlDeviceGetCount()
-            return device_count
+            return list(range(device_count))
         except (pynvml.nvml.NVMLError, OSError):
-            return 0
+            return []
```
ssheng, line 250: Should we log a warning? @sauyon

bentoml/_internal/resource.py
```diff
@@ -238,22 +245,24 @@ def from_system(cls) -> int:
...
             except Exception:  # pylint: disable=broad-except
                 pass
 
     @classmethod
-    def validate(cls, val: int):
-        if val < -1:
-            raise BentoMLConfigException(f"Invalid negative GPU resource limit {val}.")
-        if val > cls.from_system():
+    def validate(cls, val: t.List[int]):
+        if any([gpu_index < -1 for gpu_index in val]):
```
ssheng, line 259: Is -1 a valid device ID?

jjmachan:
Let me know if the change to the resource config is okay and if it is, I'll update the configuration guide too, talking about how we can configure each runner specifically

bentoml/_internal/resource.py
jjmachan: from my understanding, the -1 flag is to disable access to GPUs altogether. @sauyon can you confirm if that is indeed correct?

bentoml/_internal/resource.py
jjmachan: yeah we probably should, will add that

bentoml/_internal/resource.py
jjmachan: ahh got it. thanks!

bentoml/_internal/resource.py
jjmachan: is this a bit too confusing?
if the user gives
`nvidia.com/gpu: 3`  then it is the number of resources
``nvidia.com/gpu: [3]` is the specific GPU number

jjmachan: I can add more info in the doc section and this type error can point to that section of our doc

bentoml/_internal/resource.py
sauyon: I don't think we should support it as an element of the list even if it is; we should just parse -1 into an empty list.

sauyon (approved):
I think this pretty much looks good to me minus the one comment.

bentoml/_internal/resource.py
jjmachan: got it. updated the logic and changed the tests accordingly

tests/unit/_internal/test_strategy.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def unvalidated_get_resource(x: t.Dict[str, t.Any], y: str):
+    return resource_get_resource(x, y, validate=False)
+
+
+def test_default_gpu_strategy(monkeypatch):
+    monkeypatch.setattr(strategy, "get_resource", unvalidated_get_resource)
+    DefaultStrategy.get_worker_count(GPURunnable, {"nvidia.com/gpu": 2}) == 2
+    DefaultStrategy.get_worker_count(GPURunnable, {"nvidia.com/gpu": 0}) == 0
```
sauyon, line 21: Huh, I'm surprised this works. In theory the scheduler should be finding no supported resources and falling back to using the CPU, right?

tests/unit/_internal/test_strategy.py
```diff
@@ -0,0 +1,29 @@
+import os
```
sauyon, line 1: I think this file should probably live in `_internal/runner/test_strategy.py`.

tests/unit/_internal/test_strategy.py
jjmachan: the scheduler does that if either "system" string is specified or if the field is ignored ie only uses GPUs if available and the runner supports it. 
If you manually request for the GPU it will try and use that. There is a validation step in this case so if the GPU resources specified are unavailable, it will raise an exception. I feel this is better since it fails loudly in the case where the user did explicitly set the resource.
In the test case, I've monkey-patched it not to run validation.

what do you think?

tests/unit/_internal/test_strategy.py
jjmachan: done!

tests/unit/_internal/test_strategy.py
sauyon: I'd think that if the user specified 0 their intention is to not use the resource; the strategy is supposed to check for that (https://github.com/bentoml/BentoML/blob/fdbdc140e3916bd91c6cc71cbb245a33e8105ea4/bentoml/_internal/runner/strategy.py#L66). But this is probably a strategy bug, so I think this PR is good to go.

jjmachan: ohh I see what you mean. Yeah, that is strange. I'll check this again

tests/unit/_internal/runner/test_strategy.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def unvalidated_get_resource(x: t.Dict[str, t.Any], y: str):
+    return resource_get_resource(x, y, validate=False)
+
+
+def test_default_gpu_strategy(monkeypatch):
+    monkeypatch.setattr(strategy, "get_resource", unvalidated_get_resource)
+    assert DefaultStrategy.get_worker_count(GPURunnable, {"nvidia.com/gpu": 2}) == 2
+    assert DefaultStrategy.get_worker_count(GPURunnable, {"nvidia.com/gpu": 0}) == 1
```
jjmachan, line 21: it was working as expected @sauyon . Now if resources are not specified it will default to a worker count of 1 and give a warning saying "no resource found, falling back to using a single worker" which was the logic earlier

tests/unit/_internal/test_configuration.py
```diff
@@ -29,13 +29,18 @@
 """
 
 
-def test_bentoml_configuration_runner_override():
+def get_BentoMLConfiguration_from_str(config_str: str):
```
aarnphm, line 32: nit style: keep it snake case

tests/unit/_internal/test_configuration.py
```diff
@@ -66,3 +71,22 @@ def test_bentoml_configuration_runner_override():
     assert test_runner_batching["batching"]["max_batch_size"] == 10
     assert test_runner_batching["logging"]["access"]["enabled"] is True
     assert test_runner_batching["resources"]["cpu"] == 4  # should use global
+
+
+def test_runner_gpu_configuration():
+    GPU_INDEX = """
```
aarnphm, line 77:
```suggestion
    GPU_INDEX = """\
```

tests/unit/_internal/runner/test_strategy.py
```diff
@@ -0,0 +1,31 @@
+import os
+import typing as t
+
+import bentoml
+from bentoml._internal.runner import strategy
+from bentoml._internal.resource import get_resource as resource_get_resource
```
aarnphm, line 6: why the alias? 😄

tests/unit/_internal/runner/test_strategy.py
```diff
@@ -0,0 +1,31 @@
+import os
```
aarnphm, line 1: also lets try to use `from __future__ import annotations`

tests/unit/_internal/runner/test_strategy.py
```diff
@@ -0,0 +1,31 @@
...
+
+class GPURunnable(bentoml.Runnable):
+    SUPPORTED_RESOURCES = ("nvidia.com/gpu",)
+
+
+def unvalidated_get_resource(x: t.Dict[str, t.Any], y: str):
+    return resource_get_resource(x, y, validate=False)
+
+
+def test_default_gpu_strategy(monkeypatch):
```
aarnphm, line 18: Types of monkeypatch is
```
if TYPE_CHECKING:
    from _pytest.monkeypatch import MonkeyPatch
```

aarnphm (commented):
some very very style details. 😄

tests/unit/_internal/runner/test_strategy.py
jjmachan: done!
1 Q: the if TYPE_CHECKING is a check to improve perf type checker right?

tests/unit/_internal/runner/test_strategy.py
jjmachan: oh there was no need for that

bojiang (approved):

tests/unit/_internal/runner/test_strategy.py
bojiang: Yeah it is the behavior caused by login in SchedulingStrategy. I wonder if it's correct but we may solve that in another PR

PR #2874, opened 2022-08-06
fix: `JSON(IODescriptor[JSONType]).to_http_response` returns empty body when the response is `None`.

sbouchardet:
## What does this PR address?
The default behavior of `json.dumps` is to transform the return `None` to a string `null`.
But let's say we want to set a status_code 204 (NO_CONTENT), in this case it is expected an empty body.
Then, even if the function returns `None`, an error like  `h11._util.LocalProtocolError: Too much data for declared Content-Length` is thrown because the response has a body (a string `null`).

I described this scenario in this [slack thread](https://bentoml.slack.com/archives/CKRANBHPH/p1659623872508249?thread_ts=1659476156.379999&cid=CKRANBHPH) too.

:bulb: Extra:
 -  I've noticed that the tests `test_json_encoder` had an `or` in the assert expression. To make it more assertive, I refactored it to pass the expected result as a parameter. I'm use to doing this in this kind of test, to keep the assertiveness (but feel free to undo this change if it doesn't comply with the project pattern).
 - To test the async code, I added a `loop` fixture in the `tests/unit/_internal/io/test_json.py`. But if you guys plan to increase the coverage of async code, this fixture could be in another point. Actually, I did it because I didn't find another example of how is async code tested in the project. If there is an example, please let me know to improve this test.
 
 :thinking:  This PR is updating only the JSON(IODescriptor) behavior. If it makes sense to handle the `None` returned this way, it could be replicated in other `IODescriptor` where the `json.dumps` is called 

## Before submitting:

## Who can help review?

- @yubozhao 
- @ssheng

### DIFF
--- a/bentoml/_internal/io_descriptors/json.py
+++ b/bentoml/_internal/io_descriptors/json.py
@@ -197,13 +197,17 @@ class JSON(IODescriptor[JSONType]):
         if isinstance(obj, pydantic.BaseModel):
             obj = obj.dict()
 
-        json_str = json.dumps(
-            obj,
-            cls=self._json_encoder,
-            ensure_ascii=False,
-            allow_nan=False,
-            indent=None,
-            separators=(",", ":"),
+        json_str = (
+            json.dumps(
+                obj,
+                cls=self._json_encoder,
+                ensure_ascii=False,
+                allow_nan=False,
+                indent=None,
+                separators=(",", ":"),
+            )
+            if obj is not None
+            else None
         )
 
         if ctx is not None:
--- a/tests/unit/_internal/io/test_json.py
+++ b/tests/unit/_internal/io/test_json.py
@@ -1,12 +1,13 @@
 import json
 import typing as t
-from dataclasses import dataclass
 import asyncio
+from dataclasses import dataclass
 
 import numpy as np
-from bentoml._internal.io_descriptors.json import JSON
 import pytest
 import pydantic
+
+from bentoml._internal.io_descriptors.json import JSON
 
 
 @pytest.fixture
--- a/tests/unit/_internal/io/test_json.py
+++ b/tests/unit/_internal/io/test_json.py
@@ -1,10 +1,19 @@
 import json
 import typing as t
 from dataclasses import dataclass
+import asyncio
 
 import numpy as np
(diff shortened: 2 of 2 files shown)


bentoml/_internal/io_descriptors/json.py
```diff
@@ -197,13 +197,17 @@ async def to_http_response(
...
+        json_str = (
+            json.dumps(
+                obj,
+                cls=self._json_encoder,
+                ensure_ascii=False,
+                allow_nan=False,
+                indent=None,
+                separators=(",", ":"),
+            )
+            if obj is not None
```
aarnphm, line 209:
```suggestion
            if obj
```
There could be situation when obj is an empty string, empty list, tuple

tests/unit/_internal/io/test_json.py
```diff
@@ -1,11 +1,21 @@
 import json
 import typing as t
+import asyncio
 from dataclasses import dataclass
 
 import numpy as np
 import pytest
 import pydantic
 
+from bentoml._internal.io_descriptors.json import JSON
```
aarnphm, line 10:
```suggestion
from bentoml.io import JSON
```

tests/unit/_internal/io/test_json.py
```diff
@@ -1,11 +1,21 @@
...
+
+
+@pytest.fixture
+def loop():
```
aarnphm, line 14: Since we are only using this in one function, i think a try..finally loop is sufficient enough. No need for fixture here😊

aarnphm (commented):
LGTM with some minor comments. Thanks for the fix 👍

bentoml/_internal/io_descriptors/json.py
ssheng: Empty collections are probably OK serializing them the way it is right now. However, `None` is currently not an accepted type of `to_http_response`. I will do another PR on top of this one to add None as an accepted type.

tests/unit/_internal/io/test_json.py
ssheng: I will address in a separate PR.

ssheng (approved):

PR #2884, opened 2022-08-09
feat(grpc): IO descriptors

aarnphm:
adding gRPC support for the remaining of the IO descriptor
Pandas will be handled by @Proto007

Test will be with a follow up PR.

This PR also remove `setup.cfg` and use `pyproject.toml`. We will also lock version of setuptools==63.4.0 until 64+ is stable.

since `bentoml/_version.py` is always generated and included inside the distribution, we should use that for `__version__` instead of using `importlib.metadata`. This will also avoid `PackageNotFoundError` if bentoml is not yet installed, but attributes for version were accessed before.


bentoml/_internal/io_descriptors/numpy.py
```diff
@@ -199,18 +159,25 @@ def __init__(
...
         if bytesorder and bytesorder not in ["C", "F", "A"]:
             raise BadInput(
                 f"'bytesorder' must be one of ['C', 'F', 'A'], got {bytesorder} instead."
             )
         if not bytesorder:
-            bytesorder = "C"  # default from numpy (C-order)
             # https://numpy.org/doc/stable/user/basics.byteswapping.html#introduction-to-byte-ordering-and-ndarrays
+            bytesorder = "C"  # default from numpy (C-order)
+
         self._bytesorder: t.Literal["C", "F", "A"] = bytesorder
```
parano, line 170: remove _bytesorder?

bentoml/_internal/io_descriptors/numpy.py
```diff
@@ -546,15 +359,176 @@ async def predict(input: NDArray[np.int16]) -> NDArray[Any]:
...
+            logger.debug(
+                f"'enforce_shape={self._enforce_shape}', using '{self._shape}'..."
+            )
+            shape = self._shape
+        else:
+            shapepb = field.shape  # this can be an empty list ([])
+            if shapepb:
+                shape = tuple(shapepb)
+                if self._shape:
+                    logger.warning(
```
parano, line 420: When force_shape is True, we should always use the shape specified when creating the Numpy io descriptor, not using the shape in request. Let's just raise an exception if the specified shape is different from the shape in the request if enforce shape is True. Make sure to handle the `-1` case.

bentoml/_internal/io_descriptors/numpy.py
```diff
@@ -546,15 +359,176 @@ async def predict(input: NDArray[np.int16]) -> NDArray[Any]:
...
+
+        # The same logic for 'shape' applies to 'dtype' with the additional:
+        #
+        # 2a. If 'dtype' message is not specified, then we will infer given numpy dtype from the entry.
+        if self._enforce_dtype:
+            logger.debug(
+                f"'enforce_dtype={self._enforce_dtype}', using '{self._dtype}'..."
+            )
+            dtype = self._dtype
+            field_values = npdtype_to_fieldpb_map()[dtype]
```
parano, line 434: This is not the right behavior. let's just convert the numpy array from proto first, and then check with np.can_cast. reuse the validation function above.

bentoml/_internal/io_descriptors/numpy.py
```diff
@@ -546,15 +359,176 @@ async def predict(input: NDArray[np.int16]) -> NDArray[Any]:
...
+
+        if shape and dtype in [np.float32, np.double, np.bool_]:
+            buffer = field.SerializeToString()
+            num_entries = np.prod(shape)
+            return np.frombuffer(
+                memoryview(buffer[-(dtype.itemsize * num_entries) :]),
+                dtype=dtype,
+                count=num_entries,
+                offset=0,
+            ).reshape(shape)
```
parano, line 482: add self.validate_array here

bentoml/_internal/io_descriptors/text.py
```diff
@@ -139,16 +138,29 @@ async def to_http_response(self, obj: str, ctx: Context | None = None) -> Respon
...
-        return str(request.input.string_value)
+        import ast
+
+        from ..utils.grpc import get_field
+        from ..utils.grpc import validate_content_type
+
+        # validate gRPC content type if content type is specified
+        validate_content_type(context, self)
+
+        field = ast.literal_eval(get_field(request, self))
```
parano, line 151: why do we need this?

bentoml/_internal/io_descriptors/numpy.py
aarnphm: _bytesorder can be used for np.frombuffer.

bentoml/_internal/io_descriptors/numpy.py
aarnphm: I think then we don't need to `reshape` here.

bentoml/_internal/io_descriptors/numpy.py
aarnphm: This is checked on L410 right?

bentoml/_internal/io_descriptors/text.py
aarnphm: You can an put a bytes string.

aarnphm: We should also allow a bytes field in all of the message.

bentoml/_internal/io_descriptors/numpy.py
aarnphm: No need to right? we already cast the dtype and reshape it with `shape`

PR #2886, opened 2022-08-09
feat: Add SSL support for http api servers via bentoml serve

sptowey:
## Before submitting:

- [ ] Did you write tests to cover your changes? N/A - cleared with Sean in Slack

## Who can help review?
@ssheng
@sauyon

### DIFF
--- a/bentoml/_internal/configuration/containers.py
+++ b/bentoml/_internal/configuration/containers.py
@@ -89,6 +89,15 @@ SCHEMA = Schema(
             "workers": Or(And(int, _larger_than_zero), None),
             "timeout": And(int, _larger_than_zero),
             "max_request_size": And(int, _larger_than_zero),
+            Optional("ssl"): {
+                Optional("keyfile"): Or(str, None),
+                Optional("certfile"): Or(str, None),
+                Optional("keyfile_password"): Or(str, None),
+                Optional("version"): Or(And(int, _larger_than_zero), None),
+                Optional("cert_reqs"): Or(int, None),
+                Optional("ca_certs"): Or(str, None),
+                Optional("ciphers"): Or(str, None),
+            },
             "metrics": {"enabled": bool, "namespace": str},
             "logging": {
                 # TODO add logging level configuration
--- a/bentoml/serve.py
+++ b/bentoml/serve.py
@@ -84,13 +84,27 @@ def serve_development(
     backlog: int = Provide[BentoMLContainer.api_server_config.backlog],
     bentoml_home: str = Provide[BentoMLContainer.bentoml_home],
     reload: bool = False,
-    ssl_keyfile: t.Optional[str] = None,
-    ssl_certfile: t.Optional[str] = None,
-    ssl_keyfile_password: t.Optional[str] = None,
-    ssl_version: t.Optional[int] = None,
-    ssl_cert_reqs: t.Optional[int] = None,
-    ssl_ca_certs: t.Optional[str] = None,
-    ssl_ciphers: t.Optional[str] = None,
+    ssl_keyfile: t.Optional[str] = Provide[
+        BentoMLContainer.api_server_config.ssl.keyfile
+    ],
+    ssl_certfile: t.Optional[str] = Provide[
+        BentoMLContainer.api_server_config.ssl.certfile
+    ],
+    ssl_keyfile_password: t.Optional[str] = Provide[
+        BentoMLContainer.api_server_config.ssl.keyfile_password
+    ],
+    ssl_version: t.Optional[int] = Provide[
+        BentoMLContainer.api_server_config.ssl.version
+    ],
+    ssl_cert_reqs: t.Optional[int] = Provide[
+        BentoMLContainer.api_server_config.ssl.cert_reqs
+    ],
+    ssl_ca_certs: t.Optional[str] = Provide[
+        BentoMLContainer.api_server_config.ssl.ca_certs
+    ],
+    ssl_ciphers: t.Optional[str] = Provide[
+        BentoMLContainer.api_server_config.ssl.ciphers
+    ],
 ) -> None:
     working_dir = os.path.realpath(os.path.expanduser(working_dir))
     svc = load(bento_identifier, working_dir=working_dir)  # verify service loading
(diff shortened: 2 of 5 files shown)


bentoml/_internal/configuration/containers.py
```diff
@@ -89,6 +89,15 @@ def _is_ip_address(addr: str) -> bool:
...
             "max_request_size": And(int, _larger_than_zero),
+            Optional("ssl"): {
+                Optional("keyfile"): Or(str, None),
+                Optional("certfile"): Or(str, None),
+                Optional("keyfile_password"): Or(str, None),
+                Optional("version"): Or(And(int, _larger_than_zero), None),
+                Optional("cert_reqs"): Or(int, None),
+                Optional("ca_certs"): Or(str, None),
+                Optional("ciphers"): Or(str, None),
+            },
```
aarnphm, line 100: I think this is great. You can also add these options to `BentoMLContainerClass` below.

bentoml/serve.py
```diff
@@ -84,6 +84,13 @@ def serve_development(
     backlog: int = Provide[BentoMLContainer.api_server_config.backlog],
     bentoml_home: str = Provide[BentoMLContainer.bentoml_home],
     reload: bool = False,
+    ssl_keyfile: t.Optional[str] = None,
+    ssl_certfile: t.Optional[str] = None,
+    ssl_keyfile_password: t.Optional[str] = None,
+    ssl_version: t.Optional[int] = None,
+    ssl_cert_reqs: t.Optional[int] = None,
+    ssl_ca_certs: t.Optional[str] = None,
+    ssl_ciphers: t.Optional[str] = None,
```
aarnphm, line 93: Should we use dependency injection here?

aarnphm (commented):
Thanks for the contribution. Make sure to run `make format` and `make lint` locally 🎉

bentoml/_internal/configuration/containers.py
sauyon: There's no need to, I think?

parano: there's no need to add these to BentoMLContainerClass

bentoml_cli/serve.py
```diff
@@ -69,6 +69,48 @@ def add_serve_command(cli: click.Group) -> None:
         default=".",
         show_default=True,
     )
+    @click.option(
+        "--ssl-keyfile",
+        type=click.STRING,
+        help="SSL key file",
+        default=None,
```
parano, line 76: Should we set the default value from config here? e.g.

```
        default=BentoMLContainer.api_server_config.ssl.keyfile.get(),
```

sauyon: I think we should use dependency injection like Aaron said above, we probably should try to start avoiding depending on BentoML before the click initialization so that click completion isn't horribly slow.

aarnphm: After this PR got merge, I think we can do a quick cleanup on CLI and set default value to None, and then we can access the value via DI lazily.

sptowey:
> Thanks for the contribution. Make sure to run `make format` and `make lint` locally 🎉

Whoops, I ran the linter but missed the formatter!

bentoml_cli/serve.py
sptowey: Let me know if you want me to help out with this. Sounds like you all may have it under control though?

ssheng: Since there is no default value for the SSL related configurations, doing dependency injection here is not very meaningful. I think we can leave it out.

aarnphm: It is more about user providing this option via config right?

ssheng: Correct. What I meant was that we should do config injection at the places you suggested in `serve_development` and `serve_production` instead of CLI.

bentoml/serve.py
ssheng: +1. Let's inject config here.
```
ssl_keyfile: t.Optional[str] = Provide[BentoMLContainer.api_server_config.ssl.keyfile]
```

bentoml/serve.py
```diff
@@ -178,6 +209,13 @@ def serve_production(
     host: str = Provide[BentoMLContainer.api_server_config.host],
     backlog: int = Provide[BentoMLContainer.api_server_config.backlog],
     api_workers: t.Optional[int] = None,
+    ssl_keyfile: t.Optional[str] = None,
+    ssl_certfile: t.Optional[str] = None,
+    ssl_keyfile_password: t.Optional[str] = None,
+    ssl_version: t.Optional[int] = None,
+    ssl_cert_reqs: t.Optional[int] = None,
+    ssl_ca_certs: t.Optional[str] = None,
+    ssl_ciphers: t.Optional[str] = None,
```
ssheng, line 218: And inject config here.
```
ssl_keyfile: t.Optional[str] = Provide[BentoMLContainer.api_server_config.ssl.keyfile],
```

bentoml/serve.py
sptowey: Aha, I understand now. I'd considered doing it that was to start, but wasn't entirely clear on if `Provide` would default to `None` as expected.

bentoml/serve.py
```diff
@@ -285,27 +351,52 @@ def serve_production(
...
+    ) if ssl_version is not None else None  # pylint: disable=W0106
+    api_server_watcher_args.extend(
+        ["--ssl-cert-reqs", str(ssl_cert_reqs)]
+    ) if ssl_cert_reqs is not None else None  # pylint: disable=W0106
+    api_server_watcher_args.extend(
+        ["--ssl-ca-certs", ssl_ca_certs]
+    ) if ssl_ca_certs is not None else None  # pylint: disable=W0106
+    api_server_watcher_args.extend(
+        ["--ssl-ciphers", ssl_ciphers]
+    ) if ssl_ciphers is not None else None  # pylint: disable=W0106
```
aarnphm, line 393: https://github.com/encode/uvicorn/blob/b21ecabc5bf911f571e0629438315a1e5472065c/uvicorn/config.py

Uvicorn does handle None option for some option. 

IMO, a nicer way to write None check is
```
if ssl_certfile:
   args.extend(["--ssl-certfile", ssl_certfile])
if ssl_cert_reqs:
   args.extend(...)
if ssl_version:
   ...
```

Then we can pass all of the other options straight through

aarnphm (approved):
Thanks again for the contribution. I can update the style afterward.

bentoml/serve.py
sptowey: Yes, this is a weird bit of code. It doesn't (shouldn't?) actually pass `None` into uvicorn - [here](https://github.com/bentoml/BentoML/pull/2886/commits/d23713ac8cbed5320d38cc00107de0fced680145#diff-4f4b8a5d43ba42f4f34b447f11659cf2b6f322abe52efb26d5bc4cbcb89fb90cR117) is where uvicorn actually gets the args if they exist. 
I wrote it this way to try compacting the code a bit, but running it through the formatter made that a moot point. I agree that it should be reformatted to the way you proposed - much easier to understand.

PR #2890, opened 2022-08-10
feat: add timeout option for bentoml runner config

### LINKED ISSUE #2888: Add configuration option for runner timeout
Long model inference request could timeout the current runner request. Runner timeout is currently hardcoded at 5 minutes. We should make the runner timeout value configurable. SageMaker deployment through bentoctl can generate the configuration with an updated timeout value during build.

jjmachan:
## What does this PR address?

Fixes #2888 
eg
```yaml
runners:
  timeout: 400 #seconds (global)
  pytorch_runner:
    timeout: 500 # individual timeouts for specific runners
```
by default all runner have a timeout of 300 seconds 

## Before submitting:

## Who can help review?

### DIFF
--- a/bentoml/_internal/runner/runner_handle/remote.py
+++ b/bentoml/_internal/runner/runner_handle/remote.py
@@ -81,7 +81,6 @@ class RemoteRunnerClient(RunnerHandle):
     @property
     def _client(
         self,
-        timeout_sec: float | None = None,
     ) -> ClientSession:
         import aiohttp
 
@@ -100,11 +99,9 @@ class RemoteRunnerClient(RunnerHandle):
                 return str(url.with_query(None))
 
             jar = aiohttp.DummyCookieJar()
-            if timeout_sec is not None:
-                timeout = aiohttp.ClientTimeout(total=timeout_sec)
-            else:
-                DEFAULT_TIMEOUT = aiohttp.ClientTimeout(total=5 * 60)
-                timeout = DEFAULT_TIMEOUT
+            # uses only global timeout and not runner specific timeouts
+            timeout_from_config = BentoMLContainer.config.runners.timeout.get()
+            timeout = aiohttp.ClientTimeout(total=timeout_from_config)
             self._client_cache = aiohttp.ClientSession(
                 trace_configs=[
                     create_trace_config(
--- a/bentoml/_internal/runner/runner_handle/remote.py
+++ b/bentoml/_internal/runner/runner_handle/remote.py
@@ -41,6 +41,15 @@ class RemoteRunnerClient(RunnerHandle):
     def _remote_runner_server_map(self) -> dict[str, str]:
         return BentoMLContainer.remote_runner_mapping.get()
 
+    @property
+    def runner_timeout(self) -> int:
+        "return the configured timeout for this runner."
+        runner_cfg = BentoMLContainer.runners_config.get()
+        if self._runner.name in runner_cfg:
+            return runner_cfg[self._runner.name]["timeout"]
+        else:
+            return runner_cfg["timeout"]
+
     def _close_conn(self) -> None:
         if self._conn:
             self._conn.close()
@@ -99,9 +108,7 @@ class RemoteRunnerClient(RunnerHandle):
                 return str(url.with_query(None))
 
             jar = aiohttp.DummyCookieJar()
-            # uses only global timeout and not runner specific timeouts
-            timeout_from_config = BentoMLContainer.config.runners.timeout.get()
-            timeout = aiohttp.ClientTimeout(total=timeout_from_config)
+            timeout = aiohttp.ClientTimeout(total=self.runner_timeout)
             self._client_cache = aiohttp.ClientSession(
                 trace_configs=[
                     create_trace_config(
--- a/tests/unit/_internal/test_configuration.py
+++ b/tests/unit/_internal/test_configuration.py
@@ -90,3 +90,26 @@ runners:
     bentoml_cfg = get_bentomlconfiguration_from_str(GPU_INDEX_WITH_STRING)
     # this behaviour can be confusing
     assert bentoml_cfg["runners"]["resources"] == {"nvidia.com/gpu": "[1, 2, 4]"}
+
+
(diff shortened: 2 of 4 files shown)


tests/unit/_internal/test_configuration.py
```diff
@@ -66,3 +66,26 @@ def test_bentoml_configuration_runner_override():
...
+    test_runner_2:
+        resources: system
+"""
+
+
+def test_runner_timeouts():
+    tmpfile = NamedTemporaryFile(mode="w+", delete=False)
+    tmpfile.write(RUNNER_TIMEOUTS)
+    tmpfile.flush()
+    tmpfile.close()
```
aarnphm, line 85: hmm should this be a context manager instead?

bentoml/_internal/runner/runner_handle/remote.py
```diff
@@ -100,11 +99,9 @@ def strip_query_params(url: yarl.URL) -> str:
...
 
             jar = aiohttp.DummyCookieJar()
-            if timeout_sec is not None:
-                timeout = aiohttp.ClientTimeout(total=timeout_sec)
-            else:
-                DEFAULT_TIMEOUT = aiohttp.ClientTimeout(total=5 * 60)
-                timeout = DEFAULT_TIMEOUT
+            # uses only global timeout and not runner specific timeouts
+            timeout_from_config = BentoMLContainer.config.runners.timeout.get()
+            timeout = aiohttp.ClientTimeout(total=timeout_from_config)
```
aarnphm, line 104: lets use `@inject` and `Provide` here.

aarnphm (commented):
Quick q: have you run this test on lower version of bentoml for compat layer?

tests/unit/_internal/test_configuration.py
jjmachan: maybe, I'm not sure. The only issue was I wanted to manually close() the `tmpfile` because in windows this will fail to open the file created by NamedTemporary file if it is still open with some other process which is a strange windows issue so in order to avoid that I'm manually handling it.

bentoml/_internal/runner/runner_handle/remote.py
jjmachan: @bojiang and I were discussing this and he suggested we might be better off not using provide here since this section is hot path and we can save a few lines since this is the only place where we need it the config.

jjmachan:
I have not actually but can you tell me more about this compat layer?

aarnphm:
> I have not actually but can you tell me more about this compat layer?

Oh just that running this on an older version of bentoml.

tests/unit/_internal/test_configuration.py
bojiang: It seems that we are not deleting the temp file after the test

aarnphm: thats the case right now. But opening it inside a ctx manager would ensure correct gc.

Another thing you can do is to use `unittest.TestCase`

aarnphm: But I think `tearDown` just gc everything, so it doesn't make a lot of difference.

jjmachan: yeah, we are not deleting it because of the same issue on windows. Originally once NamedTemporaryFile is closed it will automatically get deleted but we can't do that because In windows the tempfile has to be closed before BentoMLConfiguration can read from it.

but I thought it's okay since it's only in temp and will get cleanup anyways?

bojiang: According to the doc of tempfile, we need to take the responsibility to clean the temp files.

bentoml/_internal/runner/runner_handle/remote.py
```diff
@@ -100,11 +99,9 @@ def strip_query_params(url: yarl.URL) -> str:
                 return str(url.with_query(None))
 
             jar = aiohttp.DummyCookieJar()
-            if timeout_sec is not None:
-                timeout = aiohttp.ClientTimeout(total=timeout_sec)
-            else:
-                DEFAULT_TIMEOUT = aiohttp.ClientTimeout(total=5 * 60)
-                timeout = DEFAULT_TIMEOUT
+            # uses only global timeout and not runner specific timeouts
+            timeout_from_config = BentoMLContainer.config.runners.timeout.get()
```
ssheng, line 103: Should we be using the per-runner timeout here?

jjmachan: Initially decided we might skip this so that we can get the PR merged faster and take per-runner timeout later but I can look into this now. should not be that different.

bojiang (changes requested):
per-runner timeout

bentoml/_internal/runner/runner_handle/remote.py
jjmachan: udpated

tests/unit/_internal/test_configuration.py
jjmachan: I think it's fine since this is only for tests and is created in its own temp directory in pytest

bentoml/_internal/runner/runner_handle/remote.py
```diff
@@ -41,6 +41,15 @@ def __init__(self, runner: Runner):  # pylint: disable=super-init-not-called
...
 
+    @property
+    def runner_timeout(self) -> int:
+        "return the configured timeout for this runner."
+        runner_cfg = BentoMLContainer.runners_config.get()
+        if self._runner.name in runner_cfg:
+            return runner_cfg[self._runner.name]["timeout"]
+        else:
+            return runner_cfg["timeout"]
+
```
jjmachan, line 52: @bojiang This is how we get the timeout for each runner client

bojiang: Is there a case that the `runner_name` exists in the config but didn't provide a `timeout` ?

jjmachan: nope, the BentoML config will copy over the default config to the `runner_name` so that situation will not happen.

bojiang (approved):

Issue #2787, opened 2022-07-22
chore: Set `OMP_NUM_THREADS` and related env vars prior to importing `numpy`


ssheng:
`OMP_NUM_THREADS` must be set before numpy is imported for it to work, our current implementation doesn’t guarantee that.

PR #2893, opened 2022-08-11
fix(scheduling): numpy worker environs are not taking effect

bojiang:
solution:
* only allow scheduling_strategy to control the env variables
* setting up environ before importing bentoml

WARNING:
a breaking change for a not documented public API `custom scheduler`

fix: https://github.com/bentoml/BentoML/issues/2787


bentoml_cli/server/runner.py
```diff
@@ -51,20 +58,10 @@ def main(
             - file:///path/to/unix.sock
             - fd://12
         working_dir: (Optional) the working directory
-        worker_id: (Optional) if set, the runner will be started as a worker with the given ID
+        worker_id: (Optional) if set, the runner will be started as a worker with the given ID. Important: begin from 1.
```
parano, line 61: add arg doc for worker_env_map

bentoml_cli/server/runner.py
```diff
@@ -29,6 +28,13 @@
     default=None,
     help="If set, start the server as a bare worker with the given worker ID. Otherwise start a standalone server with a supervisor process.",
 )
+@click.option(
+    "--worker-env-map",
+    required=False,
+    type=click.STRING,
+    default=None,
+    help="The environment variables to pass to the worker process. The format is a JSON string, e.g. '{0: {\"CUDA_VISIBLE_DEVICES\": 0}}'.",
```
parano, line 36: should we use a dotenv file instead of JSON?  that seems easier for debugging purpose

bentoml/_internal/runner/runner.py
```diff
@@ -247,9 +247,13 @@ def scheduled_worker_count(self) -> int:
...
         )
 
-    def setup_worker(self, worker_id: int) -> None:
-        self.scheduling_strategy.setup_worker(
-            self.runnable_class,
-            self.resource_config,
-            worker_id,
-        )
+    @property
+    def scheduled_worker_env_map(self) -> dict[int, dict[str, t.Any]]:
```
parano, line 251: does Yatai need this information for scheduling runners?

bentoml_cli/server/runner.py
bojiang: The env map includes all envvars for each worker.
It seems not that easy to be represented with dotenv files.

bentoml/_internal/runner/runner.py
bojiang: the worker concept is transparent for yatai

parano: @bojiang got it, in the case of Yatai, it will just use resources available from system.

bentoml_cli/server/runner.py
ssheng: Agree that passing JSON in CLI isn't the most intuitive. Using an `.env` file, we can allow multiple arguments of key-value pairs.
```
--worker-env 0:worker_0.env
```

bojiang: @ssheng I think that's too complicated. This is not the public API.
The public API is the `bentoml serve` and `bentoml.serve`

bentoml/_internal/runner/runner.py
bojiang: yeah

ssheng (approved):

PR #2897, opened 2022-08-12
chore(container): add entrypoints for yatai

bojiang:


bentoml/_internal/bento/docker/entrypoint.sh
```diff
@@ -11,8 +11,16 @@ _is_sourced() {
 
 _main() {
 	# if first arg looks like a flag
-	if [ "${1:0:1}" = '-' ]; then
-		set -- bentoml serve --production "$@" "$BENTO_PATH"
+	if [ -z "$1" ] || [ "${1:0:1}" = '-' ]; then
```
parano, line 14: could we do this at the CLI level instead?

e.g.
```
bentoml serve BENTO --runner=MY_RUNNER
```

bojiang: It could be another PR? This PR is mainly for yatai deployment.

bentoml/_internal/bento/docker/entrypoint.sh
```diff
@@ -11,8 +11,16 @@ _is_sourced() {
...
+	if [ -z "$@" ] || [ "${1:0:1}" = '-' ]; then
+		if [[ -v YATAI_VERSION ]]; then
+			if [ "$BENTOML_SERVER_COMPONENT" == "RUNNER" ]; then
+				set -- python -m bentoml_cli.server.runner "$@" "$BENTO_PATH"
+			else
+				set -- python -m bentoml_cli.server.http_api_server "$@" "$BENTO_PATH"
+			fi
+		else
+			set -- bentoml serve --production "$@" "$BENTO_PATH"
+		fi
```
aarnphm, line 23: imo we should do generate the entrypoint depending on `build` command

bentoml/_internal/bento/docker/entrypoint.sh
parano: @bojiang I think we should just do that now for Yatai deployment, this change still requires Yatai to set an env var right? we can achieve the same via CLI arg, but a lot cleaner

bojiang:
should be included by #2920

Issue #1499, opened 2021-03-11
Add more debug/info logs runner indicating adaptive batching logic


parano:
**Is your feature request related to a problem? Please describe.**

Currently there is not too much visibility for the user to visualize or to understand how BentoML's micro-batching works. having more logs around how BentoML makes a decision and what is happening in the batching layer should help with that.

cc @bojiang 

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

parano:
In debug mode, RunnerApp should:

* log the batch size for each run
* log the latency for each run

PR #2902, opened 2022-08-13
feat: Adaptive batching size histogram metrics 

ssheng:
## What does this PR address?

This PR introduces three changes. Fixes [#1499](https://github.com/bentoml/BentoML/issues/1499).

1. Adds a histogram metric for adaptive batching size. The histogram used is an exponential histogram from 1 to `max_batch_size`. The assumption made here is that the adaptive batching size histogram buckets are filled more evenly with exponential bounds.

Screen Shot 2022-08-12 at 22 49 57

2. In addition, this PR also adds a configuration for generating request duration histogram buckets. Currently, request duration buckets use the [default buckets](https://github.com/prometheus/client_python/blob/master/prometheus_client/metrics.py#L544) from prometheus client. With the max bucket size at 10, the histogram is unable to track latency greater 10 seconds. This change adds three keys for configuring the duration histogram.
    - min: the smallest upper bound
    - max: the largest upper bound
    - factor: the exponential factor used to generate the buckets

    For example, with `min=0.005`, `max=10.0`, and `factor=2`, buckets generated is `(0.005, 0.01, 0.02, 0.04, 0.08, 0.16, 
    0.32, 0.64, 1.28, 2.56, 5.12, 10.0, INF)`. We also assume request duration follows a long tail pattern.

3. Lastly, this PR fixes the access log to output request duration in the correct time unit. Previously, the access outputs latency in milliseconds but the value is tracked in seconds. With this PR, the value is displayed accurately.

    Before:
    ```
    2022-08-12T22:51:10-0700 [INFO] [api_server:5] 127.0.0.1:64649 (scheme=http,method=POST,path=/classify,type=application/json,length=132) (status=200,type=application/json,length=18) 0.003ms (trace=270554840034756658801751152705387155180,span=17828550978915614742,sampled=0)
    ```
    After:
    ```
    2022-08-12T22:51:10-0700 [INFO] [api_server:5] 127.0.0.1:64649 (scheme=http,method=POST,path=/classify,type=application/json,length=132) (status=200,type=application/json,length=18) 3.458ms (trace=270554840034756658801751152705387155180,span=17828550978915614742,sampled=0)
    ```

## Before submitting:

## Who can help review?


bentoml/_internal/configuration/containers.py
```diff
@@ -434,5 +442,32 @@ def tracing_excluded_urls(
     remote_runner_mapping = providers.Static[t.Dict[str, str]]({})
     plasma_db = providers.Static[t.Optional["ext.PlasmaClient"]](None)
 
+    @providers.SingletonFactory
+    @staticmethod
+    def duration_buckets(
+        metrics: t.Dict[str, t.Any] = Provide[config.api_server.metrics],
+    ) -> t.Tuple[float, ...]:
```
aarnphm, line 449:
```suggestion
        metrics: dict[str, t.Any] = Provide[config.api_server.metrics],
    ) -> tuple[float, ...]:
```

bentoml/_internal/configuration/containers.py
```diff
@@ -434,5 +442,32 @@ def tracing_excluded_urls(
...
+        the Prometheus default is returned; otherwise, a set of exponential buckets
+        generated based on the configuration is returned.
+        """
+        from ..utils.metrics import DEFAULT_BUCKET
+        from ..utils.metrics import exponential_buckets
+
+        if "duration" not in metrics:
+            return DEFAULT_BUCKET
+        else:
+            duration: t.Dict[str, float] = metrics["duration"]
```
aarnphm, line 461:
```suggestion
            duration: dict[str, float] = metrics["duration"]
```
if 'duration' keys does not exist, this will throw `KeyError`

tests/unit/_internal/utils/test_metrics.py
```diff
@@ -0,0 +1,184 @@
+import math
+
+from bentoml._internal.utils.metrics import INF
+from bentoml._internal.utils.metrics import linear_buckets
+from bentoml._internal.utils.metrics import exponential_buckets
+
+
+def _assert_equal(actual: tuple[float, ...], expected: tuple[float, ...]):
```
aarnphm: TODO(@aarnphm): Look into [Hypothesis](https://hypothesis.readthedocs.io/en/latest/)

aarnphm (commented):
mainly types change.

Note for @aarnphm: Update access log for bentoml/grpc once this is merged.

bentoml/_internal/configuration/containers.py
ssheng: What's our standards on using `__future__` annotations?

bentoml/_internal/configuration/containers.py
ssheng: Will the previous if statement check for it?

bentoml/_internal/configuration/containers.py
aarnphm: We should just use it from now on right?

bentoml/_internal/configuration/containers.py
aarnphm: you are correct.

aarnphm:
Screenshot 2022-08-13 at 00 22 19

One kind request: since you are also fixing access log, can you also fix the value for `type` and `length` when it is =0?

ssheng:
> Screenshot 2022-08-13 at 00 22 19
> 
> One kind request: since you are also fixing access log, can you also fix the value for `type` and `length` when it is =0?

What should be the ideal output here? To omit them in the access log or adding quotes around them?

aarnphm:
> What should be the ideal output here? To omit them in the access log or adding quotes around them?
Probably just display 0 right? to keep it consistent.

ssheng:
@aarnphm We don't display quotes in other non-empty fields though. Are you suggesting that we retroactively add the quotes?

ssheng:
> @aarnphm We don't display quotes in other non-empty fields though. Are you suggesting that we retroactively add the quotes?

@aarnphm Thought unlikely, technically updating the log format is a backward incompatible change, if anyone pipes the log and parse it in a query. Let's discuss this in a separate thread.

aarnphm (approved):
minor comments, otherwise LGTM

aarnphm (approved):
LGTM. Awesome feature

PR #2904, opened 2022-08-14
chore(templates): cleanup debian dependency logic

aarnphm:
Cleanup debian installation logics


bentoml/_internal/bento/docker/templates/base_debian.j2
```diff
@@ -7,17 +7,8 @@ USER root
 
 ENV DEBIAN_FRONTEND=noninteractive
 RUN rm -f /etc/apt/apt.conf.d/docker-clean; echo 'Binary::apt::APT::Keep-Downloaded-Packages "true";' > /etc/apt/apt.conf.d/keep-cache
-RUN {{ common.mount_cache(__lib_apt__) }} \ 
+RUN {{ common.mount_cache(__lib_apt__) }} \
     {{ common.mount_cache(__cache_apt__) }} \
-	apt-get update -y \
-    && apt-get install -q -y --no-install-recommends --allow-remove-essential \
```
parano, line 13: separating user packages is better for caching, right?

aarnphm: actually it is the same, I tested it out locally.

PR #2908, opened 2022-08-15
fix(ci): unittest failed

bojiang:


tests/unit/_internal/test_store.py
```diff
@@ -76,9 +77,13 @@ def test_store(tmpdir: "Path"):
     DummyItem.store = store
     oldtime = datetime.now()
     DummyItem.create("test:version1")
+    time.sleep(1)
     DummyItem.create("test:otherprefix")
+    time.sleep(1)
     DummyItem.create(Tag("test", "version2"))
+    time.sleep(1)
     DummyItem.create("test:version3", creation_time=oldtime)
+    time.sleep(1)
```
ssheng, line 86: No other better options? 🤔

aarnphm: I'm wondering why this issue didn't surface earlier

ssheng: It seems pertinent to Windows so it may have to do with the OS clock API?

ssheng: I think it may be fine to mock `creation_time` here.

bojiang: I just open this PR on behalf of sauyon's suggestion. The current CI is broken, which false-positive all current PRs. But it seems that there's more issue in the unittest about tag version.
Let me hand it over to sauyon

bojiang: @ssheng BTW to me mock creation_time will also decrease the coverage literally. decide the creation_time according to real-world time is an important part of the logic

ssheng: @bojiang for unit testing the `StoreItem` creation logic, passing `creation_time` is probably not losing coverage. As discussed offline, let's get this merged and optimize the test separately.

ssheng (approved):

tests/unit/_internal/test_store.py
sauyon: I remember now we tried to fix this by overwriting latest if the timestamp is the same.

I'm really not sure what's happening here...

PR #2915, opened 2022-08-17
feat(build_config): extras_require

aarnphm:
update logs instruction when running inside docker container

specifying build options via `python`:
```yaml
python:
  extras_require:
    - tracing
    - grpc
```

### DIFF
--- a/bentoml/_internal/bento/build_config.py
+++ b/bentoml/_internal/bento/build_config.py
@@ -64,9 +64,7 @@ def _convert_python_version(py_version: str | None) -> str | None:
     target_python_version = f"{major}.{minor}"
     if target_python_version != py_version:
         logger.warning(
-            "BentoML will install the latest python%s instead of the specified version %s. To use the exact python version, use a custom docker base image. See https://docs.bentoml.org/en/latest/concepts/bento.html#custom-base-image-advanced",
-            target_python_version,
-            py_version,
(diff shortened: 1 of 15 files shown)


bentoml/_internal/bento/build_config.py
```diff
@@ -64,9 +64,7 @@ def _convert_python_version(py_version: str | None) -> str | None:
     target_python_version = f"{major}.{minor}"
     if target_python_version != py_version:
         logger.warning(
-            "BentoML will install the latest python%s instead of the specified version %s. To use the exact python version, use a custom docker base image. See https://docs.bentoml.org/en/latest/concepts/bento.html#custom-base-image-advanced",
-            target_python_version,
-            py_version,
+            f"BentoML will install the latest 'python{target_python_version}' instead of the specified 'python{py_version}'. To use the exact python version, use a custom docker base image. See https://docs.bentoml.org/en/latest/concepts/bento.html#custom-base-image-advanced",
```
parano, line 67: f string will be slower than the original code

aarnphm: really?

aarnphm: https://docs.python.org/3/howto/logging.html#optimization

parano: yup

Issue #2917, opened 2022-08-17
feat(tracing): add support for otlp exporter


dbuades:
### Feature request

Currently, traces are collected using OpenTelemetry.
However, when it comes to exporting, we can only export to a Zipkin or Jaeger server.
This feature would allow a user to export traces using the OTLP exporter directly.

### Motivation

Although Zipkin and Jaeger are widely used, the [OTLP exporter](https://opentelemetry.io/docs/reference/specification/protocol/exporter/) remains closer to the standard and is natively supported by other observability solutions, like Elastic's [APM](https://www.elastic.co/guide/en/apm/guide/current/open-telemetry.html).

### Other

A PR implementing this feature will follow.

PR #2918, opened 2022-08-17
feat(tracing): add support for otlp exporter

dbuades:
## What does this PR address?

Mainly fixes https://github.com/bentoml/BentoML/issues/2917
However, it also addresses the following concerns:
- Add `opentelemetry-exporter-jaeger-thrift` as an extra dependency in order to solve [this issue](https://bentoml.slack.com/archives/CKRANBHPH/p1660564534763419?thread_ts=1660563581.202739&cid=CKRANBHPH)
- Update OpenTelemetry to 1.12 (no incompatibilities found on my side)
- Log a warning message when `sample_rate` is not set for tracing.

## Before submitting:

- [ ] Did you write tests to cover your changes? -> No, I didn't found tests covering tracing. Open to do it in a subsequent PR though.

## Who can help review?
It was mentioned [in Slack](https://bentoml.slack.com/archives/CKRANBHPH/p1660584057611709?thread_ts=1660563581.202739&cid=CKRANBHPH) that @bojiang worked on the tracing feature.


bentoml/_internal/configuration/default_configuration.yaml
```diff
@@ -40,10 +40,13 @@ runners:
 
 tracing:
   type: zipkin
-  sample_rate: 0.0
+  sample_rate: Null
```
parano, line 43: @ssheng @bojiang what's the reason the default was 0 before?

bentoml/_internal/configuration/containers.py
```diff
@@ -374,6 +388,9 @@ def tracer_provider(
 
         if sample_rate is None:
             sample_rate = 0.0
+            logger.warning(
```
parano, line 391: 👍

parano: Looks like this will show up for all users that don't use tracing?

setup.cfg
```diff
@@ -43,13 +43,13 @@ install_requires =
...
     fs
     numpy
-    opentelemetry-api==1.9.0
-    opentelemetry-instrumentation==0.28b0
-    opentelemetry-instrumentation-aiohttp-client==0.28b0
-    opentelemetry-instrumentation-asgi==0.28b0
-    opentelemetry-sdk==1.9.0
-    opentelemetry-semantic-conventions==0.28b0
-    opentelemetry-util-http==0.28b0
+    opentelemetry-api>=1.9.0
```
parano, line 46: @bojiang what's the reason we lock the opentelemetry version before? was it to prevent installing the pre-release versions by accident?  if that's the case it should be ok to change this to `>=` now

bentoml/_internal/configuration/containers.py
dbuades: @parano You are right, I'm going to change it.

setup.cfg
aarnphm: There are some issue with opentelemetry pre-release before, couldn't remember what the issues are. Idk if that got fixed upstream.

bentoml/_internal/configuration/default_configuration.yaml
```diff
@@ -40,10 +40,13 @@ runners:
...
   excluded_urls: Null
   zipkin:
     url: Null
   jaeger:
     address: Null
     port: Null
+  otlp:
+    protocol: Null
+    url: Null
```
aarnphm, line 52: Is it possible that you can change `Null` to ~ instead?
https://symfony.com/doc/current/components/yaml/yaml_format.html#nulls
imo it looks a bit nicer 😄

bentoml/_internal/configuration/containers.py
```diff
@@ -410,6 +427,29 @@ def tracer_provider(
...
             )
+            provider.add_span_processor(BatchSpanProcessor(exporter))  # type: ignore (no opentelemetry types)
+            return provider
+        elif (
+            tracer_type == "otlp"
+            and otlp_server_protocol is not None
+            and otlp_server_url is not None
+        ):
+            if otlp_server_protocol == "grpc":
+                # pylint: disable=no-name-in-module # https://github.com/open-telemetry/opentelemetry-python-contrib/issues/290
```
aarnphm, line 438: If you want to put pylint disable you probably ned to put it inline with L439

bentoml/_internal/configuration/containers.py
```diff
@@ -410,6 +427,29 @@ def tracer_provider(
...
+                from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import (
+                    OTLPSpanExporter,  # type: ignore (no opentelemetry types)
```
aarnphm, line 440: remove type: ignore here? I think by default we enable pyright to infer from library code.

bentoml/_internal/configuration/containers.py
```diff
@@ -410,6 +427,29 @@ def tracer_provider(
...
+                )
+
+            elif otlp_server_protocol == "http":
+                # pylint: disable=no-name-in-module # https://github.com/open-telemetry/opentelemetry-python-contrib/issues/290
+                from opentelemetry.exporter.otlp.proto.http.trace_exporter import (
+                    OTLPSpanExporter,  # type: ignore (no opentelemetry types)
+                )
+
+            exporter = OTLPSpanExporter(  # type: ignore (no opentelemetry types)
```
aarnphm, line 449: same comments with `type: ignore`

aarnphm (commented):
LGTM with some minor changes

bentoml/_internal/configuration/default_configuration.yaml
parano: That’s not specific to this PR, we can revisit later

bentoml/_internal/configuration/containers.py
ssheng: Maybe we should warn if  user explicitly set the `sample_rate` to `0.0`. If the `sample_rate` is `Null` we can leave the behavior as-is.

ssheng (approved):
I'm happy to merge and update the warning change for you in a separate PR.

setup.cfg
bojiang: We have to lock the version of otlp libs because they are not well organized.

bojiang:
It seems that opentelemetry-exporter-otlp is requiring newer version of the opentelemetry-api.
We should fully test the otlp features since otlp plugins are in an early stage and have strict requirements to the api, plus our e2e doesn't cover the otlp.

bentoml/_internal/configuration/containers.py
dbuades: Thanks for the review! I don't have a problem modifying this but to be honest I re-used the same hints used for zipkin and jaeger from lines 406 and 421. Maybe @sauyon can chime in since these hints were originally added in their PR: https://github.com/bentoml/BentoML/pull/2632 .

bentoml/_internal/configuration/containers.py
dbuades: Same, got it from lines 408 and 423.

docs/source/guides/tracing.rst
```diff
@@ -2,26 +2,38 @@
 Tracing
 =======
 
-BentoML API server supports tracing with both `Zipkin <https://zipkin.io/>`_ and
-`Jaeger <https://www.jaegertracing.io/>`_. To config tracing server, user can provide a
-config YAML file specifying the tracer type and tracing server information:
+BentoML API server supports tracing with both `Zipkin <https://zipkin.io/>`_,
```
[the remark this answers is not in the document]
dbuades: good catch!

setup.cfg
dbuades: @aarnphm Unfortunately I needed to bump it since it was required by `opentelemetry-exporter-otlp`
@bojiang Would you be confortable if I lock them to their current version until the e2e tests are added? Although we don't have e2e tests, on our side we tested tracing with jaeger, zipkin and otlp exporters on these versions and didn't find any errors.

bentoml/_internal/configuration/containers.py
dbuades: @ssheng I still think that we should warn in both situations, since it took us a while to debug why the traces weren't being sent. I'm open to discussing it though.

I just pushed a fix for this, let me know what you think.

setup.cfg
```diff
@@ -92,8 +92,9 @@ console_scripts =
 
 [options.extras_require]
 tracing =
-    opentelemetry-exporter-jaeger
+    opentelemetry-exporter-jaeger-thrift
```
aarnphm, line 95: I don't think we need to explicitly specify this, since `opentelemetry-exported-jaeger` also includes `opentelemetry-exporter-jaeger-thrift` as a dependency.

dbuades: It was just in order to minimize the number of required packages since we only import `opentelemetry.exporter.jaeger.thrift` and because there was a problem raised [on Slack](https://bentoml.slack.com/archives/CKRANBHPH/p1660564534763419?thread_ts=1660563581.202739&cid=CKRANBHPH) were thrift was not properly installed.
Could you check if it works properly for you without specifying thrift?

aarnphm: I tested this out yesterday and it works on my end.

dbuades: In that case I'll remove it, thanks for testing! 👍

aarnphm:
Hmm maybe you need to rebase this onto main to remove the merge commit?

ssheng:
I have verified that the change here is compatible with the existing `jaeger` and `zipkin` tracers. The `otlp` exporter is also working as expected. I will go ahead and merge this PR. Thanks again for your contribution, @dbuades!

Following this PR, I will lock `opentelemetry` dependencies to `0.33b0` until they stabilizes. I also discovered that the service name field was not properly set as the `api_service` or the `runner` names, causing the UI to display "unknown_service" as a result. I will look to address in the same following PR. Did you encounter this problem, @dbuades?

ssheng (approved):

dbuades:
> I have verified that the change here is compatible with the existing `jaeger` and `zipkin` tracers. The `otlp` exporter is also working as expected. I will go ahead and merge this PR. Thanks again for your contribution, @dbuades!
> 
> Following this PR, I will lock `opentelemetry` dependencies to `0.33b0` until they stabilizes. I also discovered that the service name field was not properly set as the `api_service` or the `runner` names, causing the UI to display "unknown_service" as a result. I will look to address in the same following PR. However, using environment variable to set the service name is possible. Did you use the env var mechanism, @dbuades?

Thanks @ssheng !
That's a very good point for the service name. We are currently are using the env var mechanism since we use that strategy it in all of our microservices, so it felt natural to use the same strategy here. 
However, it makes total sense to infer it from the bento service name if the env var is not set. Let me know if you'd like me to help with the PR or with the review.

ssheng:
Thanks for the feedback, @dbuades! We'd love to get your review on this, https://github.com/bentoml/BentoML/pull/2928.

PR #2920, opened 2022-08-18
feat(cli): add CLI command to serve a runner

bojiang:
to arrange our serving CLI/Python APIs like this:
```
bentoml serve
bentoml serve --production
bentoml serve-grpc
bentoml serve-grpc --production

bentoml start-grpc-server
bentoml start-rest-server
bentoml start-runner-server
```
The APIs start with start are low-level APIs used in Yatai, would be hidden from the help message for now 

run bare servers from container
`docker run -it --rm -p 3000:3000 -e BENTOML_SERVE_COMPONENT=rest_server -e BENTOML_SERVE_RUNNER_MAP="asdfasdf=1 py_model.case-1.e2e=tcp://192.168.67.1:4000" general_workflow_service.case-1.e2e:tefgtebeok7iiusu`
`docker run -it --rm -p 4000:3000 -e BENTOML_SERVE_COMPONENT=runner -e BENTOML_SERVE_RUNNER_NAME=py_model.case-1.e2e general_workflow_service.case-1.e2e:awvtyqreokd4cusu`


bentoml_cli/worker/README.md
```diff
@@ -0,0 +1,3 @@
+Here are entrypoints for the bare workers. They are typically used by the supervisor and not directly by the user.
+
```
aarnphm, line 2: To run a standlone api server do
```bash
python -m bentoml_cli.worker.http_api_server ...
```

bentoml/_internal/bento/docker/entrypoint.sh
```diff
@@ -11,7 +11,7 @@ _is_sourced() {
 
 _main() {
 	# if first arg looks like a flag
-	if [ "${1:0:1}" = '-' ]; then
+	if [ -z "$@" ] || [ "${1:0:1}" = '-' ]; then
```
ssheng, line 14: Is `-z "$@"` a superset of `"${1:0:1}" = '-'`?

bentoml/serve.py
```diff
@@ -393,3 +393,202 @@ def serve_production(
...
+                port,
+            ),
+        )
+    finally:
+        if uds_path is not None:
+            shutil.rmtree(uds_path)
+
+
+@inject
+def serve_bare_api_server(
```
ssheng, line 485: "Serve" isn't technically accurate here. How about just `start_api_server` and `start_runner`?

bentoml_cli/serve.py
```diff
@@ -17,6 +17,30 @@ def add_serve_command(cli: click.Group) -> None:
 
     @cli.command()
     @click.argument("bento", type=click.STRING, default=".")
+    @click.option(
+        "--component",
+        type=click.STRING,
+        default=None,
+        envvar="BENTOML_SERVE_COMPONENT",
+        help="[Experimental] Component (`api-server` or `runner`) to serve, if not specified, will serve all components",
```
ssheng, line 25: Should it be `api_server` with a underscore? `api_server` is how it appears in other references.

bentoml_cli/utils.py
```diff
@@ -97,6 +97,7 @@ def wrapper(
             **kwargs: P.kwargs,
         ) -> t.Any:
             if config:
+                print("inject config:", config)
```
ssheng, line 100:
```suggestion
```

requirements/docs-requirements.txt
```diff
@@ -1,7 +1,7 @@
 # Docs dependencies
 -r tests-requirements.txt
-setuptools>=62.0.0
 sphinx==4.5.0
+setuptools==60.0.0
```
ssheng, line 4: What is the reason for the downgrade?

bentoml/_internal/bento/docker/entrypoint.sh
bojiang: Not really

bojiang: `command`: 
`-z "$@"` -> true 
`[ "${1:0:1}" = '-' ]` -> false

requirements/docs-requirements.txt
bojiang: ping @aarnphm

bentoml/_internal/bento/docker/entrypoint.sh
sauyon: Minor style nit, but I think `[ $# -eq 1 ]` is clearer than `[ -z "$@" ]`.

requirements/docs-requirements.txt
aarnphm: this has to do with the newer version of setuptools and it setup different kind of editables install. It is better to just lock setuptools for now.

sauyon:
I think HTTP is more accurate than REST, since our APIs can pretty easily be used to make a non-RESTful server.

bentoml_cli/utils.py
```diff
@@ -97,6 +97,7 @@ def wrapper(
...
                 load_global_config(config)
```
ssheng, line 102:
```suggestion
```

ssheng (approved):

PR #2921, opened 2022-08-18
perf(cli): lazy import

aarnphm:
This PR significantly improve the speed of our CLI. 

main-graph
The reason behind this is that we are doing a lot of weird chaining imports, which increase the load times from python to figure out what modules to import

Screenshot 2022-08-18 at 19 19 41

The second graphs show a significant improve in load time. The bottleneck are essentially all bentoml public API imports.

We added a regression test, in which will determined an upper bound for our library import speed.

The graph shows an incorrect value of the actual library imports, as the majority of the time relies on how Python resolves dependencies and find out import path.

The core library import itself are pretty lightweight. With this PR, it is around 160ms.


bentoml/__init__.py
```diff
@@ -75,9 +71,15 @@
...
     from bentoml import picklable_model
     from bentoml import pytorch_lightning
+
+    from . import io
+    from . import models
 else:
     from ._internal.utils import LazyLoader as _LazyLoader
 
+    io = _LazyLoader("io", globals(), "bentoml.io")
+    models = _LazyLoader("models", globals(), "bentoml.models")
```
parano, line 81: this change makes sense, let's create a separate PR for just this

bentoml/_internal/bento/bento.py
```diff
@@ -131,6 +121,10 @@ def _export_ext() -> str:
 
     @__fs.validator  # type:ignore # attrs validators not supported by pyright
     def check_fs(self, _attr: t.Any, new_fs: FS):
+        import fs.errors
+
+        from ..models.model import ModelStore
```
parano, line 126: This kind of code is very hard to maintain, we should avoid it

bentoml/_internal/bento/pip_pkg.py
```diff
@@ -1,408 +0,0 @@
-import os
```
parano, line 1: let's keep these two files here for now.

bentoml/_internal/utils/cattr.py
```diff
@@ -26,23 +27,100 @@ def datetime_structure_hook(dt_like: str | datetime, _: t.Any) -> datetime:
...
-    lambda cls: make_dict_unstructure_fn(
-        cls,
-        bentoml_cattr,
-        _cattrs_omit_if_default=getattr(cls, "__omit_if_default__", False),
-    ),
-)
-
-bentoml_cattr.register_structure_hook(datetime, datetime_structure_hook)
-bentoml_cattr.register_unstructure_hook(datetime, lambda dt: dt.isoformat())
+def register_global_structure_hook() -> None:
```
parano, line 30: let's don't do this change. it's better to keep the structure/unstructure hook contextual to where the class is defined

parano:
closing this for now, let's discuss offline @aarnphm @ssheng

PR #2926, opened 2022-08-19
fix(json): revert eager check

aarnphm:
This PR removes an eager check for a pydantic object.

Our JSONEncoder should already handle pydantic object lazily

https://bentoml.slack.com/archives/CKRANBHPH/p1660930201114169

### DIFF
--- a/bentoml/_internal/io_descriptors/json.py
+++ b/bentoml/_internal/io_descriptors/json.py
@@ -241,22 +241,22 @@ class JSON(IODescriptor[JSONType]):
                 pydantic_model = self._pydantic_model.parse_obj(json_obj)
                 return pydantic_model
             except pydantic.ValidationError as e:
-                raise BadInput(f"Invalid JSON input received: {e}") from None
+                raise BadInput(f"Invalid JSON input received: {e}") from e
         else:
             return json_obj
 
     async def to_http_response(
         self, obj: JSONType | pydantic.BaseModel, ctx: Context | None = None
     ):
-        try:
-            import pydantic
+        if self._pydantic_model:
+            try:
+                obj = self._pydantic_model.validate(obj)
+            except pydantic.ValidationError as e:
+                raise BadInput(f"Invalid JSON input received: {e}") from e
 
-            # This is to prevent cases where custom JSON encoder is used.
-            if isinstance(obj, pydantic.BaseModel):
-                obj = obj.dict()
-        except ImportError:
-            # We just need to pass if pydantic is not installed.
-            pass
+        # This is to prevent cases where custom JSON encoder is used.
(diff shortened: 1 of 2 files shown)


yubozhao (approved):

yubozhao (approved):

bentoml/_internal/io_descriptors/json.py
```diff
@@ -248,8 +248,15 @@ async def from_http_request(self, request: Request) -> JSONType:
     async def to_http_response(
         self, obj: JSONType | pydantic.BaseModel, ctx: Context | None = None
     ):
-        if isinstance(obj, pydantic.BaseModel):
-            obj = obj.dict()
+        try:
+            import pydantic
+
+            # This is to prevent cases where custom JSON encoder is used.
+            if isinstance(obj, pydantic.BaseModel):
```
parano, line 255: use `if self._pydantic_model`

aarnphm: This is not correct right? checking object instance is independent of whether `self._pydantic_model` is set or not.

bentoml/_internal/io_descriptors/json.py
```diff
@@ -241,14 +241,21 @@ async def from_http_request(self, request: Request) -> JSONType:
...
 
     async def to_http_response(
         self, obj: JSONType | pydantic.BaseModel, ctx: Context | None = None
     ):
-        if isinstance(obj, pydantic.BaseModel):
+        if self._pydantic_model:
+            try:
+                obj = self._pydantic_model.validate(obj)
+            except pydantic.ValidationError as e:
+                raise BadInput(f"Invalid JSON input received: {e}") from e
```
aarnphm, line 255:
```suggestion
```

bentoml/_internal/io_descriptors/json.py
```diff
@@ -241,14 +241,16 @@ async def from_http_request(self, request: Request) -> JSONType:
                 pydantic_model = self._pydantic_model.parse_obj(json_obj)
                 return pydantic_model
             except pydantic.ValidationError as e:
-                raise BadInput(f"Invalid JSON input received: {e}") from None
+                raise BadInput(f"Invalid JSON input received: {e}") from e
```
sauyon, line 244: Hm, do you know why this was `None` before?

sauyon (approved):
LGTM 👍

bentoml/_internal/io_descriptors/json.py
aarnphm: no idea.

parano: @sauyon @aarnphm https://peps.python.org/pep-0415/ that was for better exception context

PR #2929, opened 2022-08-22
fix: use a temporary file for file uploads

sauyon:
This should alleviate some of the performance issues observed due to the current implementation of formparser: using a Python list for large data (greater than megabytes) appears to have a severely detrimental (effects measured in minutes) effect on performance.

### DIFF
--- a/bentoml/_internal/utils/formparser.py
+++ b/bentoml/_internal/utils/formparser.py
@@ -10,15 +10,20 @@ from starlette.requests import Request
 from starlette.responses import Response
 from starlette.formparsers import MultiPartMessage
 from starlette.datastructures import Headers
+from starlette.datastructures import FormData
+from starlette.datastructures import UploadFile
 from starlette.datastructures import MutableHeaders
 
 from .http import set_cookies
+from ...exceptions import BadInput
 from ...exceptions import BentoMLException
 
 if TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


bentoml/_internal/utils/formparser.py
```diff
@@ -121,23 +131,48 @@ async def parse(self) -> _ItemsBody:
...
+                        )
+                    if b"filename" in options:
+                        filename = user_safe_decode(options[b"filename"], charset)
+                        multipart_file = UploadFile(
+                            filename=filename,
+                            content_type=content_type.decode("latin-1"),
+                            headers=Headers(raw=headers),  # type: ignore (incomplete starlette types)
+                        )
+                    else:
+                        multipart_file = None
```
ssheng, line 164: Is it a valid format if "filename" is not a part of the header? If yes, can we still create an `UploadFile` to buffer the bytes?

sauyon: I think technically this is the case, but given that I think this will probably slow down small requests (this is entirely a hunch based on no evidence) I'm not sure we should do that.

FWIW this is starlette's current implementation of multipart, so at the very least we should match its performance now. I think @larme's cpython rewrite should be the real solution, though.

ssheng (approved):

PR #2930, opened 2022-08-22
fix: importing mlflow:/ urls with no extra path info

tweeklab:
## What does this PR address?

When importing a MLFlow model with a models:/ URI, the downloader puts the files at the registered model path directly into the `dst_path`.  When `dst_path` is specified as `bento_model.path` in this case, the downloader simply returns that same value.  If `bentoml_model.path` had the value `/tmp/test_tmp_path`, then the subsequent rename of the downloaded dir to `mlflow_model` to turn into the equivalent of:

`shutil.move(/tmp/test_tmp_path, /tmp/test_tmp_path/mlflow_model)`, which is going to always fail.  The other URI types I've tried don't seem to do this, they all seem to create a subdirectory under `bento_model.path` as a side effect of the download.

This fix creates an extra directory to download into, so if a `models:/` URI is used, it will instead download into this directory and give the `shutil.move` something to rename.  If the directory is still there after the fact, it's cleaned up. 

## Before submitting:

- [ ] Did you write tests to cover your changes? - I can't think of a way to test this that does not require patching the downloader.  Since the downloader is imported when the code runs (to accommodate different versions of mlflow), I can't figure out how to get it patched.  Open to ideas.

## Who can help review?

### DIFF
--- a/bentoml/_internal/frameworks/mlflow.py
+++ b/bentoml/_internal/frameworks/mlflow.py
@@ -4,6 +4,7 @@ import os
 import shutil
 import typing as t
 import logging
+import tempfile
 from typing import TYPE_CHECKING
 
 import bentoml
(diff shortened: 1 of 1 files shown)


ssheng (approved):

bentoml/_internal/frameworks/mlflow.py
```diff
@@ -184,24 +185,35 @@ def import_model(
...
 
         mlflow_model_path = bento_model.path_of(MLFLOW_MODEL_FOLDER)
         # Rename model folder from original artifact name to fixed "mlflow_model"
         shutil.move(local_path, mlflow_model_path)
+        # If the temp dir we created still exists now, we never needed it because
+        # the provided mlflow url must have provided enough path information.  Just
+        # delete it.  If it's not here, it means we needed it but it's been renamed
+        # by now so we don't need to remove it.
+        if os.path.isdir(download_dir):
+            shutil.rmtree(download_dir)
```
ssheng, line 216: I'd suggest to put this logic in a `finally` block. We can merge this and add the finally block as a follow-up.

PR #2935, opened 2022-08-24
docs: add instruction for external package on Alpine

aarnphm:
This PR adds a docs section telling users to install additional packages if using Alpine
This has to do with the lack of compiled binary wheels from numpy on Alpine

### DIFF
--- a/docs/source/concepts/bento.rst
+++ b/docs/source/concepts/bento.rst
@@ -694,6 +694,24 @@ Here's a basic Docker options configuration:
    However, if you are using a standalone version of Docker, you can install
    BuildKit by following the instructions `here <https://github.com/docker/buildx#installing>`_.
 
+.. note::
+
+   If you are building your Bento container using Alpine image, make sure to add the
+   following to ``system_packages``:
+
+   .. code:: yaml
+
+      docker:
+          distro: alpine
+          system_packages:
+              - gcc
+              - gfortran
+              - build-base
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -2,8 +2,8 @@
 requires = [
   # sync with setup.py until we discard non-pep-517/518
   "wheel",
-  "setuptools>=59.0",
-  "setuptools-scm[toml]>=6.2.3",
+  "setuptools<60",
+  "setuptools-scm[toml]>=6.3.1",
```
ssheng, line 6: Why do we need to bump dependency?

docs/source/concepts/bento.rst
```diff
@@ -694,6 +694,24 @@ Here's a basic Docker options configuration:
...
+          distro: alpine
+          system_packages:
+              - gcc
+              - gfortran
+              - build-base
+              - wget
+              - freetype-dev
+              - libpng-dev
+              - openblas-dev
+
```
ssheng, line 714: The bento doc is already complicated. Is there a better place for these branch cases?

aarnphm: We can make a separate docs call `improvement`.

ssheng: As discussed, this may be a great addition for Discourse. cc: @yubozhao

PR #2945, opened 2022-08-26
qa: cleanup MLflow

aarnphm:
Cleanup some logic, qa with local postgres, mysql setup of mlflow registry.
cleanup some types as well.

### DIFF
--- a/bentoml/_internal/frameworks/mlflow.py
+++ b/bentoml/_internal/frameworks/mlflow.py
@@ -4,7 +4,6 @@ import os
 import shutil
 import typing as t
 import logging
-import tempfile
 from typing import TYPE_CHECKING
 
 import bentoml
@@ -24,6 +23,7 @@ if TYPE_CHECKING:
 
 if TYPE_CHECKING:
     import mlflow
+    import mlflow.models
 else:
     mlflow = LazyLoader(
         "mlflow",
@@ -185,6 +185,8 @@ def import_model(
         metadata=metadata,
         context=context,
     ) as bento_model:
+        import tempfile
+
         from mlflow.models import Model as MLflowModel
         from mlflow.pyfunc import FLAVOR_NAME as PYFUNC_FLAVOR_NAME
         from mlflow.models.model import MLMODEL_FILE_NAME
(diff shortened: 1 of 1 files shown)


bentoml/_internal/frameworks/mlflow.py
```diff
@@ -4,7 +4,6 @@
 import shutil
 import typing as t
 import logging
-import tempfile
```
parano, line 7: this is still required, right?

aarnphm: I imported this lazily

parano: probably no need to do that for built-in libraries

bentoml/_internal/frameworks/mlflow.py
```diff
@@ -205,19 +206,20 @@ def import_model(
...
-        # If the temp dir we created still exists now, we never needed it because
-        # the provided mlflow url must have provided enough path information.  Just
-        # delete it.  If it's not here, it means we needed it but it's been renamed
-        # by now so we don't need to remove it.
-        if os.path.isdir(download_dir):
+        finally:
+            if TYPE_CHECKING:
+                # We need to trick type checker here, since local_path
+                # will always be set
+                local_path = ""
```
ssheng, line 216: I think we have gone too far to please the type checker. 🙂 Maybe put this before the `try`?

bentoml/_internal/frameworks/mlflow.py
```diff
@@ -205,19 +206,20 @@ def import_model(
...
-
-        mlflow_model_path = bento_model.path_of(MLFLOW_MODEL_FOLDER)
-        # Rename model folder from original artifact name to fixed "mlflow_model"
-        shutil.move(local_path, mlflow_model_path)
-        # If the temp dir we created still exists now, we never needed it because
-        # the provided mlflow url must have provided enough path information.  Just
-        # delete it.  If it's not here, it means we needed it but it's been renamed
-        # by now so we don't need to remove it.
-        if os.path.isdir(download_dir):
+        finally:
```
ssheng, line 212: 👍

bentoml/_internal/frameworks/mlflow.py
aarnphm: 😄

sauyon: I think this is a true positive; it's possible if there are exceptions raised that `local_path` is not set.

ssheng (approved):

PR #2949, opened 2022-08-29
ci: temp fix for models test

sauyon:
Another temporary fix for the weird store issue that seems to occur only on actions.

### DIFF
--- a/tests/unit/test_models.py
+++ b/tests/unit/test_models.py
@@ -1,4 +1,5 @@
 import os
+import time
 import random
 import string
 from sys import version_info as pyver
@@ -48,6 +49,8 @@ def test_models(tmpdir: "Path"):
         _model_store=store,
     ) as testmodel:
         testmodel1tag = testmodel.tag
+
+    time.sleep(1)
 
     with bentoml.models.create(
         "testmodel",


tests/unit/test_models.py
```diff
@@ -49,6 +50,8 @@ def test_models(tmpdir: "Path"):
     ) as testmodel:
         testmodel1tag = testmodel.tag
 
+    time.sleep(1)
```
aarnphm, line 53: maybe we can check whether this is running in actions
```suggestion
if os.getenv("GITHUB_ACTIONS"):
   ...
```
So that when test run locally we don't have to sleep it

ssheng (approved):
I still think we can use the mocked timestamp for unit testing purpose but I'm OK with this change, too.

tests/unit/test_models.py
ssheng: Please address address @aarnphm's suggestion in a following PR if needed

sauyon: I think we shouldn't special case our tests too much, it'll just make debugging harder.

Though I should probably debug what's going on here at some point...

PR #2961, opened 2022-09-01
feat: logging customization

jiewpeng:
## What does this PR address?

Adds customisation for logging, for example to change the format of the opentelemetry trace id and span id. Thread: https://bentoml.slack.com/archives/CKRANBHPH/p1661766094533779

## Before submitting:

- [ ] Did you write tests to cover your changes? No. I am unsure how to get the logs from the `host` fixture to test the configuration of logging. Have manually tested using `bentoml serve` to check that the configuration works.

## Who can help review?

### DIFF
--- a/bentoml/_internal/configuration/containers.py
+++ b/bentoml/_internal/configuration/containers.py
@@ -543,7 +543,7 @@ class _BentoMLContainerClass:
 
     @providers.SingletonFactory
     @staticmethod
-    def logging_formats(
+    def logging_formatting(
         cfg: dict[str, t.Any] = Provide[config.logging.formatting],
     ) -> dict[str, str]:
         return cfg
--- a/bentoml/_internal/configuration/containers.py
+++ b/bentoml/_internal/configuration/containers.py
@@ -180,8 +180,6 @@ SCHEMA = Schema(
         },
         "logging": {
             "formatting": {
-                "format": str,
-                "datefmt": str,
                 "trace_id_format": str,
                 "span_id_format": str,
             }
--- a/bentoml/_internal/configuration/containers.py
+++ b/bentoml/_internal/configuration/containers.py
@@ -178,6 +178,14 @@ SCHEMA = Schema(
                 },
             },
         },
+        "logging": {
+            "formatting": {
+                "format": str,
+                "datefmt": str,
+                "trace_id_format": str,
+                "span_id_format": str,
+            }
+        },
     }
 )
 
@@ -534,6 +542,13 @@ class _BentoMLContainerClass:
                 f"'duration' configuration, '{duration}'."
             )
         return DEFAULT_BUCKET
+
+    @providers.SingletonFactory
+    @staticmethod
+    def logging_formats(
+        cfg: dict[str, t.Any] = Provide[config.logging.formatting],
+    ) -> dict[str, str]:
+        return cfg
 
 
 BentoMLContainer = _BentoMLContainerClass()
--- a/bentoml/_internal/configuration/default_configuration.yaml
+++ b/bentoml/_internal/configuration/default_configuration.yaml
@@ -51,3 +51,10 @@ tracing:
   otlp:
     protocol: Null
     url: Null
+
+logging:
+  formatting:
+    format: "%(asctime)s %(levelname_bracketed)s %(component)s %(message)s%(trace_msg)s"
+    datefmt: "%Y-%m-%dT%H:%M:%S%z"
+    trace_id_format: 032x
+    span_id_format: 016x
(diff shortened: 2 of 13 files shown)


jiewpeng:
I'm open to discussion on the following

- Default `trace_id` and `span_id` format: currently I am proposing to change the default format to hexadecimal, which will be different from bentoml's current behaviour (logging as integer).
- Exposing the logging formatting for the user to configure. I think this is useful for organizations that want to standardise their logging formats across their apps.

bentoml/_internal/configuration/containers.py
```diff
@@ -535,5 +541,12 @@ def duration_buckets(
             )
         return DEFAULT_BUCKET
 
+    @providers.SingletonFactory
+    @staticmethod
+    def logging_formats(
```
aarnphm, line 546: I prefer to make this `logging_formatting` to be consistent with the config level.

bentoml/_internal/log.py
```diff
@@ -50,6 +51,10 @@ def filter(self, record: logging.LogRecord) -> bool:
 )
 DATE_FORMAT = "%Y-%m-%dT%H:%M:%S%z"
 
+LOGGING_FORMATS = BentoMLContainer.logging_formats.get()
+TRACE_ID_FORMAT = LOGGING_FORMATS["trace_id_format"]
+SPAN_ID_FORMAT = LOGGING_FORMATS["span_id_format"]
```
aarnphm, line 56: Let's to this lazily, since this will be imported at top level.

bentoml/_internal/log.py
```diff
@@ -111,10 +116,13 @@ def trace_record_factory(*args: t.Any, **kwargs: t.Any):
...
     record.component = f"[{_component_name()}]"  # type: ignore (adding fields to record)
-    if trace_context.trace_id == 0:
+    trace_id = trace_context.trace_id
+    if trace_id in (0, None):
         record.trace_msg = ""  # type: ignore (adding fields to record)
     else:
-        record.trace_msg = f" (trace={trace_context.trace_id},span={trace_context.span_id},sampled={trace_context.sampled})"  # type: ignore (adding fields to record)
+        trace_id = format(trace_id, TRACE_ID_FORMAT)
+        span_id = format(trace_context.span_id, SPAN_ID_FORMAT)
+        record.trace_msg = f" (trace={trace_id},span={span_id},sampled={trace_context.sampled})"  # type: ignore (adding fields to record)
```
aarnphm, line 125: move the eager imports above to L122.

aarnphm (commented):
LGTM, with some minor comments.

bentoml/_internal/configuration/containers.py
jiewpeng: I could change it, but would there be a use case for someone to have one trace id format for the api server and a different format for the runner?

jiewpeng:
Have done the import lazily, but I'm just curious, is there a reason to do so. given it will run at some point anyway, provided the model is called? Based on my understanding, this delays the import of the `BentoMLContainer` for as long as possible, until the model gets invoked. Either this import is slow, in which case doing the import eagerly slows down the first request to the model, or the import is fast, and there is no difference (or it makes it slightly slower because we always run the 3 lines every time we log something, when those things do not change).

bentoml/_internal/configuration/containers.py
aarnphm: Oh I was saying just to change the naming to make it consistent

aarnphm:
> Have done the import lazily, but I'm just curious, is there a reason to do so.

This has to do with importing bentoml to global namespace is actually pretty inefficient. We will fix this some day 😄 

`logging_formatting` is a singletonFactory, meaning it will be cached once, so actually the access for the field are pretty fast, so I'm not too worried about the performance deficit if we do it lazily.

aarnphm:
can you rebase to the main branch, I saw there were some bad commits there. Otherwise this is GTG

bentoml/_internal/configuration/default_configuration.yaml
```diff
@@ -51,3 +51,8 @@ tracing:
   otlp:
     protocol: Null
     url: Null
+
+logging:
+  formatting:
+    trace_id_format: 032x
+    span_id_format: 016x
```
sauyon, line 58: Unfortunately while I think we should have done this to begin with, I don't think we can do this by default as it may break people's setups.

At some point we should work out how to handle default configuration changes, but for now I think we'll have to keep the defaults the way they were before :(

sauyon: Spoke to the team, we're ok with this change :+1:

sauyon (approved):
Looks good, thank you!

PR #2997, opened 2022-09-14
feat(xgboost): support xgboost sklearn models

### LINKED ISSUE #2983: feature: add support for XGBoost models trained using the XGBoost scikit-learn interface
### Feature request

I would like to use BentoML to save and deploy XGBoost models that were trained using the scikit-learn learn interface of the XGBoost model. 

### Motivation

I'm trying to save an XGBoost model using BentoML. I'm using the [scikit-learn interface](https://xgboost.readthedocs.io/en/stable/python/python_api.html#module-xgboost.sklearn) of the XGBoost module, so my final model is of type `xgboost.sklearn.XGBClassifier`. I noticed [here](https://github.com/bentoml/BentoML/blob/main/bentoml/_internal/frameworks/xgboost.py#L162) that BentoML expects the model to be of type `xgboost.core.Booster`, which is what you get when you use the native XGBoost module. 

I also can't save the model as an sklearn model (without failing [this validation](https://github.com/bentoml/BentoML/blob/main/bentoml/_internal/frameworks/sklearn.py#L145)).

### Other

_No response_

sauyon:
n.b.: I don't believe this supports 0.7post3; will do some digging later to identify which version we support now.

Should resolve #2983.

### DIFF
--- a/bentoml/xgboost.py
+++ b/bentoml/xgboost.py
@@ -4,12 +4,9 @@ from ._internal.frameworks.xgboost import get
 from ._internal.frameworks.xgboost import load_model
 from ._internal.frameworks.xgboost import save_model
 from ._internal.frameworks.xgboost import get_runnable
-from ._internal.frameworks.xgboost import XGBoostOptions
+from ._internal.frameworks.xgboost import XGBoostOptions as ModelOptions
 
 logger = logging.getLogger(__name__)
-
-
-ModelOptions = XGBoostOptions
 
 
 def save(tag, *args, **kwargs):
--- a/bentoml/xgboost.py
+++ b/bentoml/xgboost.py
@@ -4,8 +4,12 @@ from ._internal.frameworks.xgboost import get
 from ._internal.frameworks.xgboost import load_model
 from ._internal.frameworks.xgboost import save_model
 from ._internal.frameworks.xgboost import get_runnable
+from ._internal.frameworks.xgboost import XGBoostOptions
 
 logger = logging.getLogger(__name__)
+
+
+ModelOptions = XGBoostOptions
 
 
 def save(tag, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


bentoml/xgboost.py
```diff
@@ -4,10 +4,14 @@
 from ._internal.frameworks.xgboost import load_model
 from ._internal.frameworks.xgboost import save_model
 from ._internal.frameworks.xgboost import get_runnable
+from ._internal.frameworks.xgboost import XGBoostOptions
 
 logger = logging.getLogger(__name__)
 
 
+ModelOptions = XGBoostOptions
```
aarnphm, line 12:
```suggestion
from ._internal.frameworks.xgboost import XGBoostOptions as ModelOptions

logger = logging.getLogger(__name__)
```

aarnphm (commented):
LGTM. Should we have a way to handle difference between versioning. I imagine as well start to move versioning for framework up it will be harder to maintain different versions.

sauyon:
We should probably add version checking to all our current frameworks. Not sure that we necessarily need to worry about separating loads and things for now, though; we can just make a v1_load or something if that's necessary in the future.

aarnphm:
> We should probably add version checking to all our current frameworks. Not sure that we necessarily need to worry about separating loads and things for now, though; we can just make a v1_load or something if that's necessary in the future.

I believe a better way to approach this is via folder structure.

```bash
frameworks/v1/... -> v1.framework.load
frameworks/v2/... -> v2.framework.load
```

aarnphm (approved):
For versioning changes, I will make an issue for it.

Wondering if you can test for this. If not then we can merge this in.

sauyon:
Tests added, should be good to go now, hopefuly.

aarnphm (approved):

PR #2984, opened 2022-09-12
test(grpc): e2e + unit tests

aarnphm:
## What does this PR address?

This PR is a follow up on #2808 for tests related manners

This includes e2e, unit tests

This will also include a fix in our e2e folder structure where test model should not be
saved under bentoml local model directory if users run tests locally.

This PR also fixes multipart handles for `from_proto` and `to_proto`.

depends on #3012

PR #2999, opened 2022-09-14
feat(io/image): allow restricting mime types

sauyon:
This PR enables support for passing a list of mime types to the `Image` IO descriptor, which, alongside passing `accept_all_images=False`, will allow users to whitelist allowed image formats.

### DIFF
--- a/bentoml/_internal/io_descriptors/image.py
+++ b/bentoml/_internal/io_descriptors/image.py
@@ -4,7 +4,6 @@ import io
 import typing as t
 import functools
 from typing import TYPE_CHECKING
-from itertools import chain
 from urllib.parse import quote
 
 from starlette.requests import Request
@@ -187,10 +186,15 @@ class Image(IODescriptor[ImageType]):
         )
         self._allow_all_images = allowed_mime_types is None
 
-        for mtype in chain(self._allowed_mimes, [self._mime_type]):
+        if self._mime_type not in MIME_EXT_MAPPING:  # pragma: no cover
+            raise InvalidArgument(
+                f"Invalid Image mime_type '{mime_type}'; supported mime types are {', '.join(PIL.Image.MIME.values())} "
+            )
+
+        for mtype in self._allowed_mimes:
             if mtype not in MIME_EXT_MAPPING:  # pragma: no cover
                 raise InvalidArgument(
-                    f"Invalid Image mime_type '{mtype}'; supported mime types are {', '.join(PIL.Image.MIME.values())} "
+                    f"Invalid Image MIME in allowed_mime_types: '{mtype}'; supported mime types are {', '.join(PIL.Image.MIME.values())} "
                 )
 
         self._pilmode: _Mode | None = pilmode
@@ -252,11 +256,11 @@ class Image(IODescriptor[ImageType]):
                 else:
                     if self._allowed_mimes is None:
                         raise BadInput(
-                            f"no multipart image file (with mime type in {MIME_EXT_MAPPING.keys()} or 'image/*'), got files with content types {', '.join(found_mimes)}"
+                            f"no multipart image file (supported images are: {', '.join(MIME_EXT_MAPPING.keys())}, or 'image/*'), got files with content types {', '.join(found_mimes)}"
                         )
                     else:
                         raise BadInput(
-                            f"no multipart image file (with mime type in {self._allowed_mimes}), got files with content types {', '.join(found_mimes)}"
+                            f"no multipart image file (allowed mime types are: {', '.join(self._allowed_mimes)}), got files with content types {', '.join(found_mimes)}"
                         )
 
         elif self._allowed_mimes is None:
@@ -267,11 +271,11 @@ class Image(IODescriptor[ImageType]):
         else:
             if self._allowed_mimes is None:
                 raise BadInput(
-                    f"no multipart image file (with mime type in {MIME_EXT_MAPPING.keys()} or 'image/*'), got request with content type {mime_type}"
+                    f"unsupported mime type {mime_type}; supported mime types are: {', '.join(MIME_EXT_MAPPING.keys())}, or 'image/*'"
                 )
             else:
                 raise BadInput(
-                    f"no multipart image file (with mime type in {self._allowed_mimes}), got a request with content type {mime_type}"
+                    f"mime type {mime_type} is not allowed, allowed mime types are: {', '.join(self._allowed_mimes)}"
                 )
 
         assert bytes_ is not None
--- a/bentoml/_internal/io_descriptors/image.py
+++ b/bentoml/_internal/io_descriptors/image.py
@@ -3,11 +3,13 @@ from __future__ import annotations
 import io
 import typing as t
 from typing import TYPE_CHECKING
+from itertools import chain
 from urllib.parse import quote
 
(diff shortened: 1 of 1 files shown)


bentoml/_internal/io_descriptors/image.py
```diff
@@ -140,7 +141,9 @@ async def predict_image(f: Image) -> NDArray[Any]:
     def __init__(
         self,
         pilmode: _Mode | None = DEFAULT_PIL_MODE,
-        mime_type: str = "image/jpeg",
+        mime_type: str | t.Iterable[str] | None = None,
+        *,
+        accept_all_images: bool = True,
```
sauyon, line 146: Maybe `restrict_allowed_input_types` or something along those lines instead?

bentoml/_internal/io_descriptors/image.py
```diff
@@ -192,18 +213,47 @@ async def from_http_request(self, request: Request) -> ImageType:
...
+            else:
+                if len(found_mimes) == 0:
+                    raise BadInput("no image file found in multipart form")
+                else:
+                    raise BadInput(
+                        f"no multipart image file (with mime type in {self._accepted_mime_types}"
+                        + " or that matches 'image/*'"
+                        if self._accept_all_images
+                        else ""
+                        + "), got files with content types {', '.join(found_mimes)}"
```
sauyon, line 239: This ended up quite ugly and I wouldn't be against just factoring this out into its own if statement, I think.

bentoml/_internal/io_descriptors/image.py
aarnphm: `allow_all_mime_types`?

bentoml/_internal/io_descriptors/image.py
```diff
@@ -140,7 +141,9 @@ async def predict_image(f: Image) -> NDArray[Any]:
     def __init__(
         self,
         pilmode: _Mode | None = DEFAULT_PIL_MODE,
-        mime_type: str = "image/jpeg",
+        mime_type: str | t.Iterable[str] | None = None,
```
aarnphm, line 144: I think this should be a `t.Sequence` instead. str is also considered `t.Iterable`
```suggestion
        mime_type: str | t.Sequence[str] | None = None,
```

bentoml/_internal/io_descriptors/image.py
sauyon: This is for compatibility, because we specifically allowed only `image/*` before.

bentoml/_internal/io_descriptors/image.py
aarnphm: +1 for refactoring this out.

bentoml/_internal/io_descriptors/image.py
sauyon: I think `str` is also a sequence, but we can remove the `str` type hint.

sauyon: Actually, I think it's better for documentation to allow `str` explicitly.

sauyon: We don't want to make this only accept mutable sequences. I think the current types are fine.

bentoml/_internal/io_descriptors/image.py
aarnphm: `restrict_allowed_input_types` make sense.

bentoml/_internal/io_descriptors/image.py
aarnphm: noted.

aarnphm:
tests are failing. Probably you have also to add unit test and fix the e2e once the gRPC PR got merged.

bentoml/_internal/io_descriptors/image.py
```diff
@@ -190,20 +216,66 @@ def openapi_responses(self) -> OpenAPIResponse:
...
+                if isinstance(val, UploadFile):
+                    found_mimes.append(val_content_type)
+
+                if self._allowed_mimes is None:
+                    if (
+                        val_content_type in MIME_EXT_MAPPING
+                        or val_content_type.startswith("image/")
+                    ):
+                        bytes_ = await val.read()
+                        break
```
ssheng, line 238: Shouldn't this be a reject case with `BadInput`?

bentoml/_internal/io_descriptors/image.py
```diff
@@ -190,20 +216,66 @@ def openapi_responses(self) -> OpenAPIResponse:
...
-                f"`{self._mime_type}` or `image/*`, got {content_type} instead"
-            )
+            if self._allowed_mimes is None:
+                raise BadInput(
+                    f"no multipart image file (with mime type in {MIME_EXT_MAPPING.keys()} or 'image/*'), got request with content type {mime_type}"
+                )
+            else:
+                raise BadInput(
+                    f"no multipart image file (with mime type in {self._allowed_mimes}), got a request with content type {mime_type}"
+                )
```
ssheng, line 268: Are they still "multipart" image files under the `else` condition?

sauyon: Oops, I failed to update this copy-paste.

bentoml/_internal/io_descriptors/image.py
sauyon: This is our old behavior if `allowed_mime_types` is not specified.

bentoml/_internal/io_descriptors/image.py
```diff
@@ -129,42 +150,44 @@ async def predict_image(f: Image) -> NDArray[Any]:
...
     """
 
-    MIME_EXT_MAPPING: t.Dict[str, str] = {}
-
     def __init__(
         self,
         pilmode: _Mode | None = DEFAULT_PIL_MODE,
         mime_type: str = "image/jpeg",
+        *,
+        allowed_mime_types: t.Iterable[str] | None = None,
```
sauyon, line 165: @aarnphm changed this to have a separate parameter for allowed mimes.

bentoml/_internal/io_descriptors/image.py
```diff
@@ -190,20 +220,66 @@ def openapi_responses(self) -> OpenAPIResponse:
...
+                    f"mime type {mime_type} is not allowed, allowed mime types are: {', '.join(self._allowed_mimes)}"
+                )
+
+        assert bytes_ is not None
+
+        if isinstance(bytes_, str):
+            bytes_ = bytes(bytes_, "UTF-8")
+
         try:
             return PIL.Image.open(io.BytesIO(bytes_))
```
aarnphm, line 280: after we open the image, should we also run [verify](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.verify) ?

sauyon: I think we should leave this lazy and with as little overhead as we can; users can do that if they want. At some point maybe we can turn on some sort of option to enable automatic verification.

sauyon:
@aarnphm @ssheng I think this should be good to go in now.

aarnphm:
maybe we need test for this :)

aarnphm (approved):

aarnphm:
I will fix the broken CI on #2984 since we have a lot of changes wrt e2e tests there.

PR #3012, opened 2022-09-18
fix(grpc): docker message

aarnphm:
This PR fixes message output after successfully built Docker image where tags are None

### DIFF
--- a/bentoml_cli/containerize.py
+++ b/bentoml_cli/containerize.py
@@ -6,8 +6,11 @@ import logging
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    F = t.Callable[..., t.Any]
     from click import Group
+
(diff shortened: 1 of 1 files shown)


bojiang (approved):

bentoml_cli/containerize.py
```diff
@@ -6,9 +6,12 @@
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    F = t.Callable[..., t.Any]
     from click import Group
 
+    from bentoml._internal.bento import BentoStore
+
+    F = t.Callable[..., t.Any]
```
bojiang, line 13: where is this used?

aarnphm: it is not, but I will remove it in a separate PR.

PR #3017, opened 2022-09-20
fix(e2e): exception test for image IO

aarnphm:


sauyon:
Uh, this looks like a real issue with the descriptor; let me look into that...

aarnphm:
> Uh, this looks like a real issue with the descriptor; let me look into that...

I check, apparently PIL does accept application/pdf

sauyon:
Oh, glad we fixed this then lol

sauyon (approved):

bojiang:
@aarnphm The docs just say it only supports writing rather than reading

bojiang:
revert request

sauyon:
Let's leave this one in until we actually fix the bug so CI isn't failing

bojiang:
will take that

PR #3020, opened 2022-09-20
revert(3017): filter write-only mime type for Image IO

bojiang:
pdf reading is not supported by PIL

### DIFF
--- a/bentoml/_internal/io_descriptors/image.py
+++ b/bentoml/_internal/io_descriptors/image.py
@@ -57,6 +57,10 @@ ImageType = t.Union["PIL.Image.Image", "ext.NpNDArray"]
 DEFAULT_PIL_MODE = "RGB"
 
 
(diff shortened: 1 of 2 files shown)


bentoml/_internal/io_descriptors/image.py
```diff
@@ -72,7 +76,7 @@ def initialize_pillow():
         )
 
     PIL.Image.init()
-    MIME_EXT_MAPPING = {v: k for k, v in PIL.Image.MIME.items()}  # type: ignore (lazy constant)
+    MIME_EXT_MAPPING = {v: k for k, v in PIL.Image.MIME.items() if k not in PIL_WRITE_ONLY_FORMATS}  # type: ignore (lazy constant)
```
sauyon, line 79: We should support write-only formats for sending, though.

sauyon (approved):

aarnphm (approved):

PR #3024, opened 2022-09-20
feat(io/image): allow writeable mimes as output

sauyon:
@aarnphm you'll probably have to update your exceptions PR if that's in progress.


aarnphm (commented):
Ah you need to update test for this change as well 😄

sauyon:
No, this should pass with the current tests, the failures are because I can't type.

aarnphm (approved):

PR #3025, opened 2022-09-20
chore: migrate to pyproject.toml

aarnphm:
Seems like pyproject.toml is what the Steering council is pushing forward (With the
merge of [tomllib(PEP680)](https://peps.python.org/pep-0680/) to stdlib). We will once
settle all changes related to python packaging.

This PR also introduces a `bentoml[extras]` that enables users to install a feature-full
variant of BentoML


bentoml/_internal/configuration/__init__.py
```diff
@@ -4,23 +4,17 @@
...
-    import importlib.metadata as importlib_metadata
-except ModuleNotFoundError:
-    import importlib_metadata
-
-import yaml
-
-try:
-    import bentoml._version as version_mod
+    from ..._version import __version__
+    from ..._version import version_tuple
```
aarnphm, line 12: Since we are setting version inside `pyproject.toml` to be dynamic, we can do poor man  version compare here.

bentoml/_internal/configuration/__init__.py
```diff
@@ -150,10 +144,3 @@ def load_global_config(bentoml_config_file: t.Optional[str] = None):
     )
 
     BentoMLContainer.config.set(bentoml_configuration.as_dict())
-
-
-def save_global_config(config_file_handle: t.IO[t.Any]):
-    from ..configuration.containers import BentoMLContainer
-
-    content = yaml.safe_dump(BentoMLContainer.config)
-    config_file_handle.write(content)
```
aarnphm, line 159: I believe we probably don't want to expose this to end user, hence the removal.

pyproject.toml
```diff
@@ -1,7 +1,120 @@
 [build-system]
-requires = ["setuptools<=60", "setuptools_scm[toml]>=6.2", "wheel"]
+requires = ["setuptools", "setuptools_scm[toml]>=6.2", "wheel"]
```
aarnphm: note that `pyproject.toml` support is from `setuptools>=63`

bentoml/_internal/configuration/__init__.py
sauyon: I'm not sure what this means.

bentoml/_internal/configuration/__init__.py
sauyon: Why not?

aarnphm: We can have this back once we have config versioning.

sauyon: Why are you removing this? Is this even related to this PR?

bentoml/_internal/configuration/__init__.py
sauyon: Can you explain?

bentoml/_internal/configuration/__init__.py
aarnphm: not really, let me revert all the changes in here.

bentoml/_internal/configuration/__init__.py
aarnphm: I just realized that this change is not needed, but I think for the sake of simplicity, we should just import __version__ from _version instead of importing the module to fix types

sauyon (approved):
LGTM now. I assume the generated file changes are just weird dependency things?

aarnphm:
Yep

PR #3026, opened 2022-09-20
ci(experimental): Enable 3.11 rc

aarnphm:
We will probably should start enable 3.11 rc CI to start support


sauyon:
Can we force trigger all the jobs?

aarnphm:
I'm thinking to just run all general CI jobs nonetheless

sauyon:
Shouldn't we just add all the CI files to the things that it checks? We don't really want to run CI on things like README changes.

aarnphm:
Eh, I will close this for now since 3.11 is not fully supported by many of our dependent libraries. We will run this back when 3.11 is out officially.

I have a 3.11 RC installed on my system, and it seems to work fine. Haven't run any perf benchmark yet, so...

PR #3030, opened 2022-09-21
draft: add get_output_signatures_v2

bojiang:
image


bentoml/_internal/utils/tensorflow.py
```diff
@@ -221,6 +221,19 @@ def get_input_signatures_v2(
 
     return []
 
+def get_output_signatures_v2(
```
aarnphm, line 224: Im thinking whether we should also bump the api_version for tf as well 🤔

PR #3031, opened 2022-09-21
docs: updates on blocks that failed to render

aarnphm:
There are some docs block that are not render

also start migrating :code: to double backtick for inline


larme:
@aarnphm do we have a guideline for when to use `:code:` and when to use double backquotes?

aarnphm:
I think it is a syntactic choice. double backtick feels more natural to markdown.

larme (approved):
lgtm

PR #3018, opened 2022-09-20
feat(configuration): override options with envvar

bojiang:
Some use cases:

### 1. override one value

assign a gpu for a runner

> ```BENTOML_CONFIG_OPTIONS='runners.pytorch_mnist.resources."nvidia.com/gpu"=1' bentoml serve pytorch_mnist_demo --production```

equal to
```
runners:
    pytorch_mnist:
        resources:
            nvidia.com/gpu: 1
```

### 2. override multiple values (divide with space or new line)

assign two specified gpus to a runner
> ```BENTOML_CONFIG_OPTIONS='runners.pytorch_mnist.resources."nvidia.com/gpu"=[0,2]' bentoml serve pytorch_mnist_demo --production```

or 

> ```BENTOML_CONFIG_OPTIONS='runners.pytorch_mnist.resources."nvidia.com/gpu"[0]=0 runners.pytorch_mnist.resources."nvidia.com/gpu"[1]=2' bentoml serve pytorch_mnist_demo --production```

equal to
```
runners:
    pytorch_mnist:
        resources:
            nvidia.com/gpu:
                - 0
                - 2
```

or 

```
runners:
    pytorch_mnist:
        resources:
            nvidia.com/gpu: [0, 2]
```

PR #3041, opened 2022-09-27
fix(configuration): should validate config early

aarnphm:
revert some of changes before #3018


ssheng (approved):
LGTM. Some unit tests are still failing.

PR #3043, opened 2022-09-27
dependencies: ignore broken pypi combination

aarnphm:
Currently protobuf 3.19.5 is pretty much broken, and the latest version that is usuable
and not break stub is 3.19.4

this means that we can't also have to restrict grpcio not to use 1.48.2

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,7 @@ dependencies = [
   "opentelemetry-instrumentation-aiohttp-client==0.33b0",
   "opentelemetry-instrumentation-asgi==0.33b0",
   "opentelemetry-sdk>=1.9.0",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -40,7 +40,7 @@ dependencies = [
   "opentelemetry-instrumentation-aiohttp-client==0.33b0",
   "opentelemetry-instrumentation-asgi==0.33b0",
   "opentelemetry-sdk>=1.9.0",
-  "opentelemetry-semantic-conventions==0.33b0",
+  "opentelemetry-semantic-conventions==0.34b0",
```
ssheng, line 43: Why do we only bump `opentelemetry-semantic-conventions`? Should we bump all other opentelemetry dependencies as well?

ssheng (approved):

PR #3044, opened 2022-09-27
feat: support multipart upload for large bento and model

yetone:


bentoml/_internal/bento/build_dev_bentoml_whl.py
```diff
@@ -28,9 +28,8 @@ def build_bentoml_editable_wheel(target_path: str) -> None:
         return
 
     try:
-        from build.env import IsolatedEnvBuilder
-
         from build import ProjectBuilder
```
aarnphm, line 31: I think there is some isort formatter bug for this specific file. I couldn't figure out why this is happening.

aarnphm (approved):

bentoml/_internal/yatai_rest_api_client/schemas.py
```diff
@@ -4,6 +4,7 @@
 from typing import Dict
 from typing import List
 from typing import Type
+from typing import Literal
```
aarnphm, line 7: This won't work for <3.7

bentoml/_internal/yatai_rest_api_client/schemas.py
```diff
@@ -175,7 +179,9 @@ class BentoSchema(ResourceSchema):
     presigned_download_url: str
     manifest: BentoManifestSchema
 
-    presigned_urls_deprecated: Optional[bool] = attr.field(default=None)
+    transmission_strategy: TransmissionStrategy | None = attr.field(default=None)
```
aarnphm, line 182:
```suggestion
    transmission_strategy: t.Optional[TransmissionStrategy] = attr.field(default=None)
```

bentoml/_internal/yatai_rest_api_client/schemas.py
```diff
@@ -175,7 +179,9 @@ class BentoSchema(ResourceSchema):
...
+    upload_id: str | None = attr.field(default=None)
```
aarnphm, line 183:
```suggestion
    upload_id: t.Optional[str] = attr.field(default=None)
```

bentoml/_internal/yatai_rest_api_client/schemas.py
```diff
@@ -164,6 +165,9 @@ class BentoManifestSchema:
     runners: Optional[List[BentoRunnerSchema]] = attr.field(factory=list)
 
 
+TransmissionStrategy = Literal["presigned_url", "proxy"]
```
aarnphm, line 168: To fit with runtime type checking for attrs.
```suggestion
if TYPE_CHECKING:
    TransmissionStrategy = t.Literal["presigned_url", "proxy"]
else:
    TransmissionStrategy = str
```

bentoml/_internal/yatai_rest_api_client/schemas.py
aarnphm, line 7:
```suggestion
```

yetone: @aarnphm Do we support <3.7? But we use Literal a lot in our code.

https://github.com/bentoml/BentoML/blob/main/bentoml/grpc/types.py#L81

aarnphm: Sorry i meant <3.8

bentoml_cli/models.py
```diff
@@ -244,9 +244,15 @@ def pull(model_tag: str, force: bool):  # type: ignore (not accessed)
...
         help="Forced push to yatai even if it exists in yatai",
     )
-    def push(model_tag: str, force: bool):  # type: ignore (not accessed)
+    @click.option(
+        "-t",
+        "--threads",
+        default=10,
+        help="Number of threads to use for upload",
+    )
+    def push(model_tag: str, force: bool, threads: int = 10):  # type: ignore (not accessed)
```
aarnphm, line 253: Ah I think you don't have to specify the defaults here, since you already pass it to `click.option`.
```suggestion
    def push(model_tag: str, force: bool, threads: int):  # type: ignore (not accessed)
```

bentoml/_internal/yatai_rest_api_client/schemas.py
yetone: @aarnphm Understand, which are just annotation

aarnphm: hence the block below so that we don't actually import `t.Literal` in runtime

aarnphm (dismissed):
LGTM

aarnphm (approved):

aarnphm (approved):

PR #3047, opened 2022-09-28
fix(containerize): kaniko support

aarnphm:
Move the syntax version to 1.2 to be compatible with kaniko support.


bentoml/_internal/bento/docker/templates/base.j2
```diff
@@ -2,7 +2,9 @@
 {# users can use these values #}
 {% import '_macros.j2' as common %}
 {% set bento__entrypoint = bento__entrypoint | default(expands_bento_path("env", "docker", "entrypoint.sh", bento_path=bento__path)) %}
-# syntax = docker/dockerfile:1.4.3
+# 1.2.1 is the current docker frontend that both buildkitd and kaniko supports.
+# Think of the syntax frontend as a IDL parser.
```
aarnphm, line 6:
```suggestion
```

PR #3050, opened 2022-09-29
feat: cascade health checks to runner

quandollar:
- Added `is_ready` method to `RunnerHandle`'s `init` file, then implemented it in the local and remote runners
- Added new method to `Runner` to check that `RunnerHandle` is ready
- Added new method to HTTP service app's `on_startup` method to check the Runner in loop until it returns true

### DIFF
--- a/src/bentoml/_internal/configuration/containers.py
+++ b/src/bentoml/_internal/configuration/containers.py
@@ -171,7 +171,7 @@ SCHEMA = Schema(
                 "max_message_length": Or(int, None),
                 "maximum_concurrent_rpcs": Or(int, None),
             },
-            "runner_readiness_probe": {"check_runners": bool},
+            "health_probe": {"check_runners": bool},
         },
         "runners": {
             **RUNNER_CFG_SCHEMA,
--- a/src/bentoml/_internal/configuration/containers.py
+++ b/src/bentoml/_internal/configuration/containers.py
@@ -171,7 +171,11 @@ SCHEMA = Schema(
                 "max_message_length": Or(int, None),
                 "maximum_concurrent_rpcs": Or(int, None),
             },
-            "health_probe": {"check_runners": bool},
+            "runner_probe": {
+                "enabled": bool,
+                "timeout": int,
+                "period": int,
+            },
         },
         "runners": {
             **RUNNER_CFG_SCHEMA,
--- a/src/bentoml/_internal/configuration/containers.py
+++ b/src/bentoml/_internal/configuration/containers.py
@@ -171,6 +171,7 @@ SCHEMA = Schema(
                 "max_message_length": Or(int, None),
                 "maximum_concurrent_rpcs": Or(int, None),
             },
+            "runner_readiness_probe": {"check_runners": bool},
         },
         "runners": {
             **RUNNER_CFG_SCHEMA,
--- a/src/bentoml/_internal/configuration/default_configuration.yaml
+++ b/src/bentoml/_internal/configuration/default_configuration.yaml
@@ -38,6 +38,8 @@ api_server:
     metrics:
       host: 0.0.0.0
       port: 3001
+  runner_readiness_probe:
+    check_runners: True
 
 runners:
   batching:
--- a/src/bentoml/_internal/configuration/default_configuration.yaml
+++ b/src/bentoml/_internal/configuration/default_configuration.yaml
@@ -38,8 +38,10 @@ api_server:
     metrics:
       host: 0.0.0.0
       port: 3001
-  health_probe:
-    check_runners: True
+  runner_probe: # configure whether the API server's health check endpoints (readyz, livez, healthz) also check the runners
+    enabled: true
+    timeout: 1
+    period: 10
 
 runners:
   batching:
--- a/src/bentoml/_internal/configuration/default_configuration.yaml
+++ b/src/bentoml/_internal/configuration/default_configuration.yaml
@@ -38,7 +38,7 @@ api_server:
     metrics:
       host: 0.0.0.0
       port: 3001
-  runner_readiness_probe:
+  health_probe:
     check_runners: True
 
 runners:
--- a/src/bentoml/_internal/runner/runner.py
+++ b/src/bentoml/_internal/runner/runner.py
@@ -5,6 +5,7 @@ import logging
 from typing import TYPE_CHECKING
 
 import attr
+from simple_di import inject
 from simple_di import Provide
 
 from ..tag import validate_tag_str
@@ -241,6 +242,7 @@ class Runner:
     def destroy(self):
         object.__setattr__(self, "_runner_handle", DummyRunnerHandle())
 
+    @inject
     async def runner_handle_is_ready(
         self,
         timeout: int = Provide[BentoMLContainer.api_server_config.runner_probe.timeout],
--- a/src/bentoml/_internal/runner/runner.py
+++ b/src/bentoml/_internal/runner/runner.py
@@ -5,6 +5,7 @@ import logging
 from typing import TYPE_CHECKING
 
 import attr
+from simple_di import Provide
 
 from ..tag import validate_tag_str
 from ..utils import first_not_none
--- a/src/bentoml/_internal/runner/runner.py
+++ b/src/bentoml/_internal/runner/runner.py
@@ -240,6 +240,9 @@ class Runner:
     def destroy(self):
         object.__setattr__(self, "_runner_handle", DummyRunnerHandle())
 
+    async def runner_handle_is_ready(self) -> bool:
+        return await self._runner_handle.is_ready()
+
     @property
     def scheduled_worker_count(self) -> int:
         return self.scheduling_strategy.get_worker_count(
--- a/src/bentoml/_internal/runner/runner.py
+++ b/src/bentoml/_internal/runner/runner.py
@@ -241,8 +241,11 @@ class Runner:
     def destroy(self):
         object.__setattr__(self, "_runner_handle", DummyRunnerHandle())
 
-    async def runner_handle_is_ready(self) -> bool:
-        return await self._runner_handle.is_ready()
+    async def runner_handle_is_ready(
+        self,
+        timeout: int = Provide[BentoMLContainer.api_server_config.runner_probe.timeout],
+    ) -> bool:
+        return await self._runner_handle.is_ready(timeout)
 
     @property
     def scheduled_worker_count(self) -> int:
--- a/src/bentoml/_internal/server/grpc_app.py
+++ b/src/bentoml/_internal/server/grpc_app.py
@@ -89,11 +89,6 @@ class GRPCAppFactory:
         on_startup.append(self.wait_for_runner_ready)
         return on_startup
 
-        raise Exception(
-            "Runners %s failed to be ready within %s seconds"
-            % (self.bento_service.runners, timeout)
-        )
-
     @property
     def on_shutdown(self) -> list[t.Callable[[], None]]:
         on_shutdown = [self.bento_service.on_grpc_server_shutdown]
--- a/src/bentoml/_internal/server/grpc_app.py
+++ b/src/bentoml/_internal/server/grpc_app.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import time
 import typing as t
 import asyncio
 import logging
@@ -21,7 +20,7 @@ if TYPE_CHECKING:
     from ..service import Service
     from .grpc.servicer import Servicer
 
-    OnStartup = list[t.Callable[[], None | t.Coroutine[t.Any, t.Any, None]]]
+    OnStartup = list[t.Callable[[], t.Union[None, t.Coroutine[t.Any, t.Any, None]]]]
 
 
 class GRPCAppFactory:
@@ -51,28 +50,27 @@ class GRPCAppFactory:
         check_interval: int = 5,
     ):
         if BentoMLContainer.api_server_config.health_probe.check_runners.get():
-            logger.debug(
-                "Waiting for runners {!r} to be ready...".format(
-                    self.bento_service.runners
-                )
+            logger.info(
+                "Waiting for runners %r to be ready...", self.bento_service.runners
             )
 
             while True:
                 try:
-                    if all(
-                        await asyncio.gather(
-                            *(
-                                runner.runner_handle_is_ready()
-                                for runner in self.bento_service.runners
-                            )
-                        )
-                    ):
+                    runner_statuses = (
+                        runner.runner_handle_is_ready()
+                        for runner in self.bento_service.runners
+                    )
+
+                    runners_ready = all(await asyncio.gather(*runner_statuses))
+
+                    if runners_ready:
                         break
-                    else:
-                        time.sleep(check_interval)
                 except ConnectionError as e:
-                    logger.debug("[%s] Retrying ..." % e)
-                    time.sleep(check_interval)
+                    logger.debug("[%s] Retrying ...", e)
+
+                await asyncio.sleep(check_interval)
+
+            logger.info("All runners ready; continuing initialization.")
 
     @property
     def on_startup(self) -> OnStartup:
--- a/src/bentoml/_internal/server/grpc_app.py
+++ b/src/bentoml/_internal/server/grpc_app.py
@@ -23,11 +23,6 @@ if TYPE_CHECKING:
     from .grpc.servicer import Servicer
 
     OnStartup = list[t.Callable[[], None | t.Coroutine[t.Any, t.Any, None]]]
-    from grpc import aio
-else:
-    from bentoml.grpc.utils import import_grpc
-
-    _, aio = import_grpc()
 
 
 class GRPCAppFactory:
@@ -74,7 +69,7 @@ class GRPCAppFactory:
                     break
                 else:
                     time.sleep(check_interval)
-            except (ConnectionError, socket.timeout, aio.AioRpcError) as e:
+            except (ConnectionError, socket.timeout) as e:
                 logger.debug("[%s] Retrying ..." % e)
                 time.sleep(check_interval)
         raise RuntimeError(
--- a/src/bentoml/_internal/server/grpc_app.py
+++ b/src/bentoml/_internal/server/grpc_app.py
@@ -47,9 +47,11 @@ class GRPCAppFactory:
     async def wait_for_runner_ready(
         self,
         *,
-        check_interval: int = 5,
+        check_interval: int = Provide[
+            BentoMLContainer.api_server_config.runner_probe.period
+        ],
     ):
-        if BentoMLContainer.api_server_config.health_probe.check_runners.get():
+        if BentoMLContainer.api_server_config.runner_probe.enabled.get():
             logger.info(
                 "Waiting for runners %r to be ready...", self.bento_service.runners
             )
--- a/src/bentoml/_internal/server/grpc_app.py
+++ b/src/bentoml/_internal/server/grpc_app.py
@@ -62,7 +62,6 @@ class GRPCAppFactory:
                         runner.runner_handle_is_ready()
                         for runner in self.bento_service.runners
                     )
-
                     runners_ready = all(await asyncio.gather(*runner_statuses))
 
                     if runners_ready:
--- a/src/bentoml/_internal/server/grpc_app.py
+++ b/src/bentoml/_internal/server/grpc_app.py
@@ -23,6 +23,11 @@ if TYPE_CHECKING:
     from .grpc.servicer import Servicer
 
     OnStartup = list[t.Callable[[], None | t.Coroutine[t.Any, t.Any, None]]]
+    from grpc import aio
+else:
+    from bentoml.grpc.utils import import_grpc
+
+    _, aio = import_grpc()
 
 
 class GRPCAppFactory:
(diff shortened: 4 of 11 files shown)


bentoml/_internal/server/http_app.py
```diff
@@ -264,6 +264,8 @@ def on_startup(self) -> list[t.Callable[[], None]]:
         else:
             for runner in self.bento_service.runners:
                 on_startup.append(runner.init_client)
+        for runner in self.bento_service.runners:
+            asyncio.get_event_loop().run_until_complete(runner.runner_handle_is_ready())
```
sauyon, line 268: This doesn't actually check the result of the `is_ready` call, it just runs it.

You'll probably want to have a separate async function to call this in a loop instead; maybe using `asyncio.gather` to run all the requests to the runners in parallel.

quandollar: Thanks again. I fixed this section in the latest commits.

sauyon: 👍

bentoml/_internal/server/http_app.py
```diff
@@ -264,6 +264,15 @@ def on_startup(self) -> list[t.Callable[[], None]]:
...
             for runner in self.bento_service.runners:
                 on_startup.append(runner.init_client)
+
+        async def wait_for_runner_ready():
+            ready_status = False
+            while not ready_status:
+                ready_status = all(await asyncio.gather(*(runner.runner_handle_is_ready() for runner in self.bento_service.runners)))
+                await asyncio.sleep(0.1)
+        
+        on_startup.append(lambda: asyncio.run(wait_for_runner_ready()))
```
aarnphm, line 274: This won't work since you are calling `asyncio.run` inside a running event loops
```python
        async def wait_runner_ready() -> None:
            ready_status = False
            while not ready_status:
                ready_status = all(
                    await asyncio.gather(
                        *(
                            runner.runner_handle_is_ready()
                            for runner in self.bento_service.runners
                        )
                    )
                )

        on_startup.append(
            lambda: asyncio.ensure_future(
                wait_runner_ready(), loop=asyncio.get_running_loop()
            )
        )
```

aarnphm, line 274: This won't work since you are calling `asyncio.run` inside a running event loop.
```suggestion
        async def wait_runner_ready() -> None:
            ready_status = False
            while not ready_status:
                ready_status = all(
                    await asyncio.gather(
                        *(
                            runner.runner_handle_is_ready()
                            for runner in self.bento_service.runners
                        )
                    )
                )

        on_startup.append(
            lambda: asyncio.ensure_future(
                wait_runner_ready(), loop=asyncio.get_running_loop()
            )
        )
```

aarnphm, line 274: This won't work because we are already inside a running event loop
```suggestion
        on_startup.append(
            lambda: asyncio.ensure_future(
                wait_runner_ready(), loop=asyncio.get_running_loop()
            )
        )
```
I don't think you need to sleep on L272 either.

quandollar: Thanks. I implemented the suggested change and got the below message from VS Code. I'm looking into this but wanted to share in case you have any thoughts.
```
Argument of type "() -> Task[None]" cannot be assigned to parameter "__object" of type "() -> None" in function "append"
  Type "() -> Task[None]" cannot be assigned to type "() -> None"
    Function return type "Task[None]" is incompatible with type "None"
```

I'll remove sleep on line 272.

sauyon: We should probably sleep so as not to send a bunch of spam packets until the runner is ready; honestly I'd think about increasing it to something like 5s to mirror the kubernetes ready check intervals.

sauyon: Re: `ensure_future`: it should be `get_event_loop().run_until_complete(...)`; I was wrong about the deprecation, that's just outside an event loop.

aarnphm: `run_until_complete` won't work since the loop is already runner.

sauyon: Oh, our type annotations are just wrong and this should work fine:
```suggestion
        on_startup.append(wait_for_runner_ready)
```

We should also extend the `on_startup` types to allow coroutines as well as functions.

Is there actually a way to run an async function inside of a non-async function inside a running event loop?

Also, why is it that we aren't failing e2e tests?

aarnphm: you have to call `run_sync`

aarnphm: re for the question, i don't believe there is such thing

quandollar: @sauyon
1. Could you give me pointer on how to properly extend `on_startup` to return multiple types? It looks like [Union](https://docs.python.org/3/library/typing.html#typing.Union) is the solution. I imported `Union` and tried `def on_startup(self) -> Union[list[t.Callable[[], None]], Coroutine]` but got the following message:
```Declared return type, "list[() -> None] | Coroutine", is partially unknown```.

2. It looks like after we do the above, we also need to do the same for `on_startup` in `base_app.py`.

3. Side question on how typing is implemented in the code base: In ` -> list[t.Callable[[], None]]`, I understand that `t.Callable[[], None]` allows us to return a function, but what does `list[]` do? Does it allow us to return a list of functions?

@aarnphm 
Confirming that you were referring to `run_sync` from [Tornado](https://www.tornadoweb.org/en/stable/ioloop.html)? If so, should we be concerned about this docs' statement that "The IOLoop interface is now provided primarily for backwards compatibility; new code should generally use the asyncio event loop interface directly."?

Would the proper implementation of `run_sync` look something like this?
```
import tornado.ioloop

on_startup.append(ioloop.IOLoop.run_sync(wait_for_runner_ready))
```

File: bentoml/_internal/server/grpc_app.py [hunk 3]
```diff
@@ -52,7 +55,49 @@ def on_startup(self) -> OnStartup:
...
+                )
+
+        import yaml
+
+        start_time = time.time()
+        print("Waiting for runners %s to be ready.." % self.bento_service.runners)
+        with open(
+            "bentoml/_internal/configuration/default_configuration.yaml", "r"
+        ) as f:
+            timeout = yaml.load(f)["runners"]["timeout"]
```
aarnphm, line 77: get timeout via `BentoMLContainer.runners.timeout`
```python
from simple_di import Provide, inject

...
@inject
def on_startup(self, timeout: int = Provide[BentoMLContainer.runners.timeout]):
...
```

Also on_startup should be sync not async.

bentoml/_internal/server/grpc_app.py
```diff
@@ -52,7 +55,49 @@ def on_startup(self) -> OnStartup:
...
+                ):
+                    break
+                else:
+                    time.sleep(5)
+            except (ConnectionError, socket.timeout):
+                print("Retrying ...")
+                time.sleep(5)
+            else:
+                on_startup.append(wait_for_runner_ready)
+                return on_startup
```
aarnphm, line 96: You can define a `check_interval` variable here, and put it in the parameter and we can add a separate fields for user to configure this in the future.

bentoml/_internal/server/http_app.py
```diff
@@ -264,9 +267,26 @@ def on_startup(self) -> list[t.Callable[[], None]]:
...
+                    runner.runner_handle_is_ready()
+                    for runner in self.bento_service.runners
+                )
+            )
+        )
+
+        if ready_status:
+            return PlainTextResponse("\n", status_code=200)
+        else:
+            raise HTTPException(500)
```
aarnphm, line 288: For this I think you can just put the ready_status logic into the if branch, and no need to have an else here, you can just raise directly after the if block:
```python
if all(...):
   return PlainTextResponse
raise ...
```

bentoml/_internal/server/grpc_app.py
```diff
@@ -52,8 +86,14 @@ def on_startup(self) -> OnStartup:
...
                 on_startup.append(runner.init_client)
 
+        on_startup.append(self.wait_for_runner_ready)
         return on_startup
 
+        raise Exception(
+            "Runners %s failed to be ready within %s seconds"
+            % (self.bento_service.runners, timeout)
+        )
+
```
aarnphm, line 96:
```suggestion
```

bentoml/_internal/server/http_app.py
quandollar: Thanks Aaron. That does look cleaner. Just pushed.

bentoml/_internal/server/http_app.py
```diff
@@ -272,17 +274,32 @@ def on_startup(self) -> list[t.Callable[[], None]]:
...
-            await asyncio.gather(
-                *(
-                    runner.runner_handle_is_ready()
-                    for runner in self.bento_service.runners
-                )
-            )
-        ):
-            return PlainTextResponse("\n", status_code=200)
+        runner_timeout = BentoMLContainer.runners_config.timeout.get()
+        check_interval = 5
```
aarnphm, line 278: note: allow user to configure check_interval in configuration @aarnphm

bentoml/_internal/server/http_app.py
```diff
@@ -272,17 +274,32 @@ def on_startup(self) -> list[t.Callable[[], None]]:
...
+                else:
+                    time.sleep(check_interval)
+            except (ConnectionError, socket.timeout, HTTPException) as e:
+                logger.debug("[%s] Retrying ..." % e)
+                time.sleep(check_interval)
 
-        raise HTTPException(500)
+        raise RuntimeError(
+            f"Timed out waiting {runner_timeout} seconds for runners to be ready."
+        )
```
aarnphm, line 302: I think we can return HTTPException here instead of raising RuntimeError

bentoml/_internal/server/grpc_app.py
```diff
@@ -52,6 +86,7 @@ def on_startup(self) -> OnStartup:
             for runner in self.bento_service.runners:
                 on_startup.append(runner.init_client)
 
+        on_startup.append(self.wait_for_runner_ready)
```
sauyon, line 89: @aarnphm are we leaving this this way for now?

bentoml/_internal/server/grpc_app.py
```diff
@@ -42,6 +45,37 @@ def __init__(
...
+                    )
+                ):
+                    break
+                else:
+                    time.sleep(check_interval)
+            except (ConnectionError, socket.timeout) as e:
+                logger.debug("[%s] Retrying ..." % e)
+                time.sleep(check_interval)
+        raise RuntimeError(
+            f"Timed out waiting {runner_timeout} seconds for runners to be ready."
```
sauyon, line 76: Why are we using runner timeout for the startup timeout? I don't think we should timeout by default, but if we do, the timeout for this and runner requests should definitely be separate fields.

bentoml/_internal/server/http_app.py
quandollar: Note that this code diff is outdated. As of commit `26a349b`, I added `HTTPException` to the try-catch block's exceptions -- Let me know if that looks better. 

My understanding was that we should leave `RuntimeError` in here in case the runners take too long for some reason. Could you clarify for me why we should remove it?

bentoml/_internal/server/grpc_app.py
quandollar: I'll let @aarnphm chime in since he authored this and can speak to it a lot better than I can.

@sauyon For my learning, what do you suggest to be the behavior instead of "timeout by default"? And could you clarify "timeout for this and runner requests should definitely be separate fields" for me?

aarnphm: I think timeout for this make sense, as if the runners failed to start, we should just raise RuntimeError or at least return a FAILED_PRECONDITION

quandollar: Per our discussion, I removed timeout in a recent commit.

bentoml/_internal/runner/runner_handle/remote.py
```diff
@@ -217,7 +217,10 @@ def run_method(
         )
 
     async def is_ready(self) -> bool:
-        async with self._client.get(f"{self._addr}/readyz") as resp:
+        import aiohttp
+
+        timeout = aiohttp.ClientTimeout(total=5)
```
aarnphm, line 222: This is 5s, not 5m?

quandollar: Thanks for reviewing, Aaron. I did mean to write 5s as I recall Seoyeon saying that the timeout doesn't need to be long, but I could be wrong...@sauyon Could you confirm?

aarnphm: I think timeout should fit with k8s timeout @sauyon ?

quandollar: Right. Then I think we should be good to go since k8s performs a liveness probe every 5 seconds per [their docs](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/).

tests/e2e/bento_server_http/tests/test_meta.py
```diff
@@ -26,7 +26,7 @@ async def test_api_server_meta(host: str) -> None:
     status, _, _ = await async_request("GET", f"http://{host}/docs.json")
     assert status == 200
     status, _, _ = await async_request("GET", f"http://{host}/readyz")
-    assert status == 200
+    assert status == 500
```
quandollar, line 29: This is wrong. I need to wait first for runners to get ready and not just change it to 500. Working on this fix.

tests/e2e/bento_server_http/tests/test_meta.py
```diff
@@ -26,8 +27,9 @@ async def test_api_server_meta(host: str) -> None:
...
     status, _, body = await async_request("GET", f"http://{host}/metrics")
+    await asyncio.sleep(20)
```
aarnphm, line 32: I suggest to make a separate test, call it `test_readiness` and remove the async_request to readyz
```python
@pytest.mark.asyncio
async def test_readiness(host: str):
    await async_request("GET", f"http://{host}/readyz", assert_status=500)
    await asyncio.sleep(10)
    status, _, _ = await async_request("GET", f"http://{host}/readyz")
    assert status == 200
```

quandollar: Thanks for catching that, Aaron. I'll make the update.

bentoml/_internal/configuration/default_configuration.yaml
```diff
@@ -38,6 +38,8 @@ api_server:
     metrics:
       host: 0.0.0.0
       port: 3001
+  runner_readiness_probe:
```
ssheng, line 41: We can just call this `readyz` maybe?

bentoml/_internal/configuration/default_configuration.yaml
```diff
@@ -38,6 +38,8 @@ api_server:
...
+    check_runners: True
```
ssheng, line 42: Should we have a configurable timeout?

sauyon: Plan is to do this in a separate PR; spoke to Quan and he'll open an issue!

bentoml/_internal/configuration/default_configuration.yaml
quandollar: I checked in with @sauyon and we'll keep this name more broad, like `health_probe` since the check isn't named `readyz` on the grpc side.

bentoml/_internal/configuration/default_configuration.yaml
quandollar: Thanks, Seoyeon. I'll add changes to this PR per our chat and open an issue for this after.

aarnphm: nit:
```yaml
probe:
  health:
   ...
```

bentoml/_internal/runner/runner_handle/remote.py
```diff
@@ -216,5 +216,12 @@ def run_method(
             ),
         )
 
+    async def is_ready(self) -> bool:
+        import aiohttp
+
+        timeout = aiohttp.ClientTimeout(total=5)
```
aarnphm, line 222: Good to have a note up here link to the k8s timeout

tests/e2e/bento_server_http/tests/test_meta.py
```diff
@@ -25,15 +27,26 @@ async def test_api_server_meta(host: str) -> None:
...
+async def test_runner_readiness(host: str) -> None:
+    timeout = 300
+    start_time = time.time()
+    status = ""
+    while (time.time() - start_time) < timeout:
+        status, _, _ = await async_request("GET", f"http://{host}/readyz")
+        await asyncio.sleep(20)
+        if status == 200:
+            break
+    assert status == 200
```
aarnphm, line 47: I think you can also test the exception by just sending a request right after, and see whether it returns 500

I believe you don't have to do a while loop, shouldn't asyncio.sleep be sufficient enough?

quandollar: @aarnphm Right, so this is the part that's puzzling me. Seoyeon and I discussed adding a while loop because with asyncio.sleep alone, the request still returns 500. Even after adding the while loop, it still returns 500 meaning the timeout of 300 was reached. I wonder if there's something preventing all the runners from becoming ready.

bentoml/_internal/configuration/default_configuration.yaml
quandollar: Given your different suggestions, I'll leave the config as-is (currently `health_probe:` `check_runners:`) for now.

Added an issue to make the timeout configurable [here](https://github.com/bentoml/BentoML/issues/3077).

quandollar:
@yetone It's nice to meet you! I recently joined BentoML as an MLH fellow/intern and am working on this PR to add the runner readiness probe feature. Seoyeon suggested that I reach out to you -- Could I get your help disabling this probe on the yatai side?

src/bentoml/_internal/configuration/default_configuration.yaml
```diff
@@ -38,6 +38,8 @@ api_server:
     metrics:
       host: 0.0.0.0
       port: 3001
+  health_probe:
+    check_runners: True
```
aarnphm, line 42: very nit pick:
```yaml
```suggestion
  probe:
    health:
      check_runners: true
```
```

File: src/bentoml/_internal/configuration/containers.py [hunk 19]
```diff
@@ -171,6 +171,7 @@ def _is_ip_address(addr: str) -> bool:
                 "max_message_length": Or(int, None),
                 "maximum_concurrent_rpcs": Or(int, None),
             },
+            "health_probe": {"check_runners": bool},
```
aarnphm, line 174: same with configuration here.

src/bentoml/_internal/server/grpc_app.py
```diff
@@ -42,6 +43,35 @@ def __init__(
         self.bento_service = bento_service
         self.enable_metrics = enable_metrics
 
+    @inject
+    async def wait_for_runner_ready(
+        self,
+        *,
+        check_interval: int = 5,
+    ):
```
aarnphm, line 51:
```suggestion
    async def wait_for_runner_ready(self, *, check_interval: int = 5):
```

src/bentoml/_internal/server/http_app.py
```diff
@@ -267,6 +269,19 @@ def on_startup(self) -> list[t.Callable[[], None]]:
         on_startup.extend(super().on_startup)
         return on_startup
 
+    async def readyz(self, _: "Request") -> "Response":
+        if BentoMLContainer.api_server_config.health_probe.check_runners.get():
+            runner_statuses = (
+                runner.runner_handle_is_ready() for runner in self.bento_service.runners
+            )
+
```
aarnphm, line 277: remove whitespace :)

src/bentoml_cli/worker/http_api_server.py
```diff
@@ -113,6 +113,11 @@ def main(
     component_context.component_index = worker_id
     configure_server_logging()
 
+    if worker_id is None:
+        # worker ID is not set; this server is running in standalone mode
+        # and should not be concerned with the status of its runners
+        BentoMLContainer.config.health_probe.check_runners.set(False)
```
aarnphm, line 119: Should also include this for grpc_api_server.py

aarnphm (commented):
Some final comments. Otherwise gtg.

src/bentoml/_internal/configuration/default_configuration.yaml
quandollar: Seoyeon and I discussed and went a different route. Let us know if you have any questions.

src/bentoml/_internal/server/grpc_app.py
```diff
@@ -42,6 +43,37 @@ def __init__(
...
+            )
+
+            while True:
+                try:
+                    runner_statuses = (
+                        runner.runner_handle_is_ready()
+                        for runner in self.bento_service.runners
+                    )
+
+                    runners_ready = all(await asyncio.gather(*runner_statuses))
```
sauyon, line 66: For consistency.

```suggestion
                    )
                    runners_ready = all(await asyncio.gather(*runner_statuses))
```

src/bentoml/_internal/runner/runner.py
```diff
@@ -240,6 +241,12 @@ def init_client(self):
     def destroy(self):
         object.__setattr__(self, "_runner_handle", DummyRunnerHandle())
 
+    async def runner_handle_is_ready(
```
sauyon, line 244: We forgot to add this so `Provide` actually works.

```suggestion
    @inject
    async def runner_handle_is_ready(
```

src/bentoml/_internal/runner/runner.py
```diff
@@ -5,6 +5,7 @@
 from typing import TYPE_CHECKING
 
 import attr
+from simple_di import Provide
```
sauyon, line 8: For below.

```suggestion
from simple_di import inject
from simple_di import Provide
```

sauyon:
@aarnphm codecov is broken again lol

aarnphm (approved):

aarnphm (approved):

PR #3062, opened 2022-10-03
chore(deps): ensure grpc resolve dependencies correctly

aarnphm:
Dependencies resolution for pip-tools apparently will try to lock to the latest version
instead of depending on previous dependent packages.

The expected behaviour would be:

- grpcio == 1.42 -> (grpcio-health-checking should lock to 1.42)

However, this will lock grpcio-health-checking to latest.

This could be a bug, but imminent fix for this is to specify the boundary.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -112,8 +112,8 @@ grpc = [
   # since grpcio >=1.49.0 uses protbuf >=3.20.0, which breaks compatibility.
   # We can't use 1.48.2 since it depends on 3.19.5
   "grpcio>=1.41.0, <1.49, !=1.48.2",
-  "grpcio-health-checking",
-  "grpcio-reflection",
+  "grpcio-health-checking>=1.41.0, <1.49, !=1.48.2",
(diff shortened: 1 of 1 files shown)


aarnphm:
cc @sauyon

sauyon (approved):

sauyon:
Wait, why are we locking grpcio?

aarnphm:
> Wait, why are we locking grpcio?

We don't. The dependency resolver ignores grpcio version. Maybe we should file a bug to pip.

for example: if grpcio==1.42.0 -> grpcio-health-checking should be 1.42.0 by default (since grpcio-health-checking depends on grpcio). However, in this case, grpcio-health-checking for some reason is then set to the latest version.

I will look into this, but for now this is the immediate patch.

sauyon:
Does the dependency resolver ignore `grpcio==1.42.0`?

sauyon (commented):
What does protobuf>3.20 break again?

aarnphm:
protobuf > 3.20 breaks everything, because it try to install protobuf 4 instead of 3

PR #3068, opened 2022-10-05
fix(pytest): export plugin lazily via environment variables

aarnphm:
Instead of exporting pytest plugins via setuptools, we should do it via PYTEST_PLUGINS
to avoid interfere with users who already using pytest

Using environment variables also unblocks https://bentoml.slack.com/archives/CKRANBHPH/p1665006750560559


aarnphm:
cc @sauyon

sauyon (approved):

PR #3069, opened 2022-10-05
chore: move startup after health checker is ready

aarnphm:
Move user on_startup to after health checking servicer

### DIFF
--- a/bentoml/_internal/server/grpc/server.py
+++ b/bentoml/_internal/server/grpc/server.py
@@ -98,8 +98,6 @@ class Server:
     async def startup(self) -> None:
         from bentoml.exceptions import MissingDependencyException
 
-        # Running on_startup callback.
-        await self.servicer.startup()
         # register bento servicer
         services.add_BentoServiceServicer_to_server(
             self.servicer.bento_servicer, self.server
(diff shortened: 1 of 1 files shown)


aarnphm:
cc @sauyon

bentoml/_internal/server/grpc/server.py
```diff
@@ -134,6 +132,9 @@ async def startup(self) -> None:
             await self.servicer.health_servicer.set(
                 service, pb_health.HealthCheckResponse.SERVING  # type: ignore (no types available)
             )
+        # Running on_startup callback.
+        await self.servicer.startup()
```
sauyon, line 136: Do we want to do startup before or after the user services are added? I feel like we still want to run startup code before user services can be called.

aarnphm: I think we should run before, but that means the readyz will also be called before adding user service.

sauyon: Sure? I don't think we need to put those services up beforehand.

But maybe it's more correct to move the wait outside of on_startup anyway.

aarnphm: maybe the wait can just live here instead of on startup?

aarnphm: I think we can close this, and then move the readiness from runner to here?

PR #3074, opened 2022-10-06
infra: Bazel

aarnphm:
Bazel.


aarnphm:
I will address MLflow PR separately

aarnphm:
TODO: setup FOSSA_API_KEY in secrets.

bojiang (changes requested):
It looks great, but greatness is not the only thing we need to think about to accept a PR. I have concerns about the maintenance.

aarnphm (commented):
This PR can refactor into three smaller ones.
- one that includes all CI and bazel migration
- then the one that only concerns grpc-client refactoring based on that bazel PR and docs
- finally a PR that includes all of the generated reuqirements.txt for hermetic build.

sauyon:
There seems to be a lot of unrelated changes in this PR that would be useful without bazel, can we factor those out?

aarnphm:
I'm going to do that now.

src/bentoml/_internal/frameworks/pytorch.py
```diff
@@ -75,9 +75,7 @@ def load_model(
 
     weight_file = bentoml_model.path_of(MODEL_FILENAME)
     with Path(weight_file).open("rb") as file:
-        model: "torch.nn.Module" = torch.load(
-            file, map_location=device_id, pickle_module=cloudpickle
-        )
+        model = torch.load(file, map_location=device_id)
```
aarnphm, line 78: @bojiang I dont' think we can use cloudpickle here, since pytorch actually use `pickle_module.Unpickler` inside torch.load, which cloudpickle doesn' have a `Unpickler` module.

bojiang: @aarnphm which version of PyTorch are you using? It works in my test

aarnphm: im using the latest version.

PR #3075, opened 2022-10-07
fix(mlflow): cleanup if temporary models:/ exists

phildamore-phdata:
## What does this PR address?

In https://github.com/bentoml/BentoML/pull/2930 I submitted a fix for importing MLFlow models using a models:/ URI.  These imports are different, because the mlflow client does not add an extra new directory for the download.  This means that the directory which gets renamed to `mlflow_model` is incorrect and causes the import to explode.  To fix that I added an extra sacrificial temporary directory into which the model is downloaded (`download_dir` in the code).  For most cases, this directory is left behind and needs to be cleaned up.  For `models:/` imports, this directory gets renamed.  Therefore, the rmtree of that directory needs to be conditional or it will explode for `models:/` downloads.  Sometime beteween verions 1.0.5 and 1.0.6 this if condition was removed, re-breaking `models:/` downloads.

This PR re-fixes that, tries to comment it to describe why it's there, and adds a test to ensure the `models:/` imports don't break anymore.  For this to work I had to create a registered model version in the first test when the model is created in the first place.

      those accordingly? Here are [documentation guidelines](https://github.com/bentoml/BentoML/tree/main/docs) and [tips on writting docs](https://github.com/bentoml/BentoML/tree/main/docs#writing-documentation). (No updates needed)

### DIFF
--- a/bentoml/_internal/frameworks/mlflow.py
+++ b/bentoml/_internal/frameworks/mlflow.py
@@ -213,8 +213,12 @@ def import_model(
             mlflow_model_path = bento_model.path_of(MLFLOW_MODEL_FOLDER)
             # Rename model folder from original artifact name to fixed "mlflow_model"
             shutil.move(local_path, mlflow_model_path)  # type: ignore (local_path is bound)
-            # Remove the tempdir
-            shutil.rmtree(download_dir)
+            # Remove the tempdir if it still exists.
+            # NOTE for models:/ uri downloads, the download_dir itself is actually renamed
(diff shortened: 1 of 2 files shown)


aarnphm (commented):
I'm wondering if we should set up e2e tests for mlflow to test out all the import_model scenarios 🤔

aarnphm:
Also don't worry about the unit tests, that is a bug on our own test. I will have a PR to address this bug specifically

aarnphm (approved):

PR #3080, opened 2022-10-10
feat(grpc): channelz

aarnphm:
Add channelz support to allow easy live server tracing

This is useful to debug the server if user wish to debug the rpc.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -127,6 +127,7 @@ grpc = [
   "grpcio>=1.41.0, <1.49, !=1.48.2",
   "grpcio-health-checking>=1.41.0, <1.49, !=1.48.2",
   "grpcio-reflection>=1.41.0, <1.49, !=1.48.2",
+  "grpcio-channelz>=1.41.0, <1.49, !=1.48.2",
   "opentelemetry-instrumentation-grpc==0.33b0",
 ]
 # We kept for compatibility with previous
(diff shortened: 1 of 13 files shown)


src/bentoml_cli/worker/grpc_api_server.py
```diff
@@ -39,6 +39,13 @@
     is_flag=True,
     help="Enable reflection.",
 )
+@click.option(
+    "--enable-channelz",
+    type=click.BOOL,
+    is_flag=True,
+    help="Enable reflection.",
```
jjmachan, line 46:
```suggestion
    help="Enable Channelz. See https://github.com/grpc/proposal/blob/master/A14-channelz.md.",
```

pyproject.toml
```diff
@@ -111,6 +111,8 @@ all = [
   "bentoml[io-image]",
   "bentoml[io-pandas]",
   "bentoml[grpc]",
+  "bentoml[grpc-reflection]",
+  "bentoml[grpc-channelz]",
```
jjmachan, line 115: is it better to have all of grpc dependencies together? or will it be too large?

aarnphm: better to separate out, similar to how we approach tracing dependencies.

jjmachan (approved):

PR #3083, opened 2022-10-10
fix: flaky CI

aarnphm:
This PR ensures interceptors are registered correctly instead of MK


bojiang:
It seems that the e2e get unstable after https://github.com/bentoml/BentoML/pull/2984

sauyon (approved):

PR #3089, opened 2022-10-11
fix(prometheus): ensure metrics are lazily loaded

aarnphm:
This PR address the issue inside the container, where if the user service uses `bentoml.metrics`, the service is imported before `PROMETHEUS_MULTIPROC_DIR` is created.

### DIFF
--- a/src/bentoml/metrics.py
+++ b/src/bentoml/metrics.py
@@ -20,9 +20,9 @@ def __dir__() -> list[str]:
     return dir(metrics_client.prometheus_client)
 
 
-class LazyObject:
+class _LazyMetric:
     def __init__(self, attr: t.Any):
-        self.__bentoml_metrics_attr__ = attr
+        self._attr = attr
         self._proxy = None
         self._initialized = False
         self._args: tuple[t.Any, ...] = ()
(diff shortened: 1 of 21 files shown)


bojiang:
@aarnphm Why do we need this?

aarnphm:
When users use `bentoml.metrics` inside service.py, this would currently fail, since the multiproc dir is not yet initialized (inside container)

src/bentoml/metrics.py
```diff
@@ -13,8 +20,34 @@ def __dir__() -> list[str]:
...
+        self._kwargs: dict[str, t.Any] = {}
+
+    def __call__(self, *args: t.Any, **kwargs: t.Any) -> t.Any:
+        # This is where we lazy load the proxy object.
+        self._args = args
+        self._kwargs = kwargs
+        return self
+
+    def __getattr__(self, item: t.Any) -> t.Any:
+        if self._proxy is None:
```
bojiang, line 38:
```suggestion
    def __getattr__(self, item: t.Any) -> t.Any:
        if item in {"_attr", "_proxy", ...}:
           raise AttibuteError()
        if self._proxy is None:
```

src/bentoml/metrics.py
```diff
@@ -13,8 +20,36 @@ def __dir__() -> list[str]:
...
+    @inject
+    def _load_proxy(
+        self,
+        metrics_client: PrometheusClient = Provide[BentoMLContainer.metrics_client],
+    ) -> None:
+        self._proxy = getattr(metrics_client, self._attr)(*self._args, **self._kwargs)
+        self._initialized = True
+
+
 def __getattr__(item: t.Any):
```
bojiang, line 54: shall we check the typo for users

bojiang:
@aarnphm The test is still failling

bojiang:
It seems the e2e tests are skipped. Will continue the review when it's enabled

sauyon (approved):

PR #3096, opened 2022-10-14
perf: logging optimization

aarnphm:
Ensure that all logger call are using %s format

See https://docs.python.org/3.10/howto/logging.html#optimization

To prevent from changing this behaviour, the linter rules for logging lazy format is
enabled via pylint.


bojiang:
That would be a major improvement!

sauyon (approved):
👍

PR #3101, opened 2022-10-17
feat: allow nested `requirements.txt` files

### LINKED ISSUE #3102: feature: allow nested `requirements.txt` files
### Feature request

I would like to be able to have a `requirements.txt` file which references other requirements files with `-r`.

### Motivation

I would like to be able to you `pip-compile-multi` to manage the dependencies of my models. Often there are different dependencies at serve and train time with some overlap. Therefore, I'd like to be able to define the common dependencies in one file, the serving specific ones in another file, and the training specific ones in a third file.

ie.

**model.txt**
```
scikit-learn
```

**train.txt**
```
-r model.txt
pandas
```

**serve.txt**
```
-r model.txt
pydantic
```

### Other

_No response_

judahrand:
## What does this PR address?

This is arguably a bug fix? Not sure... somewhere between a bug fix and a feature... I'd expected this to 'just work'.

Fixes #3102

## Before submitting:

### DIFF
--- a/src/bentoml/_internal/bento/build_config.py
+++ b/src/bentoml/_internal/bento/build_config.py
@@ -9,6 +9,7 @@ import subprocess
 from sys import version_info
 from shlex import quote
 from typing import TYPE_CHECKING
+from pathlib import Path
 
 import fs
 import attr
@@ -22,6 +23,7 @@ from .gen import generate_dockerfile
 from ..utils import bentoml_cattr
 from ..utils import resolve_user_filepath
 from ..utils import copy_file_to_fs_folder
+from ..utils import find_nested_requirements_txts
 from .docker import DistroSpec
 from .docker import get_supported_spec
 from .docker import SUPPORTED_CUDA_VERSIONS
@@ -544,15 +546,25 @@ fi
             f.write(install_sh)
 
         if self.requirements_txt is not None:
-            requirements_txt_file = resolve_user_filepath(
-                self.requirements_txt, build_ctx
+            base_txt, nested_txts = find_nested_requirements_txts(
+                Path(
+                    resolve_user_filepath(self.requirements_txt, build_ctx),
+                )
             )
             copy_file_to_fs_folder(
-                requirements_txt_file,
+                str(base_txt.absolute()),
                 bento_fs,
                 py_folder,
                 dst_filename="requirements.txt",
             )
+            for txt in nested_txts:
+                dst_path = txt.relative_to(base_txt.parent)
+                copy_file_to_fs_folder(
+                    str(txt.absolute()),
+                    bento_fs,
+                    fs.path.combine(py_folder, str(dst_path.parent)),
+                    dst_filename=dst_path.name,
+                )
         elif self.packages is not None:
             with bento_fs.open(fs.path.join(py_folder, "requirements.txt"), "w") as f:
                 f.write("\n".join(self.packages))
--- a/src/bentoml/_internal/bento/build_config.py
+++ b/src/bentoml/_internal/bento/build_config.py
@@ -549,6 +549,14 @@ fi
                 resolve_user_filepath(self.requirements_txt, build_ctx),
                 include_nested=True,
             )
+            # We need to make sure that we don't write any file references
+            # back into the final `requirements.txt` file. We've already
+            # resolved them and included their contents so we can discard
+            # them.
+            for option_line in requirements_txt.options:
+                option_line.options.pop("constraints", None)
+                option_line.options.pop("requirements", None)
+
             with bento_fs.open(
                 fs.path.combine(py_folder, "requirements.txt"), "w"
             ) as f:
--- a/src/bentoml/_internal/bento/build_config.py
+++ b/src/bentoml/_internal/bento/build_config.py
@@ -9,7 +9,6 @@ import subprocess
 from sys import version_info
 from shlex import quote
 from typing import TYPE_CHECKING
-from pathlib import Path
 
 import fs
 import attr
@@ -18,12 +17,12 @@ import psutil
 import fs.copy
 from dotenv import dotenv_values  # type: ignore
 from pathspec import PathSpec
+from pip_requirements_parser import RequirementsFile
 
 from .gen import generate_dockerfile
 from ..utils import bentoml_cattr
 from ..utils import resolve_user_filepath
 from ..utils import copy_file_to_fs_folder
-from ..utils import find_nested_requirements_txts
 from .docker import DistroSpec
 from .docker import get_supported_spec
 from .docker import SUPPORTED_CUDA_VERSIONS
@@ -546,25 +545,14 @@ fi
             f.write(install_sh)
 
         if self.requirements_txt is not None:
-            base_txt, nested_txts = find_nested_requirements_txts(
-                Path(
-                    resolve_user_filepath(self.requirements_txt, build_ctx),
-                )
+            requirements_txt = RequirementsFile.from_file(
+                resolve_user_filepath(self.requirements_txt, build_ctx),
+                include_nested=True,
             )
-            copy_file_to_fs_folder(
-                str(base_txt.absolute()),
-                bento_fs,
-                py_folder,
-                dst_filename="requirements.txt",
-            )
-            for txt in nested_txts:
-                dst_path = txt.relative_to(base_txt.parent)
-                copy_file_to_fs_folder(
-                    str(txt.absolute()),
-                    bento_fs,
-                    fs.path.combine(py_folder, str(dst_path.parent)),
-                    dst_filename=dst_path.name,
-                )
+            with bento_fs.open(
+                fs.path.combine(py_folder, "requirements.txt"), "w"
+            ) as f:
+                f.write(requirements_txt.dumps(preserve_one_empty_line=True))
         elif self.packages is not None:
             with bento_fs.open(fs.path.join(py_folder, "requirements.txt"), "w") as f:
                 f.write("\n".join(self.packages))
--- a/src/bentoml/_internal/utils/__init__.py
+++ b/src/bentoml/_internal/utils/__init__.py
@@ -310,7 +310,7 @@ def find_nested_requirements_txts(
     """Recursively search the provided requirements.txt file to find nested
     requirements files specified with `-r`.
     """
-    sub_requirements_txt_files: t.Set[Path] = set()
+    sub_requirements_txt_files: set[Path] = set()
     with base_requirements_txt.open("r") as f:
         for line in f.readlines():
             if line.startswith("-r"):
--- a/src/bentoml/_internal/utils/__init__.py
+++ b/src/bentoml/_internal/utils/__init__.py
@@ -303,6 +303,24 @@ def resolve_user_filepath(filepath: str, ctx: t.Optional[str]) -> str:
     raise FileNotFoundError(f"file {filepath} not found")
 
 
+def find_nested_requirements_txts(
+    base_requirements_txt: Path,
+) -> t.Tuple[Path, t.Set[Path]]:
+    """Recursively search the provided requirements.txt file to find nested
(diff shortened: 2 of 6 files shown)


src/bentoml/_internal/utils/__init__.py
```diff
@@ -303,6 +304,24 @@ def resolve_user_filepath(filepath: str, ctx: t.Optional[str]) -> str:
     raise FileNotFoundError(f"file {filepath} not found")
 
 
+def find_nested_requirements_txts(
+    base_requirements_txt: Path,
+) -> t.Tuple[Path, t.Set[Path]]:
+    """Recursively search the provided requirements.txt file to find nested
+    requirements files specified with `-r`.
+    """
+    sub_requirements_txt_files: t.Set[Path] = set()
```
aarnphm, line 313:
```suggestion
    sub_requirements_txt_files: set[Path] = set()
```

src/bentoml/_internal/utils/__init__.py
```diff
@@ -303,6 +304,24 @@ def resolve_user_filepath(filepath: str, ctx: t.Optional[str]) -> str:
     raise FileNotFoundError(f"file {filepath} not found")
 
 
+def find_nested_requirements_txts(
+    base_requirements_txt: Path,
+) -> t.Tuple[Path, t.Set[Path]]:
```
aarnphm, line 309:
```suggestion
) -> tuple[Path, set[Path]]:
```

aarnphm (commented):
for this feature I feel that we should use a requirements parser https://github.com/sarugaku/requirementslib or https://github.com/nexB/pip-requirements-parser

src/bentoml/_internal/utils/__init__.py
```diff
@@ -279,6 +279,7 @@ def copy_file_to_fs_folder(
     src_fs = fs.open_fs(dir_name)
     dst_filename = file_name if dst_filename is None else dst_filename
     dst_path = fs.path.join(dst_folder_path, dst_filename)
+    dst_fs.makedir(dst_folder_path, recreate=True)
```
aarnphm, line 282: I don't think this is correct.

src/bentoml/_internal/bento/build_config.py
```diff
@@ -544,15 +546,25 @@ def write_to_bento(self, bento_fs: FS, build_ctx: str) -> None:
...
                 dst_filename="requirements.txt",
             )
+            for txt in nested_txts:
+                dst_path = txt.relative_to(base_txt.parent)
+                copy_file_to_fs_folder(
+                    str(txt.absolute()),
+                    bento_fs,
+                    fs.path.combine(py_folder, str(dst_path.parent)),
+                    dst_filename=dst_path.name,
+                )
```
aarnphm, line 567: I propose that for requirements.txt alone, we flatten all requirements into one requirements.txt.

src/bentoml/_internal/utils/__init__.py
judahrand: That's only valid syntax in Python 3.9+ isn't it?

src/bentoml/_internal/utils/__init__.py
judahrand: That's only valid syntax in Python 3.9+ isn't it?

tests/e2e/bento_server_http/bentofile.yaml
```diff
@@ -1,6 +1,6 @@
 service: service:svc
 python:
-  requirements_txt: ./requirements.txt
+  requirements_txt: ./requirements/service.txt
```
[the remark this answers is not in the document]
judahrand: Why?

judahrand:
> for this feature I feel that we should use a requirements parser https://github.com/sarugaku/requirementslib or https://github.com/nexB/pip-requirements-parser

Why? Do we need to introduce an additional dependency for this fix/feature? Do we gain anything?

judahrand:
Happy to be wrong on any of my questions but I'm not sure I understand your reasons for the proposed changes.

src/bentoml/_internal/utils/__init__.py
sauyon: Type annotations aren't evaluated with `from __future__ import annotations`, so we use the new syntax for type annotations.

src/bentoml/_internal/bento/build_config.py
```diff
@@ -544,15 +546,25 @@ def write_to_bento(self, bento_fs: FS, build_ctx: str) -> None:
...
             )
             copy_file_to_fs_folder(
-                requirements_txt_file,
+                str(base_txt.absolute()),
                 bento_fs,
                 py_folder,
                 dst_filename="requirements.txt",
             )
+            for txt in nested_txts:
+                dst_path = txt.relative_to(base_txt.parent)
```
sauyon, line 561: What does `dst_path` resolve to for `-r ../../requirements.txt`?

judahrand: It throws an exception. I should have noted somewhere that there is a requirement that nested files live at the same level or below the base file. To me this feels like a reasonable limitation.

src/bentoml/_internal/utils/__init__.py
aarnphm: we are using `from __future__` import

aarnphm:
> > for this feature I feel that we should use a requirements parser https://github.com/sarugaku/requirementslib or https://github.com/nexB/pip-requirements-parser
> 
> Why? Do we need to introduce an additional dependency for this fix/feature? Do we gain anything?

The reason for my proposal is that when user pass in their requirements file
```
--extra-index-url ...
```

Another contributor will also open a PR to support this feature in requirements.txt, and they might write their own parser. I think it is probably nicer if we use a parser and then that contributor can just use the same parser to add the feature.

judahrand:
> The reason for my proposal is that when user pass in their requirements file
> 
> ```
> --extra-index-url ...
> ```
> 
> Another contributor will also open a PR to support this feature in requirements.txt, and they might write their own parser. I think it is probably nicer if we use a parser and then that contributor can just use the same parser to add the feature.

For your example, it would just work wouldn't it? `--extra-index-url` doesn't introduce an external filesystem dependency. Are there other examples which would currently not work?

aarnphm:
> > for this feature I feel that we should use a requirements parser https://github.com/sarugaku/requirementslib or https://github.com/nexB/pip-requirements-parser
> 
> Why? Do we need to introduce an additional dependency for this fix/feature? Do we gain anything?

The reason for my proposal is that when user pass in their requirements file
```
--extra-index-url ...
```

Another contributor will also open a PR to support this feature in requirements.txt, and they might write their own parser. I think it is probably nicer if we use a parser, and that contributor can just use the same parser to add the feature.

aarnphm:
`--extra-index-url` parse the URL right, which is not related to filesystem dependency.

tests/e2e/bento_server_http/bentofile.yaml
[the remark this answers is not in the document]
aarnphm: tbh this is more of a unit test cases, but we didn't write any unit tests for build config yet.
This e2e tests are more for testing HTTP functionality of bento server, not build requirements necessarily.

judahrand:
> `--extra-index-url` parse the URL right, which is not related to filesystem dependency.

That's exactly my point. I'm not sure what other feature would be missing currently?

aarnphm:
Right now, we only support a very minimal requirements.txt, and hence pretty much all of the advanced features from requirements.txt https://pip.pypa.io/en/stable/reference/requirements-file-format/#requirements-file-format are not yet supported

src/bentoml/_internal/bento/build_config.py
sauyon: Yep, that makes sense to me. Can you catch that and throw a more user-friendly exception here that explains that we don't support it, then?

src/bentoml/_internal/utils/__init__.py
```diff
@@ -303,6 +304,24 @@ def resolve_user_filepath(filepath: str, ctx: t.Optional[str]) -> str:
...
+def find_nested_requirements_txts(
+    base_requirements_txt: Path,
+) -> tuple[Path, set[Path]]:
+    """Recursively search the provided requirements.txt file to find nested
+    requirements files specified with `-r`.
+    """
+    sub_requirements_txt_files: set[Path] = set()
+    with base_requirements_txt.open("r") as f:
+        for line in f.readlines():
+            if line.startswith("-r"):
```
sauyon, line 316: Maybe something like this so we support both `--requirement` and whitespace.

```suggestion
            if re.match(r"^\s*(-r|--requirement)", line):
```

src/bentoml/_internal/utils/__init__.py
```diff
@@ -303,6 +304,24 @@ def resolve_user_filepath(filepath: str, ctx: t.Optional[str]) -> str:
...
+                base, nested = find_nested_requirements_txts(
+                    base_requirements_txt.parent / line.split("-r")[1].strip(),
```
sauyon, line 318: In case there's `-r` inside a filename. Maybe you'll need another branch for `--requirement` too.

```suggestion
                    base_requirements_txt.parent / line.split("-r", 1)[1].strip(),
```

sauyon (commented):
While I also think eventually we may want to switch to using a library for this just using a regex for now is probably fine, since it's not too much overhead and we're only supporting this one feature. Might be worth double checking their implementation to ensure that we aren't missing some other syntax, though.

tests/e2e/bento_server_http/bentofile.yaml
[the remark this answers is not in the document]
judahrand: Yeah, I agree that it is something that should really be included in unit testing rather than the e2e tests. However, I don't see the utility in introducing an entirely new e2e test (which will increase the 'test time') to cover it. Really at some point the build requirements should be unit tested and then this e2e test could go back to how it was -- this isn't something I think should be tackled in this PR though.

I can if you really want but it doesn't feel necessary.

src/bentoml/_internal/bento/build_config.py
judahrand: Now that I've switched to using `pip-requirements-parser` as @aarnphm suggested this should no longer be a limitation as the parser is reading all nested files and essentially concating them all into a single file (with a small modification to make sure that we remove other file references).

I think the only thing which might not work not is `--find-links` with a path to an HTML file. However, that seems like a super niche edge case.

judahrand:
Now that I've switched to using pip-requirements-parser as @aarnphm suggested this should no longer be a limitation as the parser is reading all nested files and essentially concating them all into a single file (with a small modification to make sure that we remove other file references).

I think the only thing which might not work not is --find-links with a path to an HTML file. However, that seems like a super niche edge case.

judahrand:
I don't think that the current test failures are actually related to my changes. They seem to be a combination of CodeCov upload failures and tests which are also failing in `main`. Are these able to be fixed?

tests/e2e/bento_server_http/bentofile.yaml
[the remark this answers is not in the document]
aarnphm: Me and @sauyon discussed that this change is ok for this PR. We will add unit tests for build config specifically to test out these use case.

aarnphm:
> I don't think that the current test failures are actually related to my changes. They seem to be a combination of CodeCov upload failures and tests which are also failing in `main`. Are these able to be fixed?

I have a PR to fix the tests, for that one to get merged first, and then I will rebase and update your changes into the main branch.

aarnphm:
gg

> Now that I've switched to using `pip-requirements-parser` as @aarnphm suggested the only thing in a `requirements.txt` file which might not work not is `--find-links` with a path to an HTML file. However, that seems like a super niche edge case.

Thanks for the reminder. I will follow up with a PR for documentation about this behaviour.

aarnphm (approved):

PR #3114, opened 2022-10-18
chore: add options to install pydantic with `bentoml[io-json]`

aarnphm:
Add ability to install optional pydantic with `bentoml[io-json]`

### DIFF
--- a/docs/source/concepts/bento.rst
+++ b/docs/source/concepts/bento.rst
@@ -559,6 +559,14 @@ validation, specify the desired variant in the under ``python.packages`` field:
            packages:
            - "bentoml[aws]"
 
+   .. tab-item:: JSON IO
+
+      .. code-block:: yaml
+
+         python:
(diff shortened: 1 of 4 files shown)


docs/source/concepts/bento.rst
```diff
@@ -591,7 +607,7 @@ validation, specify the desired variant in the under ``python.packages`` field:
            packages:
            - "bentoml[tracing-zipkin]"
 
-   .. tab-item:: OpenTelemetry Protocol
+   .. tab-item:: OTLP
```
jjmachan, line 610: Why did you change it to the shorter name here?

jjmachan (approved):

docs/source/concepts/bento.rst
aarnphm: It is actually the name of the protocol, it should be OTLP :)

PR #3127, opened 2022-10-21
fix(cli): expands working_dir with default path

quandollar:
We should be able to run the command `bentoml serve` outside of the directory containing bentofile.yaml, for example, by running `bentoml serve ..` in a subdirectory. This does not work today. These are WIP changes made during my pairing with @sauyon today.


aarnphm:
You also have to change L72 for serve-http

aarnphm (approved):

PR #3132, opened 2022-10-24
fix: dispatcher for early requests

sauyon:
Attempts to train the optimizer using 5 requests, during which no requests are denied.

Does this kind of approach make any sense @bojiang?

### DIFF
--- a/src/bentoml/_internal/marshal/dispatcher.py
+++ b/src/bentoml/_internal/marshal/dispatcher.py
@@ -50,19 +50,19 @@ class Optimizer:
         self.o_stat: collections.deque[tuple[int, float, float]] = collections.deque(
             maxlen=self.N_KEPT_SAMPLE
         )  # to store outbound stat data
-        self.o_a = min(2, max_latency * 2 / 3)
-        self.o_b = min(1, max_latency * 1 / 3)
+        self.o_a = min(2, max_latency * 2.0 / 30)
+        self.o_b = min(1, max_latency * 1.0 / 30)
 
         self.wait = 0.01  # the avg wait time before outbound called
 
         self._refresh_tb = TokenBucket(2)  # to limit params refresh interval
-        self.outbound_counter = 0
+        self._outbound_counter = 0
 
     def log_outbound(self, n: int, wait: float, duration: float):
         if (
-            self.outbound_counter <= self.N_SKIPPED_SAMPLE
+            self._outbound_counter <= self.N_SKIPPED_SAMPLE
         ):  # skip inaccurate info at beginning
-            self.outbound_counter += 1
+            self._outbound_counter += 1
             return
 
         self.o_stat.append((n, duration, wait))
@@ -99,8 +99,6 @@ class CorkDispatcher:
         * implement CORK algorithm to cork & release calling of wrapped function
     The wrapped function should be an async function.
     """
-
-    N_TRAINING_REQUESTS = 5  # number of requests to wait for
 
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


src/bentoml/_internal/marshal/dispatcher.py
```diff
@@ -172,14 +174,55 @@ async def controller(self):
         """
         A standalone coroutine to wait/dispatch calling.
         """
+        decay = 0.95  # the decay rate of wait time
+
+        while self.optimizer.outbound_counter <= self.N_TRAINING_REQUESTS:
+            # early loop for when the optimizer has not been trained yet
+            try:
+                async with self._wake_event:  # block until there's any request in queue
+                    await self._wake_event.wait_for(self._queue.__len__)
```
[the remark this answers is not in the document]
aarnphm: that makes sense.

aarnphm:
I load test this setup with yolov5 model with locust on EC2, and seems like I didn't see as much 429 as b4.

src/bentoml/_internal/marshal/dispatcher.py
```diff
@@ -172,14 +174,55 @@ async def controller(self):
...
+                    await asyncio.sleep(self.tick_interval)
+                    continue
+                if n * (wn + dt + (a or 0)) <= self.optimizer.wait * decay:
+                    await asyncio.sleep(self.tick_interval)
+                    continue
+
+                n_call_out = min(
+                    self.max_batch_size,
+                    n,
+                )
```
aarnphm, line 204: quick style things

```suggestion
                n_call_out = min(self.max_batch_size, n)
```

src/bentoml/_internal/marshal/dispatcher.py
```diff
@@ -118,7 +120,7 @@ def __init__(
         """
         self.max_latency_in_ms = max_latency_in_ms / 1000.0
         self.fallback = fallback
-        self.optimizer = Optimizer()
+        self.optimizer = Optimizer(self.max_latency_in_ms)
```
larme, line 123: should we rename this property from `max_latency_in_ms` to `max_latency_in_seconds`?

sauyon: Yes, we should. I'll make a quick change.

bojiang (changes requested):
These are ancient codes, and the effects and side effects of the changes will not be readily apparent, but have a large impact on performance. I recommend doing a careful review and some marginal case testing.

bojiang:
@sauyon what's happening in the checks?

sauyon:
Seems like something is failing related to coverage. @aarnphm?

src/bentoml/_internal/marshal/dispatcher.py
```diff
@@ -42,27 +42,27 @@ class Optimizer:
...
         """
         assume the outbound duration follows duration = o_a * n + o_b
         (all in seconds)
         """
         self.o_stat: collections.deque[tuple[int, float, float]] = collections.deque(
             maxlen=self.N_KEPT_SAMPLE
         )  # to store outbound stat data
-        self.o_a = 2
-        self.o_b = 1
+        self.o_a = min(2, max_latency * 2 / 3)
```
aarnphm, line 53:
```suggestion
        self.o_a = min(2, max_latency * 2 / 30)
```

bojiang:
Discussed in slack. The init training step need to avoid large batch which may cause our user wait forever.

src/bentoml/_internal/marshal/dispatcher.py
aarnphm: I will make this change in my docs PR.

parano: @aarnphm why docs change? this should not be a user-facing change right?

aarnphm (approved):

aarnphm:
cc @bojiang can you take a look and accept this?

src/bentoml/_internal/marshal/dispatcher.py
```diff
@@ -172,6 +172,174 @@ async def controller(self):
...
+                # don't try to be smart here, just serve the first few requests
+                if self._sema.is_locked():
+                    await asyncio.sleep(self.tick_interval)
+                    continue
+
+                n_call_out = 1
+                # call
+                self._sema.acquire()
+                inputs_info = tuple(self._queue.pop() for _ in range(n_call_out))
+                self._loop.create_task(self.outbound_call(inputs_info))
```
bojiang, line 206: @sauyon  `create_task` doesn't mean `outbound_counter` will increase instantly. More requests will come into the loop. Do you expect this to happen?

bojiang:
@aarnphm Have you tested the new changes over yolov5?

src/bentoml/_internal/marshal/dispatcher.py
sauyon: I'm basically waiting for the requests to finish before exiting this loop instead of having separate loops to wait for the requests to finish before triggering an optimizer refresh.

The main issue I'm working around here is the case that the runner errors; I'm not actually sure that was a real issue but I'm not sure how the training behaves in that case.

aarnphm:
> @aarnphm Have you tested the new changes over yolov5?

yes I did.

src/bentoml/_internal/marshal/dispatcher.py
sauyon: I should probably check it again before we try to call out, though.

sauyon:
Closing because we need to make a new one to merge while @bojiang is offline.

PR #3135, opened 2022-10-24
fix(multipart): use field names in request

sauyon:


src/bentoml/_internal/io_descriptors/multipart.py
```diff
@@ -238,13 +238,23 @@ async def from_http_request(self, request: Request) -> dict[str, t.Any]:
...
-        )
-        return dict(zip(self._inputs, reqs))
+        form_values = await populate_multipart_requests(request)
+
+        res = {}
+        for field, descriptor in self._inputs.items():
+            if field not in form_values:
+                break
+            res[field] = descriptor.from_http_request(form_values[field])
+        else:
```
aarnphm, line 248: I think a note for `for..else` loop would be nice for the reader
```suggestion
        else:  # NOTE: This is similar to goto, when there is no break.
```

aarnphm (dismissed):

aarnphm:
would be nice to have a test case for this in our e2e test @sauyon

src/bentoml/_internal/io_descriptors/multipart.py
aarnphm: we can also link https://docs.python.org/3/tutorial/controlflow.html#break-and-continue-statements-and-else-clauses-on-loops this as well for reference for the unheard.

PR #3148, opened 2022-10-26
fix(templates): ensure dependencies are met for CUDA templates

fortunto2:
## What does this PR address?

The new image build Kaniko Dockerfile is broken. When using the CUDA Debian-based  template.
1. software-properties-common - must be run before the add-apt-repository command is called on the next command.
2. pipefail command - Kaniko is also not perceived

Use latest bentoml 1.0.7 (26 oktober 2022) (installed from pip install git+https://github.com/bentoml/BentoML.git)
Yatai v1.0.0-d7-6c68b8e

### DIFF
--- a/src/bentoml/_internal/bento/docker/templates/base_debian.j2
+++ b/src/bentoml/_internal/bento/docker/templates/base_debian.j2
@@ -10,5 +10,5 @@ RUN rm -f /etc/apt/apt.conf.d/docker-clean; echo 'Binary::apt::APT::Keep-Downloa
 RUN {{ common.mount_cache(__lib_apt__) }} \
     {{ common.mount_cache(__cache_apt__) }} \
     apt-get update -y \
-    && apt-get install -q -y --no-install-recommends --allow-remove-essential ca-certificates gnupg2 bash build-essential software-properties-common {% if __options__system_packages is not none %}{{ __options__system_packages | join(' ') }}{% endif -%}
+    && apt-get install -q -y --no-install-recommends --allow-remove-essential ca-certificates gnupg2 bash build-essential {% if __options__system_packages is not none %}{{ __options__system_packages | join(' ') }}{% endif -%}
 {% endblock %}
--- a/src/bentoml/_internal/bento/docker/templates/base_debian.j2
+++ b/src/bentoml/_internal/bento/docker/templates/base_debian.j2
@@ -10,5 +10,5 @@ RUN rm -f /etc/apt/apt.conf.d/docker-clean; echo 'Binary::apt::APT::Keep-Downloa
 RUN {{ common.mount_cache(__lib_apt__) }} \
     {{ common.mount_cache(__cache_apt__) }} \
     apt-get update -y \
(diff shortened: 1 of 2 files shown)


src/bentoml/_internal/bento/docker/templates/cuda_debian.j2
```diff
@@ -4,7 +4,6 @@
 
 RUN {{ common.mount_cache(__lib_apt__) }} \
     {{ common.mount_cache(__cache_apt__) }} \
-    set -euxo pipefail && \
```
aarnphm, line 7: revert this
```
set -eux
```

fortunto2: Yes, worked! Thanks

src/bentoml/_internal/bento/docker/templates/base_debian.j2
```diff
@@ -10,5 +10,5 @@ RUN rm -f /etc/apt/apt.conf.d/docker-clean; echo 'Binary::apt::APT::Keep-Downloa
 RUN {{ common.mount_cache(__lib_apt__) }} \
     {{ common.mount_cache(__cache_apt__) }} \
     apt-get update -y \
-    && apt-get install -q -y --no-install-recommends --allow-remove-essential ca-certificates gnupg2 bash build-essential {% if __options__system_packages is not none %}{{ __options__system_packages | join(' ') }}{% endif -%}
+    && apt-get install -q -y --no-install-recommends --allow-remove-essential ca-certificates gnupg2 bash build-essential software-properties-common {% if __options__system_packages is not none %}{{ __options__system_packages | join(' ') }}{% endif -%}
```
aarnphm, line 13: let's not add this into base_debian.j2 and keep it at cuda

if users need to add software-properties-common to the bento you can specify it with `docker.system_packages`

aarnphm:
Canm you try again from main with yatai 1.0.0 instead of d7?

src/bentoml/_internal/bento/docker/templates/cuda_debian.j2
```diff
@@ -4,11 +4,11 @@
...
+    set -eux pipefail && \
```
aarnphm, line 7:
```suggestion
    set -eux && \
```

src/bentoml/_internal/bento/docker/templates/base_debian.j2
aarnphm: can you explain why we need to add `software-properties-common` to base_debian.j2?

I prefer not to install unnecessary packages for the container that doesn't require this.

fortunto2:
> Canm you try again from main with yatai 1.0.0 instead of d7?

while it is difficult to do, we have it deployed in kubernetes, I will definitely try to update later when we can update

aarnphm (approved):
thanks for the contribution.

fortunto2:
> thanks for the contribution.

received! thank you for a great product

PR #3152, opened 2022-10-27
fix: async bento client methods

sauyon:
Fixes the newly changed async/sync bento client methods, and makes from_url synchronous using the stdlib `HTTPConnection` so that it can be used in both async and non-async contexts.

### DIFF
--- a/src/bentoml/client.py
+++ b/src/bentoml/client.py
@@ -31,8 +31,14 @@ class Client(ABC):
 
         for name, api in self._svc.apis.items():
             if not hasattr(self, name):
-                setattr(self, name, functools.partial(self._sync_call, _bentoml_api=api))
-                setattr(self, f"async_{name}", functools.partial(self._call, _bentoml_api=api))
+                setattr(
+                    self, name, functools.partial(self._sync_call, _bentoml_api=api)
+                )
+                setattr(
+                    self,
+                    f"async_{name}",
+                    functools.partial(self._call, _bentoml_api=api),
+                )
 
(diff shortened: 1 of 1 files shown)


src/bentoml/client.py
```diff
@@ -27,15 +31,19 @@ def __init__(self, svc: Service, server_url: str):
...
+        return asyncio.run(self.async_call(bentoml_api_name, inp))
 
     async def async_call(
-        self, api_name: str, inp: t.Any = None, **kwargs: t.Any
+        self, bentoml_api_name: str, inp: t.Any = None, **kwargs: t.Any
     ) -> t.Any:
-        return self._call(inp, _bentoml_api=self._svc.apis[api_name])
+        return await self._call(inp, _bentoml_api=self._svc.apis[bentoml_api_name])
+
+    def _sync_call(self, inp: t.Any = None, *, _bentoml_api: InferenceAPI, **kwagrs: t.Any):
```
aarnphm, line 45: where is this function getting called?

sauyon: It gets used for the `.<API name>` methods.

aarnphm: ah got it. I didn't see the `setattr`.

aarnphm (dismissed):

aarnphm:
o the formatter check is failing @sauyon

sauyon:
Ah, you know what, my hooks all disappeared when I deleted my repo :sweat_smile:

ssheng (approved):

ssheng (approved):

PR #3153, opened 2022-10-27
feat: exemplar support for metrics

aarnphm:
This PR adds the ability to include exemplars to metrics to support OpenMetrics
specification.

### DIFF
--- a/src/bentoml/_internal/server/http/instruments.py
+++ b/src/bentoml/_internal/server/http/instruments.py
@@ -135,7 +135,6 @@ class HTTPTrafficMetricsMiddleware:
                     "trace_id": trace_context.trace_id,
                     "span_id": trace_context.span_id,
                     "request_id": trace_context.request_id,
-                    "sampled": trace_context.sampled,
                 }.items(),
             )
         )
@@ -276,7 +275,6 @@ class RunnerTrafficMetricsMiddleware:
                     "trace_id": trace_context.trace_id,
                     "span_id": trace_context.span_id,
                     "request_id": trace_context.request_id,
-                    "sampled": trace_context.sampled,
                 }.items(),
             )
         )
--- a/src/bentoml/_internal/server/http/instruments.py
+++ b/src/bentoml/_internal/server/http/instruments.py
@@ -8,6 +8,7 @@ from typing import TYPE_CHECKING
 from simple_di import inject
 from simple_di import Provide
 
+from ...context import trace_context
 from ...context import component_context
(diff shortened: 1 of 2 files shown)


src/bentoml/_internal/server/http/instruments.py
```diff
@@ -146,7 +162,7 @@ async def wrapped_send(message: "ext.ASGIMessage") -> None:
                         service_name=component_context.bento_name,
                         service_version=component_context.bento_version,
                         http_response_code=STATUS_VAR.get(),
-                    ).inc()
+                    ).inc(exemplar=exemplar)
```
ssheng, line 165: How do we control the sampling? Does the current implementation include all requests as examplars?

src/bentoml/_internal/server/http/instruments.py
```diff
@@ -124,6 +125,21 @@ async def __call__(
             await response(scope, receive, send)
             return
 
+        # NOTE: Exemplars are reference to data outside
+        # of the metrics set. This is often use to link
+        # traces to given metrics.
```
sauyon, line 130:
```suggestion
        # NOTE: Exemplars are references to data outside
        # of the metrics set. This is often use to link
        # traces to given metrics.
```

src/bentoml/_internal/server/http/instruments.py
```diff
@@ -124,6 +125,21 @@ async def __call__(
...
+            filter(
+                lambda tup: tup[1] is not None,
+                {
+                    "trace_id": "%s" % trace_context.trace_id,
+                    "span_id": "%s" % trace_context.span_id,
+                    "request_id": "%s" % trace_context.request_id,
+                    "sampled": "%s" % trace_context.sampled,
+                }.items(),
+            )
+        )
```
sauyon, line 141: I'm somewhat unsure what's happening here, it seems like you're casting all the values into strings with `%` instead of `str`, and then filtering out `None`, but nothing should be `None`, since `"%s" % None` should be `'None'`.

```suggestion
        exemplar: dict[str, str] = {
            key: val for key, val in [
                "trace_id": "%s" % trace_context.trace_id,
                 "span_id": "%s" % trace_context.span_id,
                 "request_id": "%s" % trace_context.request_id,
                 "sampled": "%s" % trace_context.sampled,
            ] if val is not None
        }
```

sauyon (changes requested):
One question. Also, have you responded to Sean above?

aarnphm:
Closed as irrelevant for now.

PR #3157, opened 2022-10-28
fix: framework ci

aarnphm:
fix broken fastai ci


sauyon (approved):
fine as a temporary fix

PR #3160, opened 2022-10-28
chore: deprecate buildx args

aarnphm:
This PR revert `containerize` to use `docker build` with BuildKit.

`buildx` is currently not yet supported on some CI vendor.

Signed-off-by: Aaron Pham <29749331+aarnphm@users.noreply.github.com>

```bash
$ bentoml containerize --help
Usage: bentoml containerize [OPTIONS] BENTO_TAG

  Containerizes given Bento into a ready-to-use Docker image.

  BENTO is the target BentoService to be containerized, referenced by its name
  and version in format of name:version. For example: "iris_classifier:v1.2.0"

  'bentoml containerize' command also supports the use of the 'latest' tag which will automatically use the last built
  version of your Bento.

  You can provide a tag for the image built by Bento using the '--docker-image-tag' flag. Additionally, you can
  provide a '--push' flag, which will push the built image to the Docker repository specified by the image tag.

  You can also prefixing the tag with a hostname for the repository you wish to push to. e.g. 'bentoml containerize
  IrisClassifier:latest --push --tag repo-address.com:username/iris' would build a Docker image called
  'username/iris:latest' and push that to docker repository at repo-address.com.

  By default, the 'containerize' command will use the current credentials provided by Docker daemon.

  'bentoml containerize' also uses Docker Buildx as backend, in place for normal 'docker build'. By doing so, BentoML
  will leverage Docker Buildx features such as multi-node builds for cross-platform images, Full BuildKit capabilities
  with all of the familiar UI from 'docker build'.

Options:
  -t, --docker-image-tag TEXT    Name and optionally a tag (format: 'name:tag'), defaults to bento tag.
  --add-host TEXT                Add a custom host-to-IP mapping (format: 'host:ip').
  --build-arg TEXT               Set build-time variables.
  --cache-from TEXT              External cache sources (e.g., 'user/app:cache', 'type=local,src=path/to/dir').
  --iidfile TEXT                 Write the image ID to the file.
  --label TEXT                   Set metadata for an image.
  --network TEXT                 Set the networking mode for the 'RUN' instructions during build (default 'default').
  --no-cache                     Do not use cache when building the image.
  --output TEXT                  Output destination (format: 'type=local,dest=path').
  --platform TEXT                Set target platform for build.
  --progress [auto|tty|plain]    Set type of progress output ('auto', 'plain', 'tty'). Use plain to show container
                                 output.
  --pull                         Always attempt to pull all referenced images.
[...]

Issue #3146, opened 2022-10-26
bug: failed to containerize when using mlflow


SohelKabir:
Trying to integrate Mlflow with my current bentoml workflow and following this example
`https://github.com/bentoml/BentoML/tree/main/examples/mlflow/pytorch`
But getting error when i try to deploy model with docker, 

when i run  `bentoml containerize mlflow_pytorch_mnist_demo:latest`

```Building docker image for Bento(tag="mlflow_pytorch_mnist_demo:3utxjn2vbgxh5gbc")...
ERROR: failed to solve: executor failed running [/bin/sh -c bash <<EOF
set -euxo pipefail

if [ -f /home/bentoml/bento/env/conda/environment.yml ]; then
   set pip_interop_enabled to improve conda-pip interoperability. Conda can use
   pip-installed packages to satisfy dependencies.
  echo "Updating conda base environment with environment.yml"
  /opt/conda/bin/conda config --set pip_interop_enabled True
  /opt/conda/bin/conda env update -n base -f /home/bentoml/bento/env/conda/environment.yml
  /opt/conda/bin/conda clean --all
fi
EOF]: exit code: 1
Failed building docker image: Command '['docker', 'buildx', 'build', '--progress', 'auto', '--tag', 'mlflow_pytfile', 'env\\docker\\Dockerfile', '--load', '.']' returned non-zero exit status 1.
```
### To reproduce

Bug recreation steps:
Clone the repo `https://github.com/bentoml/BentoML/tree/main/examples/mlflow/pytorch` 
Goto the folder `examples/mlflow/pytorch`
`python mnist.py`
`bentoml build`
`bentoml containerize mlflow_pytorch_mnist_demo:latest`

P.S. `bentoml serve service.py:svc`  works fine`

### Environment

bentoml version 1.0.7
Python version  3.9.12
Docker Engine 20.10.17

aarnphm:
Hi there, can you send the full error output?

SohelKabir:
Full error given alread, that was all @aarnphm

SohelKabir:
![image](https://user-images.githubusercontent.com/33397287/198088425-81fdd1dd-2187-4e83-a05d-7a376108888b.png)

aarnphm:
This issue has been addressed on main and will be included in 1.0.8

aarnphm:
Please try out 1.0.8 release, as this issue has been fixed. Thanks a bunch.

SohelKabir:
[...]

PR #3164, opened 2022-10-31
feat(container): builder implementation

aarnphm:
# BEP: better containerization supports

This PR introduces a new API that allows users to build Bento into OCI-compliant images
with their build tools of choice. This PR includes enhancement to our current
containerization feature.

### Introduce a new Python API: `bentoml.container`

To build a container from a file, use `bentoml.container.build`

```python

import bentoml

bentoml.container.build("iris_classifier:latest", backend="podman", features=["grpc","grpc-reflection"], **kwargs)
```

This PR introduces the concept of an OCI backend interface, which is a light wrapper to
construct arguments that can be parsed directly to the given build tools CLI.

To access default backend, use `bentoml.container.get_backend`:

```python
builder = bentoml.container.get_backend("docker")
```

To register your custom backend builder, use `bentoml.container.register_backend`:

```python
def build_arg_parser_fn(*, arg1:str | None = None, **kwargs) -> list[str]:
	# parsing arguments to CLI args
	...

def healthcheck_fn() -> bool:
	# check whether the builder is healthy
	...

bentoml.container.register_backend('nerdctl', binary="/usr/local/bin/nerdctl",
buildkit_support=True,construct_build_arg=build_arg_parser_fn, health=healthcheck_fn)

b = bentoml.container.get_backend("nerdctl")
```

To check if your builder is healthy, use `bentoml.container.health`:

```python
assert bentoml.container.health("my_backend")
```

Currently, supported backends are:

- buildctl ✅ (standalone buildkit builder)
- kaniko 🚧 (n/a)
- Docker ✅
- nerdctl ✅ (docker equivalent but with containerd, requires BuildKit for build features)
- Podman ✅ 
- Buildah ✅ 
- Buildx ✅  (docker build + some additional features)

### Backward compatibility

This PR introduces a new API, which means the old containerize function will be
deprecated.

```python
bentoml.bentos.containerize(**kwargs)  # deprecated
```

### BuildKit interop

BentoML leverages BuildKit for a more extensive feature set. However, This has come with
a lot of friction for migration purposes as well as restrictions to use with other build tools (such as podman, buildah). This PR introduces BuildKit with BentoML to be an opt-out. To disable BuildKit, pass `DOCKER_BUILDKIT=0` to `bentoml containerize`, similar to how you disable buildkit with `docker build`:
```bash
DOCKER_BUILDKIT=0 bentoml containerize ...
```

> NOTE that all Bento container will now be following OCI spec instead of Docker spec. The difference is that in OCI spec, there is no SHELL argument.

Now, the generated Dockerfile in the bento will include a Dockerfile that is compatible with all of the build tools. We encourage users always to use `bentoml containerize`. 

### CLI improvement

To better support different backends, `bentoml containerize` will be more agnostic when
it comes to parsing options. This is rather an improvement from #3160 where we will provide `buildx` via `--backend=buildx` for backward compatibility:

```bash
bentoml containerize --help

Usage: bentoml containerize [OPTIONS] BENTO:TAG

  Containerizes given Bento into an OCI-compliant container, with any of your favorite OCI builder.

  BENTO is the target BentoService to be containerized, referenced by its name
  and version in format of name:version. For example: "iris_classifier:v1.2.0"

  ``bentoml containerize`` command also supports the use of the ``latest`` tag
  which will automatically use the last built version of your Bento.

  You can provide a tag for the image built by Bento using the
  ``--image-tag`` flag.

  You can also prefixing the tag with a hostname for the repository you wish
  to push to.

  .. code-block:: bash

     $ bentoml containerize iris_classifier:latest -t repo-address.com:username/iris

  This would build an image called ``username/iris:latest`` and push it to docker repository at repo-address.com.

  By default, the ``containerize`` command will use the current credentials provided by each backend implementation.

  ``bentoml containerize`` also leverage BuildKit features such as multi-node
  builds for cross-platform images, more efficient caching, and more secure builds.
  BuildKit will be enabled by default by each backend implementation.

Options:
  -t, --image-tag, --docker-image-tag TEXT
                                  Name and optionally a tag (format: 'name:tag') for building container, defaults to
                                  the bento tag.
  --backend TEXT                  Define builder backend. Available: docker, buildah, buildx, podman
  --opt ARG=VALUE[,ARG=VALUE]     Define options for given backend. (format: '--opt target=foo --opt build-arg=foo=BAR
                                  --opt iidfile:/path/to/file --opt no-cache')
  --enable-features FEATURE[,FEATURE]
                                  Enable additional BentoML features. Available features are: io-pandas, grpc,
                                  tracing, grpc-reflection, tracing-otlp, aws, tracing-jaeger, grpc-channelz, all, io-
                                  image, tracing-zipkin, io-json.
  --add-host [HOST:IP,]           (Equivalent to --opt add-host=host:ip): Add a custom host-to-IP mapping.
  --build-arg [KEY=VALUE,]        (Equivalent to --opt build-arg=FOO=bar): Set build-time variables.
  --cache-from [NAME|type=TYPE[,KEY=VALUE]]
                                  (Equivalent to --opt cache-from=user/app:cache): External cache sources (e.g.:
                                  'type=local,src=path/to/dir').
  --iidfile PATH                  (Equivalent to --opt iidfile=/path/to/file): Write the image ID to the file.
  --label [NAME|[KEY=VALUE],]     (Equivalent to --opt label=io.container.capabilities=baz): Set metadata for an
                                  image. This is equivalent to LABEL in given Dockerfile.
  --network default|VALUE         (Equivalent to --opt network=default): Set the networking mode for the 'RUN'
                                  instructions during build (default 'default').
  --no-cache                      (Equivalent to --opt no-cache): Do not use cache when building the image.
  --output [PATH,-,type=TYPE[,KEY=VALUE]
                                  (Equivalent to --opt output=type=local,dest=/path/to/dir): Output destination.
  --platform VALUE[,VALUE]        (Equivalent to --opt platform=linux/amd64,linux/arm/v7): Set target platform for
                                  build.
  --progress VALUE                (Equivalent to --opt progress=auto): Set type of progress output. Use plain to show
                                  container output if BENTOML_DEBUG is set.
  --pull                          (Equivalent to --opt pull): Always attempt to pull all referenced images.
  --secret [type=TYPE[,KEY=VALUE]
                                  (Equivalent to --opt secret=id=aws,src=$HOME/.aws/crendentials): Secret to expose to
                                  the build.
  --ssh default|<id>[=<socket>|<key>[,<key>]]
                                  (Equivalent to --opt ssh=default=$SSH_AUTH_SOCK): SSH agent socket or keys to expose
                                  to the build. See
                                  https://docs.docker.com/engine/reference/commandline/buildx_build/#ssh
  --target VALUE                  (Equivalent to --opt target=release-stage): Set the target build stage to build.
  --allow ENTITLEMENT             (Equivalent to --opt allow=network.host): Allow extra privileged entitlement (e.g.,
                                  'network.host', 'security.insecure'). [Only available with '--backend=buildx']
  --build-context name=VALUE      (Equivalent to --opt build-context=project=path/to/project/source): Additional build
                                  contexts (e.g., name=path). [Only available with '--backend=buildx']
  --builder TEXT                  (Equivalent to --opt builder=default): Override the configured builder instance.
                                  Same as 'docker buildx --builder' [Only available with '--backend=buildx']
  --cache-to [NAME|type=TYPE[,KEY=VALUE]]
                                  (Equivalent to --opt cache-to=type=registry,ref=user/app): Cache export destinations
                                  (e.g., 'user/app:cache', 'type=local,dest=path/to/dir'). [Only available with '--
                                  backend=buildx']
  --load                          (Equivalent to --opt load): Shorthand for '--output=type=docker'. Note that '--push'
                                  and '--load' are mutually exclusive. [Only available with '--backend=buildx']
  --push                          (Equivalent to --opt push): Shorthand for '--output=type=registry'. Note that '--
                                  push' and '--load' are mutually exclusive. [Only available with '--backend=buildx']
  --cgroup-parent TEXT            (Equivalent to --opt cgroup-parent=cgroupv2): Optional parent cgroup for the
                                  container. [Only available with '--backend=buildx']
  --ulimit <type>=<soft limit>[:<hard limit>]
                                  (Equivalent to --opt ulimit=nofile=1024:1024): Ulimit options. [Only available with
                                  '--backend=buildx']
  --no-cache-filter TEXT          (Equivalent to --opt no-cache-filter): Do not cache specified stages. [Only
                                  available with '--backend=buildx']
  --metadata-file TEXT            (Equivalent to --opt metadata-file=/path/to/file): Write build result metadata to
                                  the file. [Only available with '--backend=buildx']
  --shm-size TEXT                 (Equivalent to --opt shm-size=8192Mb): Size of '/dev/shm'. [Only available with '--
                                  backend=buildx']
  -q, --quiet                     Suppress all warnings and info logs
  --verbose, --debug              Generate debug information
  --do-not-track                  Do not send usage info
  -h, --help                      Show this message and exit.

```

Users should now use `--opt` to pass corresponding args for the given builder.

Before:

```bash
bentoml containerize --label io.container=bar iris_classifier:latest
```

Now:

```bash
bentoml containerize --opt label=io.container=bar iris_classifier:latest
```

Note that we also accept `--opt label:io.container=baz`

This means BentoML will maintain backward compatibility options for CLI args until BentoML 2.0 is
released. We encourage users to migrate when possible.

To change different backend, pass in `--backend`:
```bash
bentoml containerize --backend nerdctl --opt build-arg=FOO=bar ...
```
or via envvar `BENTOML_CONTAINERIZE_BACKEND`:
```bash
BENTOML_CONTAINERIZE_BACKEND=nerdctl bentoml containerize ...
```

## TODO:
A few items to be implemented as a follow up:

- [ ] add options to parse args to the given builder:
    Builder can also specify args, hence we should also allow userrr to configure this.

- [x] Supports for Kaniko:
    In the case of Yatai, it can just call containerize without having to build from the bento.


src/bentoml/_internal/container/__init__.py
```diff
@@ -0,0 +1,334 @@
...
+        backend: The name of the backend.
+
+    Returns:
+        True if the backend is healthy, False otherwise.
+    """
+    try:
+        return get_backend(backend).health()
+    except Exception as e:  # pylint: disable=broad-except
+        logger.error(
+            f"Backend '%s' failed to health check: %s",
```
jjmachan, line 169: is it okay to use f-strings in loggers?

aarnphm: oops I forgot to remove this 😄

aarnphm, line 169:
```suggestion
            "Backend '%s' failed to health check: %s",
```

src/bentoml/_internal/container/kaniko.py
```diff
@@ -0,0 +1,103 @@
...
+
+BUILD_CMD = []
+
+
+def find_binary() -> str | None:
+    # We will check if /kaniko/executor exists
+    if os.path.exists("/kaniko/executor"):
+        return "/kaniko/executor"
+    return shutil.which("executor")
+
```
jjmachan, line 36: this will check for kaniko in the root filesystem right but Kaniko is designed to be run as a container?

aarnphm: I think I make an assumption here that kaniko will be ran inside a container.

L35 are mostly act as a fallback when I was debugging and building executor from a container and move it to a different path

src/bentoml/_internal/container/dockerfile/templates/miniconda_alpine.j2
```diff
@@ -1,6 +1,6 @@
 {% extends "base_alpine.j2" %}
 {% block SETUP_BENTO_COMPONENTS %}
-{{ common.setup_conda(__python_version__, bento__path) }}
+{{ common.setup_conda(__conda_python_version__, bento__path, enable_buildkit=__enable_buildkit__) }}
```
aarnphm, line 3: To address #3146

src/bentoml_cli/containerize.py
```diff
@@ -1,360 +1,636 @@
 from __future__ import annotations
 
+import os
```
aarnphm, line 3: This file mainly contains backward compatibility layer for our current containerize behaviour.

jjmachan (approved):
👍🏽

src/bentoml/container.py
```diff
@@ -0,0 +1,466 @@
...
+                )
+            logger.info(
+                "Writting Containerfile for '%s' to '%s'.", bento.tag, output_path
+            )
+            shutil.copyfile(final_containerfile, output_path)
+            return
+        # otherwise we will just write this to stderr.
+        with open(final_containerfile, "r") as f:
+            sys.stderr.write(f.read())
+
```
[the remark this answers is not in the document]
aarnphm: This is for user who wants to just see the dockerfile

jjmachan (approved):

aarnphm:
hmm the transformers PR failed, which is not related to this PR 😢

PR #3167, opened 2022-10-31
qa: retry env to be NoneType

aarnphm:

### DIFF
--- a/src/bentoml/_internal/runner/runner_handle/remote.py
+++ b/src/bentoml/_internal/runner/runner_handle/remote.py
@@ -171,7 +171,7 @@ class RemoteRunnerClient(RunnerHandle):
             ) as resp:
                 body = await resp.read()
         except aiohttp.ClientOSError as e:
-            if os.getenv("BENTOML_RETRY_RUNNER_REQUESTS").lower() == "true":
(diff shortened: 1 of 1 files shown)


src/bentoml/_internal/runner/runner_handle/remote.py
```diff
@@ -171,7 +171,8 @@ async def async_run_method(
             ) as resp:
                 body = await resp.read()
         except aiohttp.ClientOSError as e:
-            if os.getenv("BENTOML_RETRY_RUNNER_REQUESTS").lower() == "true":
+            retry_env = os.environ.get("BENTOML_RETRY_RUNNER_REQUESTS")
+            if retry_env is not None and retry_env.lower() == "true":
```
sauyon, line 175: just use the default for getenv: `getenv("var", "")`

aarnphm, line 175:
```suggestion
            if os.getenv("BENTOML_RETRY_RUNNER_REQUESTS", "").lower() == "true"::
```

sauyon (approved):

ssheng (approved):

PR #3170, opened 2022-11-01
fix(templates): ensure to use python3 and pip3

aarnphm:
Ensure our install.sh uses pip3 and python3

For distro such as amazonlinux, the docker container includes both python2 and python3.
python is often aliased with python2, hence this PR explicitly state the distinction.


sauyon (approved):
LGTM assuming you've checked that these names exist in all the images we support.

PR #3175, opened 2022-11-02
fix(sdk): montioring log output

bojiang:


src/bentoml/_internal/log.py
```diff
@@ -110,6 +110,8 @@ def _component_name():
 
 def trace_record_factory(*args: t.Any, **kwargs: t.Any):
     record = default_factory(*args, **kwargs)
+    if record.name in ("bentoml_monitor_data", "bentoml_monitor_schema"):
```
bojiang, line 113: Shall we apply filters rather than this?

aarnphm (dismissed):

aarnphm (approved):

PR #3189, opened 2022-11-04
refactor(sdk): separate default monitoring impl

bojiang:


src/bentoml/_internal/monitoring/api.py
```diff
@@ -30,257 +23,126 @@
...
 
-    PRESERVED_COLUMNS = (COLUMN_TIME, COLUMN_RID) = ("timestamp", "request_id")
+    PRESERVED_COLUMNS: tuple[str, ...] = ()
 
     def __init__(
         self,
         name: str,
-        log_path: str,
-        log_config_file: str | None = None,
+        **_: t.Any,
```
aarnphm, line 48: Lets' not do this, it should be using __init_subclass__?

bojiang: why?

aarnphm: isn't _ the same?

aarnphm (approved):

PR #3199, opened 2022-11-07
fix: CI on main

aarnphm:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,14 +104,18 @@ jobs:
           black --check src examples tests
           black --check --pyi typings
           isort --check .
-      - name: Lint check
+      - name: Get current branch.
+        id: get-current-branch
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -104,14 +104,18 @@ jobs:
           black --check src examples tests
           black --check --pyi typings
           isort --check .
-      - name: Lint check
+      - name: Get current branch.
+        id: get-current-branch
+        run: echo "branch=${GITHUB_REF##*/}" >> $GITHUB_OUTPUT
+      - name: (buf) Linting proto files.
+        if: ${{ steps.get-current-branch.output.branch != 'main' }}
```
sauyon, line 111: Why aren't we running this on main?

aarnphm: Because it will definitely fail right?

aarnphm: maybe not the proto linting

PR #3222, opened 2022-11-10
fix(sdk): monitor type hint for Python3.7.3

bojiang:


sauyon:
We did this one in #3221

PR #3229, opened 2022-11-12
fix(json): make sure to parse a list of dict for_sample

aarnphm:


src/bentoml/_internal/io_descriptors/json.py
```diff
@@ -280,7 +280,7 @@ def openapi_example(self):
                 self.sample
             ):
                 return self.sample.dict()
-            elif isinstance(self.sample, str):
+            elif isinstance(self.sample, (str, list)):
```
mqk, line 283: Shouldn't this addition of `list` happen further down on l. 292? We want to support `list` in addition to `dict`, not in addition to `str`.

aarnphm: You have to dump the list with JSON. This line ensures list of a complex datatype can be serialized.

bojiang (approved):

src/bentoml/_internal/io_descriptors/json.py
bojiang: @aarnphm Could you provide some examples here?

aarnphm: similar to the one that @mqk provides:
```python
@svc.api(input=JSON.from_sample([{"a": 1, "b": 2}, {"a": 2, "b": 4}]), output=JSON())
```

PR #3231, opened 2022-11-13
fix(docs): missing `table` tag

nyongja:
Add table closing tag on instruction 4


aarnphm (approved):
Thanks for spotting this.

PR #3253, opened 2022-11-16
fix: multipart for client

sauyon:
Never tested this after refactor to use aiohttp...


aarnphm (approved):

aarnphm:
@sauyon you need to configure your git hooks for formatting 😄

aarnphm (approved):

PR #3257, opened 2022-11-17
feat(sdk): implement otlp monitoring exporter

bojiang:
Now bentoml monitoring can work with any collection libs that support otlp

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -156,6 +156,11 @@ tracing-jaeger = ["opentelemetry-exporter-jaeger==1.14.0"]
 tracing-zipkin = ["opentelemetry-exporter-zipkin==1.14.0"]
 tracing-otlp = ["opentelemetry-exporter-otlp==1.14.0"]
 
+monitoring = [
+  "bentoml[monitoring-otlp]",
+]
+monitoring-otlp = ["opentelemetry-exporter-otlp==1.14.0"]
+
 [tool.setuptools_scm]
 write_to = "src/bentoml/_version.py"
 git_describe_command = "git describe --dirty --tags --long --first-parent"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,10 +39,11 @@ dependencies = [
   # Since there are discrepancies among API and instrumentation packages,
   # we should always pin the set version of Opentelemetry suite
   "opentelemetry-api==1.14.0",
+  "opentelemetry-sdk==1.14.0",
+  "opentelemetry-exporter-otlp-proto-http==1.14.0",
   "opentelemetry-instrumentation==0.35b0",
   "opentelemetry-instrumentation-aiohttp-client==0.35b0",
   "opentelemetry-instrumentation-asgi==0.35b0",
-  "opentelemetry-sdk==1.14.0",
   "opentelemetry-semantic-conventions==0.35b0",
   "opentelemetry-util-http==0.35b0",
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -156,6 +156,11 @@ tracing-jaeger = ["opentelemetry-exporter-jaeger==1.14.0"]
 tracing-zipkin = ["opentelemetry-exporter-zipkin==1.14.0"]
 tracing-otlp = ["opentelemetry-exporter-otlp==1.14.0"]
 
+monitoring = [
+  "bentoml[monitoring-otlp]",
+]
```
aarnphm, line 161: Hmm, idk if we should add this. I prefer explicitly declare external dependencies for these.

 The reason we have a `bentoml[tracing]` is for backward compatibility.

src/bentoml/_internal/monitoring/api.py
```diff
@@ -253,7 +253,12 @@ def monitor(
     Context manager for monitoring.
 
     :param name: name of the monitor
-    :param monitor_class: class of the monitor
+    :param monitor_class: class of the monitor, can be a string or a class
```
aarnphm, line 256: can you convert this to Google-style docstring?

src/bentoml/_internal/monitoring/otlp.py
```diff
@@ -0,0 +1,220 @@
...
+import os
+import random
+import typing as t
+import logging
+import datetime
+import collections
+import logging.config
+from typing import TYPE_CHECKING
+
+from opentelemetry.sdk._logs import LoggerProvider
```
aarnphm, line 12: Why do we have to use a private library like this? Don't opentelemetry has a public SDK for logs?

aarnphm: Seems like they dont 😢

aarnphm:
Another thing can you also add the additional dependency to `--extra-requires`, which is defined here https://github.com/bentoml/BentoML/blob/bd547bc5405a151aeed5091ce69f6af1a4e6ee4d/src/bentoml/bentos.py#L441?

aarnphm:
Also remember to add `monitoring-otlp` to `all` as well

aarnphm (approved):

src/bentoml/_internal/monitoring/otlp.py
```diff
@@ -0,0 +1,220 @@
...
+            resource=Resource.create(
+                {
+                    "service.name": f"{component_context.bento_name}:{self.name}",
+                    "service.instance.id": "{component_context.bento_version}",
+                }
+            ),
+        )
+        set_logger_provider(self.logger_provider)
+
+        if self.endpoint is not None:
```
aarnphm, line 141: We shouldn't override user env if they set themselves right?

bojiang: It has the same behavior as open telemetry implementation now.

judahrand: This isn't the behavior now - I don't know if it was at the time: https://github.com/open-telemetry/opentelemetry-python/blob/ea15f72f6678ff23c86511c095094cb3e4bafc44/exporter/opentelemetry-exporter-otlp-proto-http/src/opentelemetry/exporter/otlp/proto/http/_log_exporter/__init__.py#L75-L94

PR #3260, opened 2022-11-19
chore(stub): `import filetype`

aarnphm:
Add filetype stubs since from_sample from file and image are using it.

Signed-off-by: Ubuntu <29749331+aarnphm@users.noreply.github.com>


larme (approved):
LGTM

PR #3261, opened 2022-11-19
chore: generate plain dockerfile without buildkit syntax

aarnphm:
The generated Dockerfile inside Bento will now be stripped down without BuildKit syntax.

This is for purpose of serving users who wants to build directly using this Dockerfile.

By default `bentoml containerize` is the recommended way and it uses BuildKit by default.

To opt-out specify `DOCKER_BUILDKIT=0`

Signed-off-by: Ubuntu <29749331+aarnphm@users.noreply.github.com>

### DIFF
--- a/src/bentoml/_internal/bento/build_config.py
+++ b/src/bentoml/_internal/bento/build_config.py
@@ -219,7 +219,7 @@ class DockerOptions:
         dockerfile_path = fs.path.combine(docker_folder, "Dockerfile")
 
         # NOTE that by default the generated Dockerfile won't have BuildKit syntax.
-        # By default, BentoML will use BuildKit. To opt-out specify DOCKER_BUILDKIT=0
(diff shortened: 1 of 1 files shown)


src/bentoml/_internal/bento/build_config.py
```diff
@@ -218,9 +218,13 @@ def write_to_bento(self, bento_fs: FS, build_ctx: str, conda: CondaOptions) -> N
         bento_fs.makedirs(docker_folder, recreate=True)
         dockerfile_path = fs.path.combine(docker_folder, "Dockerfile")
 
+        # NOTE that by default the generated Dockerfile won't have BuildKit syntax.
+        # By default, BentoML will use BuildKit. To opt-out specify DOCKER_BUILDKIT=0
```
ssheng, line 222:
```suggestion
        # By default, BentoML containerization will use BuildKit. To opt-out specify DOCKER_BUILDKIT=0
```

ssheng (approved):

PR #3264, opened 2022-11-21
chore: refactor from psutil constant

aarnphm:
This PR refactors out psutil constant to prepare for #3074

Signed-off-by: Aaron Pham (EC2) <29749331+aarnphm@users.noreply.github.com>


bojiang:
Sorry but I think it ridiculous that introducing a build tool requires project dependency changes.

src/bentoml/_internal/utils/__init__.py
```diff
@@ -48,10 +48,29 @@
 
 C = t.TypeVar("C")
 T = t.TypeVar("T")
-_T_co = t.TypeVar("_T_co", covariant=True, bound=t.Any)
 
 rich_console = Console(theme=None)
 
+# lightweight port from psutil._common.py to avoid circular dependency in bazel
```
bojiang, line 54: Sorry but I think it ridiculous that introducing a build tool requires project dependency changes.

bojiang: There should be some other ways to do that with bazel right?

src/bentoml/testing/server.py
```diff
@@ -208,9 +211,10 @@ def run_bento_server_container(
     ]
     if config_file is not None:
         cmd.extend(["--env", "BENTOML_CONFIG=/home/bentoml/bentoml_config.yml"])
-        cmd.extend(
-            ["-v", f"{os.path.abspath(config_file)}:/home/bentoml/bentoml_config.yml"]
-        )
+        # NOTE: We don't set abspath path if running with bazel
+        if not run_in_bazel():
```
bojiang, line 215: Can we avoid this? Is this because we want to maintain two ways to run tests?

bojiang (changes requested):
Recommend removing `BENTOML_IN_BAZEL` from the codebase. It's hard to maintain if the source depends on build tools.

src/bentoml/_internal/utils/__init__.py
aarnphm: Actually, it is not because of bazel. I ran into the issue before because of a bug on my end. Let me refactor the comment.

aarnphm, line 54:
```suggestion
# lightweight port from psutil._common.py.
```

src/bentoml/testing/server.py
aarnphm: I like just to do it one way with bazel.

bojiang: Yeah. Let's chat about this tmr.

bojiang: Yeah. The source code should not touch `pip install` things. This is an unnecessary coupling. 
Let build tools/ test tools to prepare the environment.

aarnphm: there are some places that we need to change for bazel to work. Things like abspath or relpath wouldn't work inside bazel since it depends on symlink for incremental testing.

PR #3298, opened 2022-12-02
fix: unset `NVIDIA_VISIBLE_DEVICES` when cuda image is used

aarnphm:
User know need to setup NVIDIA_VISIBLE_DEVICES now as we unset this be default when using cuda to prevent a known memory leak.

```python
docker run -e NVIDIA_VISIBLE_DEVICES=all ...
```

### BREAKING CHANGE

If user are running container then one might have to provide additional `-e` flag.


yetone:
LGTM

yetone (approved):

PR #3305, opened 2022-12-04
fix(monitoring): wrong format key for formatter

aarnphm:
format should be an empty string instead of an empty space, otherwise it will fail the regex search for logging format

### DIFF
--- a/src/bentoml/_internal/monitoring/default.py
+++ b/src/bentoml/_internal/monitoring/default.py
@@ -44,8 +44,8 @@ handlers:
     filename: "{schema_filename}"
 formatters:
   bentoml_json:
-    class: pythonjsonlogger.jsonlogger.JsonFormatter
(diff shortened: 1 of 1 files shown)


src/bentoml/_internal/monitoring/default.py
```diff
@@ -44,8 +44,8 @@
     filename: "{schema_filename}"
 formatters:
   bentoml_json:
-    class: pythonjsonlogger.jsonlogger.JsonFormatter
-    format: " "
+    class: pythonjsonlogger.tsonlogger.JsonFormatter
```
aarnphm, line 47:
```suggestion
    class: pythonjsonlogger.jsonlogger.JsonFormatter
```

bojiang (changes requested):
Need more details for this change

PR #3316, opened 2022-12-06
feat: grpc servicer implementation per version

aarnphm:


src/bentoml/_internal/server/grpc/servicer/v1alpha1/__init__.py
```diff
@@ -1,116 +1,41 @@
...
-
-    from bentoml.grpc.v1 import service_pb2 as pb
-    from bentoml.grpc.v1 import service_pb2_grpc as services
-    from bentoml.grpc.types import Interceptors
-    from bentoml.grpc.types import AddServicerFn
-    from bentoml.grpc.types import ServicerClass
-    from bentoml.grpc.types import BentoServicerContext
 
-    from ...service.service import Service
+    from bentoml.grpc.types import BentoServicerContext
```
sauyon, line 24: Why is only this one absolute?

src/bentoml/testing/grpc/__init__.py
```diff
@@ -39,17 +40,38 @@
 ]
 
 
-def randomize_pb_ndarray(shape: tuple[int, ...]) -> pb.NDArray:
+def create_bento_servicer(
```
sauyon, line 43: Wouldn't it make more sense to just take two arguments instead of returning a function?

src/bentoml/_internal/server/grpc/servicer/v1alpha1/__init__.py
aarnphm: oops updated now

src/bentoml/testing/grpc/__init__.py
aarnphm: yes

src/bentoml/_internal/service/service.py
```diff
@@ -221,10 +221,10 @@ def on_grpc_server_shutdown(self) -> None:
         pass
 
     @property
-    def grpc_servicer(self) -> Servicer:
-        from ..server.grpc_app import GRPCAppFactory
+    def grpc_servicer(self) -> GrpcServicerFactory:
```
sauyon, line 224: Can we just return a servicer instead?

sauyon (approved):
🎉

PR #3317, opened 2022-12-06
fix(monitor): disable validate for Formatter

bojiang:


src/bentoml/_internal/monitoring/default.py
```diff
@@ -45,7 +45,8 @@
 formatters:
   bentoml_json:
     class: pythonjsonlogger.jsonlogger.JsonFormatter
-    format: " "
+    format: "()"
+    validate: false
```
aarnphm, line 49: Can u leave a in line comment for the docs of this validate from python docs?

aarnphm (approved):

aarnphm:
I'm ok with this hack for now. Eventually, we will want to validate the config.

PR #3318, opened 2022-12-06
feat: `from_sample` docstring

aarnphm:
Add from_sample docstring for all IODescriptor

### DIFF
--- a/src/bentoml/_internal/io_descriptors/base.py
+++ b/src/bentoml/_internal/io_descriptors/base.py
@@ -173,7 +173,7 @@ class IODescriptor(ABC, _OpenAPIMeta, t.Generic[IOType]):
         return None
 
     @classmethod
-    def from_spec(cls, spec: dict[str, t.Any]) -> Self:
+    def from_spec(cls, spec: dict[str, t.Any]) -> t.Self:
         raise NotImplementedError
 
     @abstractmethod
--- a/src/bentoml/_internal/io_descriptors/base.py
+++ b/src/bentoml/_internal/io_descriptors/base.py
@@ -1,14 +1,14 @@
 from __future__ import annotations
 
 import typing as t
+import functools
 from abc import ABC
 from abc import abstractmethod
-from typing import TYPE_CHECKING
 from functools import update_wrapper
 
 from ...exceptions import InvalidArgument
 
-if TYPE_CHECKING:
+if t.TYPE_CHECKING:
     from types import UnionType
 
     import pyarrow
@@ -48,6 +48,26 @@ def from_spec(spec: dict[str, t.Any]) -> IODescriptor[t.Any]:
     return IO_DESCRIPTOR_REGISTRY[spec["id"]].from_spec(spec)
 
 
+def append_to_last_docstring(docstring: str):
+    def wrapper(f: F) -> F:
+        f.__doc__ += docstring
+        return f
+
+    return wrapper
+
+
+FROM_SAMPLE_NOTES = """\
+
+        .. note::
+
+            ``from_sample`` does not take positional arguments.
+"""
+
+append_from_sample_notes = functools.partial(
+    append_to_last_docstring, FROM_SAMPLE_NOTES
+)
+
+
 class _OpenAPIMeta:
     @abstractmethod
     def openapi_schema(self) -> Schema | Reference:
@@ -77,7 +97,7 @@ class IODescriptor(ABC, _OpenAPIMeta, t.Generic[IOType]):
     endpoint IO descriptor types in BentoServer.
     """
 
-    if TYPE_CHECKING:
+    if t.TYPE_CHECKING:
         # Populated by subclasses. Makes pyright happy.
         def __init__(self, **kwargs: t.Any) -> None:
             ...
@@ -107,21 +127,15 @@ class IODescriptor(ABC, _OpenAPIMeta, t.Generic[IOType]):
 
         cls._sample: IOType | None = None
 
-        from_sample_docstring = cls._from_sample.__doc__
-        if not from_sample_docstring:
-            base_docstring = IODescriptor[t.Any]._from_sample.__doc__
-            assert base_docstring is not None
-            from_sample_docstring = base_docstring.format(name=cls.__name__)
-
         def _() -> F:
             def impl(
                 cl_: type[IODescriptor[t.Any]], sample: t.Any, **kwargs: t.Any
             ) -> IODescriptor[t.Any]:
                 klass = cl_(**kwargs)
-                object.__setattr__(klass, "_sample", klass._from_sample(sample))
+                klass.sample = klass._from_sample(sample)
                 return klass
 
-            impl.__doc__ = from_sample_docstring
+            impl.__doc__ = cls._from_sample.__doc__
             return update_wrapper(impl, cls._from_sample)
 
         cls.from_sample = classmethod(_())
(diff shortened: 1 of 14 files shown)


src/bentoml/_internal/io_descriptors/base.py
```diff
@@ -64,27 +66,51 @@ def openapi_responses(self) -> dict[str, t.Any]:
...
 class IODescriptor(ABC, _OpenAPIMeta, t.Generic[IOType]):
     """
     IODescriptor describes the input/output data format of an InferenceAPI defined
     in a :code:`bentoml.Service`. This is an abstract base class for extending new HTTP
     endpoint IO descriptor types in BentoServer.
     """
 
-    __slots__ = ("_args", "_kwargs", "_proto_fields", "_mime_type", "descriptor_id")
+    if TYPE_CHECKING:
+        # Populated by subclasses. Makes IDE happy.
```
sauyon, line 90: What IDE is the problem, and what's the error?

src/bentoml/_internal/io_descriptors/base.py
```diff
@@ -64,27 +66,51 @@ def openapi_responses(self) -> dict[str, t.Any]:
...
+    __slots__ = (
+        "_mime_type",
+        "proto_fields",
+        "descriptor_id",
+        "from_sample",
+        "_sample",
+    )
 
-    descriptor_id: str | None
+    HTTP_METHODS = ("POST",)
```
sauyon, line 102: hm, do we actually use this?

src/bentoml/_internal/io_descriptors/base.py
```diff
@@ -64,27 +66,51 @@ def openapi_responses(self) -> dict[str, t.Any]:
...
+DEFAULT_FROM_SAMPLE_DOCSTRING = """\
+Creates an instance of the :class:`{name}` from given sample. Note that ``from_sample`` does not
+take positional arguments.
+
+Args:
+    sample: The sample to create the instance from.
+    kwargs: Additional keyword arguments to pass to the constructor.
+
+Returns:
+    :class:`{name}`: The instanc
```
sauyon, line 78: Incomplete?

src/bentoml/_internal/io_descriptors/base.py
```diff
@@ -93,10 +119,29 @@ def __init_subclass__(cls, *, descriptor_id: str | None = None):
...
 
-    if TYPE_CHECKING:
+        if proto_fields is None:
+            proto_fields = tuple()
+        cls.proto_fields = proto_fields
+        cls._sample: IOType | None = None
 
-        def __init__(self, **kwargs: t.Any) -> None:
-            ...
+        from_sample_docstring = inspect.getdoc(cls._from_sample)
```
sauyon, line 127: Let's not do this? We can just have a top-level `from_sample` doc.

sauyon (commented):
Preliminary comments, will get back to this in a bit.

src/bentoml/_internal/io_descriptors/base.py
aarnphm: oops.

aarnphm, line 78:
```suggestion
    :class:`{name}`: The IODescriptor instance.
```

src/bentoml/_internal/io_descriptors/base.py
aarnphm: pyright.

src/bentoml/_internal/io_descriptors/base.py
```diff
@@ -71,18 +72,24 @@ class IODescriptor(ABC, _OpenAPIMeta, t.Generic[IOType]):
...
+    HTTP_METHODS = ("POST",)
 
     _mime_type: str
-    _rpc_content_type: str = "application/grpc"
-    _proto_fields: tuple[ProtoField]
-    _sample: IOType | None = None
-    _args: t.Sequence[t.Any]
-    _kwargs: dict[str, t.Any]
+    descriptor_id: str | None
+    proto_fields: tuple[ProtoField]
```
sauyon, line 92: This class is exposed to users, we shouldn't make these public.

src/bentoml/_internal/io_descriptors/json.py
```diff
@@ -203,6 +203,47 @@ def __init__(
             )
 
     def _from_sample(self, sample: JSONType) -> JSONType:
+        """
+        Create a class:`JSON` IO Descriptor from given inputs.
+
+        Args:
+            sample: Given JSONType, which can be either dict, str, list.
```
sauyon, line 210: I don't think "JSONType" is helpful here.

sauyon (commented):
Quick initial pass.

src/bentoml/_internal/io_descriptors/base.py
aarnphm: But `descriptor_id` is already public right?

aarnphm:
cc @sauyon PTAL when you have bandwidth.

src/bentoml/_internal/io_descriptors/base.py
sauyon: Oh, yeah, it is.

src/bentoml/_internal/io_descriptors/base.py
```diff
@@ -47,6 +48,26 @@ def from_spec(spec: dict[str, t.Any]) -> IODescriptor[t.Any]:
...
+        return f
+
+    return wrapper
+
+
+FROM_SAMPLE_NOTES = """\
+
+        .. note::
+
+            ``from_sample`` does not take positional arguments.
```
sauyon, line 63: What does this mean?

aarnphm: This means `from_sample` will only accept kwargs arguments.

src/bentoml/_internal/io_descriptors/base.py
sauyon: If we're going to make `proto_fields` mandatory for gRPC we should probably make it a class arg like `descriptor_id`, I think.

sauyon: Maybe revert making it public, anyway; it'd technically be a breaking change.

Maybe that doesn't matter since gRPC isn't out yet, though?

src/bentoml/_internal/io_descriptors/numpy.py
```diff
@@ -471,6 +471,11 @@ def _from_sample(self, sample: ext.NpNDArray | t.Sequence[t.Any]) -> ext.NpNDArr
            @svc.api(input=input_spec, output=NumpyNdarray())
            async def predict(input: NDArray[np.int16]) -> NDArray[Any]:
                return await runner.async_run(input)
+
+        Raises:
+            :class:`BentoMLException`: If given sample is a type ``numpy.generic``. This exception
+                                       will also be raised if we failed to create a ``np.ndarray``
+                                       from given sample.
```
sauyon, line 478: If we're going to document this we should probably make it a better exception.

src/bentoml/_internal/io_descriptors/base.py
aarnphm: got it

sauyon (commented):
Few last things, sorry!

src/bentoml/_internal/io_descriptors/numpy.py
aarnphm: got it. Maybe I can put the exception PR up 😄

sauyon (approved):
:+1:

PR #3320, opened 2022-12-07
fix: remove assertion for dtype

aarnphm:
remove assertion check for `_dtype`, as pd.read_json and read_csv does accept dtype with bool for auto infer types


sauyon (approved):
Also deprecates passing non-dataframe input to `DataFrame.from_sample`.

PR #3099, opened 2022-10-14
feat: support batch inference

qu8n:
## What does this PR address?

BentoML supports serving models online through HTTP and gRPC but needs more support for batch and streaming inference. Integrating with Apache Spark helps bridge the gap. This PR involves allowing users to create Spark UDFs with BentoML-packaged ML models in different types of environment.

Fixes #890

## Before submitting:

- [x] Does the Pull Request follow [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/#summary) naming? Here are [GitHub's
      guide](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request) on how to create a pull request.
- [x] Does the code follow BentoML's code style, both `make format` and `make lint` script have passed ([instructions](https://github.com/bentoml/BentoML/blob/main/DEVELOPMENT.md#style-check-auto-formatting-type-checking))?
- [x] Did you read through [contribution guidelines](https://github.com/bentoml/BentoML/blob/main/CONTRIBUTING.md#ways-to-contribute) and follow [development guidelines](https://github.com/bentoml/BentoML/blob/main/DEVELOPMENT.md#start-developing)?
- [ ] Did your changes require updates to the documentation? Have you updated
      those accordingly? Here are [documentation guidelines](https://github.com/bentoml/BentoML/tree/main/docs) and [tips on writting docs](https://github.com/bentoml/BentoML/tree/main/docs#writing-documentation).
- [ ] Did you write tests to cover your changes?

PR #3321, opened 2022-12-07
feat: bentoserver client

quandollar:
This PR is related to PR #3099 and reflects step 2 of the internal Notion doc `Batch Inference Design`. Users can launch a bento server via a new API `bentoml.serve()` which returns a class that (1) manages the spawned server process and (2) creates a client for making requests to a BentoML server.


src/bentoml/bentos.py
```diff
@@ -419,3 +425,103 @@ def containerize(bento_tag: Tag | str, **kwargs: t.Any) -> bool:
     except Exception as e:  # pylint: disable=broad-except
         logger.error("Failed to containerize %s: %s", bento_tag, e)
         return False
+
+
+def serve(
+    bento: str,
+    production: bool = False,
+    port: int = BentoMLContainer.http.port.get(),
+    host: str = BentoMLContainer.http.host.get(),
```
aarnphm, line 434: Use `Provide` from `simple_di`

src/bentoml/bentos.py
```diff
@@ -4,9 +4,11 @@
 
 from __future__ import annotations
 
+import shutil
 import typing as t
 import logging
 from typing import TYPE_CHECKING
+from subprocess import Popen
```
aarnphm, line 11: nit:
```suggestion
import subprocess
```

src/bentoml/bentos.py
```diff
@@ -20,6 +22,10 @@
 from ._internal.configuration.containers import BentoMLContainer
 
 if TYPE_CHECKING:
+    from subprocess import Popen
```
aarnphm, line 25:
```suggestion
```

src/bentoml/bentos.py
```diff
@@ -419,3 +425,103 @@ def containerize(bento_tag: Tag | str, **kwargs: t.Any) -> bool:
...
+    ssl_ciphers: str | None = None,
+    # GRPC-specific args
+    enable_reflection: bool = BentoMLContainer.grpc.reflection.enabled.get(),
+    enable_channelz: bool = BentoMLContainer.grpc.channelz.enabled.get(),
+    max_concurrent_streams: int
+    | None = BentoMLContainer.grpc.max_concurrent_streams.get(),
+) -> Server:
+    """Launch a BentoServer and returns a client that exposes all APIs defined in target service"""
+
+    if type not in ["HTTP", "GRPC"]:
```
aarnphm, line 456: lets make this lower case,

src/bentoml/bentos.py
```diff
@@ -419,3 +425,103 @@ def containerize(bento_tag: Tag | str, **kwargs: t.Any) -> bool:
...
         logger.error("Failed to containerize %s: %s", bento_tag, e)
         return False
+
+
+def serve(
+    bento: str,
+    production: bool = False,
+    port: int = BentoMLContainer.http.port.get(),
+    host: str = BentoMLContainer.http.host.get(),
+    type: str = "HTTP",
```
aarnphm, line 435: avoid redefining reserved keywords. We should already have a lint rule for this.

```suggestion
    server_type: t.Literal["http", "grpc"] = "HTTP",
```

src/bentoml/bentos.py
```diff
@@ -419,3 +425,103 @@ def containerize(bento_tag: Tag | str, **kwargs: t.Any) -> bool:
...
+    max_concurrent_streams: int
+    | None = BentoMLContainer.grpc.max_concurrent_streams.get(),
+) -> Server:
+    """Launch a BentoServer and returns a client that exposes all APIs defined in target service"""
+
+    if type not in ["HTTP", "GRPC"]:
+        raise ValueError('Server type must either be "HTTP" or "GRPC"')
+
+    args = [
+        str(shutil.which("bentoml")),
```
aarnphm, line 460: lets not do this. lets do
```python
"-m", "bentoml"
```
and pass the executable `sys.executable` to subprocess.Popen

src/bentoml/bentos.py
```diff
@@ -419,3 +425,103 @@ def containerize(bento_tag: Tag | str, **kwargs: t.Any) -> bool:
...
+            args.extend(["--enable-channelz", str(enable_channelz)])
+        if max_concurrent_streams is not None:
+            args.extend(["--max-concurrent-streams", str(max_concurrent_streams)])
+
+    process = Popen(args)
+
+    return Server(process, host, port)
+
+
+class Server:
```
aarnphm, line 506: Why do we need this Server class?

note that bentoml.bentos is a public API.

src/bentoml/client.py
```diff
@@ -64,6 +64,23 @@ async def _call(
     ) -> t.Any:
         raise NotImplementedError
 
+    @staticmethod
+    def wait_until_server_is_ready(host: str, port: int, timeout: int) -> None:
```
aarnphm, line 68: This should just invoke `/readyz`?

aarnphm (commented):
initial review

src/bentoml/bentos.py
sauyon: We can probably do `if type.lower() not in ["http", "grpc"]`.

src/bentoml/client.py
sauyon: Probably.

src/bentoml/bentos.py
sauyon: Yeah, I told Quan to put it here temporarily. You can see why we need the class from its methods, though ;).

I wasn't really sure what a good place to put this was, something like `bentoml._internal.serve.Server`?

aarnphm: can we just call serve_http and serve_grpc in this case?

src/bentoml/client.py
```diff
@@ -64,6 +64,23 @@ async def _call(
     ) -> t.Any:
```
aarnphm, line 64: Should we also add an health for client?

src/bentoml/bentos.py
quandollar: Thanks, Aaron. For my notes, could you help me understand why this change is better? My understanding was that `from <x> import <y>` would have better performance than `import <x>`. Or is it because the improvement there would be trivial and `import <x>` is cleaner/more readable?

src/bentoml/bentos.py
quandollar: Thanks for catching this.

src/bentoml/bentos.py
quandollar: Resolving this and moving this question to Slack.

sauyon (approved):

sauyon:
:tada:

PR #3332, opened 2022-12-08
fix(grpc): handle backward protocol version

aarnphm:
Handle File changes from v1alpha1 to v1

refactor from #3256 

### DIFF
--- a/src/bentoml/_internal/io_descriptors/file.py
+++ b/src/bentoml/_internal/io_descriptors/file.py
@@ -5,6 +5,7 @@ import os
 import typing as t
 import logging
 from typing import TYPE_CHECKING
+from functools import lru_cache
 
 from starlette.requests import Request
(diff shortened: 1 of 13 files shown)


src/bentoml/_internal/io_descriptors/file.py
```diff
@@ -19,25 +20,25 @@
 from ...exceptions import InvalidArgument
 from ...exceptions import BentoMLException
 from ...exceptions import MissingDependencyException
+from ...grpc.utils import import_generated_stubs
```
bojiang, line 23: Is this importing lazy?

bojiang (approved):

src/bentoml/_internal/io_descriptors/file.py
aarnphm: yes.

PR #3333, opened 2022-12-08
feat: protobuf shim

aarnphm:
provides a light shim layer to support both protobuf 3 and protobuf 4

see https://developers.google.com/protocol-buffers/docs/news/2022-05-06 for the breaking changes.

This means bentoml will now include two sets of stubs (and so on when protobuf version moves up to 5,6++)


bojiang (approved):

aarnphm:
CI breakage is irrelevant from this PR.

PR #3336, opened 2022-12-09
fix: serve missing logic from #3321

aarnphm:

### DIFF
--- a/src/bentoml/_internal/server/server.py
+++ b/src/bentoml/_internal/server/server.py
@@ -1,31 +1,58 @@
-"""
-Server class for getting the Bento client and managing server process
-"""
-
 from __future__ import annotations
 
+import logging
+import traceback
 import subprocess
+from typing import TYPE_CHECKING
+
+import attr
+
+from ..utils import cached_property
+
+if TYPE_CHECKING:
+    from types import TracebackType
 
 
-class Server:
-    def __init__(self, process: subprocess.Popen[bytes], host: str, port: int) -> None:
-        self._process = process
-        self._host = host
-        self._port = port
+logger = logging.getLogger(__name__)
+
+
+@attr.frozen
+class ServerHandle:
+    process: subprocess.Popen[bytes]
+    host: str
+    port: int
+    timeout: int = attr.field(default=10)
+
+    @cached_property
+    def client(self):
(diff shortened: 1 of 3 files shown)


src/bentoml/_internal/server/server.py
```diff
@@ -1,31 +1,58 @@
...
+        self,
+        exc_type: type[BaseException],
+        exc_value: BaseException,
+        traceback_type: TracebackType,
+    ):
+        try:
+            self.stop()
+        except Exception as e:  # pylint: disable=broad-except
+            logger.error(f"Error stopping server: {e}", exc_info=e)
+        traceback.print_exception(exc_type, exc_value, traceback_type)
```
sauyon, line 58: Should this be in the except block?

src/bentoml/bentos.py
```diff
@@ -427,76 +427,86 @@ def containerize(bento_tag: Tag | str, **kwargs: t.Any) -> bool:
...
+        if host is None:
+            host = BentoMLContainer.http.host.get()
+        if port is None:
+            port = BentoMLContainer.http.port.get()
+            ssl_args.update(
+                ssl_keyfile_password=ssl_keyfile_password,
+                ssl_version=ssl_version,
+                ssl_cert_reqs=ssl_cert_reqs,
+                ssl_ciphers=ssl_ciphers,
+            )
```
sauyon, line 473: Er, why is this inside `if port is None`?

src/bentoml/bentos.py
```diff
@@ -427,76 +427,86 @@ def containerize(bento_tag: Tag | str, **kwargs: t.Any) -> bool:
...
         "--backlog",
         str(backlog),
+        *construct_ssl_args(**ssl_args),
     ]
     if production:
         args.append("--production")
-    if reload:
-        args.extend(["--reload", str(reload)])
+    elif reload:
+        args.append("--reload")
```
sauyon, line 499: Let's change this back; if the user specifies both we should let the CLI warn them.

src/bentoml/bentos.py
aarnphm: 😮‍💨 oopsie

sauyon (approved):

PR #3378, opened 2022-12-22
fix(containerize): fs path format on windows

bojiang:
fix https://bentoml.slack.com/archives/CKRANBHPH/p1671623024894159

image


sauyon (approved):
👍

PR #3396, opened 2023-01-06
feat: conda env for bentos in bentostore

jjmachan:
## Env Manager

The env manager allows you to run `bentoml serve` in an isolated environment. This PR adds support for creating conda environment

### Usage
```
> bentoml serve --env conda <bento>
```
The created environment is created under `$BENTOML_HOME/env` if `<bento>` is loaded to bentostore. For project-level invocation the created environment is ephemeral. 

To clean up the environment created, `rm -rf $BENTOML_HOME/env`

## Before submitting:

### DIFF
--- a/docs/source/guides/envmanager.rst
+++ b/docs/source/guides/envmanager.rst
@@ -1,43 +1,50 @@
-================
+===================
 Environment Manager
-================
+===================
 
-*Not supported on Windows*
+:bdg-info:`Note:` This feature is currently only supported on UNIX/MacOS.
 
-The Environment Manager creates an isolated environment based on the
-dependencies from the bentofile.yaml to run bentoml-cli commands. This means you
-can use the environment manager to create an environment and use it to run
-``bentoml serve`` which will then serve your bento in the isolated environment
-that is created. 
+Environment manager is a utility that helps create an isolated environment to
+run the BentoML CLI. Dependencies are pulled from your defined
+``bentofile.yaml`` and the environment is built upon request. This means by
+passing ``--env`` to supported CLI commands (such as :ref:`bentoml serve
+<reference/cli:serve>`), such commands will then be run in an sandbox
+environment that mimics the behaviour during production.
 
 .. code-block:: bash
 
    » bentoml serve --env conda iris_classifier:latest
 
-.. tip::
-
-    Depending on the environment, the env-manager will use different tools
-    behind the scenes to resolve and install the dependencies. The logs for
-    these tools are hidden by env-manager default. To view them, run bentoml 
-    in debug mode with ``--debug``/``--verbose`` flag.
-
-Caching
-~~~~~~~~~~~~
-
-There are 2 types of environment that can be created for you based on the serving
-target you invoke ``bentoml serve`` with. 
-(checkout :ref:`bentoml serving reference <reference/cli:serve>` for more
-information on different serving targets.)
+This creates and isolated conda environment from the dependencies in the bento
+and runs ``bentoml serve`` from that environment.
 
 
-1. Persistant environments - If the serving target is a Bento-Tag for a bento in
-   the bento store, the created environment will be persistant and will be saved 
-   to ``$BENTOML_HOME/env``. For subsequent invocations can use the same
-   environment.
-2. Ephimeral environments - In all other cases, ie if the serving target is the 
-   import path of a ‘bentoml.Service’ instance or a folder containing a valid 
-   ‘bentofile.yaml’ build file, the created environment will be removed after
-   the bentoml command has been complete.
+BentoML CLI Commands that support Environment Manager
+    - :ref:`serve <reference/cli:serve>`
+    - :ref:`serve-grpc <reference/cli:serve-grpc>`
+
+Supported Environments
+    - conda
+
+
+Caching strategies
+==================
+
+Currently, there are two types of environments that are supported by the
+environment manager:
+
+1. Persistent environment: If the given target is a Bento, then the created
+   environment will be stored locally to ``$BENTOML_HOME/env``. Such an
+   environment will then be cached and later used by subsequent invocations.
+
+ 2. Ephemeral environment: In cases where the given target is not a Bento
+     (import path to ``bentoml.Service``, project directory containing a valid
+     ``bentofile.yaml``), the environment will be created and cleanup up on
+     demand.
+
+.. note:: The current implementation will try to install the given dependencies
+   before running the CLI command. Therefore, the environment startup will be a
+   blocking call.  the bentoml command has been complete.
 
 .. note::
    You can run ``rm -rf $BENTOML_HOME/env`` to clear the cache.
--- a/docs/source/guides/envmanager.rst
+++ b/docs/source/guides/envmanager.rst
@@ -1,0 +1,43 @@
+================
+Environment Manager
+================
+
+*Not supported on Windows*
+
+The Environment Manager creates an isolated environment based on the
+dependencies from the bentofile.yaml to run bentoml-cli commands. This means you
+can use the environment manager to create an environment and use it to run
+``bentoml serve`` which will then serve your bento in the isolated environment
+that is created. 
+
+.. code-block:: bash
+
+   » bentoml serve --env conda iris_classifier:latest
+
+.. tip::
+
+    Depending on the environment, the env-manager will use different tools
+    behind the scenes to resolve and install the dependencies. The logs for
+    these tools are hidden by env-manager default. To view them, run bentoml 
+    in debug mode with ``--debug``/``--verbose`` flag.
+
+Caching
+~~~~~~~~~~~~
+
+There are 2 types of environment that can be created for you based on the serving
+target you invoke ``bentoml serve`` with. 
+(checkout :ref:`bentoml serving reference <reference/cli:serve>` for more
+information on different serving targets.)
+
+
+1. Persistant environments - If the serving target is a Bento-Tag for a bento in
+   the bento store, the created environment will be persistant and will be saved 
+   to ``$BENTOML_HOME/env``. For subsequent invocations can use the same
+   environment.
+2. Ephimeral environments - In all other cases, ie if the serving target is the 
+   import path of a ‘bentoml.Service’ instance or a folder containing a valid 
+   ‘bentofile.yaml’ build file, the created environment will be removed after
+   the bentoml command has been complete.
+
+.. note::
+   You can run ``rm -rf $BENTOML_HOME/env`` to clear the cache.
--- a/docs/source/guides/envmanager.rst
+++ b/docs/source/guides/envmanager.rst
@@ -18,6 +18,10 @@ environment that mimics the behaviour during production.
 This creates and isolated conda environment from the dependencies in the bento
 and runs ``bentoml serve`` from that environment.
 
+.. note:: The current implementation will try to install the given dependencies
+   before running the CLI command. Therefore, the environment startup will be a
+   blocking call.
+
 
 BentoML CLI Commands that support Environment Manager
     - :ref:`serve <reference/cli:serve>`
@@ -37,14 +41,10 @@ environment manager:
    environment will be stored locally to ``$BENTOML_HOME/env``. Such an
    environment will then be cached and later used by subsequent invocations.
 
- 2. Ephemeral environment: In cases where the given target is not a Bento
-     (import path to ``bentoml.Service``, project directory containing a valid
-     ``bentofile.yaml``), the environment will be created and cleanup up on
-     demand.
-
-.. note:: The current implementation will try to install the given dependencies
-   before running the CLI command. Therefore, the environment startup will be a
-   blocking call.  the bentoml command has been complete.
+2. Ephemeral environment: In cases where the given target is not a Bento (import
+   path to ``bentoml.Service``, project directory containing a valid
+   ``bentofile.yaml``), the environment will be created and cleanup up on
+   demand.
 
 .. note::
    You can run ``rm -rf $BENTOML_HOME/env`` to clear the cache.
--- a/src/bentoml/_internal/configuration/__init__.py
+++ b/src/bentoml/_internal/configuration/__init__.py
@@ -113,7 +113,6 @@ def set_debug_mode(enabled: bool) -> None:
     )
 
 
-@lru_cache(maxsize=1)
 def get_debug_mode() -> bool:
     if DEBUG_ENV_VAR in os.environ:
         return os.environ[DEBUG_ENV_VAR].lower() == "true"
--- a/src/bentoml/_internal/configuration/__init__.py
+++ b/src/bentoml/_internal/configuration/__init__.py
@@ -113,6 +113,7 @@ def set_debug_mode(enabled: bool) -> None:
     )
 
 
+@lru_cache(maxsize=1)
 def get_debug_mode() -> bool:
     if DEBUG_ENV_VAR in os.environ:
         return os.environ[DEBUG_ENV_VAR].lower() == "true"
--- a/src/bentoml/_internal/configuration/containers.py
+++ b/src/bentoml/_internal/configuration/containers.py
@@ -176,9 +176,9 @@ class _BentoMLContainerClass:
         )
         bentos = os.path.join(home, "bentos")
         models = os.path.join(home, "models")
-        env = os.path.join(home, "envs")
+        envs = os.path.join(home, "envs")
 
-        validate_or_create_dir(home, bentos, models, env)
+        validate_or_create_dir(home, bentos, models, envs)
         return home
 
     @providers.SingletonFactory
--- a/src/bentoml/_internal/configuration/containers.py
+++ b/src/bentoml/_internal/configuration/containers.py
@@ -29,6 +29,8 @@ from ...exceptions import BentoMLConfigException
 from ..utils.unflatten import unflatten
 
 if TYPE_CHECKING:
+    from fs.base import FS
+
     from .. import external_typing as ext
     from ..models import ModelStore
     from ..utils.analytics import ServeInfo
@@ -160,7 +162,6 @@ class BentoMLConfiguration:
 
 @dataclass
 class _BentoMLContainerClass:
-
     config = providers.Configuration()
 
     @providers.SingletonFactory
@@ -194,6 +195,13 @@ class _BentoMLContainerClass:
     @staticmethod
     def env_store_dir(bentoml_home: str = Provide[bentoml_home]):
         return os.path.join(bentoml_home, "envs")
+
+    @providers.SingletonFactory
+    @staticmethod
+    def env_store(bentoml_home: str = Provide[bentoml_home]) -> FS:
+        import fs
+
+        return fs.open_fs(os.path.join(bentoml_home, "envs"))
 
     @providers.SingletonFactory
     @staticmethod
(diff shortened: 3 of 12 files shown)


src/bentoml_cli/env_manager.py
```diff
@@ -0,0 +1,78 @@
...
+        if re.match(r"^--env=?.*", arg):
+            indices_to_remove.append(i)
+            if "=" not in arg:
+                indices_to_remove.append(i + 1)
+
+    new_cmd_args: list[str] = []
+    for i, arg in enumerate(cmd_args):
+        if i not in indices_to_remove:
+            new_cmd_args.append(arg)
+    return new_cmd_args
```
aarnphm, line 34: Should this be handled by `click.option`?

src/bentoml_cli/env_manager.py
```diff
@@ -0,0 +1,78 @@
...
+                from_bento_store = True
+                bento_path = bento._fs.getsyspath("")
+            except NotFound:
+                # env created will be ephemeral
+                from_bento_store = False
+            bento_env = EnvManager.get_environment(
+                env_name=bento_tag,
+                env_type=env,
+                from_bento_store=from_bento_store,
+                bento_path=bento_path,
```
aarnphm, line 68:
```suggestion
                bento_path=bento.path or "",
```
There is a property `path` for bento which is the equivalent of this.

If you need to get other path then you can do
```python
path = bento.path_of("/")
```

src/bentoml/_internal/env_manager/manager.py
```diff
@@ -0,0 +1,127 @@
...
+import os
+import typing as t
+import logging
+import subprocess
+from tempfile import NamedTemporaryFile
+from distutils.spawn import find_executable as which
+
+import fs
+import attrs
+from rich.status import Status
```
aarnphm, line 12: We are not depending on rich as a dependency, I suggest not doing this as rich is pretty slow.

src/bentoml/_internal/env_manager/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .cli import env_manager
+from .cli import remove_env_arg
```
aarnphm, line 2: wrong import?

src/bentoml/_internal/env_manager/manager.py
```diff
@@ -0,0 +1,127 @@
+from __future__ import annotations
+
+import os
+import typing as t
+import logging
+import subprocess
+from tempfile import NamedTemporaryFile
+from distutils.spawn import find_executable as which
```
aarnphm, line 8: ugh why can't we just use `shutil.which`?

src/bentoml/_internal/env_manager/manager.py
```diff
@@ -0,0 +1,127 @@
...
+def decode(msg: bytes) -> str:
+    if msg:
+        return msg.decode("utf-8")
+    return ""
+
+
+def run_script_subprocess(
+    script_file_path: t.Union[str, os.PathLike[str]], capture_output: bool
+):
+    shell_path = which("bash")
```
aarnphm, line 33: Does this mean we are depending on bash?

How does this work on Windows?

src/bentoml/_internal/env_manager/manager.py
```diff
@@ -0,0 +1,127 @@
...
+    if shell_path is None:
+        raise BentoMLException("Unable to locate a valid shell")
+
+    result = subprocess.run(
```
aarnphm, line 37: should we use `subprocess.check_output` here instead of this?

src/bentoml/_internal/env_manager/manager.py
```diff
@@ -0,0 +1,127 @@
...
+        [shell_path, "-euxo", "pipefail", script_file_path],
+        capture_output=capture_output,
+    )
+    if result.returncode != 0:
+        if result.stderr and result.stdout:
+            logger.debug(decode(result.stdout))
+            logger.error(decode(result.stderr))
```
aarnphm, line 44: I think that we should always log error no matter what.

src/bentoml/_internal/env_manager/manager.py
```diff
@@ -0,0 +1,127 @@
...
+@attrs.define
+class EnvManager:
+    env_type: str
+    env_name: str
+    from_bento_store: bool
+    bento_path: t.Optional[str]
+    _env_fs: FS = attrs.field(init=False)
+
+    def __attrs_post_init__(self):
+        env_home = fs.open_fs("/tmp/env_manager", create=True)
```
aarnphm, line 61: you should use fs.temp_fs instead. This won't work on Windows.

src/bentoml_cli/env_manager.py
jjmachan: the env_manager has to clear this one up before passing the args to the subroutine.

src/bentoml_cli/env_manager.py
jjmachan: ahh, thanks for pointing that out

src/bentoml/_internal/env_manager/manager.py
jjmachan: as discussed windows support will be added in the following PR

src/bentoml/_internal/env_manager/manager.py
jjmachan: changed to an env_store in $BENTO_HOME

src/bentoml/_internal/env_manager/manager.py
jjmachan: yep, changed

src/bentoml/_internal/env_manager/manager.py
jjmachan: but we are depending on rich already right? the idea was to put in a spinner but since we already have rich I thought  we will use the one already there

src/bentoml/_internal/env_manager/manager.py
jjmachan: its the same right as `subprocess.run()` and manually checking the return codes right?

aarnphm: do we need to manually check the return code?

src/bentoml/_internal/env_manager/manager.py
aarnphm: what is the decision on creating an env_store under $BENTOML_HOME?

src/bentoml/_internal/env_manager/manager.py
```diff
@@ -0,0 +1,189 @@
...
+            with NamedTemporaryFile(mode="w", delete=False) as script_file:
+                conda_env_path = self._env_fs.getsyspath(self.env_name)
+                python_version = get_python_version_from_bento(self.bento_path)
+                script_file.write(
+                    f"conda create -p {conda_env_path} python={python_version} --yes"
+                    + "\n"
+                )
+
+                # install conda deps
+                from bentoml._internal.bento.build_config import (
```
aarnphm, line 146: nit: relative import

src/bentoml_cli/env_manager.py
```diff
@@ -0,0 +1,118 @@
...
+        type=click.Choice(["conda"]),
+        default=None,
+        help="Environment to run the command in",
+        show_default=True,
+    )
+    @functools.wraps(func)
+    def wrapper(*args: t.Any, **kwargs: t.Any) -> t.Any:
+        env = kwargs.pop("env")
+        if env is not None:
+            rich_console.print(f"loading {env} environment...")
```
aarnphm, line 105: lets just use click.echo here.

aarnphm (commented):
fwiw for all newer type we should use https://peps.python.org/pep-0604/ for consistency.

There are some typo here and there. I will dm you on slack for more.

src/bentoml_cli/env_manager.py
```diff
@@ -0,0 +1,118 @@
...
+                indices_to_remove.append(i + 1)
+
+    new_cmd_args: list[str] = []
+    for i, arg in enumerate(cmd_args):
+        if i not in indices_to_remove:
+            new_cmd_args.append(arg)
+    return new_cmd_args
+
+
+def get_environment(bento_identifier: str, env: str) -> EnvManager:
```
aarnphm, line 43:
```suggestion
@inject
def get_environment(bento_identifier: str, env: str. bento_store=Provide[BentoMLContainer.bento_store]) -> EnvManager:
```

jjmachan:
After talking @aarnphm, other than these comments, need to make the following changes
- [ ] `run_script` should only write to find right before execution to minimize scope for script injection attacks
- [ ] Plan out the interface for `conda`, `docker` and other environments
- [ ] Add a caching mechanism to optimize disk usage and env init times. Only create new envs if the dependencies in `bentofile.yaml` changed.

src/bentoml/_internal/env_manager/manager.py
```diff
@@ -0,0 +1,189 @@
...
+            "-euxo",
+            "pipefail",
+        ]
+    result = subprocess.run(
+        [shell_path, *safer_bash_args, script_file_path],
+        capture_output=capture_output,
+    )
+    if result.returncode != 0:
+        logger.debug(decode(result.stdout))
+        logger.error(decode(result.stderr))
```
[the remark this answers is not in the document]
jjmachan: I don't know, I think the first one is more readable :swe

src/bentoml/_internal/env_manager/manager.py
```diff
@@ -0,0 +1,189 @@
...
+    shell_path = which("bash")
+    if shell_path is None:
+        raise BentoMLException("Unable to locate a valid shell")
+
+    safer_bash_args: list[str] = []
+
+    # but only work in bash
+    if debug_mode:
+        safer_bash_args = [
+            "-euxo",
```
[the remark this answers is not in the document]
jjmachan: removed pipe and make `x` only for debug mode

src/bentoml/_internal/env_manager/manager.py
jjmachan: not all envs are ephemeral. If loading from bento then it is perisitant.

src/bentoml/_internal/env_manager/manager.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def create_conda_env(self) -> str:
+        """
+        Create a new conda env with self.env_name
+        """
+        conda_exe = os.environ.get("CONDA_EXE")
+        if conda_exe is None:
+            raise BentoMLException(
+                "conda executable not found! Make sure conda is installed and that `CONDA_EXE` is set."
+            )
```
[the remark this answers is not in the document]
jjmachan: this is an excellent suggestion! I didn't realize `CONDA_EXE` was not universal, seems like only conda and miniconda implements it. but when I checked `CONDA_PREFIX` is also not universal and is not present if the env is deactivated in conda/miniconda. I think we might have to check both in this case.

maybe mambda/micromaba should be separate env as well. Because if all of them are installed, envManager will use the first one it gets, and no way for the user to configure that operation. That being said, I don't think that is a big problem either.

docs/source/guides/envmanager.rst
```diff
@@ -0,0 +1,43 @@
...
+Environment Manager
+================
+
+*Not supported on Windows*
+
+The Environment Manager creates an isolated environment based on the
+dependencies from the bentofile.yaml to run bentoml-cli commands. This means you
+can use the environment manager to create an environment and use it to run
+``bentoml serve`` which will then serve your bento in the isolated environment
+that is created.
```
aarnphm, line 11: some grammar
```suggestion
Environment manager is a utility that helps create an isolated environment to run the BentoML CLI. Dependencies are pulled from your defined ``bentofile.yaml`` and the environment is built upon request. 

This means by passing ``--env`` to supported CLI commands (such as :ref:``<add_link_to_bentoml_serve>``), such commands will then be run in an sandbox environment that mimics the behaviour during production.
```

docs/source/guides/envmanager.rst
```diff
@@ -0,0 +1,43 @@
...
+.. code-block:: bash
+
+   » bentoml serve --env conda iris_classifier:latest
+
+.. tip::
+
+    Depending on the environment, the env-manager will use different tools
+    behind the scenes to resolve and install the dependencies. The logs for
+    these tools are hidden by env-manager default. To view them, run bentoml 
+    in debug mode with ``--debug``/``--verbose`` flag.
```
aarnphm, line 22: I think this is not needed to be here.

```suggestion
```

docs/source/guides/envmanager.rst
```diff
@@ -0,0 +1,43 @@
...
+
+Caching
+~~~~~~~~~~~~
```
aarnphm, line 25:
```suggestion
Caching strategies
==================
```

docs/source/guides/envmanager.rst
```diff
@@ -0,0 +1,43 @@
...
+
+There are 2 types of environment that can be created for you based on the serving
+target you invoke ``bentoml serve`` with. 
+(checkout :ref:`bentoml serving reference <reference/cli:serve>` for more
+information on different serving targets.)
```
aarnphm, line 30:
```suggestion
Currently, there are two types of environments that are supported by the environment manager:
```

docs/source/guides/envmanager.rst
```diff
@@ -0,0 +1,43 @@
...
+
+
+1. Persistant environments - If the serving target is a Bento-Tag for a bento in
+   the bento store, the created environment will be persistant and will be saved 
+   to ``$BENTOML_HOME/env``. For subsequent invocations can use the same
+   environment.
```
aarnphm, line 36:
```suggestion
1. Persistent environment: If the given target is a Bento, then the created environment will be stored locally to ``$BENTOML_HOME/env``. Such an environment will then be cached and later used by subsequent invocations.
```

docs/source/guides/envmanager.rst
```diff
@@ -0,0 +1,43 @@
...
+2. Ephimeral environments - In all other cases, ie if the serving target is the 
+   import path of a ‘bentoml.Service’ instance or a folder containing a valid 
+   ‘bentofile.yaml’ build file, the created environment will be removed after
+   the bentoml command has been complete.
```
aarnphm, line 40:
```suggestion
2. Ephemeral environment: In cases where the given target is not a Bento (import path to ``bentoml.Service``, project directory containing a valid ``bentofile.yaml``), the environment will be created and cleanup up on demand.

.. note::

	The current implementation will try to install the given dependencies before running the CLI command. Therefore, the environment startup will be a blocking call.
```

src/bentoml/_internal/configuration/__init__.py
```diff
@@ -113,6 +113,7 @@ def set_debug_mode(enabled: bool) -> None:
     )
 
 
+@lru_cache(maxsize=1)
```
aarnphm, line 116: defintely should not cache this. 

should revert.

src/bentoml_cli/env_manager.py
```diff
@@ -0,0 +1,127 @@
...
+                is_ephemeral=True,
+            ).environment
+        elif bento_path_fs.isfile(DEFAULT_BENTO_BUILD_FILE):
+            # path to a bento project
+            raise NotImplementedError(
+                "Serving bento project in an environment is not supported now."
+            )
+        else:
+            raise BentoMLException(
+                f"EnvManager failed to create environment from path {bento_path_fs}. When loading from a path, it must be either a Bento containing bento.yaml or a project directory containing bentofile.yaml"
```
aarnphm, line 78:
```suggestion
                f"EnvManager failed to create an environment from path {bento_path_fs}. When loading from a path, it must be either a Bento or a project directory containing 'bentofile.yaml'."
```

src/bentoml_cli/env_manager.py
```diff
@@ -0,0 +1,127 @@
...
+        try:
+            bento = bento_store.get(bento_identifier)
+            return EnvManager.from_bento(
+                env_type=env,
+                bento=bento,
+                is_ephemeral=False,
+            ).environment
+        except BentoNotFound:
+            # service definition
+            bento_path_fs = os.path.join(os.getcwd(), DEFAULT_BENTO_BUILD_FILE)
```
aarnphm, line 90:
```suggestion
```

src/bentoml_cli/env_manager.py
```diff
@@ -0,0 +1,127 @@
...
+            from rich.status import Status
+
+            bento_identifier = kwargs["bento"]
+            spinner_status = Status(f"Loading {env} environment")
+            if not get_debug_mode():
+                spinner_status.start()
+                bento_env = get_environment(bento_identifier, env)
+                spinner_status.stop()
+            else:
+                bento_env = get_environment(bento_identifier, env)
```
aarnphm, line 117: I think we should just use the spinner as a context manager regardless of debug mode or not.

jjmachan: ohh that case the logs just get messed up. The spinner is only so that the user can see something is happening and the process is still alive

src/bentoml/_internal/configuration/__init__.py
jjmachan: ok

aarnphm: If you just evict the last one, then what is the point of caching 😄 

Also it is just O(1) access, no need to cache.

src/bentoml_cli/env_manager.py
aarnphm: but here you are only shows it during debug right? Shouldn't we show it regardless?

jjmachan: That would have been ideal but the spinner does not render properly if there are a lot of logs dumped to STDOUT which is true in this case, which is why we turn it off in debug mode (conda and pip put a  log of info)

aarnphm:
Hi @jjmachan, any updates on this PR?

src/bentoml/_internal/env_manager/manager.py
```diff
@@ -0,0 +1,62 @@
...
+
+class EnvManager:
+    @inject
+    def __init__(
+        self,
+        env_name: str,
+        env_type: str,
+        is_ephemeral: bool,
+        bento: Bento,
+        env_store_dir: str = Provide[BentoMLContainer.env_store_dir],
```
aarnphm, line 25:
```suggestion
        env_store: str = Provide[BentoMLContainer.env_store],
```

src/bentoml/_internal/env_manager/envs.py
```diff
@@ -0,0 +1,160 @@
...
+    # path to bento's /env dir
+    bento: Bento
+    env_exe: str = attr.field(init=False)
+
+    def __attrs_post_init__(self):
+        self.env_exe = self.get_executable()
+        if self.env_fs.exists(self.name):
+            return self.env_fs.getsyspath(self.name)
+        else:
+            self.create()
```
aarnphm, line 44:
```suggestion
        if not self.env_fs.exists(self.name):
            self.create()
```

src/bentoml/_internal/configuration/containers.py
```diff
@@ -175,8 +176,9 @@ def bentoml_home() -> str:
         )
         bentos = os.path.join(home, "bentos")
         models = os.path.join(home, "models")
+        env = os.path.join(home, "envs")
```
aarnphm, line 179:
```suggestion
        envs = os.path.join(home, "envs")
```

src/bentoml/_internal/configuration/containers.py
```diff
@@ -175,8 +176,9 @@ def bentoml_home() -> str:
...
 
-        validate_or_create_dir(home, bentos, models)
+        validate_or_create_dir(home, bentos, models, env)
```
aarnphm, line 181:
```suggestion
        validate_or_create_dir(home, bentos, models, envs)
```

aarnphm (approved):

Issue #3399, opened 2023-01-06
bug: Invalid dockerfile templates while containerize


aarnphm:
### Describe the bug

the `FileNotFoundError` is raised since at `containerize` time, we don't have access to this templates.

The templates is already included in the generated Dockerfile

```prolog
» b containerize iris_classifier:latest --opt platform=linux/amd64,linux/arm64 --backend buildx
Building OCI-compliant image for iris_classifier:bd355oelc2e5xgxi with buildx

Traceback (most recent call last):
  File "/Users/aarnphm/.pyenv/versions/3.10.8/bin/bentoml", line 8, in <module>
    sys.exit(cli())
  File "/Users/aarnphm/.pyenv/versions/3.10.8/lib/python3.10/site-packages/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "/Users/aarnphm/.pyenv/versions/3.10.8/lib/python3.10/site-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/Users/aarnphm/.pyenv/versions/3.10.8/lib/python3.10/site-packages/click/core.py", line 1659, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/aarnphm/.pyenv/versions/3.10.8/lib/python3.10/site-packages/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/aarnphm/.pyenv/versions/3.10.8/lib/python3.10/site-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/Users/aarnphm/workspace/bentoml/aarnphm-fork/src/bentoml_cli/utils.py", line 268, in wrapper
    return func(*args, **kwargs)
  File "/Users/aarnphm/workspace/bentoml/aarnphm-fork/src/bentoml_cli/utils.py", line 217, in wrapper
    return func(*args, **kwargs)
  File "/Users/aarnphm/workspace/bentoml/aarnphm-fork/src/bentoml_cli/utils.py", line 196, in wrapper
    return func(*args, **kwargs)
  File "/Users/aarnphm/workspace/bentoml/aarnphm-fork/src/bentoml_cli/utils.py", line 55, in wrapper
    return _f(*args, **transformed)
  File "/Users/aarnphm/workspace/bentoml/aarnphm-fork/src/bentoml_cli/containerize.py", line 557, in containerize
    result = container.build(
[...]

PR #3400, opened 2023-01-06
fix: #3399 during `containerize`

aarnphm:
See #3399 for more details.


bojiang (changes requested):
Is it possible to leave them frozen and just create a replica of the BentoInfo object with `dockerfile_template = None`

bojiang (dismissed):

bojiang (approved):

PR #3416, opened 2023-01-11
fix: regression f-string

aarnphm:


aarnphm:
introduced by the logging refactor for CLI @sauyon 😄

sauyon (approved):

Issue #3420, opened 2023-01-12
bug: models validation error incorrectly prints classes during runner instantiation


characat0:
### Describe the bug

During runner instantiation, the `models` argument is expected to be a list of `bentoml.Model`, if any is different, the program raises a ValueError that is supposed to include the types that are not subtypes of `bentoml.Model`, but the condition filters out every element that is not an instance of `bentoml.Model`, resulting in the opposite behavior.

### To reproduce

**Current behavior**
```py
from bentoml import Runner, Runnable
Runner(Runnable, models=[1, 'a', {}])
```

outputs
```sh
Using lowercased runnable class name 'runnable' for runner.
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "...\BentoML\src\bentoml\_internal\runner\runner.py", line 130, in __init__
    raise ValueError(
ValueError: models must be a list of 'bentoml.Model'. Got set() instead.
```

### Expected behavior

```sh
Using lowercased runnable class name 'runnable' for runner.
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "...\BentoML\src\bentoml\_internal\runner\runner.py", line 130, in __init__
    raise ValueError(
ValueError: models must be a list of 'bentoml.Model'. Got {<class 'dict'>, <class 'str'>, <class 'int'>} instead.
```

#### Environment variable

```bash
BENTOML_DEBUG=''
BENTOML_QUIET=''
BENTOML_BUNDLE_LOCAL_BUILD=''
BENTOML_DO_NOT_TRACK=''
BENTOML_CONFIG=''
BENTOML_CONFIG_OPTIONS=''
BENTOML_PORT=''
BENTOML_HOST=''
BENTOML_API_WORKERS=''
```

#### System information

`bentoml`: 1.0.12.post40+g213b584a
`python`: 3.10.4
`platform`: Windows-10-10.0.22621-SP0
`is_window_admin`: False
<details><summary><code>pip_packages</code></summary>

<br>

```
aiohttp==3.8.3
aiosignal==1.3.1
alabaster==0.7.12
anyio==3.6.2
appdirs==1.4.4
asgiref==3.6.0
astroid==2.13.2
asttokens==2.2.1
async-timeout==4.0.2
attrs==22.2.0
Babel==2.11.0
backcall==0.2.0
backoff==2.2.1
beautifulsoup4==4.11.1
[...]

PR #3421, opened 2023-01-12
fix(runner): log correct error types during model validation

characat0:
## What does this PR address?

Fixes https://github.com/bentoml/BentoML/issues/3420

This is a small change but helps debugging (at least for the project I was working on).


aarnphm (approved):

aarnphm:
Thanks for the catch.

PR #3428, opened 2023-01-15
integration: Triton Inference Server

aarnphm:
To create a new triton runner, create with `bentoml.triton.Runner`

```python
import bentoml

triton_runner = bentoml.triton.Runner("triton-runners", model_repository="/path/to/model_repository")
svc = bentoml.Service("service", runners=[triton_runner])
```

Model repository can also accept s3 path:

```python
triton_runner = bentoml.triton.Runner("triton-runners", model_repository="s3://path/to/model_repository")
```

An example of inference API with this triton runner

```python
@svc.api(
    input=bentoml.io.NumpyNdarray.from_sample(
        np.array([[1, 2, 3, 4]]), enforce_dtype=False
    ),
    output=bentoml.io.NumpyNdarray.from_sample(np.array([1])),
)
def triton_infer(input_data: NDArray[np.float16]) -> NDArray[np.int16]:
    # Create the data for the two input tensors. Initialize the first
    # to unique integers and the second to all ones.
    iris_res = iris_clf_runner.run(input_data)
    res_kwargs = triton_runner.txt2img.run(IRIS_NAME=iris_res)
    return iris_res
```

Note that each attribute of the `triton_runner` includes the name of all given models under model_repository

Add the following to your `bentofile.yaml`:

```yaml
service: "service.py:svc"
include:
  - "*.py"
  - "model_repository/"
docker:
  base_image: nvcr.io/nvidia/tritonserver:22.12-py3
```

> tritonserver are currently only supported with `--production` tag. Make sure to have `tritonserver` binary available in PATH if running locally.

To pass triton arguments to `serve` do it via `--triton-options ARG=VALUE`

```bash
bentoml serve --production --triton-options allow-metrics=False --triton-options grpc-port=8002
```


.gitignore
```diff
@@ -151,3 +151,6 @@ composer.lock
 vendor
 *.pkl
 *.ckpt
+
+# tritonserver directory binary
+opt
```
sauyon, line 156:
```suggestion
/opt
```

maybe?

src/bentoml/_internal/bento/bento.py
```diff
@@ -198,9 +200,16 @@ def create(
         for model in svc.models:
             models.add(model)
         # Add all models required by service runners
+        has_triton_runner = False
         for runner in svc.runners:
-            for model in runner.models:
-                models.add(model)
+            # only add models if given runner is not a TritonRunner
```
sauyon, line 205: Why do we need this? Are we specifying models in the triton runner?

src/bentoml/_internal/bento/bento.py
```diff
@@ -370,6 +381,8 @@ class BentoRunnerInfo:
 
     @classmethod
     def from_runner(cls, r: Runner) -> BentoRunnerInfo:
+        if LazyType["TritonRunner"]("bentoml.triton.Runner").isinstance(r):
+            return cls(name=r.name, runnable_type=r.runnable_class.__name__)
```
sauyon, line 385: Is this necessary? Shouldn't the triton runner just have an empty list for `models` and `None` for config?

src/bentoml/triton.py
```diff
@@ -0,0 +1,430 @@
...
+            logger.warning(
+                "Converting runner name '%s' to lowercase: '%s'", name, lname
+            )
+
+        # TODO: Support configuration (P1)
+
+        # The logic below mimic the behaviour of Exportable.import_from
+        # To determine models inside given model repository.
+        try:
+            parsed_model_repository = fs.opener.parse(model_repository)
```
sauyon, line 81: Hm, why are we using `fs` here? I thought we were just passing this through to triton?

src/bentoml/_internal/bento/bento.py
aarnphm: triton runners models is just a list of string, not actual model in the store atm.

src/bentoml/triton.py
aarnphm: we need to find out a way to assign the model inside the store as attributes to the runner.

src/bentoml/_internal/runner/runner_handle/remote.py
```diff
@@ -270,3 +288,178 @@ async def is_ready(self, timeout: int) -> bool:
...
+        inputs: list[tritongrpcclient.InferInput] = []
+
+        if len(args) > 0:
+            for (infer_input, arg) in zip(model_metadata.inputs, input_params.args):
+                InferInput = tritongrpcclient.InferInput(
+                    infer_input.name,
+                    arg.shape,
+                    tritongrpcclient.np_to_triton_dtype(arg.dtype),
+                )
+                InferInput.set_data_from_numpy(arg)
```
sauyon, line 434: Why don't we just create `InferInput`s in the data containers? That feels more natural.

src/bentoml/_internal/server/grpc/servicer/v1/__init__.py
```diff
@@ -81,20 +82,48 @@ async def Call(  # type: ignore (no async types) # pylint: disable=invalid-overr
...
+
                 res = await api.output.to_proto(output)
                 # TODO(aarnphm): support multiple proto fields
                 response = pb.Response(**{api.output._proto_fields[0]: res})
             except BentoMLException as e:
                 log_exception(request, sys.exc_info())
+                if output is not None:
+                    # more descriptive errors if output is available
+                    logger.error(
+                        "Function '%s' has 'input=%s,output=%s' as IO descriptor, and returns 'result=%s'",
```
sauyon, line 94: Hm, is this useful information for all of these exceptions? Seems like we only really want to print this when the result type is wrong.

src/bentoml/_internal/runner/runner.py
```diff
@@ -24,116 +26,172 @@
...
+        validate_tag_str(value)
+    except ValueError as e:
+        # TODO: link to tag validation documentation
+        raise ValueError(
+            f"Runner name '{value}' is not valid; it must be a valid BentoML Tag name."
+        ) from e
+
+
+@attr.define(slots=False, frozen=True)
+class RunnerMeta(ABC):
```
sauyon, line 77:
```suggestion
class AbstractRunner(ABC):
```

src/bentoml/_internal/bento/bento.py
```diff
@@ -373,8 +373,8 @@ def from_runner(cls, r: Runner) -> BentoRunnerInfo:
         return cls(
             name=r.name,
             runnable_type=r.runnable_class.__name__,
-            models=[str(model.tag) for model in r.models],
-            resource_config=r.resource_config,
+            models=[str(model.tag) for model in r.models] if r.models else [],
+            resource_config=getattr(r, "resource_config", None),
```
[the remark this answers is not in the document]
aarnphm: noted.

src/bentoml/_internal/server/grpc/servicer/v1/__init__.py
aarnphm: I think it is for each exception.

src/bentoml_cli/serve.py
```diff
@@ -123,6 +127,14 @@ def add_serve_command(cli: click.Group) -> None:
         help="Ciphers to use (see stdlib 'ssl' module)",
         show_default=True,
     )
+    @click.option(
+        "--triton-options",
+        help="Trition Inference Server options",
+        required=False,
+        multiple=True,
+        callback=opt_callback,
+        metavar="ARG=VALUE[,ARG=VALUE]",
```
aarnphm, line 136:
```suggestion
        metavar="ARG=VALUE[,VALUE]",
```

src/bentoml_cli/start.py
```diff
@@ -228,6 +231,14 @@ def start_http_server(  # type: ignore (unused warning)
         default=None,
         show_default=True,
     )
+    @click.option(
+        "--triton-options",
+        help="Trition Inference Server options",
+        required=False,
+        multiple=True,
+        callback=opt_callback,
+        metavar="ARG=VALUE[,ARG=VALUE]",
```
aarnphm, line 240:
```suggestion
        metavar="ARG=VALUE[,VALUE]",
```

sauyon (approved):
🎉

PR #3461, opened 2023-01-20
docs: fix spark docs

sauyon:
Fixes formatting and the API reference not working.


aarnphm (commented):
It stills didnt' show under the API reference. Maybe you also need

```rst

.. currentmodule:: bentoml.batch

```

before the autofunction call

sauyon:
Hm, that's odd, it worked for me locally.

ssheng (approved):

ssheng:
I think we are still missing this file, @sauyon . https://docs.bentoml.org/en/latest/integrations/spark/input.csv

PR #3471, opened 2023-01-22
example: Triton Inference Server

aarnphm:
Depends on #3502


examples/triton_runner/README.md
```diff
@@ -0,0 +1,103 @@
+# Triton Inference Server integration
+
+BentoML now provides support for Triton Inference Server. Triton can be accessed via `bentoml.triton.Runner`:
```
ssheng, line 3: Let's add a few bullet points on why users would want to use Triton. A few suggestions.
- User is already using Triton. They can unify model management with other frameworks.
- Performance gaps. While BentoML is constantly improving, under some conditions Triton performs better.
- Adding pre-process logic or multli-model inference graph

examples/triton_runner/README.md
```diff
@@ -0,0 +1,103 @@
...
+docker:
+  base_image: nvcr.io/nvidia/tritonserver:22.12-py3
+```
+
+> tritonserver are currently only supported with `--production` tag. Make sure to have `tritonserver` binary available in PATH if running locally.
+
+To pass triton arguments to `serve` do it via `--triton-options ARG=VALUE[, VALUE]`
+
+```bash
+bentoml serve --production --triton-options log-verbose=True
```
ssheng, line 50: `--triton-options` should be specific to the Triton runner as opposed to the `serve` command, right? If we have two Triton runners, which one will `--triton-options` affect? Should we put all the Triton options as input arguments to the Triton runner?

ssheng (commented):
What is `examples/triton_runner/BUILD`?

examples/triton_runner/BENCHMARK.md
```diff
@@ -0,0 +1,38 @@
...
+|               | GRAM            | 46.7%       | 63.04%     |
+|               | GPU             | 17%         | 21%        |
+
+| Framework   |         | # of runners | Average (s) | Median | P90 | P95 | RPS         | Failures       |
+| ----------- | ------- | ------------ | ----------- | ------ | --- | --- | ----------- | -------------- |
+| TorchScript | BentoML | 1            | 3.161733884 | 3.2    | 3.7 | 3.8 | 15.63056076 | 0.006666905846 |
+|             | Triton  | 1            | 2.943633996 | 2.9    | 3.2 | 3.2 | 16.78060288 | 0              |
+|             |         |              |             |        |     |     |             |                |
+|             | BentoML | 2            | 3.711332685 | 3.5    | 5.2 | 5.4 | 13.27519003 | 0              |
+|             | Triton  | 1            | 2.993086328 | 3.1    | 3.7 | 3.9 | 16.50230972 | 0              |
```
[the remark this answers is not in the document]
aarnphm: Yes you are correct.

examples/triton_runner/README.md
aarnphm: Currently `--triton-options` only concerns options that can be used by multiple triton runners. Options such as `allow-metrics`, `metrics-port`, `grpc-port` that is specific to each runner are not yet supported.

In the case of BentoML I believe this is fine since the supervisor will manage the gRPC port.

wrt metrics, it is disabled by default for now.

examples/triton_runner/README.md
aarnphm: > What is `examples/triton_runner/BUILD`?

ah I have some bazel rules I tested with locally during the weekend 😄

examples/triton_runner/README.md
sauyon: Hm, I think this deserves a little more thought; we should definitely support per-runner options.

Can we just add these to the runner config maybe?

examples/triton_runner/README.md
```diff
@@ -0,0 +1,217 @@
...
+BentoML now provides support for Triton Inference Server.
+
+### Why do you want to use Triton?
+
+1. You are current Triton users trying to look for unification among frameworks
+   other than those supported by Triton.
+2. Current BentoML users who want to reduce performance gaps between C++ and
+   Python. While BentoML is rapidly improving, Triton provides better
+   performance under given conditions.
+3. Current Triton users who are looking for reusable pre-processing logic for
```
bojiang, line 12: Shall we also add a section like:
"We recommend continuing to use the BentoML runner when:"?

aarnphm:
cc @larme PTAL

examples/triton/tensorflow/helpers.py
```diff
@@ -0,0 +1,409 @@
...
+    PILImage = LazyLoader("PILImage", globals(), "PIL.Image")
+    cv2 = LazyLoader("cv2", globals(), "cv2")
+
+WORKING_DIR = Path(__file__).parent
+MODEL_TYPE = os.environ.get("MODEL_TYPE", "yolov5s")
+MODEL_FILE = WORKING_DIR / f"{MODEL_TYPE}.pt"
+TORCH_DEVICE = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
+
+
+def load_traced_script() -> tuple[ScriptModule, dict[str, t.Any]]:
```
larme, line 41: We don't need this in a tensorflow example? But I think for now it's easier to keep the same file for all 3 examples. Let's remove this at the final stage

examples/triton/tensorflow/build_bento.py
```diff
@@ -0,0 +1,54 @@
...
+
+    tag = "triton-integration-tensorflow"
+    if args.tag:
+        tag = f"triton-integration-tensorflow:{args.tag}"
+
+    try:
+        bentos = bentoml.get(tag)
+        if args.override:
+            bentoml.delete(tag)
+            raise bentoml.exceptions.NotFound("'--override', rebuilding bentos.")
```
larme, line 33: What's the reason we want an override switch?

larme (changes requested):

examples/triton/tensorflow/README.md
```diff
@@ -0,0 +1,217 @@
...
+
+BentoML now provides support for Triton Inference Server.
+
+### Why do you want to use Triton?
+
+1. You are current Triton users trying to look for unification among frameworks
+   other than those supported by Triton.
+2. Current BentoML users who want to reduce performance gaps between C++ and
+   Python. While BentoML is rapidly improving, Triton provides better
+   performance under given conditions.
```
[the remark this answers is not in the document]
aarnphm: This will be included in the blog

examples/triton/tensorflow/build_bento.py
aarnphm: Oh I think I should remove that. I was building a lot of models during testing so I don't want to pollute and repeat a lot of these bentos

examples/triton/tensorflow/helpers.py
```diff
@@ -0,0 +1,409 @@
...
+import time
+import typing as t
+from pathlib import Path
+
+import fs
+import attr
+import torch
+import fs.errors
+
+from bentoml._internal.utils import LazyLoader
```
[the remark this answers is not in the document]
aarnphm: I think when 3.12 is officially out and modules are lazy loaded by default, it would be nice.

I think we can remove all of the internal utils here.

examples/triton/tensorflow/setup
```diff
@@ -0,0 +1,15 @@
...
+if [ -f /etc/lsb-release ]; then
+    # In Ubuntu
+    apt-get update && apt-get install -y pkg-config libhdf5-dev ffmpeg libsm6 libxext6
+fi
+
+if [ -n "${BENTO_PATH}" ] || [ -f /.dockerenv ]; then
+    # We need to export LD_PRELOAD inside bento container.
+    SITE_PACKAGES=$(python3 -c "import site; print(site.getsitepackages()[0])")
+
+    export LD_PRELOAD="$SITE_PACKAGES/torch/lib/libgomp-d22c30c5.so.1"
```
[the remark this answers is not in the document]
aarnphm: updated

examples/triton/tensorflow/README.md
```diff
@@ -0,0 +1,217 @@
...
+export LD_PRELOAD=/lib/aarch64-linux-gnu/libgomp.so.1
+```
+
+Then run the `bentoml serve` command again.
+
+> Feel free to build your own tritonserver. See
+> [here](https://github.com/triton-inference-server/server/blob/main/docs/customization_guide/build.md)
+> for more details on building customisation.
+
+3. To build the Bento, use [`build_bento.py`](./build_bento.py):
```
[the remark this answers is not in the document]
aarnphm: I want to demonstrate our python api

examples/triton/tensorflow/README.md
```diff
@@ -0,0 +1,217 @@
...
+| torchscript_yolov5s | 1       | UNAVAILABLE: Not found: unable to load shared library: /lib/aarch64-linux-gnu/libgomp.so.1: cannot allocate memory in static TLS block |
++---------------------+---------+----------------------------------------------------------------------------------------------------------------------------------------+
+
+```
+
+Exit out of the process. Then do the following:
+
+```bash
+export LD_PRELOAD=/lib/aarch64-linux-gnu/libgomp.so.1
+```
```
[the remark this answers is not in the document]
aarnphm: I think this only happens on arm devices. I also tested this with a ec2 arm and it did run into this issue. Maybe we can add this section into a possible issue?

examples/triton/tensorflow/helpers.py
aarnphm: This is actually used in the train script.

examples/triton/tensorflow/README.md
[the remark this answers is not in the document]
larme: got it

larme (approved):
LGTM

PR #3476, opened 2023-01-25
feat: add http request object directly to BentoML request context

sauyon:
I've tried to write an example, but unfortunately it doesn't  work; the response isn't the echo that I would expect from the echo service.

### DIFF
--- a/src/bentoml/_internal/context.py
+++ b/src/bentoml/_internal/context.py
@@ -88,20 +88,25 @@ class InferenceApiContext:
 
         return InferenceApiContext(request_ctx, response_ctx)
 
-    @attr.define
     class RequestContext:
         metadata: Metadata
         headers: Metadata
+        http: starlette.requests.Request | None
 
-        def __init__(self, metadata: Metadata):
+        def __init__(
+            self,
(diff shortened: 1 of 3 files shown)


aarnphm (approved):

src/bentoml/_internal/context.py
```diff
@@ -88,20 +88,25 @@ def from_http(request: "starlette.requests.Request") -> "InferenceApiContext":
 
         return InferenceApiContext(request_ctx, response_ctx)
 
-    @attr.define
     class RequestContext:
         metadata: Metadata
         headers: Metadata
+        http: starlette.requests.Request | None
```
ssheng, line 94: Should we restructure the context to have the `http` and `grpc` group at the top level, so the syntax makes more syntactical sense? The downside is that we have to deal with backward compatibility.
- `ctx.http.request`
- `ctx.http.headers`
- `ctx.grpc.metadata`

aarnphm: I think we should, as #3511 adds a `ctx.grpc` field.

Compatibility is pretty light for this if I'm not missing anything.

PR #3482, opened 2023-01-26
fix(tensorflow): handle if results returns a list

aarnphm:
Handle cases when model also returns a list, then we process it as tuple


larme (approved):
LGTM

Issue #3483, opened 2023-01-26
bug: bentoml CLI containerize --output parameter


mtichy92:
### Describe the bug

Hello, we encounter this strange behaviour which seems like bug in CLI.
When running `bentoml containerize --opt output=type=registry --image-tag registry.gitlab.com/XXX --backend buildctl TAGGED_BENTO:TAG` it seems like the default --output parameter is not overridden and appears twice in actual command.
Either switching to --output type=registry does not help. Using default `--backend docker` does not help. Using `--push` or `--opt push` does not help (throwing `unknown flag: --push` error)

Traceback looks like this:
```
error: currently only single Exports can be specified (or Error response from daemon: exporter "registry" could not be found - when using docker)

Encountered exception while trying to building image: Command '['/opt/homebrew/bin/buildctl', 'build', '--local',
'context=/var/folders/jz/bb9zdtl57n70bp2bw1lg69km0000gn/T/tmpex446mo8fsTempFS/', '--local',
'dockerfile=/var/folders/jz/bb9zdtl57n70bp2bw1lg69km0000gn/T/tmpex446mo8fsTempFS/env/docker', '--frontend',
'dockerfile.v0', '--output', 'type=registry', '--output', 'type=docker,name=docker.io/registry.gitlab.com/XXX']' returned
non-zero exit status 1.
```

Can you please look at this? Since Yatai is far from being production ready we are trying to containerize and deploy bentos ourselves and this is blocker. Seems like not working on either of OCI backends supported.

Thank you!

### To reproduce

Steps to reproduce:

run any `bentoml containerize` command with `--opt output=type=registry`

### Environment

bentoml: 1.0.13 (also tested on 1.0.12)
python: 3.9.16
docker: 20.10.21
buildctl: 0.11.1

KimSoungRyoul:
Apart from the bug fix, wouldn't this issue be solved if you did the following?

~~~
bentoml containerize iris_classifier:betk2ie5koavizx5 --opt output=type=image,name=registry.gitlab.com/XXX,push=true  --backend buildctl

~~~

aarnphm:
I will take a look into this. Thanks for finding it.

mtichy92:
[...]

PR #3484, opened 2023-01-26
fix(containerize): remove autoconfig for buildctl

aarnphm:
People who use buildctl should probably know how to use it.

Remove an auto --output from tag

### DIFF
--- a/src/bentoml_cli/containerize.py
+++ b/src/bentoml_cli/containerize.py
@@ -573,7 +573,8 @@ def add_containerize_command(cli: Group) -> None:
                     o = subprocess.check_output(
                         [container_runtime, "load"], input=result
                     )
-                    click.echo(o.decode("utf-8").strip())
+                    if get_debug_mode():
+                        click.echo(o.decode("utf-8").strip())
                     sys.exit(0)
                 return result
 
--- a/src/bentoml_cli/containerize.py
+++ b/src/bentoml_cli/containerize.py
@@ -4,6 +4,7 @@ import sys
 import shutil
 import typing as t
 import logging
+import subprocess
 import tempfile
 import itertools
 from typing import TYPE_CHECKING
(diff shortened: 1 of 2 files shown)


sauyon:
How many releases has this been a thing? Can we leave it in unless the option is specifically passed?

sauyon:
@ssheng just an FYI that this would be a breaking change, but talking to Aaron I think this makes sense.

ssheng:
Thanks, @aarnphm and @sauyon. What was the motivation of this change?

aarnphm:
cc @ssheng https://github.com/bentoml/BentoML/issues/3483

src/bentoml_cli/containerize.py
```diff
@@ -540,14 +540,27 @@ def containerize(  # type: ignore
                         "To load image built with 'buildctl' requires one of docker, podman, nerdctl (Neither are found in PATH) and try again."
                     )
                     sys.exit(0)
-                o = subprocess.check_output([container_runtime, "load"], input=result)
-                click.echo(o.decode("utf-8").strip())
+                # NOTE: We will always use the docker image spec if docker is available.
+                # Otherwise fallback to the OCI image spec.
+                if "output" not in _memoized:
```
sauyon, line 545: What's this change do?

sauyon (commented):
o

src/bentoml_cli/containerize.py
aarnphm: deprecation warning right?. I also add the instruction for loading the tar into local container registry.

sauyon: I'm not sure why we need to warn twice if you haven't specified output?

aarnphm: We need to do that right. First warning is to say the autoconfig is deprecated. The second one is instruct how they can do it with passing in output.

ssheng (approved):

PR #3502, opened 2023-02-01
feat(triton): support HTTP client

aarnphm:
Users should be able to specify which types of TritonRunner they want to use via `tritonserver_type`

```python
import bentoml

triton_runner = bentoml.triton.Runner(name="my_triton_runner", model_repository="/path/to/model_repo", tritonserver_type="http")
```

This will use http client instead of gRPC. By default, all triton_runner will use gRPC client from tritonserver.


src/bentoml_cli/env.py
```diff
@@ -28,6 +28,7 @@
     # Only print if set
     "BENTOML_RETRY_RUNNER_REQUESTS",
     "BENTOML_CONTAINERIZE_BACKEND",
+    "BENTOML_RUNNER_MAP",
```
ssheng: Not related to this PR.

src/bentoml/triton.py
```diff
@@ -450,16 +488,109 @@ class TritonServerHandle:
...
+        except subprocess.TimeoutExpired:
+            _logger.debug("Arguments %s are available in %s", args, self._binary)
+            return True
+        except subprocess.CalledProcessError:
+            _logger.debug("Arguments %s does not exists for %s", args, self._binary)
+            return False
+        except Exception as e:  # pylint: disable=broad-except
+            _logger.error("Caught exception while validating arguments: %s", args)
+            _logger.error(e)
+            raise
```
ssheng, line 517: Maybe we don't need to check upfront. If the Triton server does not support http or grpc, just let it fail with a clear error message.

ssheng (approved):

PR #3506, opened 2023-02-01
fix(onnx): flatten providers include incorrect info

aarnphm:
This PR aims to address the flattened providers list when passed to `load_model`

if users have something like

```bash
import bentoml

runner = bentoml.onnx.get("onnx-mnist").with_options(providers=[
	("CUDAExecutionProvider", {"device_id": 1, "arena_extend_strategy": "kSameAsRequested" }), "CPUExecutionProvider"]
).to_runner()
```

This will be parsed as `["CUDAExecutionProvider", {"device_id": 1, "arena_extend_strategy": "kSameAsRequested" }, "CPUExecutionProvider"]`, which will fail our exception check.

Signed-off-by: aarnphm-ec2-dev <29749331+aarnphm@users.noreply.github.com>

### DIFF
--- a/src/bentoml/_internal/frameworks/onnx.py
+++ b/src/bentoml/_internal/frameworks/onnx.py
@@ -54,20 +54,8 @@ API_VERSION = "v2"
 logger = logging.getLogger(__name__)
 
 
-# helper methods
-def _yield_providers(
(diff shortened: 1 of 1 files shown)


larme (approved):
LGTM

PR #3510, opened 2023-02-01
chore: remove unnecessary docker health check

aarnphm:

### DIFF
--- a/src/bentoml/container.py
+++ b/src/bentoml/container.py
@@ -14,6 +14,7 @@ from typing import TYPE_CHECKING
 from simple_di import inject
 from simple_di import Provide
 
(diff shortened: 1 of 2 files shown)


src/bentoml/container.py
```diff
@@ -389,10 +387,6 @@ def build(
     """
     from ._internal.container import determine_container_tag
 
-    # Run healthcheck
```
parano, line 392: Is this useful for users using the Python API for containerizing Bento?

aarnphm: I think I demonstrated how to use these API in the triton examples.

But I don't think build should call health here.

Issue #3417, opened 2023-01-11
feat: gRPC context


ssheng:
Expose the `grpc.ServicerContext` to the service API through `InferenceApiContext`.

```
@svc.api(input=NumpyNdarray(), output=NumpyNdarray())
  def classify(input_series: np.ndarray, ctx) -> np.ndarray:
    grpc_context: grpc.ServicerContext = ctx.grpc.context
    grpc_context.set_code(200)
    grpc_context.set_trailing_metadata()
    ...
    result = iris_clf_runner.predict.run(input_series)
    return result
```

https://github.com/bentoml/BentoML/pull/3511

PR #3511, opened 2023-02-02
feat(grpc): Context

aarnphm:
Address https://github.com/bentoml/BentoML/issues/3417

### DIFF
--- a/docs/source/concepts/service.rst
+++ b/docs/source/concepts/service.rst
@@ -194,7 +194,7 @@ status codes.
 
         return result
 
-For gRPC endpoints, the `grpc.aio.ServicerContext <https://grpc.github.io/grpc/python/grpc_asyncio.html#grpc.aio.ServicerContext>`_ can be accessed through the inference context at ``ctx.grpc.context``.
+For gRPC endpoints, the `grpc.aio.ServicerContext <https://grpc.github.io/grpc/python/grpc_asyncio.html#grpc.aio.ServicerContext>`_ can be accessed through the inference context at ``ctx.request.grpc``.
 
 
 .. code-block:: python
@@ -203,7 +203,7 @@ For gRPC endpoints, the `grpc.aio.ServicerContext <https://grpc.github.io/grpc/p
     def predict(input_array: np.ndarray, ctx: bentoml.Context) -> np.ndarray:
         res = runner.run(input_array)
         accuracy = accuracy_score(res, expected)
-        context.grpc.context.set_trailing_metadata(
+        ctx.request.grpc.set_trailing_metadata(
             aio.Metadata.from_tuple((("accuracy", accuracy),))
         )
         return res
--- a/docs/source/concepts/service.rst
+++ b/docs/source/concepts/service.rst
@@ -168,10 +168,7 @@ status codes.
 
 .. code-block:: python
 
-    @svc.api(
-        input=NumpyNdarray(),
-        output=NumpyNdarray(),
-    )
+    @svc.api(input=NumpyNdarray(), output=NumpyNdarray())
     def predict(input_array: np.ndarray, ctx: bentoml.Context) -> np.ndarray:
         # get request headers
         request_headers = ctx.request.headers
@@ -196,6 +193,25 @@ status codes.
         ctx.response.headers.append("X-Custom-Header", "value")
 
         return result
+
+For gRPC endpoints, the `grpc.aio.ServicerContext <https://grpc.github.io/grpc/python/grpc_asyncio.html#grpc.aio.ServicerContext>`_ can be accessed through the inference context at ``ctx.grpc.context``.
+
+
+.. code-block:: python
+
+    @svc.api(input=NumpyNdarray(), output=NumpyNdarray())
+    def predict(input_array: np.ndarray, ctx: bentoml.Context) -> np.ndarray:
+        res = runner.run(input_array)
+        accuracy = accuracy_score(res, expected)
+        context.grpc.context.set_trailing_metadata(
+            aio.Metadata.from_tuple((("accuracy", accuracy),))
+        )
+        return res
+
+.. note::
+
+    The gRPC context is a `grpc.aio.ServicerContext <https://grpc.github.io/grpc/python/grpc_asyncio.html#grpc.aio.ServicerContext>`_ object,
+    which means that any async function such as ``send_initial_metadata`` must be called in an async API function.
 
 
 IO Descriptors
--- a/docs/source/concepts/service.rst
+++ b/docs/source/concepts/service.rst
@@ -166,52 +166,51 @@ argument added to the service API function. Both the request and response contex
 accessed through the inference context for getting and setting the headers, cookies, and
 status codes.
 
-.. code-block:: python
+.. tab-set::
 
-    @svc.api(input=NumpyNdarray(), output=NumpyNdarray())
-    def predict(input_array: np.ndarray, ctx: bentoml.Context) -> np.ndarray:
-        # get request headers
-        request_headers = ctx.request.headers
+    .. tab-item:: HTTP
+       :sync: http
 
-        result = runner.run(input_array)
+       .. code-block:: python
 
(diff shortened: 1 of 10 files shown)


docs/source/concepts/service.rst
```diff
@@ -197,6 +194,25 @@ status codes.
 
         return result
 
+For gRPC endpoints, the `grpc.aio.ServicerContext <https://grpc.github.io/grpc/python/grpc_asyncio.html#grpc.aio.ServicerContext>`_ can be accessed through the inference context at ``ctx.grpc.context``.
```
sauyon, line 197: We should probably have a chat about this; I'm not sure I want to expose either the gRPC or HTTP context / requests directly, can we do this in i.e. our IO descriptors?

src/bentoml/_internal/context.py
```diff
@@ -74,19 +74,24 @@ def mutablecopy(self) -> Metadata:
 
 @attr.define
 class InferenceApiContext:
-    request: "RequestContext"
-    response: "ResponseContext"
+    request: RequestContext = attr.field(default=None)
```
sauyon, line 77: gRPC should use these too.

src/bentoml/_internal/server/http_app.py
```diff
@@ -310,6 +310,8 @@ async def api_func(request: Request) -> Response:
             try:
                 input_data = await api.input.from_http_request(request)
                 ctx = None
+                # NOTE: function should always be set here, but check anyway.
```
sauyon, line 313:
```suggestion
```

An assert should always mean this.

src/bentoml/_internal/service/inference_api.py
```diff
@@ -29,12 +28,15 @@ def __init__(
...
-        route: Optional[str] = None,
+        name: str | None = None,
+        doc: str | None = None,
+        route: str | None = None,
     ):
         # Use user_defined_callback function variable if name not specified
-        name = user_defined_callback.__name__ if name is None else name
+        if user_defined_callback is None:
+            name = "" if name is None else name
+        else:
```
sauyon, line 38: What's this for?

aarnphm: YOu are setting the user_defined_callback in the client to None right. Then the previous logic would fail as None doesn't have a __name__?

sauyon (changes requested):
I think this is waiting on the removal of `.grpc` @aarnphm

src/bentoml/_internal/server/http_app.py
sauyon: I don't think these comments add anything, let's remove them?

aarnphm: I thought I have already removed them. Let me do that now.

docs/source/concepts/service.rst
```diff
@@ -166,36 +166,51 @@ argument added to the service API function. Both the request and response contex
...
+                       max_age=None,
+                       expires=None,
+                       path="/predict",
+                       domain=None,
+                       secure=True,
+                       httponly=True,
+                       samesite="None"
+                   )
+               ]
+               ctx.response.headers.append("X-Custom-Header", "value")
```
sauyon, line 195:
```suggestion
               ctx.response.headers["X-Custom-Header"] = "value"
```

docs/source/concepts/service.rst
```diff
@@ -166,36 +166,51 @@ argument added to the service API function. Both the request and response contex
...
+
+    .. tab-item:: gRPC
+       :sync: grpc
+
+       .. code-block:: python
+
+           @svc.api(input=NumpyNdarray(), output=NumpyNdarray())
+           def predict(input_array: np.ndarray, ctx: bentoml.Context) -> np.ndarray:
+               res = runner.run(input_array)
+               ctx.response.metadata['accuracy'] = accuracy_score(res, expected)
```
sauyon, line 207: Maybe just add a line showing that metadata or headers both work?

docs/source/concepts/service.rst
```diff
@@ -166,36 +166,51 @@ argument added to the service API function. Both the request and response contex
...
-                secure=True,
-                httponly=True,
-                samesite="None"
-            )
-        ]
-        ctx.response.headers.append("X-Custom-Header", "value")
+.. note::
 
-        return result
+    The metadata for gRPC will be sent as a trailing metadata, via ``context.send_trailing_metadata``.
```
sauyon, line 212: Hm, can we make `headers` set initial metadata instead?

docs/source/concepts/service.rst
```diff
@@ -166,36 +166,51 @@ argument added to the service API function. Both the request and response contex
...
+       :sync: http
+
+       .. code-block:: python
+
+           @svc.api(input=NumpyNdarray(), output=NumpyNdarray())
+           def predict(input_array: np.ndarray, ctx: bentoml.Context) -> np.ndarray:
+               result = runner.run(input_array)
+
+               # set response headers, cookies, and status code 
+               ctx.response.status_code = 202
```
sauyon, line 181: Do you have `status_code` support for grpc? Though we can postpone it for now, most likely we'd want to document how to implement and raise a bentomlexception instead.

src/bentoml/_internal/server/grpc/servicer/v1/__init__.py
```diff
@@ -72,20 +74,47 @@ async def Call(  # type: ignore (no async types) # pylint: disable=invalid-overr
                     validate_proto_fields(request.WhichOneof("content"), api.input),
                 )
                 input_data = await api.input.from_proto(input_proto)
+                ctx: InferenceApiContext | None = None
+                assert api.func is not None
+                # NOTE: function should always be set here, but check anyway.
```
sauyon, line 79: We can probably remove these asserts entirely now, right?

src/bentoml/_internal/server/grpc/servicer/v1/__init__.py
```diff
@@ -72,20 +74,47 @@ async def Call(  # type: ignore (no async types) # pylint: disable=invalid-overr
...
+                        else:
+                            output = await anyio.to_thread.run_sync(
+                                api.func, input_data
+                            )
+
+                if ctx is not None:
+                    # Set gRPC metadata if context is available
+                    context.set_trailing_metadata(
+                        aio.Metadata.from_tuple(tuple(ctx.response.metadata.items()))
+                    )
```
sauyon, line 113: Hm, is this right? Wouldn't you be sending the trailing metadata twice? Or are users not meant to modify the gRPC context directly? (In which case we should expose our own metadata implementation)

docs/source/concepts/service.rst
aarnphm: We can add status_code support for gRPC. User can set it to the value that gRPC follows.

src/bentoml/_internal/server/grpc/servicer/v1/__init__.py
aarnphm: I only it send it once right? only if ctx is given. Usually, if user wants to handle context, either client/server interceptors should be used.

Context are meant to be modified if needed.

aarnphm: But for the majority of users (exclude power user), context is not needed. It is the case only use if you know what you are doing.

docs/source/concepts/service.rst
sauyon: Yep, for now let's do that.

src/bentoml/_internal/server/grpc/servicer/v1/__init__.py
sauyon: No, I mean with a "normal" gRPC server, does modifying the context do anything? We're both modifying the context itself and sending the metadata manually, we should do one or the other, I think.

aarnphm:
print warning when cookie are set.

sauyon:
Closing as this is not currently planned.

PR #3514, opened 2023-02-02
test(unit): adding test cases for runners

andrehora:
The test `test_runner` misses several checks:

- This PR improves `test_runner` by adding several untested cases, both for valid and invalid runners.
- I also refactored the original test, which was too long.

In summary, the novel tests are now smaller, clearer, and more focused, which is best practice when writing tests.


aarnphm (approved):

aarnphm:
Thanks for the contribution @andrehora

PR #3518, opened 2023-02-03
docs: containerisation API

aarnphm:
Expose the description of #3164 to our containerisation docs.


src/bentoml/_internal/container/__init__.py
```diff
@@ -298,6 +298,11 @@ def health(backend: str) -> bool:
     Args:
         backend: The name of the backend.
 
+    .. note::
+
+        If given backend is a type of OCIBuilder, and the backend is not registered,
+        make sure to register it with ``bentoml.container.register_backend``.
```
sauyon, line 304: Can we just use registries on subclass like we do for schedulers / IO descriptors?

aarnphm: Yep. We can. I will make a separate PR for this.

ssheng (approved):

PR #3523, opened 2023-02-03
chore(stubs): minify and remove unused

aarnphm:


bojiang:
@aarnphm Is it possible to configure the black to keep the previous behavior about blank lines?

bojiang:
It may cause tons of conflicts with current PRs

bojiang:
Or can we configure or CI to only check the changed files? So the format style can be gradually changed.

aarnphm:
> Or can we configure or CI to only check the changed files? So the format style can be gradually changed.

I don't know about this option because the blank lines is included in the stable release for 23.

We can also just merge all running PR in first to avoid this issue.

PR #3532, opened 2023-02-07
fix: wheel directory check for notebook workflow

### LINKED ISSUE #3528: bug: unintended creation of wheel directory
### Describe the bug

When the user specifies Python options with `wheels` as empty list instead of `None`, bentoml will create the wheel directory, causing `install.sh` to fail because it tries to install `*.whl` from the wheels directory, but there are no wheel files inside. 

This is normally not an issue because the user is expected to create a `bentofile.yaml` and use the bentoml CLI to build their bentos, but in specific workflows where it is more convenient to use the bentoml Python interface to build bentos (e.g. when the user is using a notebook-centric environment like Databricks), the user may specify an empty list for wheels instead of `None`.

It is fairly simple to workaround this (we can specify `None` instead of empty list when calling `bentoml.bentos.build()`), but it would be nice if it could work with empty list as well. I can work on this, just let me know if it's ok to proceed.

```python
if self.wheels is not None:
    bento_fs.makedirs(wheels_folder, recreate=True)
    for whl_file in self.wheels:  # pylint: disable=not-an-iterable
        whl_file = resolve_user_filepath(whl_file, build_ctx)
        copy_file_to_fs_folder(whl_file, bento_fs, wheels_folder)
```

```bash
if [ -d "$WHEELS_DIR" ]; then
    echo "Installing wheels packaged in Bento.."
    pip3 install "$WHEELS_DIR"/*.whl "${PIP_ARGS[@]}"
fi
```

I believe the check should be simplified to `if self.wheels`, so it covers the case of empty list as well.

### To reproduce

_No response_

### Expected behavior

_No response_

### Environment

bentoml: 1.0.8 onwards
Python: 3.8

jiewpeng:
## What does this PR address?

Fixes #3528 

One question I have is, do I also need to make this change in `grpc-client/bentoml/_internal/bento/build_config.py`?

## Before submitting:

### DIFF
--- a/src/bentoml/_internal/bento/build_config.py
+++ b/src/bentoml/_internal/bento/build_config.py
@@ -514,7 +514,7 @@ class PythonOptions:
         build_bentoml_editable_wheel(bento_fs.getsyspath(wheels_folder))
 
         # Move over required wheel files
-        if self.wheels is not None:
+        if self.wheels:
             bento_fs.makedirs(wheels_folder, recreate=True)
             for whl_file in self.wheels:  # pylint: disable=not-an-iterable
                 whl_file = resolve_user_filepath(whl_file, build_ctx)


aarnphm:
oh no, the grpc-client is a symlink.

aarnphm (approved):

PR #3533, opened 2023-02-08
fix: don't use removed content_type in starlette UploadFile

sauyon:


aarnphm (approved):

sauyon:
Turns out we already have a test which should exercise this case.

PR #3534, opened 2023-02-08
feature(framework): diffusers

larme:
## What does this PR address?

Fixes #(issue)

## Before submitting:

### DIFF
--- a/src/bentoml/_internal/frameworks/diffusers.py
+++ b/src/bentoml/_internal/frameworks/diffusers.py
@@ -109,7 +109,7 @@ def load_model(
 
         import bentoml
         pipeline = bentoml.diffusers.load_model('my_diffusers_model:latest')
-        pipeline.(prompt)
+        pipeline(prompt)
     """  # noqa
     if not isinstance(bento_model, bentoml.Model):
         bento_model = get(bento_model)
@@ -216,7 +216,7 @@ def import_model(
             'my_sd15_model',
             "runwayml/stable-diffusion-v1-5",
             signatures={
-                "__call__": {"batchable": True},
+                "__call__": {"batchable": False},
             }
         )
     """
@@ -282,7 +282,7 @@ def get_runnable(bento_model: bentoml.Model) -> t.Type[bentoml.Runnable]:
     """
     Private API: use :obj:`~bentoml.Model.to_runnable` instead.
     """
-    signature = bento_model.info.signatures["__call__"]
+
     partial_kwargs: t.Dict[str, t.Any] = bento_model.info.options.partial_kwargs  # type: ignore
     pipeline_class: type[diffusers.DiffusionPipeline] = (
         bento_model.info.options.pipeline_class or diffusers.StableDiffusionPipeline
@@ -323,12 +323,14 @@ def get_runnable(bento_model: bentoml.Model) -> t.Type[bentoml.Runnable]:
             if torch.cuda.is_available():
                 self.pipeline.to("cuda")
 
-    def make_run_method(method_name: str, partial_kwargs):
+    def make_run_method(
+        method_name: str, partial_kwargs: dict[str, t.Any] | None
+    ) -> t.Callable[..., t.Any]:
         def _run_method(
             runnable_self: DiffusersRunnable,
             *args: t.Any,
             **kwargs: t.Any,
-        ) -> diffusers.pipelines.stable_diffusion.StableDiffusionPipelineOutput:
+        ) -> t.Any:
 
             if method_partial_kwargs is not None:
                 kwargs = dict(method_partial_kwargs, **kwargs)
--- a/src/bentoml/_internal/frameworks/diffusers.py
+++ b/src/bentoml/_internal/frameworks/diffusers.py
@@ -29,6 +29,7 @@ try:
     import diffusers
     from diffusers.utils.import_utils import is_torch_version
     from diffusers.utils.import_utils import is_xformers_available
+    from diffusers.utils.import_utils import is_accelerate_available
 except ImportError:  # pragma: no cover
     raise MissingDependencyException(
         "'diffusers' is required in order to use module 'bentoml.diffusers', install diffusers with 'pip install --upgrade diffusers transformers accelerate'. For more information, refer to https://github.com/huggingface/diffusers",
@@ -85,7 +86,7 @@ def load_model(
     pipeline_class: type[
         diffusers.pipelines.DiffusionPipeline
     ] = diffusers.StableDiffusionPipeline,
-    device_map: str | dict[str, int | str | torch.device] = "auto",
+    device_map: str | dict[str, int | str | torch.device] | None = None,
     custom_pipeline: str | None = None,
     scheduler_class: type[diffusers.SchedulerMixin] | None = None,
     torch_dtype: str | torch.dtype | None = None,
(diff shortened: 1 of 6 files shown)


src/bentoml/_internal/frameworks/diffusers.py
```diff
@@ -0,0 +1,334 @@
...
+    Returns:
+        The Diffusion model loaded as diffusers pipeline from the BentoML model store.
+
+    Example:
+
+    .. code-block:: python
+
+        import bentoml
+        pipeline = bentoml.diffusers.load_model('my_diffusers_model:latest')
+        pipeline.(prompt)
```
yubozhao, line 108:
```suggestion
        pipeline(prompt)
```

src/bentoml/_internal/frameworks/diffusers.py
```diff
@@ -0,0 +1,334 @@
...
+    with the given tag from the local BentoML model store.
+
+    Args:
+        bento_model: Either the tag of the model to get from the store, or a BentoML
+            ``~bentoml.Model`` instance to load the model from.
+
+    Returns:
+        The Diffusion model loaded as diffusers pipeline from the BentoML model store.
+
+    Example:
```
yubozhao, line 102: Could we use the stable diffusion example?

src/bentoml/_internal/frameworks/diffusers.py
larme: fxied

src/bentoml/_internal/frameworks/diffusers.py
```diff
@@ -0,0 +1,355 @@
...
+            runnable_self: DiffusersRunnable,
+            *args: t.Any,
+            **kwargs: t.Any,
+        ) -> t.Any:
+
+            if method_partial_kwargs is not None:
+                kwargs = dict(method_partial_kwargs, **kwargs)
+
+            raw_method = getattr(runnable_self.pipeline, method_name)
+            res = raw_method(*args, return_dict=False, **kwargs)
```
aarnphm, line 339: Shouldn't we allow users to set this instead of ourselves? it might be different from what users might expects.

larme: @aarnphm do you mean `return_dict`? If set to `True` then we will have some issue to pickling the output. Have you met similar option in transformers implementation?

larme: I think I will set  default to False but add an option to let user change it if they really want to

src/bentoml/_internal/frameworks/diffusers.py
```diff
@@ -0,0 +1,355 @@
...
+    if pipeline_class is None:
+        pipeline_class = diffusers.StableDiffusionPipeline
+
+    diffusion_model_dir = bento_model.path_of(DIFFUSION_MODEL_FOLDER)
+
+    if low_cpu_mem_usage is None:
+        if is_torch_version(">=", "1.9.0"):
+            low_cpu_mem_usage = True
+        else:
+            low_cpu_mem_usage = False
```
aarnphm, line 131: any specific reason why we are doing this?

src/bentoml/_internal/frameworks/diffusers.py
```diff
@@ -0,0 +1,355 @@
...
+    Private API: use :obj:`~bentoml.Model.to_runnable` instead.
+    """
+
+    partial_kwargs: t.Dict[str, t.Any] = bento_model.info.options.partial_kwargs  # type: ignore
+    pipeline_class: type[diffusers.DiffusionPipeline] = (
+        bento_model.info.options.pipeline_class or diffusers.StableDiffusionPipeline
+    )
+    scheduler_class: type[
+        diffusers.SchedulerMixin
+    ] | None = bento_model.info.options.scheduler_class
```
aarnphm, line 292: This two will fail when dumping these classes to yaml. I think you might have to write custom handler for `bentoml_cattr`.

src/bentoml/_internal/frameworks/diffusers.py
aarnphm: I think that if they are passing it into kwargs it will fail right? best we don't do it here and let the framework do it. If users want to change that they can do that themselves.

src/bentoml/_internal/frameworks/diffusers.py
```diff
@@ -0,0 +1,355 @@
...
+    scheduler_class: type[diffusers.SchedulerMixin] | None = None,
+    torch_dtype: str | torch.dtype | None = None,
+    low_cpu_mem_usage: bool | None = None,
+    enable_xformers: bool = False,
+) -> diffusers.DiffusionPipeline:
+    """
+    Load a Diffusion model and convert it to diffusers `Pipeline <https://huggingface.co/docs/diffusers/api/pipelines/overview>`_
+    with the given tag from the local BentoML model store.
+
+    Args:
```
ssheng, line 99: Let's add docs for all arguments.

src/bentoml/_internal/frameworks/diffusers.py
ssheng: What if `low_cpu_mem_usage` is passed in as `True` but torch version < 1.9.0

src/bentoml/_internal/frameworks/diffusers.py
```diff
@@ -0,0 +1,355 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+@attr.define
+class DiffusersOptions(PartialKwargsModelOptions):
+    """Options for the diffusers model."""
+
+    pipeline_class: type[diffusers.pipelines.DiffusionPipeline] | None = None
+    scheduler_class: type[diffusers.SchedulerMixin] | None = None
```
ssheng, line 51: As discussed, update the type to `str`.

src/bentoml/_internal/frameworks/diffusers.py
larme: I have slightly change this part so user can change the value.

BTW all DiffusionPipeline should accept `return_dict` so the calling will never fail

src/bentoml/_internal/frameworks/diffusers.py
larme: This code will automatically pick a suitable value for `low_cpu_mem_usage`.

If user pass `low_cpu_mem_usage=True` explicitly but torch version < 1.9.0, `diffuesrs` will raise an error. I think it's better to let `diffusers` to raise the error instead of silently converting `low_cpu_mem_usage` to `False` by ourselves. It will give user a clear warning/error and we don't override the future behavior of diffusers if they decide to change this part.

src/bentoml/_internal/frameworks/diffusers.py
larme: Currently saving won't save these options. So we may need to worry about pickling.

src/bentoml/_internal/frameworks/diffusers.py
larme: I have a discussion with aaron later and we agree that diffusers' situation is different from transformers, so currently we may keep using the class type instead of str. This is also eaiser for user to use their own pipelines.

src/bentoml/_internal/frameworks/diffusers.py
larme: done

ssheng (dismissed):

aarnphm (approved):

PR #3540, opened 2023-02-08
fix: increase check_delay to address exception in arbiter callback

aarnphm:
Screenshot 2023-02-08 at 01 08 13

### DIFF
--- a/src/bentoml/_internal/utils/circus/__init__.py
+++ b/src/bentoml/_internal/utils/circus/__init__.py
@@ -53,5 +53,9 @@ def create_standalone_arbiter(watchers: list[Watcher], **kwargs: t.Any) -> Arbit
                 watchers,
                 endpoint=f"tcp://127.0.0.1:{endpoint_port}",
                 pubsub_endpoint=f"tcp://127.0.0.1:{pubsub_port}",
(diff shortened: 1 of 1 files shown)


src/bentoml/_internal/utils/circus/__init__.py
```diff
@@ -53,5 +53,9 @@ def create_standalone_arbiter(watchers: list[Watcher], **kwargs: t.Any) -> Arbit
                 watchers,
                 endpoint=f"tcp://127.0.0.1:{endpoint_port}",
                 pubsub_endpoint=f"tcp://127.0.0.1:{pubsub_port}",
+                # NOTE: increase check delay to 5s to avoid
+                # ConfclitError raised when arbiter is running arbiter_stop command
+                # See https://github.com/bentoml/BentoML/pull/3540.
+                check_delay=5,
```
sauyon, line 59: I think this could use some more explanation, either here or in the PR description.

parano:
How do you reproduce this error?

aarnphm:
> How do you reproduce this errors?

After a long serving period, when exiting this should show up.

aarnphm:
I will probably take another look to fix this, instead of having this patch fix. Closing for now

Issue #3541, opened 2023-02-08
bug: filename always set to `<request body>` when using multipart


sauyon:
### Describe the bug

With the API

```
@svc.api(
    input=Multipart(f=File()),
    output=...
)
```

`f.name` will always be `<request body>` instead of its real name; this is a bug in the way we handle files inside the multipart descriptor.

### Environment

N/A

frostming:
No longer an issue on the new service API

PR #3542, opened 2023-02-08
fix: update formparser to new Starlette version

sauyon:
@ssheng writing unit tests is a little involved so thought I'd put this up now; wrote a Bento using multipart and it seems to work.

Discovered another issue where we set the filename to `<request body>` always when using multipart, opened #3541.

### DIFF
--- a/src/bentoml/_internal/utils/formparser.py
+++ b/src/bentoml/_internal/utils/formparser.py
@@ -4,6 +4,7 @@ import io
 import uuid
 import typing as t
 from typing import TYPE_CHECKING
+from tempfile import SpooledTemporaryFile
 
 import multipart.multipart as multipart
 from starlette.requests import Request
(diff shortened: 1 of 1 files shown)


src/bentoml/_internal/utils/formparser.py
```diff
@@ -151,7 +155,10 @@ async def parse(self) -> _ItemsBody:
                         )
                     if b"filename" in options:
                         filename = user_safe_decode(options[b"filename"], charset)
+                        tempfile = SpooledTemporaryFile(max_size=MAX_FILE_SIZE)
```
ssheng, line 158: Does this limit the max size of each part?

sauyon: No, this is just the max size stored in memory before it starts writing to disk.

ssheng (approved):

PR #3552, opened 2023-02-12
fix(tests): make sure `pytest tests` works

aarnphm:
This PR fixes a canonical problem when running tests in BentoML.

Before this PR, if you try `pytest tests`, you will run into a bunch of weird errors.

This has to do with using `import-mode=prepend`, where it will prepend given tests path
to `sys.path`.

If you are running from the GIT_ROOT, then the `$GIT_ROOT/tests` will only available in sys.path.

In a lot of our tests there are places that we do relative imports, which causes the tests to fail.

For now, this will do it. However, after running `pytest tests`, there are a lot of broken tests.

I will fix them in a separate PR.

A drawback of this approach:

- Running tests are way to long
- Devs will need to install both `pip install -r requirements/dev-requirements.txt` and `pip install -r requirements/frameworks-requirements.txt` to run the whole suite
- Tests is not cached. (which is fine for now, but should fix)


aarnphm:
cc @sauyon

sauyon:
I drafted this for you, just undraft it when it's ready.

pvardanis:
Hey guys I have a question. How can we can contribute to the library and be able to run all tests for all frameworks since there might be dependency conflicts between frameworks? I've seen that you solved in the CI/CD by running separate environments per framework, but can this be feasible for local development and testing?

aarnphm:
> Hey guys I have a question. How can we can contribute to the library and be able to run all tests for all frameworks since there might be dependency conflicts between frameworks? I've seen that you solved in the CI/CD by running separate environments per framework, but can this be feasible for local development and testing?

Hi there, this PR tries to address that issue. I have also proposed a workflow with bazel in a few previous PR which helps with this issue. But We decided that the integrations for bazel is very low on our priority list atm.

For developing the library I recommend to create a virtualenv, and `pip install -r requirements/frameworks-requirements.txt`

We are actively working on improving the developer experience for this. Feel free to reach out to me on Slack if you have any recommendation for improving this.

pvardanis:
@aarnphm I did try this and surprisingly I didn't get any conflicts. Is this the case because you don't freeze to separate framework version, or is pypi failing to do conflict resolution?

aarnphm:
Whats the conflict you are getting into? I haven't have the bandwidth to update this PR yet as it is still wip.

pvardanis:
image

It's conflicts everywhere basically. Have you considered having separate folders per framework with each folder having a separate venv? That would yield the need for having separate wheels per framework, e.g. `bentoml-tensorflow`, but I think that's fair considering that it would improve the developer experience. Each `bentoml-<framework>` wheel would have bentoml as basic requirement, then build on top.

For the end user I don't see any problem, just that they would need to install `bentoml-tensorflow` if they need tensorflow, instead of `bentoml`.

aarnphm:
I don't think packaging framework with wheels is a good practice. Inherently, using ML workflow is per-framework basis. Framework installation shouldn't be coupled with BentoML.

pvardanis:
I agree, however if you follow this approach BentoML could still be installed as a standalone without framework requirements. The user could just do `pip install bentoml`. The bentoml package would only include base classes, e.g. `Runnable`, `Service` etc. so the users can still extend it. Then, it’s only when someone would need e.g. tensorflow they would have to do `pip install bentoml bentoml-tensorflow`.

sauyon:
We've explored using something like Bazel to run tests for basically this purpose, but at the moment it's a lot of overhead for the dev team and running tests locally basically mostly works as long as you use `pip` and ignore the fact that Python's dependency management is somewhat broken.

pvardanis:
@sauyon  I see, depending on native pip flaws however is not a solution. The way you do it now is painful for an external developer to contribute I'm afraid. I honestly believe my proposal would make the whole experience much more enjoyable.

sauyon:
> @sauyon I see, depending on native pip flaws however is not a solution and definitely not a good practice.

Sure, I agree it's not the greatest, but it works for us for now.

> The way you do it as of now is painful for an external developer to contribute I'm afraid. I honestly believe my proposal combined with Poetry would make the whole experience much more enjoyable.

I'm not sure what exactly the issue with contributing is, I'm happy to help if there's something you feel like is causing friction.

We actually did try poetry a little while ago, but ran into a bunch of issues and it did nothing but impede our workflow. We might reconsider that in a few months, but it just put a sour taste in our mouths about poetry.

In any case, let's keep the comment section for this PR clean. Feel free to create an issue or a discussion if you want to continue.

aarnphm:
cherry-picked to #4056

PR #3561, opened 2023-02-14
feature(frameworks): bentoml.onnx runner accept kwargs

larme:
## What does this PR address?

Sometimes it's more nature to call onnx model with keyword arguments. For example, bert's tokenizer will output a dictionary, which the original model will call with `model(**input)`. Our runner should simulate this behavior

Fixes #(issue)

## Before submitting:

### DIFF
--- a/src/bentoml/_internal/frameworks/onnx.py
+++ b/src/bentoml/_internal/frameworks/onnx.py
@@ -26,6 +26,7 @@ if TYPE_CHECKING:
 
     from .utils.onnx import ONNXArgType
     from .utils.onnx import ONNXArgCastedType
+    from .utils.onnx import ONNXArgCastingFuncType
 
     ProvidersType = list[str | tuple[str, dict[str, t.Any]]]
 
@@ -65,7 +66,8 @@ class ONNXOptions(ModelOptions):
     input_specs: dict[str, list[dict[str, t.Any]]] = attr.field(factory=dict)
     output_specs: dict[str, list[dict[str, t.Any]]] = attr.field(factory=dict)
     providers: ProvidersType = attr.field(default=None)
-    session_options: t.Optional["ort.SessionOptions"] = attr.field(default=None)
+    session_options: "ort.SessionOptions" | None = attr.field(default=None)
+    use_kwargs_inputs: bool = attr.field(default=False)
 
 
 def get(tag_like: str | Tag) -> bentoml.Model:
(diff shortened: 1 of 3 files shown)


aarnphm (approved):

aarnphm:
cc @larme, I ran the test locally, it passed, but it still didn't run on CI. can you also fix it as well?

src/bentoml/_internal/frameworks/onnx.py
```diff
@@ -419,17 +427,31 @@ def _process_output(outs):
...
-
-            input_names: dict[str, ONNXArgCastedType] = {
-                i.name: val for i, val in zip(self.model.get_inputs(), casted_args)
-            }
-            output_names: list[str] = [o.name for o in self.model.get_outputs()]
-            raw_outs = self.predict_fns[method_name](output_names, input_names)
-            return _process_output(raw_outs)
+        if options.use_kwargs_inputs:
+
+            def _run(self: ONNXRunnable, **kwargs: ONNXArgType) -> t.Any:
```
bojiang, line 432: Does this mean we can not call `onnx_runner.run(arg1, arg2, kwarg1=xxx)` ?

aarnphm: no I don't think so.

sauyon: Yes, it does.

larme: For every onnx model we have a fixed list of inputs with their own names. Calling by args means refer each argument by its position while calling by kwargs means refer each argument by its name. Now if we mix these 2 methods, I feel that it's very easy for user to miss some arguments.

Calling by kwargs is useful when dealing with NLP models where the tokenizer output a dictionary and the dictionary is passed to the model as kwargs. In this use case there's no additional positional argument to be passed

larme:
@aarnphm Now the tests run on CI. The problem is that the new bert tests will missing the pytorch dependencies (I thought installing transformers will install torch by default, I was wrong).

aarnphm (dismissed):

src/bentoml/_internal/frameworks/onnx.py
aarnphm: Do we need more discussion about this?

larme: let's have a discussion on engineer meeting

sauyon:
@larme @aarnphm what's the status of this one?

aarnphm:
This has been on this list for a while now @larme we should circle back and get this one merge

PR #3581, opened 2023-02-18
chore: enable plugins on CI only

aarnphm:
Only enable pytest plugins via envvar so that we can run tests inside examples
without using the plugins. The plugins are meant to be used with our general
tests suite only

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -236,7 +236,7 @@ exclude = '''
 '''
 
 [tool.pytest.ini_options]
-addopts = ["-rfEX", "-pbentoml.testing.pytest.plugin"]
+addopts = ["-rfEX", "-pno:warnings"]
 python_files = ["test_*.py", "*_test.py"]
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -236,7 +236,7 @@ exclude = '''
 '''
 
 [tool.pytest.ini_options]
-addopts = ["-rfEX", "-pbentoml.testing.pytest.plugin"]
+addopts = ["-rfEX", "-pno:warnings"]
```
sauyon, line 239: Why are we doing this?

sauyon (changes requested):
Is this change motivated by an actual problem?

pyproject.toml
aarnphm: Yes. Right now if users are running tests from our examples, it would be pretty strange if they already have a models for that example locally, then have to train that model again and run through the whole example tests.

PR #3582, opened 2023-02-18
chore(test): using container API for testing

aarnphm:

### DIFF
--- a/src/bentoml/testing/server.py
+++ b/src/bentoml/testing/server.py
@@ -151,24 +151,26 @@ def containerize(
     cleanup: bool = True,
     use_grpc: bool = False,
     backend: str = "docker",
+    **attrs: t.Any,
 ) -> t.Generator[str, None, None]:
     """
     Build the docker image from a saved bento, yield the docker image tag
     """
-    from bentoml import bentos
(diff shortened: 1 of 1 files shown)


src/bentoml/testing/server.py
```diff
@@ -151,24 +151,26 @@ def containerize(
...
+        container.build(
             str(bento_tag),
-            docker_image_tag=(image_tag,),
+            backend=backend,
+            image_tag=(image_tag,),
             progress="plain",
             features=["grpc", "grpc-reflection"] if use_grpc else None,
-            labels={"testing": True, "module": __name__},
-            backend=backend,
+            label={"testing": True, "module": __name__},
```
sauyon, line 172: Wait, why did we change this in the API?

sauyon (approved):
Looks good, but a question for the old API change.

src/bentoml/testing/server.py
aarnphm: because it is actually `--label` from docker, not `--labels`.

`bentoml.bentos.containerize` provides backward compatibility from `labels` to `label`.

Issue #3570, opened 2023-02-15
bug: SSL options in bentoml_configuration.yaml are ignored


ryanheise:
### Describe the bug

The `ssl` section in `bentoml_configuration.yaml` is ignored.

### To reproduce

Create a self-signed certificate:

```bash
openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout key.key -out cert.crt
```

Create the file `bentoml_configuration.yaml` as follows:

```yaml
api_server:
  ssl:
    enabled: true
    certfile: cert.crt
    keyfile: key.key
```

Run your project with the config:

```bash
BENTOML_CONFIG=bentoml_configuration.yaml bentoml serve service:svc --production --verbose
```

THE RESULT:

```
2023-02-15T18:32:29+1100 [DEBUG] [cli] Importing service "service:svc" from working dir: "/home/ryan/whatever/bentoml/src"
2023-02-15T18:32:30+1100 [DEBUG] [cli] 'transcription_pipeline' imported from source: bentoml.Service(name="transcription_pipeline", import_str="service:svc", working_dir="/home/ryan/whatever/bentoml/src")
2023-02-15T18:32:30+1100 [DEBUG] [cli] Runner map: {}
2023-02-15T18:32:30+1100 [INFO] [cli] Prometheus metrics for HTTP BentoServer from "service:svc" can be accessed at http://localhost:3000/metrics.
2023-02-15T18:32:31+1100 [INFO] [cli] Installing handle_callback_exception to loop
2023-02-15T18:32:31+1100 [INFO] [cli] Registering signals...
2023-02-15T18:32:31+1100 [INFO] [cli] Starting master on pid 2603823
2023-02-15T18:32:31+1100 [DEBUG] [cli] Socket bound at 0.0.0.0:3000 - fd: 3
2023-02-15T18:32:31+1100 [INFO] [cli] sockets started
2023-02-15T18:32:31+1100 [DEBUG] [cli] Initializing watchers
2023-02-15T18:32:31+1100 [DEBUG] [cli] cmd: /usr/bin/python
[...]

PR #3597, opened 2023-02-21
fix(serve-cli): Make sure to use BENTOML_CONFIG value

aarnphm:
address #3570

Make sure that value from CLI uses BentoMLContainer as default.


sauyon (approved):
👍

PR #3601, opened 2023-02-22
fix: invalid option dtype=True for pd.read_csv

parano:
## What does this PR address?

Fixes #(issue)

## Before submitting:

### DIFF
--- a/src/bentoml/_internal/io_descriptors/pandas.py
+++ b/src/bentoml/_internal/io_descriptors/pandas.py
@@ -429,8 +429,7 @@ class PandasDataFrame(
                 ) from None
         self._shape = sample.shape
         self._columns = [str(i) for i in list(sample.columns)]
-        if self._dtype is None:
-            self._dtype = True  # infer dtype automatically
+        self._dtype = sample.dtypes
         return sample
 
     def _convert_dtype(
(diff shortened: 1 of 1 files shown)


src/bentoml/_internal/io_descriptors/pandas.py
```diff
@@ -527,12 +526,15 @@ async def from_http_request(self, request: Request) -> ext.PdDataFrame:
         _validate_serialization_format(serialization_format)
 
         obj = await request.body()
+        dtype = self._dtype if self._enforce_dtype else None
```
parano, line 529: this could be a breaking change for users using `from_sample` or setting `dtype` today 🤔

sauyon: I think this breaks spark, because even if enforce dtype is false we use the user `dtype` to determine output table dtype.

parano: @sauyon this is only in the from_http_request code path, does not affect the spark schema method I believe

src/bentoml/_internal/io_descriptors/pandas.py
```diff
@@ -429,8 +429,7 @@ def predict(inputs: pd.DataFrame) -> pd.DataFrame: ...
                 ) from None
         self._shape = sample.shape
         self._columns = [str(i) for i in list(sample.columns)]
-        if self._dtype is None:
-            self._dtype = True  # infer dtype automatically
+        self._dtype = sample.dtypes
```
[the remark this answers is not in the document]
parano: yes but it should be very rare - the only case that it introduces a behavior change is when the user has specified a `dtype` but with `enforce_dtype=False` - previously it will still pass through the `dtype` to `pd.read_json`, but now it won't. This is also very unlikely because enforce_dtype doesn't do anything today.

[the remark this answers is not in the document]
sauyon: Can we just do

```
if self._dtype is None:
    self._dtype = sample.dtypes
```

instead?

src/bentoml/_internal/io_descriptors/pandas.py
sauyon: Ah, right, I thought this was in the constructor.

src/bentoml/_internal/io_descriptors/pandas.py
aarnphm, line 432:
```suggestion
        if self._dtype is None:
            self._dtype = sample.dtypes
```

aarnphm: cc @parano

aarnphm (approved):

src/bentoml/_internal/io_descriptors/pandas.py
parano: LGTM!

PR #3602, opened 2023-02-22
feat: enable cork for non-batched workloads

sauyon:
Enable cork for non-batched workloads, and immediately dispatch a batch if the queue contains enough requests.

This didn't actually help much with the case I was looking at, but we should have done this to begin with anyway.


aarnphm (approved):

aarnphm:
cc @bojiang

PR #3626, opened 2023-03-01
feat: make file io descriptor allow any mime type by default

sauyon:
I think this makes more sense; the file IO descriptor shouldn't care what kind of data it receives unless the user specifies it.


aarnphm (dismissed):

sauyon:
I need to fix tests for this one.

aarnphm (approved):

aarnphm (approved):
👍

PR #3632, opened 2023-03-03
feat: add ray deploy support

parano:
## What does this PR address?

A quick POC of deploying Bento format to serve on Ray cluster

To try it out with the quickstart bento: 

```python
# demo.py
from bentoml.ray import deploy_bento

iris_classifier = deploy_bento('iris_classifier:latest')
```
```bash
serve run demo:iris_classifier
```

TODO:
- [ ] add documentation
- [ ] support adaptive batching
- [ ] support passing through Ray deployment arguments for API server and runners

Fixes #(issue)

## Before submitting:

### DIFF
--- a/src/bentoml/_internal/ray/__init__.py
+++ b/src/bentoml/_internal/ray/__init__.py
@@ -1,0 +1,112 @@
+from __future__ import annotations
+
+import typing as t
+
+import requests
+from ray.serve._private.http_util import ASGIHTTPSender
+
+import bentoml
+from bentoml import Tag
+
+from ...exceptions import BentoMLException
+from ...exceptions import MissingDependencyException
+from ..bento.bento import BentoRunnerInfo
+from .runner_handle import RayRunnerHandle
+
+try:
+    import ray
+    from ray import serve
+    from ray.runtime_env import RuntimeEnv
+    from ray.serve.deployment import Deployment
+except ImportError:  # pragma: no cover
+    raise MissingDependencyException(
+        "'ray[serve]' is required in order to use module 'bentoml.ray', install with 'pip install -U \"ray[serve]\"'. For more information, refer to https://docs.ray.io/",
+    )
+
+
+def _get_runner_deployment(
+    bento_tag: str | Tag, runner_info: BentoRunnerInfo
+) -> Deployment:
+    class RunnerDeployment:
+        def __init__(self):
+            svc = bentoml.load(bento_tag)
+            runner = next(
+                (runner for runner in svc.runners if runner.name == runner_info.name),
+                None,
+            )
+            if runner is None:
+                raise BentoMLException(
+                    f"Runner '{runner_info.name}' not found in Bento({bento_tag})"
+                )
+            runner.init_local(quiet=True)
+            self._runner_instance = runner
+
+            for method in runner.runner_methods:
+
+                async def _func(self, *args, **kwargs):
+                    return await getattr(self._runner_instance, method.name).async_run(
+                        *args, **kwargs
+                    )
+
+                setattr(RunnerDeployment, method.name, _func)
+
+    # TODO: apply RunnerMethod's max batch size configs
+    return serve.deployment(RunnerDeployment, name=f"bento-runner-{runner_info.name}")
+
+
+def _get_service_deployment(bento_tag: str, *args, **kwargs) -> Deployment:
+    bento = bentoml.get(bento_tag)
+
+    @serve.deployment(name=f"bento-svc-{bento.tag.name}", *args, **kwargs)
+    class BentoDeployment:
+        def __init__(self, **runner_deployments: t.Dict[str, Deployment]):
+            from ..server.http_app import HTTPAppFactory
+
+            svc = bentoml.load(bento_tag)
+            # Use Ray's metrics setup. Ray has already initialized a prometheus client
+            # for multi-process and this conflicts with BentoML's prometheus setup
+            self.app = HTTPAppFactory(svc, enable_metrics=False)()
+
+            for runner in svc.runners:
+                assert runner.name in runner_deployments
+                runner._set_handle(RayRunnerHandle, runner_deployments[runner.name])
+
+        async def __call__(self, request: requests.Request):
+            sender = ASGIHTTPSender()
+            await self.app(request.scope, receive=request.receive, send=sender)
+            return sender.build_asgi_response()
+
+    return BentoDeployment
+
+
+def get_bento_runtime_env(bento_tag: str | Tag) -> RuntimeEnv:
+    # TODO: create Ray RuntimeEnv from Bento env configs
+    return RuntimeEnv(
+        pip=[],
+        conda=[],
+        env_vars=[],
+    )
+
+
+def _deploy_bento_runners(bento_tag: str | Tag) -> t.Dict[str, Deployment]:
+    bento = bentoml.get(bento_tag)
+
+    # Deploy BentoML Runners as Ray serve deployments
+    runner_deployments = {}
+    for runner_info in bento.info.runners:
+        runner_deployment = _get_runner_deployment(bento_tag, runner_info).bind()
+        runner_deployments[runner_info.name] = runner_deployment
+
+    return runner_deployments
+
+
+def deploy_bento(
+    bento_tag: str | Tag, ray_deployment_kwargs: t.Dict | None = None
+) -> None:
+    # TODO: allow user to pass through compute resource configs to Ray
+    runner_deployments = _deploy_bento_runners(bento_tag)
+    ray_deployment_kwargs = ray_deployment_kwargs if ray_deployment_kwargs else {}
+
+    return _get_service_deployment(bento_tag, **ray_deployment_kwargs).bind(
+        **runner_deployments
+    )
--- a/src/bentoml/_internal/runner/runner.py
+++ b/src/bentoml/_internal/runner/runner.py
@@ -130,11 +130,14 @@ class Runner(AbstractRunner):
     )
     _runner_handle: RunnerHandle = attr.field(init=False, factory=DummyRunnerHandle)
 
-    def _init(self, handle_class: t.Type[RunnerHandle]) -> None:
+    def _set_handle(
+        self, handle_class: t.Type[RunnerHandle], *args: P.args, **kwargs: P.kwargs
+    ) -> None:
         if not isinstance(self._runner_handle, DummyRunnerHandle):
             raise StateException("Runner already initialized")
 
-        object_setattr(self, "_runner_handle", handle_class(self))
+        runner_handle = handle_class(self, *args, **kwargs)
+        object_setattr(self, "_runner_handle", runner_handle)
 
     @inject
     async def runner_handle_is_ready(
@@ -287,7 +290,7 @@ class Runner(AbstractRunner):
         from .runner_handle.local import LocalRunnerRef
 
         try:
-            self._init(LocalRunnerRef)
+            self._set_handle(LocalRunnerRef)
         except Exception as e:
             import traceback
 
@@ -299,10 +302,15 @@ class Runner(AbstractRunner):
 
             raise e
 
-    def init_client(self):
-        from .runner_handle.remote import RemoteRunnerClient
+    def init_client(
+        self, handle_class: t.Type[RunnerHandle], *args: P.args, **kwargs: P.kwargs
+    ):
+        if handle_class is None:
+            from .runner_handle.remote import RemoteRunnerClient
 
-        self._init(RemoteRunnerClient)
+            self._set_handle(RemoteRunnerClient)
+        else:
+            self._set_handle(handle_class, *args, **kwargs)
 
     def destroy(self):
         """
--- a/src/bentoml/_internal/ray/runner_handle.py
+++ b/src/bentoml/_internal/ray/runner_handle.py
@@ -1,0 +1,63 @@
+from __future__ import annotations
+
+import typing as t
+import functools
+
+from ..runner import Runner
+from ...exceptions import MissingDependencyException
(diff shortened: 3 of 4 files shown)


src/bentoml/_internal/service/service.py
```diff
@@ -30,12 +31,13 @@
     from ..bento import Bento
     from ...grpc.v1 import service_pb2_grpc as services
     from .openapi.specification import OpenAPISpecification
-
 else:
     grpc, _ = import_grpc()
 
 logger = logging.getLogger(__name__)
 
+SerializationStrategy = t.Literal["EXPORT_BENTO", "LOCAL_BENTO", "REMOTE_BENTO"]
```
aarnphm, line 39: We should make this an enum 😄

parano: I slightly prefer using string literal for this case

src/bentoml/__init__.py
```diff
@@ -125,7 +126,9 @@
     )
     xgboost = _LazyLoader("bentoml.xgboost", globals(), "bentoml.xgboost")
 
+    # Integrations
     triton = _LazyLoader("bentoml.triton", globals(), "bentoml.triton")
+    ray = _LazyLoader("bentoml.ray", globals(), "bentoml.ray")
```
aarnphm, line 131: make sure to also import `ray` in the TYPE_CHECKING block above.

src/bentoml/_internal/configuration/containers.py
```diff
@@ -177,10 +177,16 @@ def bentoml_home() -> str:
         bentos = os.path.join(home, "bentos")
         models = os.path.join(home, "models")
         envs = os.path.join(home, "envs")
+        tmp_bentos = os.path.join(home, "tmp")
```
aarnphm, line 180: Why do we need a persistent tmp bento store? seems a bit unintuitive 😄

src/bentoml/_internal/ray/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+from .runner_handle import RayRunnerHandle
+
+try:
+    import ray
+    from ray import serve
+    from ray.runtime_env import RuntimeEnv
+    from ray.serve.deployment import Deployment
+except ImportError:  # pragma: no cover
+    raise MissingDependencyException(
+        "'ray[serve]' is required in order to use module 'bentoml.ray', install with 'pip install -U \"ray[serve]\"'. For more information, refer to https://docs.ray.io/",
```
aarnphm, line 21:
```suggestion
        """'ray[serve]' is required in order to use module 'bentoml.ray', install with 'pip install -U "ray[serve]"'. See https://docs.ray.io/ for more information.""",
```

src/bentoml/_internal/ray/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+        runner_deployment = _get_runner_deployment(
+            svc, runner.name, runner_deployment_config
+        ).bind()
+        runner_deployments[runner.name] = runner_deployment
+
+    return runner_deployments
+
+
+def deployment(
+    target: str | Tag | bentoml.Service,
```
aarnphm, line 103: Should we also accept `Bento` as arg?

```python

bentoml.ray.deployment(bentoml.get("my_huggingface_model"))
```

src/bentoml/_internal/ray/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+    service_deployment_config: t.Dict[str | t.Any] | None = None,
+    runners_deployment_config_map: t.Dict[str | t.Dict[str | t.Any]] | None = None,
```
aarnphm, line 105:
```suggestion
    service_deployment_config: dict[str | t.Any] | None = None,
    runners_deployment_config_map: dict[str | dict[str | t.Any]] | None = None,
```

src/bentoml/_internal/ray/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+    """
+    Deploy a Bento or bentoml.Service to Ray
+
+    Args:
+        target:
+            A bentoml.Service instance or Bento tag string
+        service_deployment_config:
+            Ray deployment config for BentoML API server
+        runners_deployment_config:
+            Ray deployment config map for all Runners
```
aarnphm, line 116:
```suggestion
        target: A bentoml.Service instance or Bento tag string
        service_deployment_config: Ray deployment config for BentoML API server
        runners_deployment_config: Ray deployment config map for all Runners
```

src/bentoml/_internal/ray/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+    Configuring BentoML-on-Ray deployment:
+
+    .. code-block:: python
+        import bentoml
+
+        classifier = bentoml.ray.deployment(
+            'iris_classifier:latest',
+            {"route_prefix": "/hello", "num_replicas": 3, "ray_actor_options": {"num_cpus": 1}},
+            {"iris_clf": {"num_replicas": 1, "ray_actor_options": {"num_cpus": 5}}}
+        )
```
aarnphm, line 142:
```suggestion
    .. code-block:: python
       :caption: `ray_integration.py`

       import bentoml

       classifier = bentoml.ray.deployment('iris_classifier:latest')

    .. code-block:: bash

       $ serve run ray_demo:classifier

    Configure BentoML-on-Ray deployment:

    .. code-block:: python
       :caption: `deployment.py`
    
       import bentoml

       classifier = bentoml.ray.deployment(
           'iris_classifier:latest',
           {"route_prefix": "/hello", "num_replicas": 3, "ray_actor_options": {"num_cpus": 1}},
           {"iris_clf": {"num_replicas": 1, "ray_actor_options": {"num_cpus": 5}}}
       )
```

src/bentoml/_internal/ray/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+    # TODO: create Ray RuntimeEnv from Bento env configs
+    return RuntimeEnv(
+        pip=[],
+        conda=[],
+        env_vars=[],
+    )
+
+
+def _deploy_bento_runners(
+    svc: bentoml.Service, runners_deployment_config_map: t.Dict | None = None
```
aarnphm, line 84:
```suggestion
    svc: bentoml.Service, runners_deployment_config_map: dict[str, t.Any] | None = None
```

src/bentoml/_internal/ray/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+) -> t.Dict[str, Deployment]:
```
aarnphm, line 85:
```suggestion
) -> dict[str, Deployment]:
```

src/bentoml/_internal/runner/runner.py
```diff
@@ -130,11 +130,14 @@ def __getattr__(self, item: str) -> t.Any:
     )
     _runner_handle: RunnerHandle = attr.field(init=False, factory=DummyRunnerHandle)
 
-    def _init(self, handle_class: t.Type[RunnerHandle]) -> None:
+    def _set_handle(
```
aarnphm, line 133: This is a backward incompatible change. Eventhough we don't document this API, I have seen people in the community use this. Maybe we can deprecate it?

src/bentoml/_internal/runner/runner.py
```diff
@@ -130,11 +130,14 @@ def __getattr__(self, item: str) -> t.Any:
...
+        self, handle_class: t.Type[RunnerHandle], *args: P.args, **kwargs: P.kwargs
```
aarnphm, line 134:
```suggestion
        self, handle_class: type[RunnerHandle], *args: P.args, **kwargs: P.kwargs
```

src/bentoml/_internal/runner/runner.py
```diff
@@ -299,10 +302,15 @@ def init_local(self, quiet: bool = False) -> None:
 
             raise e
 
-    def init_client(self):
-        from .runner_handle.remote import RemoteRunnerClient
+    def init_client(
+        self, handle_class: t.Type[RunnerHandle], *args: P.args, **kwargs: P.kwargs
```
aarnphm, line 306:
```suggestion
        self, handle_class: type[RunnerHandle] | None, *args: P.args, **kwargs: P.kwargs
```

src/bentoml/_internal/runner/runner.py
aarnphm: You will probably also need to update triton remote handle. It is under `src/bentoml/triton.py` with this new API.

src/bentoml/_internal/service/service.py
aarnphm, line 39: If this is only a type, then we should move `t.Literal` under TYPE_CHECKING block. 

`t.Literal` is not available in 3.7 and it should always live under `t.TYPE_CHECKING`.

src/bentoml/_internal/service/service.py
```diff
@@ -120,6 +122,73 @@ class Service:
...
 
+    __serialization_strategy: SerializationStrategy = "EXPORT_BENTO"
+
+    @property
+    def serialization_strategy(self):
+        return self.__serialization_strategy
+
+    @serialization_strategy.setter
+    def serialization_strategy(self, serialization_strategy: SerializationStrategy):
+        self.__serialization_strategy = serialization_strategy
```
aarnphm, line 133: This won't work because Service is a frozen object, which `on_setattr` will raise an error when setting attributes.

src/bentoml/_internal/service/service.py
```diff
@@ -120,6 +122,73 @@ class Service:
...
+        """
+        Make bentoml.Service pickle serializable
+        """
+        import fs
+
+        from bentoml._internal.bento.bento import Bento
+        from bentoml._internal.yatai_client import yatai_client
+        from bentoml._internal.service.loader import load
+        from bentoml._internal.service.loader import load_bento_dir
+        from bentoml._internal.configuration.containers import BentoMLContainer
```
aarnphm, line 145: relative import?

aarnphm (commented):
quick initial pass

src/bentoml/_internal/configuration/containers.py
parano: These Bentos are meant for a temporary cache and not suppose to be visible to users.

src/bentoml/_internal/ray/__init__.py
parano: Yes, good idea!

src/bentoml/_internal/service/service.py
parano: will move to a global config instead

src/bentoml/_internal/runner/runner.py
parano: @aarnphm this is each runner's own implementation detail, no need to change the triton runner

parano: You probably see users using `init_local`, which is a public API

aarnphm: ok I can update the API change once this is in.

aarnphm: > You probably see users using `init_local`, which is a public API

You are correct, but because we don't provide the handle class before. Users have been using `_init`. So this is still a breaking change.

ssheng:
Runtime environment and dependencies is an interesting topic for Ray. Ray supports two modes, basically [cluster-level and on-demand](https://docs.ray.io/en/latest/ray-core/handling-dependencies.html). Their recommendations for production is to use the cluster-level option. Converting `bentofile` dependencies to Ray's runtime environment can be potentially slow launching the nodes. At the very least, we should provide options for setting up environments.

src/bentoml/_internal/runner/runner.py
parano: If user code is invoking `init_local`, this won't be a breaking change

parano:
@ssheng added batching support and left the env set up as a TODO item for now, it's ready for another look

src/bentoml/_internal/runner/runner.py
aarnphm: but if user invoke runner._init it will fail right?

src/bentoml/bentos.py
```diff
@@ -238,14 +239,13 @@ def push(
...
+    _yatai_client: YataiClient = Provide[BentoMLContainer.yatai_client],
 ):
     """Push Bento to a yatai server."""
-    from bentoml._internal.yatai_client import yatai_client
-
     bento = _bento_store.get(tag)
     if not bento:
         raise BentoMLException(f"Bento {tag} not found in local store")
-    yatai_client.push_bento(bento, force=force)
+    _yatai_client.push_bento(bento, force=force)
```
aarnphm, line 248: can we refactor this change out to a different PR?

parano: This is necessary for the unit tests

aarnphm: oh I mean the change for adding the yatai_client to `BentoMLContainer`.

parano: yes, this refactor makes it easier to test the REMOTE pulling serialization strategy

aarnphm:
So the test failure is irrelevant of this. Merging this now so that I will update the Triton runner implementation.

aarnphm (approved):

judahrand:
Are the missing docs going to be added?

parano:
Hi @judahrand - Ray support is still an experimental feature, we have yet to publish the documentation. You may follow the docstring here if you want to try out this feature: https://github.com/bentoml/BentoML/blob/main/src/bentoml/_internal/ray/__init__.py#L140

Although beaware there might be breaking API changes in future releases. This implementation also assumes user have installed all dependencies in the Ray cluster, which we will improve in future versions.

PR #3633, opened 2023-03-03
revert(pyproject): add back pytest plugins

aarnphm:
cc @sauyon

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -236,7 +236,7 @@ exclude = '''
 '''
 
 [tool.pytest.ini_options]
-addopts = ["-rfEX", "-pno:warnings"]
+addopts = ["-rfEX", "-pno:warnings", "-pbentoml.testing.pytest.plugin"]
 python_files = ["test_*.py", "*_test.py"]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -236,7 +236,7 @@ exclude = '''
 '''
 
 [tool.pytest.ini_options]
-addopts = ["-rfEX", "-pno:warnings"]
+addopts = ["-rfEX", "-pno:warnings", "-pbentoml.testing.pytest.plugin"]
```
sauyon, line 239: Why did you disable warnings here again?

sauyon (approved):

pyproject.toml
aarnphm: oh to silent useless warning from tensorflow and flax.

sauyon: Which ones? Can we fix those without silencing warnings globally?

aarnphm: Those warnings aren't related to us. It is a protobuf deprecation on Tensorflow end.

PR #3637, opened 2023-03-04
chore: align implementation for gRPC healthcheck

aarnphm:

### DIFF
--- a/src/bentoml/_internal/client/grpc.py
+++ b/src/bentoml/_internal/client/grpc.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import time
 import typing as t
+import asyncio
 import logging
 import functools
 from typing import TYPE_CHECKING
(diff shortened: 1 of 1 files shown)


src/bentoml/_internal/client/grpc.py
```diff
@@ -221,13 +222,16 @@ def _rpc_metadata(self) -> dict[str, dict[str, t.Any]]:
...
-    async def health(self, service_name: str, *, timeout: int = 30) -> t.Any:
+    async def async_health(self, service_name: str, *, timeout: int = 30) -> t.Any:
         return await self._invoke(
             method_name="/grpc.health.v1.Health/Check",
             service=service_name,
             _grpc_channel_timeout=timeout,
         )
 
+    def health(self, service_name: str, *, timeout: int = 30) -> t.Any:
+        return asyncio.run(self.async_health(service_name, timeout=timeout))
```
sauyon, line 233: Should we do something like what we do with runners and `anyio` to avoid issues with running loops?

aarnphm: we will have to spawn an anyio worker right?

PR #3646, opened 2023-03-09
fix(dispatcher): prevent possible dispatcher bad behavior

sauyon:
Previously, it was possible that if request load was fast enough and max latency set low enough, the dispatcher would enter a loop of waiting for new requests while cancelling the oldest ones while doing no work, this change seeks to fix that.

 - The optimizer values for slope and intercept are now bounded below at 0; we may want to eventually try to collect cases where the linear optimizer does this, but for now we do this to avoid the most pathological cases.
 - If the first request is in danger of being cancelled, we now stop waiting and begin processing immediately.

### DIFF
--- a/src/bentoml/_internal/marshal/dispatcher.py
+++ b/src/bentoml/_internal/marshal/dispatcher.py
@@ -318,12 +318,23 @@ class CorkDispatcher:
                         continue
                     await asyncio.sleep(self.tick_interval)
                     continue
-                if (
+
+                # we are now free to dispatch whenever we like
+                while (
+                    # if we don't already have enough requests,
                     n < self.max_batch_size
-                    and n * (wn + dt + (a or 0)) <= self.optimizer.wait * decay
+                    # we are not about to cancel the first request,
+                    and latency_0 + dt <= self.max_latency_in_ms * 0.95
+                    # and waiting will cause average latency to decrese
+                    and n * (wn + dt + a) <= self.optimizer.wait * decay
                 ):
+                    n = len(self._queue)
+                    now = time.time()
+                    wn = now - self._queue[-1][0]
+                    latency_0 = w0 + a * n + b
+
+                    # wait for additional requests to arrive
                     await asyncio.sleep(self.tick_interval)
-                    continue
 
                 if self.max_batch_size == -1:  # batching is disabled
                     n_call_out = 1


src/bentoml/_internal/marshal/dispatcher.py
```diff
@@ -351,10 +351,13 @@ async def controller(self):
                 now = time.time()
                 w0 = now - self._queue[0][0]
                 wn = now - self._queue[-1][0]
-                a = self.optimizer.o_a
-                b = self.optimizer.o_b
+                # bound optimizer values at 0; negative values may cause optimizer weirdness
+                a = max(self.optimizer.o_a, 0)
+                b = max(self.optimizer.o_b, 0)
```
larme, line 356: I think `prevent possible dispatcher bad behavior` mainly means these 2 lines right? The added while loop below is mainly for better code flow?

sauyon: No, the loop below also changes the logic very slightly; it prevents any requests from being canceled once we get into the waiting phase.

src/bentoml/_internal/marshal/dispatcher.py
```diff
@@ -363,12 +366,23 @@ async def controller(self):
...
                     continue
-                if (
+
+                # we are now free to dispatch whenever we like
+                while (
+                    # if we don't already have enough requests,
                     n < self.max_batch_size
-                    and n * (wn + dt + (a or 0)) <= self.optimizer.wait * decay
+                    # we are not about to cancel the first request,
+                    and latency_0 + dt <= self.max_latency_in_ms * 0.95
```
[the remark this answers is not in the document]
sauyon: They're different things; this is just to prevent requests from going past max latency so I put in an arbitrary limiter, the other is obviously wait time delay. I actually think delay should probably be something like .5, but that change is for another day.

src/bentoml/_internal/marshal/dispatcher.py
sauyon: Removed these, they were being done by the optimizer anyway.

bojiang:
To whom it may concern:

Sauyon and I have reached an agreement: no major modifications will be made to adaptive batching before establishing benchmark tests, and only minor changes to ensure safety will be implemented. This agreement does not affect the addition of new batching strategies.

The reason is that the adaptive batching strategy has dynamic characteristics and faces complex scenarios. At present, there are no perfect testing scenarios like those during development (in version 0.13), making it temporarily unmaintainable. The existing batching strategy has been tested over time, and making significant changes without thorough testing may lead to some edge cases. Therefore, major modifications should first ensure continuous integration (CI) and restore the maintainability of adaptive batching.
-----
@sauyon  Is this ready? Maybe let's discuss tomorrow so that we can merge this and another related PR, shall we?

sauyon:
> @sauyon Is this ready? Maybe let's discuss tomorrow so that we can merge this and another related PR, shall we?

Sounds good, when do you want to chat?

PR #3649, opened 2023-03-10
fix(triton): enable runtime options

aarnphm:
Set users to pass options via runners options in service definition

### DIFF
--- a/src/bentoml/start.py
+++ b/src/bentoml/start.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import os
 import json
-import typing as t
 import logging
 import contextlib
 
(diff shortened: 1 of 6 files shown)


src/bentoml/start.py
```diff
@@ -100,12 +98,16 @@ def start_runner_server(
                     )
                     break
                 else:
-                    triton_handle = construct_triton_handle(runner, **attrs)
+                    cli_args = runner.cli_args + [
+                        f"--http-port={runner.protocol_address.split(':')[-1]}"
+                        if runner.tritonserver_type == "http"
+                        else f"--grpc-port={runner.protocol_address.split(':')[-1]}"
```
ssheng, line 104: If users happen to specify `--http-port` or `--grpc-port`, will they override the ones defined here?

ssheng (dismissed):

src/bentoml/start.py
aarnphm: no, because it appears at the end, so our options should always override users.

ssheng (approved):

PR #3663, opened 2023-03-14
chore(dispatcher): refactor out training code

sauyon:
Refactor to make this code more reasonable.

### DIFF
--- a/src/bentoml/_internal/marshal/dispatcher.py
+++ b/src/bentoml/_internal/marshal/dispatcher.py
@@ -53,7 +53,7 @@ class Optimizer:
         self.o_a = min(2, max_latency * 2.0 / 30)
         self.o_b = min(1, max_latency * 1.0 / 30)
 
-        self.wait = 0.01  # the avg wait time before outbound called
+        self.wait = 0  # the avg wait time before outbound called
 
         self._refresh_tb = TokenBucket(2)  # to limit params refresh interval
         self.outbound_counter = 0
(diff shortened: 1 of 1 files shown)


aarnphm (approved):
Only if github has a better diff view

sauyon:
This is just a nightmare diff in general :P

src/bentoml/_internal/marshal/dispatcher.py
```diff
@@ -168,52 +168,18 @@ async def _func(data: t.Any) -> t.Any:
...
-                # call
-                self._sema.acquire()
-                inputs_info = tuple(self._queue.pop() for _ in range(n_call_out))
-                self._loop.create_task(self.outbound_call(inputs_info))
-            except asyncio.CancelledError:
-                return
-            except Exception as e:  # pylint: disable=broad-except
-                logger.error(traceback.format_exc(), exc_info=e)
+    async def train_optimizer(
+        self, num_required_reqs: int, num_reqs_to_train: int, batch_size: int, wait: 0,
```
aarnphm, line 172:
```suggestion
        self, num_required_reqs: int, num_reqs_to_train: int, batch_size: int, wait: int = 0,
```
I think this is what you mean?

sauyon: Oh, no, I just forgot to remove that argument.

aarnphm (approved):
LGTM. cc @bojiang

PR #3664, opened 2023-03-14
fix(dispatcher): set wait to 0 while training

sauyon:
This is a minor fix for an issue I introduced due to misunderstanding the wait time; because the trainers will wait a decent amount of time before releasing to batching, slow early requests could cause expected wait time to balloon. For now, we are just being conservative about batching early.


aarnphm (commented):
So this will set `self.wait=0` i assume?

sauyon:
> So this will set `self.wait=0` i assume?

Not sure what you mean, you can check out the last commit if you want to review this now (or use a local client to diff vs. the refactor, but that seems like a lot of effort.)

aarnphm:
> > So this will set `self.wait=0` i assume?
> 
> Not sure what you mean, you can check out the last commit if you want to review this now (or use a local client to diff vs. the refactor, but that seems like a lot of effort.)

I think what I mean here is this would only change the wait to 0? since it includes the code from the training PR?

bojiang:
To whom it may concern:

Sauyon and I have reached an agreement: no major modifications will be made to adaptive batching before establishing benchmark tests, and only minor changes to ensure safety will be implemented. This agreement does not affect the addition of new batching strategies.

The reason is that the adaptive batching strategy has dynamic characteristics and faces complex scenarios. At present, there are no perfect testing scenarios like those during development (in version 0.13), making it temporarily unmaintainable. The existing batching strategy has been tested over time, and making significant changes without thorough testing may lead to some edge cases. Therefore, major modifications should first ensure continuous integration (CI) and restore the maintainability of adaptive batching.

-------

@sauyon Is this minor wait time bug introduced in your last change on the dispatcher or easier changes?

sauyon:
No, this is an issue I introduced where training requests would wait for almost max latency at first, and so average wait would be inflated at the beginning. For now, just setting it to 0 for only training requests.

bojiang (dismissed):

aarnphm (approved):

aarnphm:
@sauyon sry for the ping but it seems there is merge conflict

aarnphm (approved):

src/bentoml/_internal/marshal/dispatcher.py
```diff
@@ -248,6 +248,9 @@ async def controller(self):
                 # call
                 self._sema.acquire()
                 inputs_info = tuple(self._queue.pop() for _ in range(n_call_out))
+                for info in inputs_info:
+                    # fake wait as 0 for training requests
+                    info[0] = now
```
sauyon, line 253:
```suggestion
                    info.enqueue_time = now
```

aarnphm (approved):

PR #3682, opened 2023-03-17
feat: torch 2.0

aarnphm:
This PR aims to bring Torch 2.0 support to `save_model` via `torch.compile`

### DIFF
--- a/src/bentoml/_internal/frameworks/pytorch.py
+++ b/src/bentoml/_internal/frameworks/pytorch.py
@@ -5,18 +5,19 @@ import logging
 from types import ModuleType
 from typing import TYPE_CHECKING
 from pathlib import Path
+from functools import partial
 
 import cloudpickle
 
 import bentoml
-from bentoml import Tag
 
+from ..tag import Tag
 from ..types import LazyType
 from ..models import Model
 from ..utils.pkg import get_pkg_version
 from ...exceptions import NotFound
 from ..models.model import ModelContext
-from ..models.model import PartialKwargsModelOptions as ModelOptions
+from ..models.model import PartialKwargsModelOptions
 from .common.pytorch import torch
 from .common.pytorch import PyTorchTensorContainer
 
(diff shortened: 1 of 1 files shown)


src/bentoml/_internal/frameworks/pytorch.py
```diff
@@ -43,9 +44,19 @@ def get(tag_like: str | Tag) -> Model:
...
 
+class ModelOptions(PartialKwargsModelOptions):
+    fullgraph: bool = False
+    dynamic: bool = False
+    backend: t.Union[str, t.Callable[..., t.Any]] = "inductor"
+    mode: t.Optional[str] = None
+    options: t.Optional[t.Dict[str, t.Union[str, int, bool]]] = None
+    disable: bool = False
+
+
```
larme, line 55: First, let’s call this `PytorchOptions` and only when import to `bentoml.pytorch` we rename it to `ModelOptions`.

Second, I think maybe it’s better to have single `compile_kwargs` dict in `PytorchOptions` instead of polluting the name space of `PytorchOptions`. Maybe we will have only 2 entries:

- `enable_compile`
- `compile_kwargs`

src/bentoml/_internal/frameworks/pytorch.py
```diff
@@ -43,9 +44,19 @@ def get(tag_like: str | Tag) -> Model:
...
 def load_model(
     bentoml_model: str | Tag | Model,
     device_id: t.Optional[str] = "cpu",
+    **compile_kwargs: t.Any,
```
larme, line 59: I don’t think we need do compile at `load_model` level. Let’s just save the original model and do `torch.compile` when init the runner (if user set `enable_compile=True`)

src/bentoml/_internal/frameworks/pytorch.py
```diff
@@ -195,15 +208,25 @@ def get_runnable(bento_model: Model):
...
+            load_model,
+            fullgraph=opts.fullgraph,
+            dynamic=opts.dynamic,
+            backend=opts.backend,
+            mode=opts.mode,
+            options=opts.options,
+            disable=opts.disable,
+        )
+    else:
+        _load_model = load_model
```
larme, line 223: I think we can just `model = load_model`, and then `torch.compile(model, **compile_kwargs)` if `enable_compile` is True.

larme (changes requested):

PR #3696, opened 2023-03-23
feat: implement new serve API

sauyon:
This gets around the name clash (of the module `bentoml.serve` and the function `bentoml.serve` by introducing a new API:

```python
server = bentoml.HTTPServer(bento, production=True, ...)

client = server.start(block=False)  # block is false by default
client.classify(...)

# or

with server.start() as client:
    client.classify(...)

# note
with server.start(blocking=True) as client:
    client.classify()
# will not work since the start will block
```


bojiang:
@sauyon Shall we put these under `bentoml_cli`? Any thoughts?

sauyon:
> @sauyon Shall we put these under `bentoml_cli`? Any thoughts?

I thought about this, I think these management APIs which call out to the CLI should be under `bentoml.`

src/bentoml/server.py
```diff
@@ -0,0 +1,274 @@
...
+        if env:
+            args.extend(["--env", env])
+
+        if api_workers is not None:
+            args.extend(["--api-workers", str(api_workers)])
+        if working_dir is not None:
+            args.extend(["--working-dir", str(working_dir)])
+
+        self.args = args
+        self.host = "127.0.0.1" if host == "0.0.0.0" else host
```
parano, line 89: What's the reason for this conversion from "0.0.0.0" to "127.0.0.1"?

sauyon: 0.0.0.0 isn't a valid IP address, it's just means bind all IPs, so we convert it to something sensible.

src/bentoml/bentos.py
```diff
@@ -441,81 +441,62 @@ def serve(
...
+        "bentoml.serve and bentoml.bentos.serve are deprecated; use bentoml.Server instead."
+    )
+
     if server_type == "http":
-        serve_cmd = "serve-http"
+        from .server import HTTPServer
+
         if host is None:
-            host = BentoMLContainer.http.host.get()
+            host = t.cast(str, BentoMLContainer.http.host.get())
```
aarnphm, line 453: I think we should use this as the default value in serve as well? so maybe we can change it for `host`, `port`, `api_workers`?

src/bentoml/server.py
```diff
@@ -0,0 +1,296 @@
...
+        elif isinstance(bento, Tag):
+            bento_str = str(bento)
+        else:
+            bento_str = bento
+
+        args: list[str] = [
+            sys.executable,
+            "-m",
+            "bentoml",
+            "serve-http",
```
aarnphm, line 74:
```suggestion
            serve_cmd,
```

src/bentoml/server.py
```diff
@@ -0,0 +1,296 @@
...
+            args.extend(["--api-workers", str(api_workers)])
+        if working_dir is not None:
+            args.extend(["--working-dir", str(working_dir)])
+
+        self.args = args
+        self.host = "127.0.0.1" if host == "0.0.0.0" else host
+        self.port = port
+        self.start = self._create_startmanager()
+
+    def _create_startmanager(server):  # type: ignore # not calling self self
```
aarnphm, line 101:
```suggestion
	@staticmethod
    def _create_startmanager(server: Server):
```

tests/e2e/bento_server_http/tests/test_serve.py
```diff
@@ -0,0 +1,15 @@
+import bentoml
+
+
+def test_http_server(bentoml_home: str):
```
aarnphm, line 4: lets also add test for `server.start()` as a context manager as well

aarnphm (commented):
First comments

src/bentoml/bentos.py
sauyon: This should probably be removed, I made the conscious decision to only support HTTP with `bentoml.serve` for now.

sauyon: Oh, that's not true at all. The rest are provided by injection, I think.

src/bentoml/server.py
sauyon: This isn't static.

src/bentoml/bentos.py
aarnphm: I think for all of the inject we can just make it as default, so that users can use the environment variable with Python API.

sauyon: I don't think I understand, what're you proposing?

aarnphm: We should make any value that is injectable as the default, so that if users specify these configuration via the environment variable it will work when user use the Python API + envvar?

sauyon: I don't think I understand, isn't that already handled by injection?

aarnphm: I think what i mean is 
```python
        host: str = Provide[BentoMLContainer.http.host],
        port: int = Provide[BentoMLContainer.http.port],
```

sauyon: That doesn't work because it's different for gRPC, besides, this is functionally identical.

tests/e2e/bento_server_http/tests/test_serve.py
sauyon: Should be added now!

tests/e2e/bento_server_http/tests/test_serve.py
```diff
@@ -0,0 +1,58 @@
+import bentoml
+import pytest
+
+import time
+import os
+
+
+def test_http_server(bentoml_home: str):
+    print(os.getcwd())
```
aarnphm, line 9:
```suggestion
```

aarnphm (commented):
Final thing. And then LGTM @sauyon

aarnphm:
Tests on windows still seems to be failing

sauyon:
Interesting, seems like on windows the server is exiting non-0.

ssheng:
When setting `blocking=True`, the server does not really block without a context manager.
```
server.start(blocking=True)
```

If a bento does not exist, `start` does not really throw an error.
```
 server  = HTTPServer("non_exist_bento:latest", production=True, port=3001)
    with server.start() as client:
        client.classify(np.array([[4.9, 3.0, 1.4, 0.2]]))
```

The above throws a `ConnectionRefusedError: [Errno 61] Connection refused` error instead of a bento not found error. I think it is due to use of `subprocess`, so any server error will not show.

sauyon:
> 



> When setting `blocking=True`, the server does not really block without a context manager.
> 
> ```
> server.start(blocking=True)
> ```

That's strange, let me try to reproduce. Maybe the server is failing? I should make start raise if it exits with an error, but I'm not actually sure what the correct 

> If a bento does not exist, `start` does not really throw an error.
> 
> ```
>  server  = HTTPServer("non_exist_bento:latest", production=True, port=3001)
>     with server.start() as client:
>         client.classify(np.array([[4.9, 3.0, 1.4, 0.2]]))
> ```
> 
> The above throws a `ConnectionRefusedError: [Errno 61] Connection refused` error instead of a bento not found error. I think it is due to use of `subprocess`, so any server error will not show.

Yeah, there's no logic at all that checks for bentos or anything because it's simply calling out to the CLI, and storing stdin/stdout in pipes. I can probably have stdin/stdout behavior be configurable, but it's hard to catch errors correctly because control is returned to the user immediately when non-blocking.

PR #3712, opened 2023-03-24
feat: EasyOCR

aarnphm:
We now support EasyOCR for BentoML 1.0


.github/workflows/frameworks.yml
```diff
@@ -1,24 +1,20 @@
 name: Frameworks
-
```
parano, line 2: Did you use an auto formatter script to do this?

src/bentoml/_internal/frameworks/easyocr.py
```diff
@@ -0,0 +1,216 @@
...
+    import easyocr
+except ImportError:  # pragma: no cover
+    raise MissingDependencyException(
+        "'easyocr' is required in order to use module 'bentoml.easyocr'. Install easyocr with 'pip install easyocr'."
+    )
+
+try:
+    import cloudpickle
+except ImportError:  # pragma: no cover
+    raise MissingDependencyException(
```
parano, line 30: no need to check this, cloudpickle is already a bentoml dependency

src/bentoml/_internal/frameworks/easyocr.py
```diff
@@ -0,0 +1,216 @@
...
+__all__ = ["load_model", "save_model", "get_runnable", "get", "PyTorchTensorContainer"]
+
+MODULE_NAME = "bentoml.easyocr"
+API_VERSION = "v1"
+MODEL_FILENAME = "saved_model.pkl"
+
+
+logger = logging.getLogger(__name__)
+
+device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
```
parano, line 51: remove unused?

.github/workflows/frameworks.yml
aarnphm: yes

src/bentoml/_internal/frameworks/easyocr.py
aarnphm: there might be cases where cloudpickle is uninstalled afterward?

src/bentoml/_internal/frameworks/easyocr.py
```diff
@@ -0,0 +1,208 @@
...
+    """
+
+    class EasyOCRRunnable(bentoml.Runnable):
+        SUPPORTED_RESOURCES = ("nvidia.com/gpu", "cpu")
+        SUPPORTS_CPU_MULTI_THREADING = True
+
+        def __init__(self):
+            super().__init__()
+
+            self.model = load_model(bento_model)
```
parano, line 186: should we handle assigning GPU device here?

src/bentoml/_internal/frameworks/easyocr.py
parano: That's not a concern since it's required by BentoML.. otherwise, we need to do this for every dependency

src/bentoml/_internal/frameworks/easyocr.py
```diff
@@ -0,0 +1,208 @@
...
+
+if t.TYPE_CHECKING:
+    from ..models.model import ModelSignaturesType
+
+    ListStr = list[str]
+else:
+    ListStr = list
+
+
+__all__ = ["load_model", "save_model", "get_runnable", "get", "PyTorchTensorContainer"]
```
parano, line 36: remove `PyTorchTensorContainer` here?

docs/source/frameworks/index.rst
```diff
@@ -83,7 +83,6 @@ Roadmap
 The following frameworks are supported in pre-1.0 BentoML versions and are being migrated to the new 1.0 API. In the meantime, users may use :ref:`Custom Models <frameworks/index:Custom Models>` as a workaround.
 
 - Detectron
-- EasyOCR
```
parano, line 86: also add it to the supported list above?

docs/source/frameworks/easyocr.rst
parano, line 1: Let's put a simple usage example here?

sauyon (approved):

PR #3673, opened 2023-03-15
feat(containerize): caching pip/conda installation layers

smidm:
## What does this PR address?

When creating an image from bento using `bentoml containerize` the docker image layers with Python environment (pip or conda installed) were invalidated by any source file change. This could result in ~ 10 minutes image build times (e.g. conda installation of pytorch and related packages).

This pull request defers copying of sources after `SETUP_BENTO_COMPONENTS` block where Python environment is created. Only the Python environment specification files (requirements.txt, environment.yml, ...) are copied before the conda or pip installation.

With this patch I reduced image build times from 10 minutes to 10 seconds when the Python environment is not changed.

Note: I placed the `COPY --chown={{ bento__user }}:{{ bento__user }} . ./` line outside of defined blocks, but it should be probably also included in a block of it's own to enable customization in templates. I leave this for the core team to consider.

Fixes #3577

## Before submitting:

- [x] Does the Pull Request follow [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/#summary) naming? Here are [GitHub's
      guide](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request) on how to create a pull request.
- [x] Does the code follow BentoML's code style, both `make format` and `make lint` script have passed ([instructions](https://github.com/bentoml/BentoML/blob/main/DEVELOPMENT.md#style-check-auto-formatting-type-checking))?
- [x] Did you read through [contribution guidelines](https://github.com/bentoml/BentoML/blob/main/CONTRIBUTING.md#ways-to-contribute) and follow [development guidelines](https://github.com/bentoml/BentoML/blob/main/DEVELOPMENT.md#start-developing)?
- [ ] Did you write tests to cover your changes?
  - I tested the fix manually.

PR #3713, opened 2023-03-26
fix: copy files before running `setup_script`

aarnphm:
This PR fixes a bug from #3673 where it only copies the pip layers first. This
breaks the setup_script features if setup_script requires files inside src layer

The copy layer should be moved before the setup_script is ran.


parano:
LGTM!

sauyon:
Maybe rename the title to be something more descriptive of the problem, though!

sauyon (approved):
:+1:

Issue #3674, opened 2023-03-15
Access query params in bentoml.Context


Lorenzo-N:
### Feature request

I need to retrieve the query params within an API. I saw in the documentation that I can access the request context through the bentoml.Context ctx argument, like in the following example:

```
def predict(input_array: np.ndarray, ctx: bentoml.Context) -> np.ndarray:
    # get request headers
    request_headers = ctx.request.headers
```

However, looking at the implementation I see that only the headers are contained in ctx.request.
I would like to retrieve also the query params in the context.

### Motivation

I think this feature is useful for being able to access additional model parameters that are separate from the data to be predicted.

For example in a timeseries forecasting model I would like to pass the series data in POST and a parameter in GET to specify the time interval to predict. Currently to do this I can:
- pass everything in post in json format with params and data instead of a simple numpy array
- pass the additional parameter in a header

However, I think the more correct way would be to be able to access the query params so as to pass the numpy array in POST and the additional parameter in GET.

### Other

To do this I think you just need to modify the line [context.py:104](https://github.com/bentoml/BentoML/blob/main/src/bentoml/_internal/context.py#L104) by adding `request.query_params` in the constructor and the `InferenceApiContext.RequestContext` class accordingly.

PR #3717, opened 2023-03-27
feat: add query params to request context

sauyon:
See #3674.

### DIFF
--- a/src/bentoml/_internal/context.py
+++ b/src/bentoml/_internal/context.py
@@ -97,7 +97,7 @@ class InferenceApiContext:
         def __init__(self, metadata: Metadata, query_params: Metadata):
             self.metadata = metadata
             self.headers = metadata
-            self.query_params = metadata
+            self.query_params = query_params
 
         @staticmethod
         def from_http(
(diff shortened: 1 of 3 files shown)


src/bentoml/_internal/context.py
```diff
@@ -92,16 +92,21 @@ def from_http(request: "starlette.requests.Request") -> "InferenceApiContext":
     class RequestContext:
         metadata: Metadata
         headers: Metadata
+        query_params: Metadata
 
-        def __init__(self, metadata: Metadata):
+        def __init__(self, metadata: Metadata, query_params: Metadata):
             self.metadata = metadata
             self.headers = metadata
+            self.query_params = metadata
```
Lorenzo-N, line 100: This line should be `self.query_params = query_params`

parano, line 100:
```suggestion
            self.query_params = query_params
```

sauyon: Indeed :facepalm:. Adding a test for this.

sauyon:
Sorry this took a while, was hoping to get #3696 in so I could use it for the tests, but clearly that was somewhat ambitious.

parano:
LGTM!

aarnphm (approved):

PR #3719, opened 2023-03-27
fix(type): `bentoml.container.build` should accept multiple `image_tag`

pmayd:
Correct type hint for `image_tag` so that the tuple can contain more than a single str element.

Currently, IDE (and mypy) is not accepting code like this:

```python
bentoml.container.build(
        bento_tag=model_bento.tag,
        image_tag=(
            "job-sectioning-model-opt:latest",
            f"job-sectioning-model-opt:{sys.argv[1]}"
        ),
        platform="linux/amd64"
    )
```

This is due to the fact that the type hint for `image_tag` is `tuple[str]` and not `tuple[str, ...]` for a tuple with an arbitrary length.

## What does this PR address?

Fixes #(issue)

## Before submitting:

### DIFF
--- a/src/bentoml/container.py
+++ b/src/bentoml/container.py
@@ -37,7 +37,7 @@ logger = logging.getLogger(__name__)
 def build(
     bento_tag: Tag | str,
     backend: t.Literal["docker"] = ...,
-    image_tag: tuple[str] | None = ...,
+    image_tag: tuple[str, ...] | None = ...,
     features: t.Sequence[str] | None = ...,
     *,
     file: PathType | None = ...,


sauyon:
This looks fine to me, was there a particular reason you closed this?

pmayd:
I was not sure that I understood the code correctly, my change assumed that image_tag can indeed multiple tags but I guess it is indeed only meant to be a single value, that's why I closed it?

aarnphm:
I think image_tag can accept multiple value, so your change is correct. We probably need a test for this as well.

aarnphm (approved):
I test this locally. Thanks for the fix

sauyon:
Thanks for this!

PR #3721, opened 2023-03-28
fix: log converting runner name as debug

sauyon:
I'm don't remember why we originally chose warning for this log, I feel like it's not a real issue and so should be debug (or maybe info?)

/cc @aarnphm


aarnphm:
I don't remember either. I think it makes sense to have it to be debug.

sauyon:
Closing, this should be resolved by specifying a runner name manually.

PR #3731, opened 2023-03-29
chore(dispatcher): use attr class instead of a tuple

sauyon:
Makes jobs slightly easier to read.

### DIFF
--- a/src/bentoml/_internal/marshal/dispatcher.py
+++ b/src/bentoml/_internal/marshal/dispatcher.py
@@ -8,6 +8,7 @@ import functools
 import traceback
 import collections
 
+import attr
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


sauyon:
This one has been validated only with a quick server and ~50 requests.

bojiang (changes requested):
This is on the critical path. We shall keep everything lightweight. AFAIK creating tuples is much faster than objects, maybe we should do some benchmark for the attr slots.

aarnphm:
I believe #3664 depends on this PR. What is the decision on this @bojiang @sauyon ?

bojiang (approved):

aarnphm (approved):

PR #3735, opened 2023-03-30
feat: embedded runner

larme:
## What does this PR address?

Fixes #(issue)

## Before submitting:


larme:
`serve_grpc_development` and `serve_http_development` in `bentoml/serve.py`, `grpc_dev_api_server.py` and `http_dev_api_server.py` is not needed because we are merging development and production servers. I kept them now for easier reference when changing test cases. Will clean them up when tests are ready.

src/bentoml/serve.py
```diff
@@ -319,7 +324,12 @@ def serve_http_production(
     from ._internal.configuration.containers import BentoMLContainer
 
     working_dir = os.path.realpath(os.path.expanduser(working_dir))
-    svc = load(bento_identifier, working_dir=working_dir, standalone_load=True)
+
+    standalone_load: bool = False if development_mode else True
```
sauyon, line 328: Why are we loading standalone if in development mode?

src/bentoml/serve.py
```diff
@@ -500,12 +527,41 @@ def serve_http_production(
...
+
+            # NOTE: {} is faster than dict()
+            plugins = [
+                # reloader plugin
+                {
+                    "use": "bentoml._internal.utils.circus.watchfilesplugin.ServiceReloaderPlugin",
+                    "working_dir": working_dir,
+                    "bentoml_home": bentoml_home,
+                },
+            ]
```
sauyon, line 553: Let's allow people to specify reload outside of development mode, I think that's fine.

sauyon (commented):
Looks good, but a few questions.

Do you think it would be doable to pass the embedded runners through the batching layer? Probably for another PR.

src/bentoml/serve.py
larme: The old development code load in standalone.

src/bentoml/serve.py
larme: I have test reload for production server and seems working ok. So now I apply `--reload` for both production and development.

src/bentoml/serve.py
sauyon: Hm, @parano @aarnphm do you remember why this was?

src/bentoml/serve.py
```diff
@@ -884,11 +955,13 @@ def serve_grpc_production(
     )
     scheme = "https" if BentoMLContainer.ssl.enabled.get() else "http"
 
+    close_child_stdin: bool = False if development_mode else True
```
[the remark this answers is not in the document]
larme: This is also the way in old codes. I guess in development we should allow stdin access?

src/bentoml/_internal/models/model.py
```diff
@@ -342,13 +344,22 @@ def to_runner(
...
         from ..runner import Runner
 
+        # TODO: @larme @yetone run this branch only yatai version is incompatible with embedded runner
+        yatai_version = os.environ.get("YATAI_T_VERSION")
+        if embedded and yatai_version:
+            logger.warning(
+                f"Yatai of version {yatai_version} is incompatible with embedded runner, set `embedded=False` for runner {name}"
+            )
+            embedded = False
+
```
larme, line 354: This is for old version of yatai, which will ignore `BentoRunnerInfo`'s embedded field and spawn a pod anyway. In this case, we will still use non-embedded mode for embedded runner.

When yatai is compatible with embedded field, let's add a version check here. @yetone

larme:
@sauyon please review again, thanks.

src/bentoml/_internal/models/model.py
yetone: LGTM!

larme:
will put docs and examples changes in another pr

yetone (approved):

src/bentoml/serve.py
aarnphm: I believe this is a hack around the import module issue. I don't think standalone_load makes sense

src/bentoml/serve.py
```diff
@@ -32,9 +34,9 @@
 
 SCRIPT_RUNNER = "bentoml_cli.worker.runner"
 SCRIPT_API_SERVER = "bentoml_cli.worker.http_api_server"
-SCRIPT_DEV_API_SERVER = "bentoml_cli.worker.http_dev_api_server"
+SCRIPT_DEV_API_SERVER = "bentoml_cli.worker.http_api_server"
 SCRIPT_GRPC_API_SERVER = "bentoml_cli.worker.grpc_api_server"
-SCRIPT_GRPC_DEV_API_SERVER = "bentoml_cli.worker.grpc_dev_api_server"
+SCRIPT_GRPC_DEV_API_SERVER = "bentoml_cli.worker.grpc_api_server"
```
aarnphm, line 39: remove this since it is not relevant anymore.

src/bentoml/serve.py
```diff
@@ -968,11 +750,44 @@ def serve_grpc_production(
...
+                    "use": "bentoml._internal.utils.circus.watchfilesplugin.ServiceReloaderPlugin",
+                    "working_dir": working_dir,
+                    "bentoml_home": bentoml_home,
+                },
+            ]
+
+        arbiter_kwargs["plugins"] = plugins
+        arbiter_kwargs["debug"] = True if sys.platform != "win32" else False
+        arbiter_kwargs["loggerconfig"] = SERVER_LOGGING_CONFIG
+        arbiter_kwargs["loglevel"] = "WARNING"
```
aarnphm, line 785: It seems like we are having this branch the same twice for both grpc and HTTP serve function, should we refactor out to a shared function?

aarnphm (commented):
final comments, and then it is gtg

src/bentoml/serve.py
```diff
@@ -171,123 +173,6 @@ def find_triton_binary():
     return binary
 
 
-@inject
-def serve_http_development(
```
aarnphm, line 175: Technically this is a breaking change if users are using this function. I think we should deprecate it until 1.1

I'm ok with making this breaking in 1.1

src/bentoml/serve.py
larme: done

src/bentoml/serve.py
larme: is this a public api we document?

src/bentoml_cli/serve.py
```diff
@@ -20,13 +20,21 @@ def add_serve_command(cli: click.Group) -> None:
...
         show_default=True,
     )
+    @click.option(
+        "--production",
+        type=click.BOOL,
+        help="Deprecated option, by default BentoServer will run in production mode",
+        is_flag=True,
+        default=True,
+        show_default=True,
+    )
```
larme, line 37: @aarnphm just added `--production` back for backward compatibility.

src/bentoml/serve.py
larme: after some code reading, I think `standalone_load=True` should be only used by runner/api server, so I remove this part

src/bentoml/server.py
```diff
@@ -185,7 +185,7 @@ def __init__(
         self,
         bento: str | Bento | Tag,
         reload: bool = False,
-        production: bool = False,
+        production: bool = True,
```
larme, line 188: @aarnphm I'm updating HTTPServer and GrpcServer's production default to True, to match our cli default

src/bentoml/serve.py
aarnphm: we don't, but i believe there are users that use this internal API

src/bentoml/server.py
aarnphm: sg

aarnphm (commented):
Overal LGTM, just one more thing is that maybe we can deprecate `serve_http_development` and just call `serve_http_production` with api_worker=1 underneath, so that it won't break user code.

aarnphm:
Also some of the tests are failing. I can take over this since I have a tests PR.

larme:
> Also some of the tests are failing. I can take over this since I have a tests PR.

the tests failed because we switch BentoServer's production default from False to True. The production server need quite longer to shutdown hence I bumped the timeout from 10 to 60.

larme:
@aarnphm please take a look again, thanks!

aarnphm (approved):
LGTM. Since gRPC is not yet fully released, I think it is good to break `serve_grpc_development`.

Issue #3748, opened 2023-04-09
feature: support JSON(typed_dict=TypedDict() ) Input, Output


KimSoungRyoul:
### to be 
~~~Python

class IAmSchema(BaseModel):
    a: list[float]
    b: list[float]
    c: list[float]
    d: Optional[float] # float | None



@svc.api(
    input=JSON( pydantic_model=IAmSchema ),
    output=JSON.from_sample({"qqqq":"wwww"}), 
)
async def predict_json(i_am_schema: ASchema) -> Dict[str,Any]:
    result = await runner.async_run(torch.tensor(list(i_am_schema.dict().values())))

    return {"qqqq": result}

~~~

### as is

~~~Python

class IAmSchema(t.TypedDict, total=False):
    a: list[float]
    b: list[float]
    c: list[float]
    d: NotRequried[float] # use NotRequired instead of Optional
    # Optional is rejected Spec in TypedDict (pep-0655)
    #  https://peps.python.org/pep-0655/#special-syntax-around-the-key-of-a-typeddict-item


@svc.api(
    input=JSON( 
        typeddict=IAmSchema, # <--- use TypedDict
    ),
    output=JSON.from_sample({"qqqq":"wwww"})), # <-- JSON IODescriptor cann't specify specifications in OAS3.
)
async def predict_json(i_am_schema: IAmSchema) -> Dict[str, Ant]:
     result = await runner.async_run(torch.tensor(list(i_am_schema.values()))

    return {"qqqq2222": result}


~~~

### Motivation

If we use Pydantic just to specify a type schema without validation, the logic of `xx.dict()` -> `Pydantic.parse_obj()` unnecessary layer which can affect performance. In this case, using TypedDict seems necessary.

~here is POC https://github.com/bentoml/BentoML/pull/3749~
https://github.com/bentoml/BentoML/pull/3809

I'd like to hear your thoughts on TypedDict

aarnphm:
If we want to support TypedDict then it should just be supported by JSON and using the typing system, to some extend (but parsing types are costly, esp for IO descriptor as it will affect users throughput). 

TypedDict is only available from 3.8+ so this means we will have to depends on typing_extensions.

From a quick glance from the implementation, it seems like you are just using Pydantic, so I don't see the point of having this as a part of the core library.

The core usage of Pydantic is for validation. So if we want to support TypedDict and attrs class to an extend, it should probably be under the JSON IO descriptor.

KimSoungRyoul:
[...]

PR #3749, opened 2023-04-09
[#3748] feat: add New IODescriptor "TypedJSON"

KimSoungRyoul:
https://github.com/bentoml/BentoML/issues/3748

### todo List after discussion approve?
  * [x]  under the `JSON` IO descriptor. not new IO descriptor.
  * [ ] use BentoML `openapi.Schema`  not depend on pydantic
  * [x] `import typing_extensions.TypedDict` if less than py3.8 
  * [ ] docs
  * [ ] examples

~~~Python
# example how to use 


class IrisFeatures(t.TypedDict, total=False): 
    sepal_len: float
    sepal_width: float
    petal_len: float
    petal_width: float

    # Optional Field
    request_id: t.Optional[int]


@svc.api(
    input=JSON(
        typed_dict=IrisFeatures,
        sample={"aa": 123, "bb": "sdfsdf323sf"},
    ),
    output=JSON.from_sample({"qqqq": "wwww"}),
)
async def predict_json(features: IrisFeatures) -> JSON:
    if features.get("request_id"):
       features["request_id"] = uuid()
    bentoml_logger.info(f"----------------{features}")

    return {"qqqq2222": "wwww2222"}

~~~


parano:
Hi @KimSoungRyoul, thanks for suggesting the feature and the PR!

Like @aarnphm suggested in the issue, I'd recommend adding `typed_dict` option to the `JSON` IODescriptor, as an alternative to Pydantic for schema validation. The example usage you write in the PR description looks good!

One question I have is - does `typed_dict` allow the user to define the behavior when an extra field is found in the JSON message? With pydantic, user can do something like this:
```python
       class IrisFeatures(BaseModel):
           sepal_len: float
           sepal_width: float
           petal_len: float
           petal_width: float

           # Optional field
           request_id: Optional[int]

           # Use custom Pydantic config for additional validation options
           class Config:
               extra = 'forbid'
```
Does the same `config` field work when we use `pydantic.create_model_from_typeddict` under the hood?

KimSoungRyoul:
@parano 

yes TypedDict can allow to define  extra field, when use `total=False` option



TypedDict is simply validated by mypy. (not in runtime like a pydantic)

~~~Python
# example service.py


import typing as t


class IrisFeatures(t.TypedDict, total=False): # default is total=True
    sepal_len: float
    sepal_width: float
    petal_len: float
    petal_width: float

    # Optional Field
    request_id: t.Optional[int]


@svc.api(
    input=JSON(
        typed_dict=IrisFeatures,
        sample={"aa": 123, "bb": "sdfsdf323sf"},
    ),
    output=JSON.from_sample({"qqqq": "wwww"}),
)
async def predict_json(features: IrisFeatures) -> JSON:
    # there is no error, when we run 'bentoml serve service:svc' but mypy command raise error
    print(features["not_defined_attr"])  # <-- [mypy](service.py:65: error: TypedDict "IrisFeatures" has no key "not_defined_attr"  [typeddict-item])
    
    
    # "mypy service.py" raise error, when IrisFeatures don't have option "total=True"
    typed_dict_features: IrisFeatures =  { # < -- error: Missing key "request_id" for TypedDict "IrisFeatures"  [typeddict-item]
        "sepal_len": 3.0,
        "sepal_width": 2.0,
        "petal_len": 1.0,
        "petal_width": 4.0,
        # "request_id": "~~~"
    }

    print(typed_dict_features)

    bentoml_logger.info(f"----------------{features}")

    return {"qqqq2222": "wwww2222"}

~~~


A typedDict only makes sense at the type checking.
At runtime, it is the same as a dict in python (and cannot even be checked :IMO)

src/bentoml/_internal/io_descriptors/json.py
```diff
@@ -180,6 +186,8 @@ def __init__(
         self,
         *,
         pydantic_model: t.Type[pydantic.BaseModel] | None = None,
+        typed_dict: t.Type[TypedDict] | None = None,
+        sample: JSONType | None = None,
```
aarnphm, line 190: lets not put sample here since there is no need to.
```suggestion
```

src/bentoml/_internal/io_descriptors/json.py
```diff
@@ -192,6 +200,24 @@ def __init__(
...
+            typed_dict and typed_dict is not dict
+        ):  # there is no way to check isinstance TypedDict in runtime
+            assert issubclass(
+                typed_dict, dict
+            ), "'typed_dict' must be a subclass of 'typing.TypedDict or dict'."
+
+        if typed_dict:
+            from pydantic import create_model_from_typeddict
+
+            self._typed_dict_model = create_model_from_typeddict(typed_dict)
```
aarnphm, line 218: Does this mean typed_dict depends on pydantic? it seems kinda strange that in order to use TypedDict you need to have pydantic installed.

src/bentoml/_internal/io_descriptors/json.py
```diff
@@ -1,10 +1,16 @@
...
+import sys
 import json
 import typing as t
 import logging
 import dataclasses
 
+if sys.version_info >= (3, 9, 2):
+    from typing import TypedDict
+else:
+    from typing_extensions import TypedDict
```
aarnphm, line 12: Since 3.7 will reach EOL in 2 months, it is probably best not to do this and just use `t.TypedDict` instead when BentoML drops 3.7 support

src/bentoml/_internal/io_descriptors/json.py
aarnphm: If this is the case then `typed_dict` and `pydantic_model` should be mutually exclusive then.

Might also need a import check or just lazy load it

aarnphm (changes requested):
For this PR I think we should just wait until 3.7 reaches EOL. cc @parano

parano:
> `CI / python3.7_http_e2e_tests (windows-latest) (pull_request)` is failed caused by 422 (too many request)
> 
> how can I rerun ?

Just triggered a rerun on the failed e2e test

src/bentoml/_internal/io_descriptors/json.py
sauyon: Any reason this is checking for 3.9.2?

sauyon: Sorry, read the above comment. Looks like this will have to stay in even if we drop 3.7 support? Definitely not a reason to delay this PR, I think.

aarnphm: Probably also need to maintain `typing_extensions` in pyproject.toml

KimSoungRyoul:
Hi I've been busy at work lately and haven't been paying attention.
I feel positive about reflecting this feature after dropping python 3.7. 
Also, separately, the current PR should be worked on to change to typedDict -> OAS3.0 Schema without the dependency on pydantic.
but I have a lot on my plate these days 😭 , so I'll close this PR temporarily and ask review again when the work is done.

PR #3767, opened 2023-04-18
chore(dispatcher): (re-)factor out training code

sauyon:
Fixes the refactor from earlier, which broke a name change, and also introduces a try-catch to the training code in order to surface issues in the dispatcher.


sauyon:
Tested again with a thousand quickstart requests.

aarnphm (approved):

aarnphm (approved):

Issue #3784, opened 2023-04-24
Update Open Telemetry integration packages


PeterJCLaw:
The currently pinned versions of Open Telemetry's packages are somewhat outdated, meaning that the dependencies they pull in are also somewhat outdated. In turn this can cause conflicts with other libraries.

I appreciate that Open Telemetry's packages aren't themselves great at their version spellings/matching, however it's not clear that BentoML pinning specific versions is the best solution.

A particular example of package version conflicts is that `opentelemetry-proto==1.14.0` requires `protobuf~=3.13`, yet the most recent Google Cloud packages require `protobuf` >= 4.

sauyon:
@aarnphm do you have context on why we locked the otel dep versions?

aarnphm:
Because different otlp packages have conflict dependencies. Also newer version of otlp also breaks our monitoring API, cc @bojiang

sauyon:
What conflicting dependencies?

aarnphm:
> What conflicting dependencies?

I believe it was protobuf, where opentelemetry-http-proto has a newer version of protobuf where other opentelemetry repo lock to an older protobuf version iirc

PeterJCLaw:
Looks like that's sorted now?

From a clean clone of this repo, with the following patch applied:
``` diff
diff --git a/pyproject.toml b/pyproject.toml
index ca2909c9..d7e33c0f 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,14 +39,14 @@ dependencies = [
     # OpenTelemetry is the server dependencies, rather than SDK
     # Since there are discrepancies among API and instrumentation packages,
     # we should always pin the set version of Opentelemetry suite
-    "opentelemetry-api==1.14.0",
-    "opentelemetry-sdk==1.14.0",
-    "opentelemetry-exporter-otlp-proto-http==1.14.0",
-    "opentelemetry-instrumentation==0.35b0",
-    "opentelemetry-instrumentation-aiohttp-client==0.35b0",
-    "opentelemetry-instrumentation-asgi==0.35b0",
-    "opentelemetry-semantic-conventions==0.35b0",
-    "opentelemetry-util-http==0.35b0",
+    "opentelemetry-api",
+    "opentelemetry-sdk",
+    "opentelemetry-exporter-otlp-proto-http",
+    "opentelemetry-instrumentation",
+    "opentelemetry-instrumentation-aiohttp-client",
+    "opentelemetry-instrumentation-asgi",
[...]

PR #3791, opened 2023-04-25
Unpin Open Telemetry versions now that its packages don't conflict

PeterJCLaw:
Fixes https://github.com/bentoml/BentoML/issues/3784.

Not completely sure if there's other aspects to the pinning which we've missed, though based on the discussion so far on the issue and based on this appearing to work locally this is hopefully all that's needed...


aarnphm:
cc @bojiang do we have a monitoring tests to make sure that opentelemetry will work?

aarnphm:
Hi @PeterJCLaw, We have merged a PR that addresses this issue when upgrading to a newer version of OTLP (#3833). I will close that issue as we will release a new version of BentoML early this week.

aarnphm:
One of the reasons why we lock the version of OTLP to a specific version is to avoid breakage that OTLP introduce every new version.

PR #3792, opened 2023-04-25
chore: 3.11 support

PeterJCLaw:
Hopefully this will just work given that there seems to be some minimal testing on 3.11 already (and given that https://github.com/bentoml/BentoML/pull/3774 has merged), though 3.11 isn't consistently present in the CI files.


aarnphm:
I will merge this in as gRPC is still currently in experimental state.

aarnphm (approved):

PR #3801, opened 2023-04-28
fix: correctly reraise `CancelledError`

sauyon:
This prevents the dispatcher from slowing down shutdown significantly.


larme (approved):
LGTM, there's still some delay when shutting down a production server but seems not be caused by `dispatcher.shutdown` anymore

aarnphm (approved):

PR #3809, opened 2023-04-30
feat(json): support TypedDict as schema [backlog for reference]

KimSoungRyoul:
## What does this PR address?
renewal (https://github.com/bentoml/BentoML/pull/3749)

supports TypedDict in addition to pydantic.BaseModel

note that ``typeddict`` and ``pydantic.BaseModel`` are mutually exclusive.

~~~Python
from typing import List, Union
# from typing_extensions import TypedDict # version = 3.8
from typing import TypedDict  # version > py3.8

import pandas as pd
from typing_extensions import NotRequired # version < py3.11
# from typing import NotRequired # version >= 3.11

import bentoml
from bentoml.io import JSON

iris_clf_runner = bentoml.sklearn.get("iris_clf_with_feature_names:latest").to_runner()

svc = bentoml.Service("i_am_iris_classifier_typeddict", runners=[iris_clf_runner])


class IrisFeatures(TypedDict, total=False):
    sepal_len: Union[float, int]
    sepal_width: float
    petal_len: float | int # support only >=3.10
    petal_width: float

    request_id: NotRequired[int]
    # use NotRequired instead of Optional
    # Optional is rejected Spec in TypedDict 
    #  (https://peps.python.org/pep-0655/#special-syntax-around-the-key-of-a-typeddict-item)

class Response(TypedDict):
    result: List[int]


@svc.api(input=JSON(typeddict=IrisFeatures), output=JSON(typeddict=Response))
async def classify(input_data: IrisFeatures) -> Response:
    if input_data.get("request_id") is not None:
        print("Received request ID: ", input_data["request_id"])

    input_data.pop("request_id")
    input_df = pd.DataFrame([input_data])
    result = await iris_clf_runner.predict.async_run(input_df)

    # it is same to 'return { "result": result.tolist() }'
    return Response(result=result.tolist())
~~~

Fixes #(issue)

## Before submitting:


src/bentoml/_internal/io_descriptors/json.py
```diff
@@ -6,6 +6,8 @@
 import dataclasses
 
 import attr
+from typing_extensions import TypedDict
+from typing_extensions import is_typeddict
```
aarnphm, line 10: Let's do a check here, 

```suggestion

if sys.version_info[:2] >= (3, 8):
	from typing import TypedDict, is_typeddict
else:
	from typing_extensions import TypedDict, is_typeddict
```

aarnphm: Please also add typing_extensions to `pyproject.toml`.

src/bentoml/_internal/io_descriptors/json.py
```diff
@@ -84,6 +87,7 @@ class JSON(
 
     .. code-block:: python
        :caption: `service.py`
+       # pydantic Example
```
aarnphm, line 90:
```suggestion
```

src/bentoml/_internal/io_descriptors/json.py
```diff
@@ -128,6 +132,51 @@ def classify(input_data: IrisFeatures) -> NDArray[Any]:
...
+            if input_data.get("request_id") is not None:
+                print("Received request ID: ", input_data["request_id"])
+
+            input_data.pop("request_id")
+            input_df = pd.DataFrame([input_data])
+            result = await iris_clf_runner.predict.async_run(input_df)
+
+            # it is same to 'return { "result": result.tolist() }'
+            return Response(result=result.tolist())
+
```
aarnphm, line 179: Let's not include this in the same code block.

You can either create a new tab, or create a new code block here.

src/bentoml/_internal/io_descriptors/json.py
```diff
@@ -192,6 +242,10 @@ def __init__(
                 pydantic_model, pydantic.BaseModel
             ), "'pydantic_model' must be a subclass of 'pydantic.BaseModel'."
 
+        if typeddict and is_typeddict(typeddict) is False:
```
aarnphm, line 245: nit: prefer explicit check

```suggestion
        if typeddict is not None and not is_typeddict(typeddict):
```

src/bentoml/_internal/io_descriptors/json.py
```diff
@@ -180,6 +229,7 @@ def __init__(
         self,
         *,
         pydantic_model: t.Type[pydantic.BaseModel] | None = None,
+        typeddict: t.Type[t.TypedDict] | None = None,
```
aarnphm, line 232: Another thing is is typeddict and pydantic_model should be mutually exclusive.

src/bentoml/_internal/io_descriptors/json.py
```diff
@@ -300,10 +354,10 @@ def input_type(self) -> UnionType:
         return JSONType
 
     def openapi_schema(self) -> Schema:
-        if not self._pydantic_model:
```
aarnphm, line 303: Since you are refactoring this part, maybe we can do the following
```python
if self._pydantic_model:
	return Schema(**schema.model_process_schema...)
elif self._typeddict:
	return Schema(**typed_dict_to_dict(...))
else:
	return Schema(type='object')
```

src/bentoml/_internal/io_descriptors/json.py
```diff
@@ -315,12 +369,16 @@ def openapi_schema(self) -> Schema:
...
+        elif self._pydantic_model:
+            from ..service.openapi.utils import pydantic_components_schema
 
-        from ..service.openapi.utils import pydantic_components_schema
+            return {"schemas": pydantic_components_schema(self._pydantic_model)}
+        elif self._typeddict:
+            from ..service.openapi.utils import typeddict_components_schema
 
-        return {"schemas": pydantic_components_schema(self._pydantic_model)}
+            return {"schemas": typeddict_components_schema(self._typeddict)}
```
aarnphm, line 381: same refactor comments like above

src/bentoml/_internal/service/openapi/utils.py
```diff
@@ -1,15 +1,47 @@
...
+    from typing import get_type_hints
+else:
+    from typing_extensions import get_args
+    from typing_extensions import Required
+    from typing_extensions import TypedDict
+    from typing_extensions import get_origin
+    from typing_extensions import NotRequired
+    from typing_extensions import get_type_hints
+    from typing_extensions import is_typeddict
+
```
aarnphm, line 37: Let's move this to `bentoml/_internal/types.py`. Also there is already some backport function that we include there that doesn't rely on typing_extensions.

@sauyon if we decided to introduce typing_extensions, then maybe we can also cleanup `bentoml/_internal/types.py`?

src/bentoml/_internal/service/openapi/utils.py
```diff
@@ -43,6 +75,33 @@ def pydantic_components_schema(pydantic_model: t.Type[pydantic.BaseModel]):
     return {k: Schema(**definitions[k]) for k in sorted(definitions)}
 
 
+def typeddict_components_schema(typeddict: t.Type[TypedDict]):
```
aarnphm, line 78:
```suggestion
def typeddict_components_schema(typeddict: type[TypedDict]):
```

aarnphm: I'm also ok with this section, but it seems to me that this section is something that pydantic already do.

Maybe a way to reduce coverage for this logic is to convert the user typeddict to pydantic model, and then use the logic we already have for converting pydantic model to the schema

aarnphm (changes requested):
For a feature PR like this, I would love to have included tests as well.

Great first step!

src/bentoml/_internal/service/openapi/utils.py
KimSoungRyoul: I think that it is good way to implement `typeddict_components_schema()`  
because i'm concerned about potential bugs during maintenance that will arise from implementing `typeddict_components_schema()` bentoml itself not depend on pydantic.

but it is awkward to bentoml user, when they use typeddict, should be need extra install `bentoml[pydantic]`
`nitpick:`  I think It would be nice to have pydantic as a default dependency module in bentoml instead of an extra dependency module.

IMO these is no good way or bad way, 
it just would be trade off, depend on pydantic or not 
for now, i will make an additional commit using pydantic. :)

src/bentoml/_internal/io_descriptors/json.py
KimSoungRyoul: On top of the `__init__` , I'll add the following logic
~~~python
def __init__(...):
    if pydantic_model is not None and typeddict is not None:
        raise BadInput("typeddict and pydantic_model are mutually exclusive. use only typeddict or pydantic_model")
~~~

src/bentoml/_internal/service/openapi/utils.py
aarnphm: Now I have more time to think about this, if we want to support TypedDict first class, then we shouldn't use pydantic

KimSoungRyoul:
@aarnphm please resolve completed comment

[f1ea712](https://github.com/bentoml/BentoML/pull/3809/commits/f1ea712f77f0206886faff9fea1c2dd913cf59a0)
* (pyproject.toml) add dependency "io-json = [... "type-extensions"]"
* (json.py) separate code block service.py(pydantic) and service.py(TypedDict)
* (json.py) add validate "typeddict and pydantic_model are mutually exclusive"
* (json.py) "if sys.version_info >= (3, 8):"
* (json.py) explicit check () "if typeddict is not None and not is_typeddict(typeddict)"



[615db0e](https://github.com/bentoml/BentoML/pull/3809/commits/615db0e6714f90727991036cb77486b0d7e9c85e)
* (json.py) check legacy_typeddict
  BadInput("You should use `typing_extensions.TypedDict` instead of `typing.TypedDict` with Python < 3.11.")

* (json.py) refactor openapi_schema() & openapi_components()
* (tests/unit/_internal/io/test_json.py) typeddict testcase
* (utils.py) refactor typed_dict_to_schema() -> typed_dict_to_dict()

pyproject.toml
```diff
@@ -131,7 +131,7 @@ io = [
     "bentoml[io-file]",
 ] # syntatic sugar for bentoml[io-json,io-image,io-pandas,io-file]
 io-file = ["filetype"] # Currently use for from_sample
-io-json = ["pydantic<2"] # currently we don't have support for pydantic 2.0
+io-json = ["pydantic<2", "type-extensions"] # currently we don't have support for pydantic 2.0
```
aarnphm, line 134:
```suggestion
io-json = ["pydantic<2", "typing-extensions"] # currently we don't have support for pydantic 2.0
```

src/bentoml/_internal/io_descriptors/json.py
```diff
@@ -1,15 +1,16 @@
...
 import logging
 import dataclasses
 
 import attr
-from typing_extensions import TypedDict
-from typing_extensions import is_typeddict
+from pydantic.typing import is_typeddict_special
 from starlette.requests import Request
 from starlette.responses import Response
+from pydantic.annotated_types import is_legacy_typeddict
```
aarnphm, line 13: We don't want to eager import like this? pydantic is still optional for `bentoml.io.JSON`.

src/bentoml/_internal/io_descriptors/json.py
```diff
@@ -19,11 +20,29 @@
...
 from ..service.openapi import REF_PREFIX
 from ..service.openapi import SUCCESS_DESCRIPTION
+from ..service.openapi.utils import typed_dict_to_dict
 from ..service.openapi.specification import Schema
 from ..service.openapi.specification import MediaType
 
 EXC_MSG = "'pydantic' must be installed to use 'pydantic_model'. Install with 'pip install bentoml[io-json]'."
 
+if sys.version_info >= (3, 11):
+    from typing import TypedDict
```
aarnphm, line 30: is there a change to typeddict from 3.11 onwards :(

src/bentoml/_internal/io_descriptors/json.py
```diff
@@ -19,11 +20,29 @@
...
+    from typing import is_typeddict
+
+    def is_legacy_typeddict(_: t.Any) -> t.Any:
+        return False
+
+else:
+    from typing_extensions import TypedDict
+    from typing_extensions import is_typeddict
+
+    def is_legacy_typeddict(typeddict_cls: t.Type["TypedDict"]) -> bool:  # type: ignore[valid-type]
```
aarnphm, line 40:
```suggestion
    def is_legacy_typeddict(typeddict_cls: type[t.TypedDict]) -> bool:  # type: ignore[valid-type]
```

src/bentoml/_internal/io_descriptors/json.py
```diff
@@ -128,8 +147,56 @@ def classify(input_data: IrisFeatures) -> NDArray[Any]:
...
+            if input_data.get("request_id") is not None:
+                print("Received request ID: ", input_data["request_id"])
+
+            input_data.pop("request_id")
+            input_df = pd.DataFrame([input_data])
+            result = await iris_clf_runner.predict.async_run(input_df)
+
+            # it is same to 'return { "result": result.tolist() }'
+            return Response(result=result.tolist())
+
```
aarnphm, line 196:
```suggestion

	:obj:`JSON` also supports ``typing.TypedDict`` in addition to ``pydantic.BaseModel`` as a validation schema. Note that for Python < 3.9, you can add the following code block to your code in order to use ``TypedDict``:
	
	.. code-block:: python
	
		import sys
		
		if sys.version_info[:2] >= 3.8
			from typing import TypedDict
		else:
			from typing_extensions import TypedDict
			
	.. note::
	
		Make sure to install ``bentoml`` with ``pip install 'bentoml[io-json]'`` to ensure the extra dependencies is installed.
		
	One can then add the following TypedDict to one's service definition:
	
    .. code-block:: python
        :caption: `service.py`

        from __future__ import annotations
		
		import typing as t
		import sys
        import pandas as pd

        import bentoml
        from bentoml.io import JSON

		if sys.version_info[:2] >= (3,11):
			from typing import NotRequired
		else:
			from typing_extensions import NotRequired
			

        class IrisFeatures(TypedDict, total=False):
            sepal_len: Union[float, int]
            sepal_width: float
            petal_len: Union[float, int]
            petal_width: float

            request_id: NotRequired[int]

        class Response(TypedDict):
            result: List[int]

        @svc.api(input=JSON(typeddict=IrisFeatures), output=JSON(typeddict=Response))
        async def classify(input_data: IrisFeatures) -> Response:
            if input_data.get("request_id") is not None:
                print("Received request ID: ", input_data["request_id"])

            input_data.pop("request_id")
            input_df = pd.DataFrame([input_data])
            result = await iris_clf_runner.predict.async_run(input_df)
            return Response(result=result.tolist())

```

src/bentoml/_internal/io_descriptors/json.py
```diff
@@ -180,9 +247,15 @@ def __init__(
         self,
         *,
         pydantic_model: t.Type[pydantic.BaseModel] | None = None,
+        typeddict: t.Type[t.TypedDict] | None = None,
         validate_json: bool | None = None,
         json_encoder: t.Type[json.JSONEncoder] = DefaultJsonEncoder,
     ):
+        if pydantic_model is not None and typeddict is not None:
+            raise BadInput(
+                "typeddict and pydantic_model are mutually exclusive. use only typeddict or pydantic_model"
```
aarnphm, line 256:
```suggestion
                "'typeddict' and 'pydantic_model' are mutually exclusive. Make sure to only use one of them."
```

src/bentoml/_internal/io_descriptors/json.py
```diff
@@ -192,6 +265,15 @@ def __init__(
                 pydantic_model, pydantic.BaseModel
             ), "'pydantic_model' must be a subclass of 'pydantic.BaseModel'."
 
+        if typeddict is not None and not is_typeddict(typeddict):
+            raise BadInput("'typeddict' must be inherited 'TypedDict'.")
+
+        if is_legacy_typeddict(typeddict):
+            raise BadInput(
+                "You should use `typing_extensions.TypedDict` instead of `typing.TypedDict` with Python < 3.11."
+            )
```
aarnphm, line 274: A quick refactor, imho:
```suggestion
        if typeddict is not None and not is_typeddict(typeddict):
            	raise BadInput(f"'typeddict={typeddict}' is not a subclass of '{TypedDict}'.")
```

I don't think asking user to use `typing_extensions` to create TypedDict is sensible here.

Since we are going to support 3.8, 3.9 and 3.10 for a while (at least until it reaches EOL), we then should also support "legacy" TypedDict

Also What is the difference between legacy and the "new" TypedDict may I ask?

src/bentoml/_internal/service/openapi/utils.py
aarnphm: We probably dont' want to depend on pydantic.

src/bentoml/_internal/service/openapi/utils.py
```diff
@@ -43,6 +74,33 @@ def pydantic_components_schema(pydantic_model: t.Type[pydantic.BaseModel]):
...
+    for typeddict in typeddict_set:
+        definitions[typeddict.__name__] = typed_dict_to_dict(typeddict)
+    return {k: Schema(**definitions[k]) for k in sorted(definitions)}
+
+
+def get_flat_typeddicts_from_typeddict(
+    typeddict: t.Type[TypedDict], typeddict_set: t.Set[t.Type[t.Any]]
+) -> t.Set[t.Type[t.Any]]:
+    typeddict_set.add(typeddict)
+    field_types: t.Dict[str, t.Any] = get_type_hints(typeddict, include_extras=True)
```
aarnphm, line 91: get_type_hints is very expensive, but I guess we don't have any other choice really.

src/bentoml/_internal/io_descriptors/json.py
KimSoungRyoul: In 3.11 and below, TypedDict seems to have few exception issues to consider. 

As you can see from pydantic's implementation of TypedDict, they have complicated exception handling for this issue. 
https://github.com/pydantic/pydantic/pull/2216/files#diff-bde859c4d65789b57bf14f734ec9ebabc83b30a82f6b0765a650ac18a25f0950R15
https://github.com/pydantic/pydantic/blob/v1.10.7/pydantic/annotated_types.py#L22

In implementing the TypedDict feature in this PR, I experienced a number of issues so did not use the internal features of TypedDict ( like a `__required_keys__`, `there is no way to reflect Required/NotRequired keys.`) which are the reason why Pydantic wanted to exception handle the availability of typing.TypedDict for each version(<=3.8 ,  >=3.11). 
So `sys.version_info >= (3, 11)` may not be necessary at this point. 
However, without this condition, someone trying to modify this logic would have to take into account the subtle differences in the internal implementation of TypedDict for each python version.

In fact, the exception handling below is also unnecessary at this point. 
~~~Python
if is_legacy_typeddict(typeddict):
            raise BadInput(
                "You should use `typing_extensions.TypedDict` instead of `typing.TypedDict` with Python < 3.11."
            )
~~~

it is fine to use.
~~~Python
from typing import TypedDcit
from typing_extension import NotRequried 

from typing import List, Union, TypedDict


from typing_extensions import NotRequired

class IrisFeatures(TypedDict, total=False): # it's working 
    sepal_len: Union[float, int]
    sepal_width: float
    petal_len: Union[float , int]
    petal_width: float

    request_id: NotRequired[int]


JSON(typeddict=IrisFeatures)
~~~
but i'm not sure some side effect

If we're just using TypedDict for our Open API Schema implementation like the current implementation,
"sys.version_info >= (3, 11)" is unnecessary. 

Is it appropriate to remove  that `sys.version_info >= (3, 11)` ?

src/bentoml/_internal/service/openapi/utils.py
KimSoungRyoul: IMO `openapi_components()` is only used for `/docs.json` 
it means that there is no issue to inference performance? so I use it 😭

src/bentoml/_internal/io_descriptors/json.py
aarnphm: if i understand correct, the `typing_extension` should only act as backward compat for newer feature in older python version. So I think if we want to support TypedDict, then we will have to support similar to what pydantic did.


```python
from typing import List, Union, TypedDict


from typing_extensions import NotRequired
```

> However, without this condition, someone trying to modify this logic would have to take into account the subtle differences in the internal implementation of TypedDict for each python version.

Not sure if I understand this, but I'm surprised that they introduced breaking change among TypedDict iteration of Python versions.

I don't think TypedDict can be imported here in 3.7
Also in the code you can add in line documentation that reference back to pydantic's source code (credit is due where it should be)

P/s: I just read a bit more about TypedDict, it does seems they are slightly different among Python version 😮‍💨 

p/s: +1 that we probably need to do something similar to Pydantic

src/bentoml/_internal/service/openapi/utils.py
aarnphm: The swagger UI depends on this right? so `openapi_components` path would have some penalty for actual loading the UI.

We probably just need to add documentation somewhere about this.

src/bentoml/_internal/io_descriptors/json.py
```diff
@@ -128,8 +147,57 @@ def classify(input_data: IrisFeatures) -> NDArray[Any]:
            input_df = pd.DataFrame([input_data.dict(exclude={"request_id"})])
            return iris_clf_runner.run(input_df)
 
+    :obj:`JSON` also supports ``typing.TypedDict`` in addition to ``pydantic.BaseModel`` as a validation schema. Note that for Python < 3.9, you can add the following code block to your code in order to use ``TypedDict``:
+
+    .. code-block:: python
+
+        import sys
+
+        if sys.version_info[:2] >= 3.8
```
aarnphm, line 156:
```suggestion
        if sys.version_info[:2] >= (3, 8)
```

Sorry my bad on this

src/bentoml/_internal/io_descriptors/json.py
KimSoungRyoul: > So I think if we want to support TypedDict, then we will have to support similar to what pydantic did.

If we want to support the same as pydantic, we may not need to have complex constraints like pydantic at this point

because pydantic is not only used OAS3 Schema but also Validation with type hint so they need to deeply each TypedDict feature. 

for example pydantic should raise Validation Error  in below example
https://docs.pydantic.dev/latest/usage/types/#typeddict
~~~Python
class AA(TypedDict)
      a: int
      b: str

class BB(BaseModel):
     aa: AA


BB(aa ={"a": "1", b: 1000}) # raise validation error
~~~

but this pullrequest is use TypedDict for type hint



56e3763 (here is (`>=3.11` condition ) is removed version)  
it is safe , if we only use to  OpenAPI Schema generating

src/bentoml/_internal/io_descriptors/json.py
KimSoungRyoul: can you resolve this comment? it's already reflected

src/bentoml/_internal/io_descriptors/json.py
KimSoungRyoul: [feat: remove legacy typeddict check (56e3763)](https://github.com/bentoml/BentoML/pull/3809/commits/56e3763f193481793b4a40b2a02293a8a8fcf04c)

KimSoungRyoul:
is there no any more feedback to this PullRequest? 😭

aarnphm:
https://github.com/python-attrs/cattrs/pull/364/files

src/bentoml/_internal/service/openapi/utils.py
```diff
@@ -43,6 +52,33 @@ def pydantic_components_schema(pydantic_model: t.Type[pydantic.BaseModel]):
     return {k: Schema(**definitions[k]) for k in sorted(definitions)}
 
 
+def typeddict_components_schema(typeddict: t.Type[TypedDict]):
```
aarnphm, line 55:
```suggestion
def typeddict_components_schema(typeddict: type[TypedDict]):
```

as much as possible all new code should follow PEP604

src/bentoml/_internal/service/openapi/utils.py
```diff
@@ -43,6 +52,33 @@ def pydantic_components_schema(pydantic_model: t.Type[pydantic.BaseModel]):
...
+        typeddict=typeddict, typeddict_set=set()
+    )
+    definitions: dict[str, Schema] = {}
+    for typeddict in typeddict_set:
+        definitions[typeddict.__name__] = typed_dict_to_dict(typeddict)
+    return {k: Schema(**definitions[k]) for k in sorted(definitions)}
+
+
+def get_flat_typeddicts_from_typeddict(
+    typeddict: t.Type[TypedDict], typeddict_set: t.Set[t.Type[t.Any]]
```
aarnphm, line 66:
```suggestion
    typeddict: type[TypedDict], typeddict_set: set[type[TypedDict]]
```

src/bentoml/_internal/service/openapi/utils.py
```diff
@@ -43,6 +52,33 @@ def pydantic_components_schema(pydantic_model: t.Type[pydantic.BaseModel]):
...
+) -> t.Set[t.Type[t.Any]]:
```
aarnphm, line 67:
```suggestion
) -> set[type[TypedDict]]:
```

src/bentoml/_internal/service/openapi/utils.py
```diff
@@ -43,6 +52,33 @@ def pydantic_components_schema(pydantic_model: t.Type[pydantic.BaseModel]):
     return {k: Schema(**definitions[k]) for k in sorted(definitions)}
 
 
+def typeddict_components_schema(typeddict: t.Type[TypedDict]):
+    typeddict_set = get_flat_typeddicts_from_typeddict(
+        typeddict=typeddict, typeddict_set=set()
+    )
+    definitions: dict[str, Schema] = {}
+    for typeddict in typeddict_set:
+        definitions[typeddict.__name__] = typed_dict_to_dict(typeddict)
```
aarnphm, line 61:
```suggestion
        definitions[typeddict.__name__] = typed_dict_to_dict(typeddict)
```

This actually doesn't handle the case where there are two typeddict, `user.Model` and `base.Model`. One of them will just overwrite the other.

aarnphm: You probably need to handle conflicting nametype with something like `f'{model.__module__}__{model.__qualname__}'.replace('.', '__')`, and provide a mapping to it.

src/bentoml/_internal/service/openapi/utils.py
frostming: I think it's enough to move them under `if TYPE_CHECKING` block, we can assume the version is the latest when we work with types.

aarnphm:
#3906

src/bentoml/_internal/service/openapi/utils.py
aarnphm: @frostming I'm thinking about `bentoml._internal.types` as a compat typing layer to `python` typing

KimSoungRyoul:
Hi @aarnphm 
I think we need to clean up this task.

I saw new issues related this PullRequest
* [feature: bentoml.io.AttrsClass](https://github.com/bentoml/BentoML/issues/3930)
   *  [PR: feat(descriptor): support attrs model for JSON [experimental] ](https://github.com/bentoml/BentoML/pull/3906)
*  [feature: bentoml.io.PydanticModel](https://github.com/bentoml/BentoML/issues/3929)

these issues might be refactor `io.JSON(pydantic_model=AModel)` to
* `io.JSON(typeddict=ATypedDict /* if it is merged */)`
* `io.Attrs()`
* `io.PydanticModel()`

But I guess that these issues and this  PullRequest is not a high priority within bentoml team.
I think that that why this PR and other issues has been waiting few months. 

if it's true, then it's right to keep this PullRequest now,  but if not, then I think it's right to close this PR.

aarnphm:
Hey there, thanks for staying persistent with us. I believe that we have discussed this internally, and decided that it makes more sense to separate it out, making TypedDict a superset of `bentoml.io.JSON`

Since this is very low in our roadmap right now, we don't have a specific timeline on when we will add attrs and pydantic extension of JSON (and TypedDict as well).

aarnphm:
I will keep this one open for reference. Thanks for the great work @KimSoungRyoul

aarnphm:
For future reader, we decided to have the enhancement for JSON io descriptor separated from `bentoml.io.JSON`

In the future, there will be the following 4 IO descriptors as a superset of `bentoml.io.JSON`:
- `bentoml.io.TypedDict`
- `bentoml.io.AttrsClass`
- `bentoml.io.PydanticModel`
- `bentoml.io.Dataclass`

sauyon:
Closing this one as per the above and in favor of the new IO descriptor rework; @frostming this is something we should implement in the new API, though.

Issue #3580, opened 2023-02-17
rfc: build improvement


aarnphm:
## Problem statement

From community reports and internal discussion, `bentoml build` currently has the following caveats:

- imports users' `service.py` as a module, which means all of the code in `service.py` will be invoked during the build, including every dependency imported in `service.py`.
This is not ideal as it requires users to have all the dependencies installed to build, which might not always be available in the CI/CD environment.

    Current workaround we have seen from the community is that users will have to setup their environment beforehand, to run it on CI. This means users will end up with installing dependencies twice, one for build and one during containerise.

- `bentoml build` creates a bento that includes a Dockerfile, which will be used by `containerize` to package a BentoContainer. Often times for CI, the desired behaviour is that build should be able to resolve
to the container directly.

## Proposed solutions

1. importing `service.py` during build

a. Using `--env`

- #3396 introduces `--env` argument to serve, which allow serving within conda environment. We should also be able to extend this to container, virtualenv, mamba, and so on.
`--env` can also be used during `build`, which will build the given bento with the specified environment:

```bash
bentoml build --env container
```

The behaviour is as follows:
- creates a container that contains all of the necessary dependencies defined under `bentofile.yaml` (PyPI, Conda, system packages, setup script)
- Attach the container to build the directory and build the Bento inside the container environment
    - For BuildKit supported container daemon, we can use ``--output=type=local,dest=/path/to/bentoml_home/bentos/bento_name/version`` to copy the built Bento to local machine,
    - For daemons that doesn't support BuildKit, use `cp` instead.
- The container that is used to build the Bento can also be used for `containerize` and `serve` directly.

> NOTE: #2495 mentions about this capability. All of the APIs are there to be used, so community contributions are welcome.

b. Not using `--env`

If users wish not to use `--env`, then to solve this issue, we will need to extract the Service object from `service.py` without actually importing it to the file.
- Propose that we write a custom Python parser :smile:
- Or running build in a subprocess

2. `bentoml build` directly to a container

For CI, what we can support is that `build` can also do `containerize` directly via flag `--format=container`:

```bash
bentoml build --format=container
```
by default, build will still create a Bento. (``--format=bento``)

Would love to hear more feedback and comments on this.

## Additional context

#3577 suggests that we should refactor the containerization steps so that it will cache the environment setup, and move the model copying to later steps. This fails into `--env container` proposal, where it set up all of the dependencies inside the container once.

Quasarman:
[...]

PR #3793, opened 2023-04-26
tests: migrate to use new client and server API

aarnphm:
## What does this PR address?

Migrate current `bentoml.testing` to use the recent server and client API.

Remove broken pandas series tests for now, will investigate on a
different tickets.

depends on #3843, #3844, #3845, #3846

- In addition to the test improvement, this PR also introduces a `async_request` and `request` to `HTTPClient`. This enables users to send request other that `POST` to the server.

This is useful when BentoService is mounted with a fastapi or other ASGI app.

`client.async_request` is a thin wrapper for `aiohttp.ClientSession.request`.

- all attributes that ``aio.ClientSession.post`` accepts that has prefix of `_http_*` can also be passed from `client.<endpoint>`

```python

client.predict(np.array([1,2]), _http_timeout=90, _http_headers={"Content-Type": "application/json"})

```

Signed-off-by: Aaron <29749331+aarnphm@users.noreply.github.com>

## Before submitting:

- [ ] Does the Pull Request follow [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/#summary) naming? Here are [GitHub's
      guide](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request) on how to create a pull request.
- [ ] Does the code follow BentoML's code style, both `make format` and `make lint` script have passed ([instructions](https://github.com/bentoml/BentoML/blob/main/DEVELOPMENT.md#style-check-auto-formatting-type-checking))?
- [ ] Did you read through [contribution guidelines](https://github.com/bentoml/BentoML/blob/main/CONTRIBUTING.md#ways-to-contribute) and follow [development guidelines](https://github.com/bentoml/BentoML/blob/main/DEVELOPMENT.md#start-developing)?
- [ ] Did your changes require updates to the documentation? Have you updated
      those accordingly? Here are [documentation guidelines](https://github.com/bentoml/BentoML/tree/main/docs) and [tips on writting docs](https://github.com/bentoml/BentoML/tree/main/docs#writing-documentation).
- [ ] Did you write tests to cover your changes?

PR #3814, opened 2023-05-02
feat: subprocess build

aarnphm:
## What does this PR address?

Initial implementation as a part of #3580

## Before submitting:

### DIFF
--- a/src/bentoml/_internal/bento/build_config.py
+++ b/src/bentoml/_internal/bento/build_config.py
@@ -835,9 +835,11 @@ class BentoBuildConfig:
                 raise InvalidArgument(str(e)) from e
 
     def to_yaml(self, stream: t.TextIO) -> None:
-        # TODO: Save BentoBuildOptions to a yaml file
-        # This is reserved for building interactive build file creation CLI
-        raise NotImplementedError
+        try:
+            yaml.dump(bentoml_cattr.unstructure(self), stream)
+        except yaml.YAMLError as e:
+            logger.error(e)
+            raise
 
 
 @attr.define(frozen=True)
--- a/src/bentoml/_internal/bento/build_config.py
+++ b/src/bentoml/_internal/bento/build_config.py
@@ -838,6 +838,7 @@ class BentoBuildConfig:
         try:
             yaml.dump(bentoml_cattr.unstructure(self), stream)
         except yaml.YAMLError as e:
+            logger.error("Error while deserializing BentoBuildConfig to yaml:")
             logger.error(e)
             raise
 
--- a/src/bentoml/bentos.py
+++ b/src/bentoml/bentos.py
@@ -4,8 +4,12 @@ User facing python APIs for managing local bentos and build new bentos.
 
 from __future__ import annotations
 
+import os
+import re
 import typing as t
 import logging
+import tempfile
+import subprocess
 
 from simple_di import inject
 from simple_di import Provide
@@ -273,7 +277,7 @@ def build(
     version: str | None = None,
     build_ctx: str | None = None,
     _bento_store: BentoStore = Provide[BentoMLContainer.bento_store],
-) -> "Bento":
+) -> Bento:
     """
     User-facing API for building a Bento. The available build options are identical to the keys of a
     valid 'bentofile.yaml' file.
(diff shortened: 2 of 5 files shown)


src/bentoml/bentos.py
```diff
@@ -349,21 +353,40 @@ def build(
...
+        build_args.extend(["--bentofile", bentofile_path])
+        try:
+            output = subprocess.check_output(build_args)
+        except subprocess.CalledProcessError as e:
+            logger.error("Failed to build BentoService bundle: %s", e)
+            raise
+
+    rgx = re.compile(r"^Successfully built.*")
+    tag_parse = [s for s in output.decode("utf-8").split("\n") if rgx.match(s)][0]
+    return get(tag_parse.split(" ")[-1])
```
ssheng, line 379: Maybe add a `-s` / `--short` option to `bentoml build` so only the tag name is returned so we won't have to parse with regex here.

ssheng (dismissed):

aarnphm:
I will cherry-pick this pr and fix the test in #3793 as the tests fail is related to #3793

src/bentoml/_internal/bento/build_config.py
```diff
@@ -835,9 +835,11 @@ def from_yaml(cls, stream: t.TextIO) -> BentoBuildConfig:
                 raise InvalidArgument(str(e)) from e
 
     def to_yaml(self, stream: t.TextIO) -> None:
-        # TODO: Save BentoBuildOptions to a yaml file
-        # This is reserved for building interactive build file creation CLI
-        raise NotImplementedError
+        try:
+            yaml.dump(bentoml_cattr.unstructure(self), stream)
+        except yaml.YAMLError as e:
+            logger.error(e)
```
sauyon, line 841: Maybe add a quick descriptor of where in the code this error has happened?

src/bentoml_cli/bentos.py
```diff
@@ -278,22 +285,72 @@ def push(bento_tag: str, force: bool, threads: int, context: str) -> None:  # ty
...
+            fg="blue",
         )
         click.secho(
             f"\n * Push to BentoCloud with `bentoml push`:\n    $ bentoml push {bento.tag}",
-            fg="yellow",
+            fg="blue",
         )
+
+        # We need to pop the build_ctx from sys.path to avoid conflicts.
+        if sys.path[0] == build_ctx:
```
sauyon, line 355: Hm, why? Isn't the point of doing this in a subprocess that we shouldn't care?

Also, this should probably be done in the build function itself and not here.

src/bentoml_cli/bentos.py
```diff
@@ -278,22 +285,72 @@ def push(bento_tag: str, force: bool, threads: int, context: str) -> None:  # ty
...
+
+        if output == "machine":
+            click.echo(bento.tag)
+            return
+
         click.echo(BENTOML_FIGLET)
-        click.secho(f"Successfully built {bento}.", fg="green")
+
+        # NOTE: Don't change the following message, since it is used by 'bentoml build' to parse for the bento.tag
+        click.secho(f"Successfully built {bento.tag}", fg="green")
```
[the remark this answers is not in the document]
aarnphm: You are right, since now we have a 'tag' output this is not needed to change anymore.

src/bentoml_cli/bentos.py
aarnphm: This is now the build function right? since from our python API we will just call this.

aarnphm: Also do you remember why we do L317?

src/bentoml/bentos.py
```diff
@@ -381,14 +402,21 @@ def build_bentofile(
...
-        version=version,
-        build_ctx=build_ctx,
-    ).save(_bento_store)
+    try:
+        output = subprocess.check_output(build_args)
+    except subprocess.CalledProcessError as e:
+        logger.error("Failed to build BentoService bundle: %s", e)
+        raise
+
+    return get(output.decode("utf-8").strip().split("\n")[-1])
```
sauyon, line 419: Is this split necessary?

src/bentoml_cli/bentos.py
sauyon: Sorry, to be clear, I think this cleanup should be happening where we modify the syspath and not as an extra step in the CLI.

aarnphm: This cleanup should be here right, since we modify the sys.path from the beginning, and then after the _save we should cleanup the sys.path

sauyon: Oh, I somehow forgot that that was being done. That seems odd, why's it here again?

ssheng (dismissed):
LGTM. Please coordinate with @sauyon on his comments.

src/bentoml_cli/bentos.py
aarnphm: If users do the following in service

```python
from src.path import module

```

We need to put the build context into sys.path[0] to be able to find src.path

ssheng (approved):

PR #3826, opened 2023-05-04
fix: Eliminate the exception during shutdown

### LINKED ISSUE #3813: bugs: Exception during shutdown
Issuing SIGTERM (Ctrl+C) to a runner server causes the following exception to be raised. 

```
023-05-01T23:27:09-0700 [ERROR] [cli] Exception in callback <bound method Arbiter.manage_watchers of <circus.arbiter.Arbiter object at 0x10719e2c0>>
  File "/opt/miniconda3/envs/bentoml/lib/python3.10/site-packages/tornado/ioloop.py", line 921, in _run
    val = self.callback()
  File "/opt/miniconda3/envs/bentoml/lib/python3.10/site-packages/circus/util.py", line 1038, in wrapper
    raise ConflictError("arbiter is already running %s command"
circus.exc.ConflictError: arbiter is already running arbiter_stop command
```

frostming:

## What does this PR address?

Fixes #3813

A bug that causes the runner to keep running after it is terminated. This is because the CancelledError is swallowed silently, making the tasks never be cancelled by the runner.

BTW, since `CancelledError` is subclassing `BaseException` it won't be caught by the `except Exception` block. We can remove it.
## Before submitting:

### DIFF
--- a/src/bentoml/_internal/marshal/dispatcher.py
+++ b/src/bentoml/_internal/marshal/dispatcher.py
@@ -234,6 +234,8 @@ class CorkDispatcher:
                     # fake wait as 0 for training requests
                     info.enqueue_time = now
                 self._loop.create_task(self.outbound_call(inputs_info))
+        except asyncio.CancelledError:
+            raise
         except Exception as e:  # pylint: disable=broad-except
             logger.error(traceback.format_exc(), exc_info=e)
 
@@ -267,6 +269,8 @@ class CorkDispatcher:
                     "BentoML has detected that a service has a max latency that is likely too low for serving. If many 503 errors are encountered, try raising the 'runner.max_latency' in your BentoML configuration YAML file."
                 )
             logger.debug("Dispatcher optimizer training complete.")
+        except asyncio.CancelledError:
+            raise
         except Exception as e:  # pylint: disable=broad-except
             logger.error(traceback.format_exc(), exc_info=e)
 
--- a/src/bentoml/_internal/marshal/dispatcher.py
+++ b/src/bentoml/_internal/marshal/dispatcher.py
@@ -234,8 +234,6 @@ class CorkDispatcher:
                     # fake wait as 0 for training requests
                     info.enqueue_time = now
                 self._loop.create_task(self.outbound_call(inputs_info))
-        except asyncio.CancelledError:
-            raise
         except Exception as e:  # pylint: disable=broad-except
             logger.error(traceback.format_exc(), exc_info=e)
 
@@ -269,8 +267,6 @@ class CorkDispatcher:
                     "BentoML has detected that a service has a max latency that is likely too low for serving. If many 503 errors are encountered, try raising the 'runner.max_latency' in your BentoML configuration YAML file."
                 )
             logger.debug("Dispatcher optimizer training complete.")
-        except asyncio.CancelledError:
-            raise
         except Exception as e:  # pylint: disable=broad-except
             logger.error(traceback.format_exc(), exc_info=e)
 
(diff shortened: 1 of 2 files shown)


aarnphm (approved):

aarnphm (approved):

src/bentoml/_internal/marshal/dispatcher.py
```diff
@@ -234,8 +234,6 @@ async def train_optimizer(
                     # fake wait as 0 for training requests
                     info.enqueue_time = now
                 self._loop.create_task(self.outbound_call(inputs_info))
-        except asyncio.CancelledError:
-            return
```
sauyon, line 238: Huh, did this not get replaced by `raise` earlier? If you don't re-raise it manually it'll get caught by the catch-all below, right?

sauyon:
Just tried this and if you kill the server in training you get

```

File: src/bentoml/_internal/marshal/dispatcher.py [hunk 2]
```diff
@@ -269,8 +267,6 @@ async def controller(self):
                     "BentoML has detected that a service has a max latency that is likely too low for serving. If many 503 errors are encountered, try raising the 'runner.max_latency' in your BentoML configuration YAML file."
                 )
             logger.debug("Dispatcher optimizer training complete.")
-        except asyncio.CancelledError:
-            raise
```
sauyon, line 273: I feel like this should just be a return instead of a raise?

frostming: If it's a return, only this training process will return and the subsequent ones will continue to run:

https://github.com/bentoml/BentoML/blob/6f7a32df9ca24d3299a3a6c13ca3cd381d5a3680/src/bentoml/_internal/marshal/dispatcher.py#L247-L263

frostming:
> Just tried this and if you kill the server in training you get

@sauyon Thanks for that info. I tested on Python 3.7(which I didn't do previously), it looks good after I reraise the exception.

sauyon:
Looks better to me! Should we be worried about the `return` on [L316](https://github.com/bentoml/BentoML/blob/9f89fe5c5d8d2f8a6aae8d2d057116b6d13c8a63/src/bentoml/_internal/marshal/dispatcher.py#L316) and on [L180](https://github.com/bentoml/BentoML/blob/9f89fe5c5d8d2f8a6aae8d2d057116b6d13c8a63/src/bentoml/_internal/marshal/dispatcher.py#LL180C9-L180C9)?

frostming:
> Looks better to me! Should we be worried about the `return` on [L316](https://github.com/bentoml/BentoML/blob/9f89fe5c5d8d2f8a6aae8d2d057116b6d13c8a63/src/bentoml/_internal/marshal/dispatcher.py#L316) and on [L180](https://github.com/bentoml/BentoML/blob/9f89fe5c5d8d2f8a6aae8d2d057116b6d13c8a63/src/bentoml/_internal/marshal/dispatcher.py#LL180C9-L180C9)?

L180 is intentional to return a fallback value. L316 doesn't make a difference since it is a top async task on its own, but IMO it has a slight mismatch in semantic: a cancellation should not be a normal return.

sauyon (approved):
Ok, this LGTM! Not sure why I didn't reraise in the training code and added the raise elsewhere...

frostming:
> Ok, this LGTM! Not sure why I didn't reraise in the training code and added the raise elsewhere...

NVM, it may come from a refactor when the logic is extracted into a function

sauyon:
@aarnphm are unit tests broken?

aarnphm:
Oh we need to install Pillow in our tests. Let me make a quick PR for that

aarnphm (approved):

PR #3829, opened 2023-05-05
feat: allow starting server with bentoml.Service instance

parano:
## What does this PR address?

This PR made the following use case possible, this is especially helpful in a project development environment where no Bento was built yet.

```python
import bentoml

# import the Service defined in `/clip_api_service/service.py` file
from clip_api_service.service import svc 

# start a server:
server = bentoml.HTTPServer(svc)
server.start(blocking=False)
client = server.get_client()
client.rank(..)
```

Similarly, users can also put a launch script within a service definition file and run it as a regular Python program:

```python
# service.py
svc = bentoml.Service(...)

if __name__ == "__main__":
    server = bentoml.HTTPServer(svc)
    server.start(blocking=True)
```

```bash
python service.py
```

See discussion here: https://github.com/bentoml/BentoML/discussions/3560#discussioncomment-5711687

Fixes #(issue)

## Before submitting:

### DIFF
--- a/src/bentoml/_internal/service/service.py
+++ b/src/bentoml/_internal/service/service.py
@@ -247,13 +247,20 @@ class Service:
             if import_module == "__main__":
                 if hasattr(sys.modules["__main__"], "__file__"):
                     import_module = sys.modules["__main__"].__file__
+                else:
+                    raise BentoMLException(
+                        "Failed to get service import origin, bentoml.Service object defined interactively in console or notebook is not supported"
+                    )
 
             for name, value in vars(sys.modules[self._caller_module]).items():
                 if value is self:
                     object.__setattr__(self, "_import_str", f"{import_module}:{name}")
                     break
+
             if not self._import_str:
-                raise BentoMLException("Failed to get service import origin")
+                raise BentoMLException(
+                    "Failed to get service import origin, bentoml.Service object must be assigned to a variable at module level"
+                )
 
         return self._import_str, self._working_dir
 
@@ -286,14 +293,14 @@ class Service:
         if self.bento:
             return f'bentoml.Service(tag="{self.tag}", ' f'path="{self.bento.path}")'
 
-        if self._caller_module != "__main__":
+        try:
             import_str, working_dir = self.get_service_import_origin()
             return (
                 f'bentoml.Service(name="{self.name}", '
                 f'import_str="{import_str}", '
                 f'working_dir="{working_dir}")'
             )
-        else:
+        except BentoMLException:
             return (
                 f'bentoml.Service(name="{self.name}", '
                 f'runners=[{",".join([r.name for r in self.runners])}])'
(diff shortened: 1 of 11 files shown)


src/bentoml/_internal/service/service.py
```diff
@@ -226,6 +230,40 @@ def __init__(
...
+        """
+        Returns the module name and working directory of the service
+        """
+        if not self._import_str:
+            import_module = self._caller_module
+            if import_module == "__main__":
+                if hasattr(sys.modules["__main__"], "__file__"):
+                    import_module = sys.modules["__main__"].__file__
+
+            for name, value in vars(sys.modules[self._caller_module]).items():
```
parano, line 251: This is a bit hacky, ideally, we should get the pickle serialization to work properly for bentoml.Service objects.

aarnphm: This seems fine actually. I don't think it is that hacky.

parano: Added a few error messages to show the user the limit of this approach. Basically we require the user to put the Service object at module's top level for it to work. E.g. using a function that returns a Service instance may not work properly here.

src/bentoml/_internal/service/service.py
```diff
@@ -226,6 +230,47 @@ def __init__(
...
+                    )
+
+            for name, value in vars(sys.modules[self._caller_module]).items():
+                if value is self:
+                    object.__setattr__(self, "_import_str", f"{import_module}:{name}")
+                    break
+
+            if not self._import_str:
+                raise BentoMLException(
+                    "Failed to get service import origin, bentoml.Service object must be assigned to a variable at module level"
```
aarnphm, line 262: nit:

```suggestion
                    "Failed to get service import origin, bentoml.Service object must be loaded eagerly (it should be available in module scope)."
```

src/bentoml/_internal/service/service.py
```diff
@@ -226,6 +230,47 @@ def __init__(
...
+                )
+
+        return self._import_str, self._working_dir
+
+    def is_service_importable(self) -> bool:
```
aarnphm, line 267: Let's make this a property?

src/bentoml/_internal/service/service.py
parano: `must be loaded eagerly (it should be available in module scope)."` is not technically correct here. To simplify put, it should be defined and assigned at the module level (not in/under module scope, that's a bit different..)

src/bentoml/server.py
```diff
@@ -40,7 +42,7 @@ class Server(ABC):
 
     def __init__(
         self,
-        bento: str | Bento | Tag,
+        bento: str | Bento | Tag | Service,
```
ssheng, line 45: Will we need bentofile.yaml in some cases now or in the future, because Service + bentofile.yaml = Bento.

parano: currently, we only use `bentofile.yaml` for getting the service instance when running `bentoml serve` in a folder containing a `bentofile.yaml`. The problem with adding `bentofile.yaml` path here in this API is that we can not guarantee the working directory is the same as the current Python process. I feel for Python API, the most intuitive usage is still importing the service object directly, instead of using a string like "service.py:svc" or passing in the "bentofile.yaml"

ssheng: That makes sense. Adding this API will also limit us from using data in bentofile.yaml during `serve`. We don't use it today but that's an assumption made with `bentoml serve`.

ssheng (approved):

aarnphm (approved):

PR #3833, opened 2023-05-05
improvement: monitoring and tests

aarnphm:
- makes monitoring deps optional (depends on #3840)
- update to 1.17 for now
- Support registering Monitor class via MONITOR_REGISTRY
- fix imports for OTLP monitor 1.17
- added e2e tests from examples/monitoring/task_classification


ssheng (dismissed):

sauyon (approved):

aarnphm:
@bojiang PTAL.

pyproject.toml
```diff
@@ -152,10 +152,10 @@ tracing = [
...
     "bentoml[tracing-zipkin]",
 ]
-tracing-jaeger = ["opentelemetry-exporter-jaeger==1.14.0"]
-tracing-zipkin = ["opentelemetry-exporter-zipkin==1.14.0"]
-tracing-otlp = ["opentelemetry-exporter-otlp==1.14.0"]
-monitor-otlp = ["opentelemetry-exporter-otlp-proto-http==1.14.0"]
+tracing-jaeger = ["opentelemetry-exporter-jaeger==1.17.0"]
+tracing-zipkin = ["opentelemetry-exporter-zipkin==1.17.0"]
+tracing-otlp = ["opentelemetry-exporter-otlp==1.17.0"]
+monitor-otlp = ["opentelemetry-exporter-otlp-proto-http==1.17.0"]
```
ssheng, line 158: I thought we couldn't upgrade this to `1.17.0`

aarnphm: I have updated to 1.17 API. The breaking change relies on the `set_logger_provider` import

src/bentoml/_internal/monitoring/base.py
```diff
@@ -0,0 +1,209 @@
...
+    - export_schema
+    - export_data
+    to export the schema and data to the desired format.
+    """
+
+    PRESERVED_COLUMNS: tuple[str, ...] = ()
+
+    monitor_type: str | None
+
+    def __init_subclass__(cls, *, monitor_type: str | None = None):
```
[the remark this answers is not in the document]
aarnphm: so that we can register it to the MONITOR_REGISTRY, which is used under the api.py

Since we expose the MonitorBase on monitoring, user can write their own like so

```python

class CustomMonitor(MonitorBase, monitor_type='my-monitor'): ...
```

Obv this is opt-in, as the previous behaviour should still work as expected.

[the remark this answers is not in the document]
bojiang: I think the previous version has the same effect, and easy to understand.

bojiang (changes requested):
Didn't see the value of `__init_subclass__` refactoring.
Normal classes and inheritance should be enough, and we might not need the extra functionality provided by __init_subclass__. I make it harder to trace.

src/bentoml/_internal/monitoring/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+    "MonitorBase",
+    "DefaultMonitor",
+    "NoOpMonitor",
+]
+
+if _is_otlp_available:
+    __all__.append("OTLPMonitor")
+
+
+def __getattr__(item: str) -> t.Any:
```
bojiang, line 35: Great!

src/bentoml/_internal/monitoring/otlp.py
```diff
@@ -128,13 +135,23 @@ def __init__(
         self._will_export_schema = False
 
     def _init_logger(self) -> None:
+        from opentelemetry.sdk.resources import SERVICE_NAME
```
bojiang, line 138: How this works?

aarnphm: This follows our change within our tracer API in the BentoMLContainer

bojiang (approved):

PR #3853, opened 2023-05-10
refactor(client): move async and sync implementation

aarnphm:
added auto-generated docs per function

add supports for client.kwargs parsing

```python
client.predict(input_series=np.array([1,2,3,4]))
```

Client now also support parsing kwargs for the server api.

update protocol interface and use asyncio.run during destructor

fix(client): create a client type protocol

Move common to main metaclass


sauyon:
Closing in favor of #4116.

PR #3861, opened 2023-05-13
improve: more explicit exception printing when configuration failed to register

aarnphm:
Print more explicit exception when default configuration failed to register

I ran into this when running service from a directory with the file name
`schema.py`, which indirectly interfere with the `schema` module. The error
message was not very helpful, so I added a more explicit message to help with
debugging.

Note that the reason for this collision is that we are still importing
`service.py` into sys.path, so if the directory contains a file named
`schema.py`, it will be imported instead of the `schema` module

We probably need to figure out a way for parsing service.py without importing
it.

### DIFF
--- a/src/bentoml/_internal/configuration/helpers.py
+++ b/src/bentoml/_internal/configuration/helpers.py
@@ -11,6 +11,7 @@ import yaml
 import schema as s
 
 from ..utils import LazyLoader
+from ...exceptions import BentoMLException
 from ...exceptions import BentoMLConfigException
 
 if TYPE_CHECKING:
@@ -113,7 +114,9 @@ def get_default_config(version: int) -> dict[str, t.Any]:
             % (version, e)
         ) from e
     except Exception as err:
-        raise BentoMLException("Unexpected error while setting configuration:\n") from err
+        raise BentoMLException(
+            "Unexpected error while setting configuration:\n"
+        ) from err
     finally:
         return config
 
--- a/src/bentoml/_internal/configuration/helpers.py
+++ b/src/bentoml/_internal/configuration/helpers.py
@@ -113,9 +113,7 @@ def get_default_config(version: int) -> dict[str, t.Any]:
             % (version, e)
         ) from e
     except Exception as err:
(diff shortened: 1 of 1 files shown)


src/bentoml/_internal/configuration/helpers.py
```diff
@@ -105,17 +105,19 @@ def get_default_config(version: int) -> dict[str, t.Any]:
...
     except s.SchemaError as e:
         raise BentoMLConfigException(
             "Default configuration for version %d does not conform to given schema:\n%s"
             % (version, e)
-        ) from None
-    return config
+        ) from e
+    except Exception as err:
+        logger.error("Unexpected error while setting configuration:\n")
+        logger.error(err)
```
sauyon, line 117: Why are we both printing and raising here?

aarnphm: hmm, i think that I want to show the exception here instead of printing.

src/bentoml/_internal/configuration/helpers.py
```diff
@@ -105,17 +105,19 @@ def get_default_config(version: int) -> dict[str, t.Any]:
...
+        raise
```
aarnphm, line 118:
```suggestion
        raise BentoMLException("Unexpected error while setting configuration:\n") from err
```

src/bentoml/_internal/configuration/helpers.py
```diff
@@ -105,17 +105,17 @@ def get_default_config(version: int) -> dict[str, t.Any]:
...
         raise BentoMLConfigException(
             "Default configuration for version %d does not conform to given schema:\n%s"
             % (version, e)
-        ) from None
-    return config
+        ) from e
+    except Exception as err:
+        raise BentoMLException("Unexpected error while setting configuration:\n") from err
+    finally:
+        return config
```
sauyon, line 121: Does this finally actually do anything? Isn't it the same as just having return on the inside?

frostming: Yes, this is incorrect, the config will be always returned and no exception will be raised.

PR #3862, opened 2023-05-15
feat(transformers): support parsing args

aarnphm:
Some model such as GPT-J does support positional args, so we should be able to
support it for load_model


aarnphm:
Quick note here is we aren't currently support passing args and kwargs into get_runnable, since I'm not sure if all args can be write to metadata file.

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -356,18 +356,22 @@ def load_model(
...
-def load_model(bento_model: str | Tag | Model, **kwargs: t.Any) -> t.Any:
+def load_model(bento_model: str | Tag | Model, *args: t.Any, **kwargs: t.Any) -> t.Any:
     """
     Load the Transformers model from BentoML local modelstore with given name.
 
     Args:
         bento_model: Either the tag of the model to get from the store,
                      or a BentoML :class:`~bentoml.Model` instance to load the
                      model from.
+        args: Additional model args to be parse into the model if the object is a Transformers PreTrained protocol.
```
ssheng, line 373:
```suggestion
        args: Additional model args to be passed into the model if the object is a Transformers PreTrained protocol.
```

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -463,8 +467,11 @@ def load_model(bento_model: str | Tag | Model, **kwargs: t.Any) -> t.Any:
         if "_pretrained_class" in bento_model.info.metadata:
             with open(bento_model.path_of(PRETRAINED_PROTOCOL_NAME), "rb") as f:
                 protocol: PreTrainedProtocol = cloudpickle.load(f)
-            return protocol.from_pretrained(bento_model.path, **kwargs)
+            return protocol.from_pretrained(bento_model.path, *args, **kwargs)
         else:
+            assert (
+                len(args) == 0
+            ), "Additional args are not supported for pipeline. Make sure to only use kwargs instead."
```
ssheng, line 474:
```suggestion
            ), "Positional args are not supported for pipeline. Make sure to only use kwargs instead."
```

ssheng (approved):

ssheng (approved):

PR #3864, opened 2023-05-15
feat: support traffic.max_concurrency for api server and runner

frostming:

## What does this PR address?

This PR adds support for `traffic` configuration for both API server and runners.

The previous `timeout` setting is moved under `traffic` field plus a new configuration `max_concurrency`, which defaults to infinite.

## Before submitting:


src/bentoml/_internal/configuration/v1/__init__.py
```diff
@@ -64,7 +64,10 @@
 }
 _API_SERVER_CONFIG = {
     "workers": s.Or(s.And(int, ensure_larger_than_zero), None),
-    "timeout": s.And(int, ensure_larger_than_zero),
```
aarnphm, line 67: This is a breaking change. We should include a migration or deprecation warning below.

src/bentoml/_internal/runner/runner_handle/remote.py
```diff
@@ -77,9 +77,9 @@ def runner_timeout(self) -> int:
         "return the configured timeout for this runner."
         runner_cfg = BentoMLContainer.runners_config.get()
         if self._runner.name in runner_cfg:
-            return runner_cfg[self._runner.name]["timeout"]
+            return runner_cfg[self._runner.name].get("traffic", {})["timeout"]
```
aarnphm, line 80: This will raise KeyError if the default is an empty dict?

frostming: traffic dict is optional but the timeout value isn't, so it is guaranteed to have a `timeout` value in the dict.

tests/unit/_internal/configuration/test_containers.py
```diff
@@ -145,14 +145,16 @@ def test_runner_gpu_configuration(
...
         resources: system
 """
     bentoml_cfg = container_from_file(RUNNER_TIMEOUTS)
     runner_cfg = bentoml_cfg["runners"]
-    assert runner_cfg["timeout"] == 50
-    assert runner_cfg["test_runner_1"]["timeout"] == 100
-    assert runner_cfg["test_runner_2"]["timeout"] == 50
+    assert runner_cfg["traffic"]["timeout"] == 50
+    assert runner_cfg["test_runner_1"]["traffic"]["timeout"] == 100
+    assert runner_cfg["test_runner_2"]["traffic"]["timeout"] == 50
```
aarnphm, line 160: would be nice to also have a test for the migration components when the users pass in the old configuration value and test the warning.

aarnphm (approved):
I think the test is pretty minuscule, but this is LGTM

aarnphm:
Also I suspect the tests might got cancelled because of the timeout. So you might need to increase the default timeout for our github actions.

src/bentoml/_internal/configuration/v1/__init__.py
```diff
@@ -64,7 +64,10 @@
...
+    s.Optional("traffic"): {
```
bojiang, line 67: I suggest to not implement this as a breaking change.

bojiang:
Here we see it will be a breaking change over our configuration.
Since there is just an adjustment over the configuration structure, we may not bother our users.
We may make the original `timeout` optional and do some adapting in configuration

src/bentoml/_internal/configuration/v1/__init__.py
```diff
@@ -279,4 +285,21 @@ def migration(*, override_config: dict[str, t.Any]):
...
+    for key in override_config:
+        if key.startswith("runners."):
+            runner_name = key.split(".")[1]
+            if any(key.schema == runner_name for key in _RUNNER_CONFIG):
+                continue
+            rename_fields(
+                override_config,
+                current=f"runners.{runner_name}.timeout",
+                replace_with=f"runners.{runner_name}.traffic.timeout",
+            )
```
frostming, line 304: Yes, we already did some auto migrations @bojiang

src/bentoml/_internal/configuration/v1/__init__.py
aarnphm: This is included under the migration layer.

aarnphm (approved):

PR #3868, opened 2023-05-16
chore(type): cleanup

aarnphm:
Currently, subprocess build doesn't respect current users environment. If custom
BENTOML_HOME is set, and in the service there is a import_model logics, it will
fail to find the model from the custom BENTOML_HOME and it will try to create a
default bentoml_home instead.

### DIFF
--- a/src/bentoml/_internal/bento/build_config.py
+++ b/src/bentoml/_internal/bento/build_config.py
@@ -8,7 +8,6 @@ import logging
 import subprocess
 from sys import version_info
 from shlex import quote
-from typing import TYPE_CHECKING
 
 import fs
 import attr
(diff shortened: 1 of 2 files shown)


sauyon:
```
❯ env -i A=b python
Python 3.11.3 (main, Apr  5 2023, 15:52:25) [GCC 12.2.1 20230201] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.environ
environ({'A': 'b', 'LC_CTYPE': 'C.UTF-8'})
>>> import subprocess
>>> print(subprocess.check_output("env").decode("utf-8"))
A=b
LC_CTYPE=C.UTF-8
LINES=48
COLUMNS=182
```

Not sure why this is happening, can you check yourself?

sauyon (approved):
Haha, I was very confused about this one for a little while there.

PR #3870, opened 2023-05-16
chore(cli): remove path from `bentoml list`

aarnphm:
this is a chore, but adding path for models when users do `bentoml models list`
is synonymous to when `bentoml list` also shows path


sauyon:
Let's remove the path from both; I thought we discussed that we shouldn't be displaying path?

sauyon (approved):

PR #3876, opened 2023-05-20
docs: Update Document For Example

khanh41:
## Update documentation
- Add `</table>` to `examples/custom_runner/nltk_pretrained_model/README.md`
- Update right input of example : `https://docs.bentoml.org/en/latest/concepts/service.html#json`
- Update document for clearly and remove not have parameters: `https://docs.bentoml.org/en/latest/concepts/service.html#composite-types`


docs/source/concepts/service.rst
```diff
@@ -367,30 +367,35 @@ logic:
 
 .. code-block:: python
 
-    import typing as t
+    from typing import Any
```
aarnphm, line 370: This is just a semantic change. It shouldn't be here.

aarnphm (changes requested):
one tiny comment

docs/source/concepts/service.rst
khanh41: i think we should use `from typing import Any` instead `import typing as t` for clean code. I don't understand what's `t` when read the code below. using `from typing import Any` making it clear and explicit where the Any type is coming from.

aarnphm: I think we did some check a while ago (or someone did this I don't remember exactly), but explicit import from typing is slower than the alias import, because you actually don't eagerly access the module `typing`.

aarnphm, line 370:
```suggestion
    from __future__ import annotations
    from typing import Any
```

aarnphm (dismissed):

aarnphm (approved):

PR #3878, opened 2023-05-22
feat: cleaner APIs for testing Service implementation

parano:
## What does this PR address?

Before this PR, testing a `bentoml.Service` is a bit cumbersome and requires multiple steps:
```python
from clip_api_service import svc

for runner in svc.runners:
   runner.init_local()

# Test python object input
svc.apis["encode"].func( TEST_INPUT )

# Test HTTP request input
inference_api = svc.apis["encode"]
inference_api.func( inference_api.input.from_http_request( TEST_REQUEST) )
```

With changes in this PR:
```python
from clip_api_service import svc

svc.init_local()

# Test python object input
svc.apis.encode( TEST_INPUT )

# Test HTTP request input
svc.apis.encode( svc.apis.encode.input.from_http_request(TEST_REQUEST) )
```

The changes are backwards compatible, the previous usage are still supported.  We should add this usage to the new docs cc @jianshen92 

This was previously undocumented and very hard for the user to discover, it's been asked many times in the community chat:

https://bentoml.slack.com/archives/CKRANBHPH/p1660156096973409?thread_ts=1660155873.466689&cid=CKRANBHPH

https://bentoml.slack.com/archives/CKRANBHPH/p1676023254529259?thread_ts=1675955112.059749&cid=CKRANBHPH

https://bentoml.slack.com/archives/CKRANBHPH/p1682726861687219?thread_ts=1674577682.060979&cid=CKRANBHPH

We should also document how to set up testing for async APIs:

```python
import asyncio
import pytest

pytest_plugins = ('pytest_asyncio',)

@pytest.mark.asyncio
async def test_async_api():
    from clip_api_service import svc
    svc.init_local()  # TODO: add something like `svc.mock_runners`?
    result = await svc.apis.encode( TEST_INPUT )
    ...
```

Fixes #(issue)

## Before submitting:

### DIFF
--- a/src/bentoml/_internal/service/inference_api.py
+++ b/src/bentoml/_internal/service/inference_api.py
@@ -176,6 +176,9 @@ class InferenceAPI:
                 "Reserved API route: '{}' is reserved for infra endpoints".format(route)
             )
 
+    def __call__(self, *args: t.Any, **kwargs: t.Any) -> t.Any:
+        return self.func(*args, **kwargs)
+
 
 def _InferenceAPI_dumper(dumper: yaml.Dumper, api: InferenceAPI) -> yaml.Node:
     return dumper.represent_dict(
--- a/src/bentoml/_internal/service/service.py
+++ b/src/bentoml/_internal/service/service.py
@@ -8,6 +8,7 @@ import logging
 import importlib
 from typing import TYPE_CHECKING
 from functools import partial
+from collections import UserDict
 
 import attr
 
@@ -34,10 +35,21 @@ if TYPE_CHECKING:
     from ..bento import Bento
     from ...grpc.v1 import service_pb2_grpc as services
     from .openapi.specification import OpenAPISpecification
+
+    APIDictBase = UserDict[str, InferenceAPI]
 else:
     grpc, _ = import_grpc()
+    APIDictBase = UserDict
 
 logger = logging.getLogger(__name__)
+
+
+class APIDict(APIDictBase):
+    def __getattr__(self, name):
+        try:
+            return self.__getitem__(name)
+        except KeyError:
+            raise AttributeError(f"InferenceAPI '{attr}' not found")
 
 
 def add_inference_api(
@@ -113,7 +125,7 @@ class Service:
     grpc_handlers: list[grpc.GenericRpcHandler] = attr.field(init=False, factory=list)
 
     # list of APIs from @svc.api
-    apis: t.Dict[str, InferenceAPI] = attr.field(init=False, factory=dict)
+    apis: APIDict = attr.field(init=False, factory=APIDict)
 
     # Tag/Bento are only set when the service was loaded from a bento
     tag: Tag | None = attr.field(init=False, default=None)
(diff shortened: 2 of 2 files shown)


src/bentoml/_internal/service/inference_api.py
```diff
@@ -176,6 +176,9 @@ def _validate_route(route: str):
                 "Reserved API route: '{}' is reserved for infra endpoints".format(route)
             )
 
+    def __call__(self, *args: t.Any, **kwargs: t.Any) -> t.Any:
+        return self.func(*args, **kwargs)
```
aarnphm, line 180: Idk if this is a good idea.

aarnphm: How do we handle ctx in this case?

src/bentoml/_internal/service/service.py
```diff
@@ -34,12 +35,23 @@
...
 
 logger = logging.getLogger(__name__)
 
 
+class APIDict(APIDictBase):
+    def __getattr__(self, name):
+        try:
+            return self.__getitem__(name)
+        except KeyError:
+            raise AttributeError(f"InferenceAPI '{attr}' not found")
```
aarnphm, line 52: Is this purely for convenience?

I think that asking people to write `svc.apis['encode']` is also fine

Even though `svc.apis.encode` is nice.

src/bentoml/_internal/service/service.py
```diff
@@ -435,6 +447,15 @@ def add_grpc_interceptor(
...
         self.grpc_handlers.extend(handlers)
 
+    def init_local(self):
+        """
+        Initiate this service instance in local mode, which will load all runners in
+        current Python process. This should be used for one-off inference task, testing
+        and debugging purpose only.
+        """
+        for runner in self.runners:
+            runner.init_local(quiet=True)
```
aarnphm, line 457: I don't think this is a good idea. Maybe we can make a runners_map and then allow users to do something like

```python
svc = bentoml.Service(runners=[runners_1, runners_2])

svc.runner_1.init_local()
svc.runner_2.init_local()
```

parano: users can already do `svc.runners[0].init_local()`, this is just for making it easier to test things in a notebook/dev environment

src/bentoml/_internal/service/service.py
parano: yes just for convenience, although I came up with another different idea for this, will follow up

src/bentoml/_internal/service/inference_api.py
parano: user will need to pass in the context in that case 🤔

aarnphm: hmm, that seems pretty hacky

src/bentoml/_internal/service/service.py
aarnphm: right, but I feel like users shouldn't be exposed to these types of API,  since they will "shoot" themselves in the foot.

aarnphm:
@parano any updates on this?

ssheng:
Is it possible to invoke the API directly without having to go through `.apis` indirection?

```python
svc.apis.encode( TEST_INPUT )

svc.encode(TEST_INPUT)
```

aarnphm:
> Is it possible to invoke the API directly without having to go through `.apis` indirection?
> 
> ```python
> svc.apis.encode( TEST_INPUT )
> 
> svc.encode(TEST_INPUT)


+1 for `svc.apis.encode`

frostming:
> Is it possible to invoke the API directly without having to go through `.apis` indirection?
> 
> ```python
> svc.apis.encode( TEST_INPUT )
> 
> svc.encode(TEST_INPUT)
> ```

This would be in the roadmap of new IO descriptor to add another Client implementation that doesn't need to start the server in a subprocess.

frostming:
With the new IO descriptor, we can do this:


```python
from bentoml_io.client.testing import TestingClient

with TestingClient(service=svc) as client:
    result = await client.classify(input_series=[[1,2,3,4]])
```

Closing this now.

PR #3882, opened 2023-05-23
fix(transformers): make sure to set correct GPU device on set framework

aarnphm:
cc @bojiang

### DIFF
--- a/src/bentoml/_internal/frameworks/transformers.py
+++ b/src/bentoml/_internal/frameworks/transformers.py
@@ -25,6 +25,8 @@ from ..models.model import ModelOptions
 from ..models.model import ModelSignature
 
 if t.TYPE_CHECKING:
+    import torch
+    import tensorflow as tf
     import cloudpickle
     from transformers.models.auto.auto_factory import (
         _BaseAutoModelClass as BaseAutoModelClass,
@@ -49,6 +51,10 @@ if t.TYPE_CHECKING:
     TupleAutoModel = tuple[BaseAutoModelClass, ...]
 
     class PreTrainedProtocol(t.Protocol):
+        @property
+        def framework(self) -> str:
+            ...
+
         def save_pretrained(self, save_directory: str, **kwargs: t.Any) -> None:
             ...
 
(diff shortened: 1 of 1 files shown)


src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -886,19 +900,56 @@ def __init__(self):
...
+                        with tf.device(
+                            "/device:CPU:0"
+                            if available_gpus in ("", "-1")
+                            else f"/device:GPU:{available_gpus}"
+                        ):
+                            self.model = t.cast(
+                                transformers.TFPreTrainedModel,
+                                load_model(bento_model, **kwargs),
+                            )
+                    else:
```
bojiang, line 939: shall we check if it is "JAX" here?

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -886,19 +900,56 @@ def __init__(self):
...
+                            transformers.PreTrainedModel,
+                            load_model(bento_model, **kwargs),
+                        ).to(
+                            torch.device(
+                                "cuda" if available_gpus not in ("", "-1") else "cpu"
+                            )
+                        )
+                        torch.set_default_tensor_type("torch.cuda.FloatTensor")
+                    elif "tf" == bento_model.info.metadata["_framework"]:
+                        with tf.device(
```
bojiang, line 930: shall we delay the current tf importing to here?

src/bentoml/_internal/frameworks/transformers.py
aarnphm: it is

src/bentoml/_internal/frameworks/transformers.py
aarnphm: tf and torch is already lazy loaded

src/bentoml/_internal/frameworks/transformers.py
aarnphm: since the possible runtime are just torch, tf and flax

bojiang (approved):

PR #3886, opened 2023-05-23
feat(configuration): allow overwrite load_config with default override_config_dict

aarnphm:
This PR allows user to override the default configuration with given dict when
using `load_config`

Currently, we allow this behaviour via `BENTOML_CONFIG_OPTIONS`, which is still
the current behaviour

This PR add an additional enhancement via `load_config`:

```python
from bentoml import load_config

load_config(override_config={"api_server": {"timeout": 36000000}, "runners": {"llm-dolly-v2-runner": {'timeout': 36000000}}})
```

### DIFF
--- a/src/bentoml/_internal/configuration/__init__.py
+++ b/src/bentoml/_internal/configuration/__init__.py
@@ -134,7 +134,10 @@ def get_quiet_mode() -> bool:
     return False
 
 
-def load_config(bentoml_config_file: str | None = None):
+def load_config(
+    bentoml_config_file: str | None = None,
+    override_config: dict[str, t.Any] | None = None,
+):
     """Load global configuration of BentoML"""
 
     from .containers import BentoMLContainer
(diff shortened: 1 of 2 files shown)


src/bentoml/_internal/configuration/__init__.py
```diff
@@ -134,7 +134,10 @@ def get_quiet_mode() -> bool:
     return False
 
 
-def load_config(bentoml_config_file: str | None = None):
+def load_config(
+    bentoml_config_file: str | None = None,
+    override_config: dict[str, t.Any] | None = None,
```
sauyon, line 139:
```suggestion
    config_override: dict[str, t.Any] | None = None,
```

sauyon (commented):
I think calling it `config_override` is slightly more natural, but other than that LGTM!

aarnphm:
discussed with @ssheng about this. This is not suited for the upcoming configuration rework.

PR #3889, opened 2023-05-24
feat: use uvloop

sauyon:
Set our dependency to `uvicorn[standard]` in order to use the cpython `uvloop` instead of the standard `asyncio` async runtime.

EDIT: We should A/B compare this, if we have time.


aarnphm (approved):

aarnphm:
@sauyon This doesn't seem like a free update, since the test is now consistently failing.

PR #3891, opened 2023-05-25
feat: initial rust implementation

sauyon:
@aarnphm ;)


aarnphm:
Love this very much. But probably we can merge this in once everyone else PR is in, since this essentially move the project structure to rust implementation.

sauyon:
Actually got should handle the merge just fine.

But I don't know whether we actually want to merge this anytime soon, anyway, probably want to confirm everything's working.

gRPC tests are failing, for example... ;)

sauyon:
Actually got should handle the merge just fine.

But I don't know whether we actually want to merge this anytime soon, anyway, probably want to confirm everything's working.

gRPC tests are failing, for example... ;)

sauyon:
After that gets fixed I should factor out both the tag and exception changes, I think.

sauyon:
Closing as discussed.

PR #3896, opened 2023-05-27
feat(tracking): enable telemetry to track serve from Server API

aarnphm:
This PR add a telemetry field into our serve metrics to enable checking if users
are running serve from the new Server API or from the CLI. Since we are copying
over the environment to the subprocess, it will still respect
BENTOML_DO_NOT_TRACK

This is usefull for tracking OpenLLM telemetry.


src/bentoml_cli/serve.py
```diff
@@ -2,21 +2,61 @@
...
+
+
 def add_serve_command(cli: click.Group) -> None:
     from bentoml.grpc.utils import LATEST_PROTOCOL_VERSION
     from bentoml._internal.log import configure_server_logging
     from bentoml_cli.env_manager import env_manager
     from bentoml._internal.configuration.containers import BentoMLContainer
 
+    # NOTE: This flag is used for tracking when users are using the new server API.
+    from_server_api = click.option(
```
parano, line 56: Could we just use an env bar instead is it not exposed neither in the api nor in the cli args?

aarnphm: that works as well.

src/bentoml/server.py
```diff
@@ -189,6 +197,9 @@ def get_client(self) -> Client | None:
         pass
 
     def stop(self) -> None:
+        # NOTE: User shouldn't manually set this since this envvar will be managed by BentoML.
+        os.environ[BENTOML_SERVE_FROM_SERVER_API] = str(False)
```
ssheng, line 201: delete the env var instead of setting to false?

src/bentoml/_internal/utils/analytics/schemas.py
```diff
@@ -190,6 +190,7 @@ class ServeInitEvent(EventMeta):
     serve_id: str
     production: bool
     serve_from_bento: bool
+    serve_from_server_api: bool
```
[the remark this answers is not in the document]
aarnphm: I think right now we only have two option, serving from CLI and API.

src/bentoml/server.py
aarnphm: 👍

src/bentoml/server.py
```diff
@@ -305,18 +316,24 @@ def client(self) -> HTTPClient | None:
 
     def get_client(self) -> HTTPClient | None:
         if self.process is None:
+            # NOTE: if the process is None, we reset this envvar
+            os.environ[BENTOML_SERVE_FROM_SERVER_API] = str(False)
```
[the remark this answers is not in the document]
aarnphm: I refactor it into one function that we can manage the process correctly.

But unfortunately I don't think we can avoid this.

src/bentoml/server.py
```diff
@@ -184,11 +192,50 @@ def __exit__(
 
         return _Manager()
 
-    @abstractmethod
     def get_client(self) -> Client | None:
```
aarnphm, line 195: @sauyon do you think that this is technically a breaking change?

ssheng (approved):

src/bentoml/server.py
sauyon: I don't believe so; it should only make things that didn't work before work now, right?

aarnphm: I guess only case it breaks is user implementing their own server. But since we never expose it in docs anywhere i think it should be ok

PR #3898, opened 2023-05-28
feat(save_model): support parsing bentoml.Tag

aarnphm:
Before, it is kinda awkard to provide a `bentoml.Tag` to save_model

```python
tag: bentoml.Tag
bentoml.transformers.save_model(str(tag), ...)
```

Now, with this PR, save_model can also accept `bentoml.Tag`

```python
tag: bentoml.Tag
bentoml.transformers.save_model(tag, ...)
```


sauyon:
Er, why'd you delete the template?

aarnphm:
> Er, why'd you delete the template?

I don't think it is useful anymore. iirc we did discuss it some time ago about removing the template since it is not really up-to-date anw.

I feel like we can just copy from the existing file when adding a new framework, all things considered.

sauyon (approved):
Sure, that seems fine.

I think having a template is nice so we have placeholders everywhere necessary to prevent possible copy paste errors, but I'm not implementing frameworks anyway so I'm not fussed.

PR #3902, opened 2023-05-28
fix(serialization): loads gpu torch tensor on cpu-only machine and revert DefaultContainer optimization

larme:
### fix cuda pytorch tensor unpickling in cpu-only enviornment
Try to fix the problem when unpickling a PyTorch tensor with its device set to "cuda" in a cpu-only environment (see this issue: https://github.com/pytorch/pytorch/issues/16797 for more information). This is crucial for BentoML's distribution runner because the distribution runner may run in a pod with GPU device while the api server run in cpu-only pods. Then when the runner pickle the tensor and send it to api server, the api server will meet unpickling errors.

The current strategy is to call normal unpickling function (`pickle.loads`) and only fallback to special torch tensor handling unpickler when a `RuntimeError` is raised. The other way is to call the special torch tensor handling unpickler everytime we need to unpickle bytes. The advantage of current strategy is it will handle torch tensor on device other than cuda better (the reason why https://github.com/pytorch/pytorch/pull/49920 is not accepted by PyTorch people), and it's faster than the second strategy when unpickling data other than torch tensor. The disadvantage of the current strategy is that if the `RuntimeError` is raised, catching the exception will have some performance cost. To measure the performance, I made some codes for benchmark. The repos is at https://github.com/larme/pytorch_unpickler_benchmark and a sample result is at https://docs.google.com/spreadsheets/d/17i4fXPpcamTIqzg_mMBRnyY7FzQIuXiQc7dTp_iRCkA/edit?usp=sharing

### revert `DefaultContainer` optimization

The `pep574_dumps` is originally intended for optimizing numpy ndarray/pandas dataframe serialization. In these situations the `main_bytes` part will only contain some metadata. That's why putting these bytes in header will not cause trouble. In the meantime the `concat_buffer_bytes` contains out-of-band buffers that need no computation during deserialization, which will save computation resource especially when the payload is large. 

However after optimizing ndarray/dataframe container I also want to use this pair of functions to optimize `DefaultContainer` because it may also contain large ndarray. However there's also the possiblity that `DefaultContainer` may be a dictionary/list containing a large object that has no out-of-band buffer (e.g. PIL Image or PyTorch tensor) and a small numpy ndarray. In that case `concat_buffer_bytes` will be small and `main_bytes` will be huge, hence we barely save any time while having a large header (that may cause errors).

Since `DefaultContainer` has so many possible combinations, I think we should just stick with plain pickle.dumps and pickle.loads (with pytorch cuda tensor fix) now.

### DIFF
--- a/src/bentoml/_internal/runner/container.py
+++ b/src/bentoml/_internal/runner/container.py
@@ -10,6 +10,7 @@ from ..types import LazyType
 from ..utils import LazyLoader
 from ..utils.pickle import pep574_dumps
 from ..utils.pickle import pep574_loads
+from ..utils.pickle import fixed_torch_loads
 
 SingleType = t.TypeVar("SingleType")
 BatchType = t.TypeVar("BatchType")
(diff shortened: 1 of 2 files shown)


aarnphm (approved):
Should we add a document notice on our pytorch docs?

aarnphm:
Hmm, it seems it degrade by quite a bit no?

larme:
> Hmm, it seems it degrade by quite a bit no?

switched to another strategy that degrades the performance about 0.03ms per unpickling. I think this will be acceptable.

larme:
The `pep574_dumps` is originally intended for optimizing numpy ndarray/pandas dataframe serialization. In these situations the `main_bytes` part will only contain some metadata. That's why putting these bytes in header will not cause trouble. In the meantime the `concat_buffer_bytes` contains out-of-band buffers that need no computation during deserialization, which will save computation resource especially when the payload is large. 

However after optimizing ndarray/dataframe container I also want to use this pair of functions to optimize `DefaultContainer` because it may also contain large ndarray. However there's also the possiblity that `DefaultContainer` may be a dictionary/list containing a large object that has no out-of-band buffer (e.g. PIL Image or PyTorch tensor) and a small numpy ndarray. In that case `concat_buffer_bytes` will be small and `main_bytes` will be huge, hence we barely save any time while having a large header (that may cause errors).

Since `DefaultContainer` has so many possible combinations, I think we should just stick with plain pickle.dumps and pickle.loads (with pytorch cuda tensor fix) now. This fix has some coupling with PyTorch tensor loading fix so I fix these 2 things in the same pr

cc @bojiang @frostming

aarnphm (approved):

PR #3908, opened 2023-05-30
chore(cli): cleanup and remove unused

aarnphm:
Some house keeping stuff for type warning and `# type: ignore`.

using CLI.command(name=...) to remove redundant reportUnusedFunction
in the local scope.

## What does this PR address?

Fixes #(issue)

## Before submitting:


sauyon:
Hm, this seems like it's only a problem because it's being done in a function, right? What's the reason for that?

aarnphm:
> Hm, this seems like it's only a problem because it's being done in a function, right? What's the reason for that?

Yes, since we are lazy load the CLI to avoid side effect from importing bentoml (IIRC), but otherwise this is mainly to address that.

frostming:
This change seems anti-pattern. Maybe we could configure the linter(pyright I guess) to ignore reportUnusedFunction in the scope.

aarnphm:
> This change seems anti-pattern. Maybe we could configure the linter(pyright I guess) to ignore reportUnusedFunction in the scope.

For pyright they don't have support for per-file ignore, like ruff. It is just either exclude they whole module or do line-by-line `# type: ignore`, which is not ideal

sauyon:
> Yes, since we are lazy load the CLI to avoid side effect from importing bentoml (IIRC), but otherwise this is mainly to address that.

I think this is what we should solve; why are we lazy loading `bentoml` in the CLI? Don't we load it anyways?

> For pyright they don't have support for per-file ignore, like ruff. It is just either exclude they whole module or do line-by-line `# type: ignore`, which is not ideal

Why is it not ideal? This seems fine.

aarnphm:
I don't think we should lazy load bentoml. I suspect the reason we have to put it into the function to solve the cycling chain import.

sauyon:
Let me quickly try something.

PR #3909, opened 2023-05-30
imports: LazyModule

aarnphm:
migrate `__init__` to LazyModule, which is an extension of LazyLoader

This allows use to load attributes from given module lazily, or as late as
possible, without sacrificing the import speed.

LazyModule supports IDE completion fully, and LSP should work nicely with type
hint.

LazyModule is largely based on Transformers' implementation, and have been
tested on OpenLLM.


src/bentoml/_internal/frameworks/fastai.py
```diff
@@ -140,7 +140,7 @@ def save_model(
               :obj:`~bentoml.Tag` name.
         learner: :obj:`~fastai.learner.Learner` to be saved.
         signatures: Signatures of predict methods to be used. If not provided, the signatures default to
-                    ``predict``. See :obj:`~bentoml.types.ModelSignature` for more details.
+                    ``predict``. See :obj:`~bentoml._internal.models.model.ModelSignature` for more details.
```
sauyon, line 143: Did this not work before?

pyproject.toml
```diff
@@ -260,6 +260,7 @@ exclude = [
     "tests/proto",
 ]
 target-version = "py310"
+extend-exclude = ["src/bentoml/utils.py"]
```
sauyon, line 263: Should this be here?

requirements/tests-requirements.txt
```diff
@@ -3,7 +3,7 @@ pandas
 pydantic
 scikit-learn
 yamllint==1.32.0
-black[jupyter]==22.12.0
+black[jupyter]==23.3.0
```
sauyon, line 6: Maybe just make a PR to update black?

src/bentoml/_internal/context.py
```diff
@@ -73,7 +73,7 @@ def mutablecopy(self) -> Metadata:
 
 
 @attr.define
-class InferenceApiContext:
+class Context:  # NOTE: This is the InferenceApiContext
```
sauyon, line 76: Why the rename?

sauyon:
Looks like there are a lot of things going on here, can we split those up?

src/bentoml/_internal/frameworks/fastai.py
aarnphm: I don't think this shows on doc no?

aarnphm: and also it doens't make sense to have a `bentoml.types`, since `bentoml.types` can only work under the `TYPE_CHECKING` anw.

pyproject.toml
aarnphm: oh no, i forgot to update this

aarnphm, line 263:
```suggestion
```

src/bentoml/_internal/context.py
aarnphm: For the LazyModule import attribute to work properly

frostming:
I have another approach for this, using the module-level `__getattr__`, maybe PR soon.

src/bentoml/_internal/frameworks/fastai.py
sauyon: Hm, that feels like it shouldn't be the case.

sauyon: Isn't the real problem that the bentoml.types module doesn't have any docs?

aarnphm:
Closed and will revisit some other time, as this is super low in the list.

PR #3910, opened 2023-05-30
feat(attr): intial support of Attrs model

aarnphm:
A simplier support for Attrs that is based on #3906


sauyon:
Closing in favor of IO descriptor rework.

PR #3914, opened 2023-05-31
feat: support ipv6 addresses for serve

sauyon:
No longer listens on ipv6 catchall by default.

### DIFF
--- a/src/bentoml/_internal/configuration/helpers.py
+++ b/src/bentoml/_internal/configuration/helpers.py
@@ -1,8 +1,8 @@
 from __future__ import annotations
 
+import ipaddress
 import logging
 import os
-import socket
 import typing as t
 from functools import singledispatch
 from typing import TYPE_CHECKING
@@ -161,7 +161,7 @@ def ensure_iterable_type(typ_: type) -> t.Callable[[t.MutableSequence[t.Any]], b
 def is_valid_ip_address(addr: str) -> bool:
     """Check if given string is a valid IP address."""
     try:
-        _ = socket.inet_aton(addr)
+        _ = ipaddress.ip_address(addr)
         return True
-    except socket.error:
+    except ValueError:
         return False
--- a/src/bentoml/serve.py
+++ b/src/bentoml/serve.py
@@ -22,6 +22,7 @@ from bentoml._internal.log import SERVER_LOGGING_CONFIG
 from ._internal.configuration.containers import BentoMLContainer
 from ._internal.runner.runner import Runner
 from ._internal.utils import is_async_callable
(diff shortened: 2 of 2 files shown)


aarnphm (dismissed):

src/bentoml/server.py
```diff
@@ -431,5 +432,6 @@ def get_client(self) -> GrpcClient | None:
             GrpcClient.wait_until_server_ready(
                 host=self.host, port=self.port, timeout=self.timeout
             )
-            self._client = GrpcClient.from_url(f"{self.host}:{self.port}")
+            client_host = "localhost" if self.host in ["0.0.0.0", "::"] else self.host
```
aarnphm, line 435:
```suggestion
            client_host = "localhost" if self.host in ["0.0.0.0", "::"] else self.host
```
gRPC actually support ipv6 by default. so `::3000` should work

sauyon: There are a bunch of reasons that wouldn't work, but importantly this is more about the fact that `::` and `0.0.0.0` are not valid addresses.

aarnphm (approved):

aarnphm:
@sauyon hmm seems like the monitoring tests is failing on Windows :)

sauyon:
@aarnphm can you look at this one again?

sauyon:
@bojiang it seems that `AF_INET6` doesn't give us ipv4 for free on Windows; I don't think it's worth taking the effort to fix this for now.

sauyon:
Actually, just reverted serving by default on `::`; this should be fine now!

aarnphm (approved):

aarnphm:
@sauyon seems like the windows tests are still failing

aessiane:
Hey guys! Having this working would be terrific for our team, do you have an idea of how likely this is to be merged?

sauyon:
Have been ignoring this because of Windows CI bugs, but will bump this one!

aarnphm:
@sauyon if the windows bug is hard to fix then should we just warn windows users that ipv6 won't work for now?

sauyon:
Oddly it's not even that ipv6 doesn't work; it seems to work just fine both v4 and v6. Something's wrong with the test.

aarnphm:
Ok lets just merge then?

aarnphm (approved):

aarnphm:
One day we will fix our tests for once.

sauyon:
I'm working on it now; if it's not done by next release we can probably merge this, yeah.

sauyon:
@aarnphm let's merge this now? dropped test work, clearly..

PR #3916, opened 2023-05-31
use def instead of `functools.partial`

sauyon:


aarnphm (dismissed):

aarnphm:
@sauyon reminder to setup your git pre-commit

sauyon:
one of these days....

aarnphm (approved):

PR #3927, opened 2023-06-02
docs: document bentoml.types

sauyon:
At some point we should probably move these to `bentoml.models` or something, but that would be breaking.

### DIFF
--- a/src/bentoml/types.py
+++ b/src/bentoml/types.py
@@ -1,7 +1,4 @@
-from typing import TYPE_CHECKING
+from ._internal.models.model import ModelSignature
+from ._internal.models.model import ModelSignatureDict
 
-if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


aarnphm (approved):

src/bentoml/types.py
```diff
@@ -1,7 +1,4 @@
-from typing import TYPE_CHECKING
+from ._internal.models.model import ModelSignature
+from ._internal.models.model import ModelSignatureDict
```
aarnphm, line 2: actually this won't work, maybe we need to import TypedDict from `typing_extensions` and just make ModelSignatureDict a type that you can just use to initialize with

aarnphm (approved):

aarnphm (approved):

PR #3947, opened 2023-06-10
fix(strategy): handle special case when assigning device 0

aarnphm:

## What does this PR address?

Fixes #(issue)

## Before submitting:

### DIFF
--- a/src/bentoml/_internal/resource.py
+++ b/src/bentoml/_internal/resource.py
@@ -224,6 +224,9 @@ class NvidiaGpuResource(Resource[t.List[int]], resource_id="nvidia.com/gpu"):
 
         try:
             if isinstance(spec, int):
+                if spec == 0:
+                    # NOTE: this is a special case for the default GPU placement.
+                    return [0]
                 if spec < -1:
                     raise ValueError
                 return list(range(spec))


src/bentoml/_internal/resource.py
```diff
@@ -224,6 +224,9 @@ def from_spec(cls, spec: int | str | list[int | str]) -> list[int]:
 
         try:
             if isinstance(spec, int):
+                if spec == 0:
+                    # NOTE: this is a special case for the default GPU placement.
+                    return [0]
                 if spec < -1:
                     raise ValueError
                 return list(range(spec))
```
larme, line 232: here we actually allow -1. Should we allow -1? For CUDA_VISIBLE_DEVICES we can set to -1 to indicated that we want to disable gpu. But for our case, if users did specify gpu in resource configuration, that means they want to use GPU. So maybe -1 should also be disallowed?

aarnphm: with -1, torch will also respect that as well. I think user might get surprised if we don't have the same behaviour

aarnphm: I think we should

aarnphm:
closed as for assigning single GPU, it should be `'runners.resource."nvidia.com/gpu"=[0]'`

The -1 behaviour should be addressed in a separate PR @larme

PR #3961, opened 2023-06-14
feat: Model specification in bentofile config

frostming:

## What does this PR address?

1. Add `models` config to `bentofile.yaml`, where each item can either be a string model tag, or a dictionary with `tag`, `filter` and `alias` keys:
   - `tag`: the model tag
   - `filter`: filter string in the same syntax as bentocloud, applicable when the version is `latest` or not given
   - `alias`: if set, one can use `bentoml.models.get(alias)` to get the model.
2. Allow to run `bentoml models pull` without tag argument, in which case the models defined in the build config will be pulled
3. No longer save models with bento
   - When building a bento, the `models` folder won't be created
   - When exporting a bento, the models will be copied from the model store into the artifact
   - When importing a bento from an artifact, the models will be copied into the model store and `models/` will be removed.
   - When serving a bento, the bento-local model store will be first read for compatibilty, and then the global model store.
   - When pulling a bento, the models will be created into the model store.

## Before submitting:

### DIFF
--- a/src/bentoml/_internal/bento/bento.py
+++ b/src/bentoml/_internal/bento/bento.py
@@ -13,6 +13,7 @@ import yaml
 import fs.osfs
 import fs.errors
 import fs.mirror
+import fs.walk
 from fs.copy import copy_file
 from cattr.gen import override
 from cattr.gen import make_dict_structure_fn
@@ -27,7 +28,7 @@ from ..types import PathType
 from ..utils import bentoml_cattr
 from ..utils import copy_file_to_fs_folder
 from ..utils import normalize_labels_value
-from ..models import ModelStore
+from ..models import ModelStore, copy_model
 from ..runner import Runner
 from ...exceptions import InvalidArgument
 from ...exceptions import BentoMLException
@@ -123,7 +124,7 @@ class Bento(StoreItem):
 
     _info: BentoInfo
 
-    _model_store: ModelStore
+    _model_store: t.Optional[ModelStore] = None
     _doc: t.Optional[str] = None
 
     @staticmethod
@@ -133,12 +134,11 @@ class Bento(StoreItem):
     @__fs.validator  # type:ignore # attrs validators not supported by pyright
     def check_fs(self, _attr: t.Any, new_fs: FS):
         try:
-            new_fs.makedir("models", recreate=True)
-        except fs.errors.ResourceReadOnly:
-            # when we import a tarfile, it will be read-only, so just skip the step where we create
-            # the models folder.
-            pass
-        self._model_store = ModelStore(new_fs.opendir("models"))
+            models = new_fs.opendir("models")
+        except fs.errors.ResourceNotFound:
+            return
+        else:
+            self._model_store = ModelStore(models)
 
     def __init__(self, tag: Tag, bento_fs: "FS", info: "BentoInfo"):
         self._tag = tag
@@ -211,12 +211,6 @@ class Bento(StoreItem):
             for runner in svc.runners:
                 for model in runner.models:
                     models.add(model)
-
-        bento_fs.makedir("models", recreate=True)
-        bento_model_store = ModelStore(bento_fs.opendir("models"))
-        for model in models:
-            logger.info('Packing model "%s"', model.tag)
-            model.save(bento_model_store)
 
         # create ignore specs
         specs = BentoPathSpec(build_config.include, build_config.exclude)
@@ -312,6 +306,40 @@ class Bento(StoreItem):
 
         return res
 
+    def export(
+        self,
+        path: str,
+        output_format: t.Optional[str] = None,
+        *,
+        protocol: t.Optional[str] = None,
+        user: t.Optional[str] = None,
+        passwd: t.Optional[str] = None,
+        params: t.Optional[t.Dict[str, str]] = None,
+        subpath: t.Optional[str] = None,
+    ) -> str:
+        # copy the models to fs
+        models_dir = self._fs.makedir("models", recreate=True)
+        model_store = ModelStore(models_dir)
+        global_model_store = BentoMLContainer.model_store.get()
+        for model in self.info.models:
+            copy_model(
+                model.tag,
+                src_model_store=global_model_store,
+                target_model_store=model_store,
+            )
+        try:
+            return super().export(
+                path,
+                output_format,
+                protocol=protocol,
+                user=user,
+                passwd=passwd,
+                params=params,
+                subpath=subpath,
+            )
+        finally:
+            self._fs.removetree("models")
+
     @property
     def path(self) -> str:
         return self.path_of("/")
@@ -340,6 +368,7 @@ class Bento(StoreItem):
     def save(
         self,
         bento_store: "BentoStore" = Provide[BentoMLContainer.bento_store],
+        model_store: "ModelStore" = Provide[BentoMLContainer.model_store],
     ) -> "Bento":
         try:
             self.validate()
@@ -348,7 +377,20 @@ class Bento(StoreItem):
 
         with bento_store.register(self.tag) as bento_path:
             out_fs = fs.open_fs(bento_path, create=True, writeable=True)
+            if self._model_store is not None:
+                # Move models to the global model store, if any
+                for model in self._model_store.list():
+                    copy_model(
+                        model.tag,
+                        src_model_store=self._model_store,
+                        target_model_store=model_store,
+                    )
+                self._model_store = None
             fs.mirror.mirror(self._fs, out_fs, copy_if_newer=False)
+            try:
+                out_fs.removetree("models")
+            except fs.errors.ResourceNotFound:
+                pass
             self._fs.close()
             self.__fs = out_fs
 
--- a/src/bentoml/_internal/bento/bento.py
+++ b/src/bentoml/_internal/bento/bento.py
@@ -165,9 +165,9 @@ class Bento(StoreItem):
         build_config: BentoBuildConfig,
         version: t.Optional[str] = None,
         build_ctx: t.Optional[str] = None,
+        model_store: ModelStore = Provide[BentoMLContainer.model_store],
     ) -> Bento:
         from ..service.loader import import_service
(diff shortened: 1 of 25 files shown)


aarnphm:
cc @Haivilo

src/bentoml/_internal/yatai_rest_api_client/yatai.py
```diff
@@ -387,3 +388,18 @@ def download_model(
         resp = self.session.get(url, stream=True)
         self._check_resp(resp)
         return resp
+
+    def get_latest_model(
```
aarnphm, line 392: I feel like since the bento cloud client PR is going to get merged soon, we should also include these in there.

src/bentoml_cli/models.py
```diff
@@ -162,6 +167,20 @@ def delete(delete_targets: str, yes: bool) -> None:  # type: ignore (not accesse
...
+            for bento in bento_store.list():
+                if bento._model_store is not None:
+                    continue
+                if any(model.tag == tag for model in bento.info.models):
+                    in_use.append(bento.tag)
+            if in_use:
+                raise BentoMLException(
+                    f"Model {tag} is being used by the following bentos and can't be deleted:\n  "
+                    + "\n  ".join(map(str, in_use))
+                )
```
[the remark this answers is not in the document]
frostming: Nope, it's required due to this change, to prevent a model from being deleted when it's still used by a bento, as it is no longer stored with the bento(changed by this PR).

src/bentoml/_internal/yatai_rest_api_client/yatai.py
frostming: No problem, I can resolve the conflicts after it is merged.

src/bentoml/_internal/configuration/containers.py
```diff
@@ -172,6 +174,7 @@ def to_dict(self) -> providers.ConfigDictType:
 @dataclass
 class _BentoMLContainerClass:
     config = providers.Configuration()
+    model_aliases = providers.Static({})
```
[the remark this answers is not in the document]
frostming: To allow giving models an alias that is usually shorter to reference to the model. So this is possible:

```python
model = bentoml.models.get(alias)
```

src/bentoml_cli/models.py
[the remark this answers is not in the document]
aarnphm: got it make sense

src/bentoml/_internal/yatai_rest_api_client/yatai.py
aarnphm: sg if that is not too much of a hassle for you ❤️

src/bentoml/_internal/configuration/containers.py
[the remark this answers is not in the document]
aarnphm: ah i see. Hmm 🤔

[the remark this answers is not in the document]
sauyon: This seems fine to me.

aarnphm:
cc @frostming when you have time to fix the merge conflict 😅 as #3991 has been merged

src/bentoml/_internal/bento/bento.py
```diff
@@ -306,6 +316,40 @@ def from_fs(cls, item_fs: FS) -> Bento:
...
+        # copy the models to fs
+        models_dir = self._fs.makedir("models", recreate=True)
+        model_store = ModelStore(models_dir)
+        global_model_store = BentoMLContainer.model_store.get()
+        for model in self.info.models:
+            copy_model(
+                model.tag,
+                src_model_store=global_model_store,
+                target_model_store=model_store,
+            )
```
larme, line 339: if we assume all models are available in local model store (not sure about this when we have model repositories), then maybe we can symlink the models and export should follow symlink. That saves copy time for larger models.

Maybe have another pr to do this so we can merge this pr first?

src/bentoml/_internal/bento/build_config.py
```diff
@@ -705,6 +705,38 @@ def _converter(value: OptionsCls | dict[str, t.Any]) -> options_type:
     return _converter
 
 
+@attr.frozen
+class ModelSpec:
+    tag: str
+    filter: t.Optional[str] = None
```
larme, line 711: can filter be a list? We may have multiple filters

src/bentoml_cli/models.py
```diff
@@ -233,9 +253,37 @@ def import_from(model_path: str) -> None:  # type: ignore (not accessed)
...
+        help="Path to bentofile. Default to 'bentofile.yaml'",
+    )
+    def pull(model_tag: str | None, force: bool, context: str, bentofile: str):  # type: ignore (not accessed)
+        """Pull Model from a yatai server. If model_tag is not provided,
+        it will pull models defined in bentofile.yaml.
+        """
+        if model_tag is not None:
+            cloud_client.pull_model(model_tag, force=force, context=context)
+            return
+
```
larme, line 270: should we assert that user not providing both model_tag and bentofile (by using `-f bentofile.yaml`? or warn them if they provide both, we will just download model_tag

tests/unit/_internal/service_loader/bento_with_models/service.py
```diff
@@ -0,0 +1,9 @@
+import bentoml
+
+
+test_model = bentoml.models.get("testmodel")
+# get by alias
+another_model = bentoml.models.get("another")
+svc = bentoml.Service(
+    "test-bento-service-with-models", models=[test_model, another_model]
+)
```
larme, line 9: Do we need another service file that explicitly have no model at all? Then when we load the bento, we should still have one model because the modelspec in bentofile.yaml

larme (changes requested):
added some feedback

src/bentoml/_internal/bento/bento.py
frostming: Do you mean adding symlinks into the exported tarball? I think the exported product should be self-contained and can start a bento without cloud access

src/bentoml/_internal/bento/bento.py
```diff
@@ -198,19 +200,22 @@ def create(
...
-        # Add all models required by the service
-        for model in svc.models:
-            models.add(model)
-        # Add all models required by service runners
-        for runner in svc.runners:
-            for model in runner.models:
+        resolved_aliases: t.Dict[Tag, str] = {}
+        if build_config.models:
+            for model_spec in build_config.models:
+                model = model_store.get(model_spec.tag)
```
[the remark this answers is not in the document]
frostming: We already have this message: https://github.com/bentoml/BentoML/blob/501362511ea9e438e0e64b224fa94226e709ba7e/src/bentoml/_internal/store.py#L146-L149
It should be good enough?

src/bentoml/_internal/bento/bento.py
```diff
@@ -123,7 +124,7 @@ class Bento(StoreItem):
 
     _info: BentoInfo
 
-    _model_store: ModelStore
+    _model_store: t.Optional[ModelStore] = None
```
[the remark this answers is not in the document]
frostming: It might be needed by attrs to correctly evaluate the field types.
Not sure though, will have a test later.

src/bentoml/_internal/bento/build_config.py
frostming: By design it is the same as the "filter" input on BentoCloud, where you can connect multiple filters by ","

tests/unit/_internal/service_loader/bento_with_models/service.py
frostming: No, this is only for testing service loading. The said testing point should be covered by https://github.com/bentoml/BentoML/blob/6d0839fe54572270f3842597e827a422b268a2c1/tests/unit/_internal/bento/test_bento.py#L321-L336

src/bentoml/_internal/bento/bento.py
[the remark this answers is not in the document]
aarnphm: I think for Bento it is fine, because we don't serialize it to file.

We only need to use the normal types for any path that serialize to config file, because of support for <3.10

src/bentoml/_internal/bento/bento.py
larme: I think the current implementation will copy model to the existing bento's `models/` and then export the bento.

But if you make symlinks to models, the `Exportable`'s export function will still resolve the symlink to the real file and package them inside the bento

src/bentoml/_internal/bento/build_config.py
larme: that's great

tests/unit/_internal/service_loader/bento_with_models/service.py
larme: got it

src/bentoml/_internal/bento/bento.py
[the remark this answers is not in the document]
larme: got it

src/bentoml/_internal/bento/bento.py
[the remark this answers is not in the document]
frostming: So it is only required for cattr? Will change it.

src/bentoml/_internal/bento/bento.py
frostming: I understand now, but we needs to consider compatibility on Windows.

larme: ye, I think let's do it in another pr.

src/bentoml/_internal/bento/bento.py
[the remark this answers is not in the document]
aarnphm: yes

larme (approved):
good works!

aarnphm (approved):
I guess that documentation will follow up in another PR @Sherlock113?

Sherlock113:
I think the docs was already added in this pr https://github.com/bentoml/BentoML/pull/3971. @aarnphm

PR #3964, opened 2023-06-15
feat(bentocloud): implement deployment CLI

Haivilo:
## What does this PR address?
It implements the CLI interface for bentocloud deployment.

Fixes #(issue)

## Before submitting:

### DIFF
--- a/src/bentoml/_internal/cloud/client.py
+++ b/src/bentoml/_internal/cloud/client.py
@@ -29,6 +29,8 @@ from .schemas import BentoWithRepositoryListSchema
 from .schemas import CompleteMultipartUploadSchema
 from .schemas import ModelWithRepositoryListSchema
 from .schemas import PreSignMultipartUploadUrlSchema
+from .schemas import ClusterListSchema
+from .schemas import ClusterFullSchema
 from ...exceptions import YataiRESTApiClientError
 from ..configuration import BENTOML_VERSION
 
@@ -418,7 +420,9 @@ class YataiRESTApiClient:
         self._check_resp(resp)
         return schema_from_json(resp.text, ModelWithRepositoryListSchema)
 
-    def get_deployment_list(self, cluster_name: str, params: dict[str,str|int] | None = None) -> DeploymentListSchema | None:
+    def get_deployment_list(
+        self, cluster_name: str, params: dict[str, str | int] | None = None
+    ) -> DeploymentListSchema | None:
         url = urljoin(self.endpoint, f"/api/v1/clusters/{cluster_name}/deployments")
         resp = self.session.get(url, params=params)
         if self._is_not_found(resp):
@@ -489,3 +493,21 @@ class YataiRESTApiClient:
             return None
         self._check_resp(resp)
         return schema_from_json(resp.text, DeploymentSchema)
+
+    def get_cluster_list(
+        self, params: dict[str, str | int] | None = None
+    ) -> ClusterListSchema | None:
+        url = urljoin(self.endpoint, "/api/v1/clusters")
+        resp = self.session.get(url, params=params)
+        if self._is_not_found(resp):
+            return None
+        self._check_resp(resp)
+        return schema_from_json(resp.text, ClusterListSchema)
+
+    def get_cluster(self, cluster_name: str) -> ClusterFullSchema | None:
+        url = urljoin(self.endpoint, f"/api/v1/clusters/{cluster_name}")
+        resp = self.session.get(url)
+        if self._is_not_found(resp):
+            return None
+        self._check_resp(resp)
+        return schema_from_json(resp.text, ClusterFullSchema)
--- a/src/bentoml/_internal/cloud/client.py
+++ b/src/bentoml/_internal/cloud/client.py
@@ -49,7 +49,11 @@ class YataiRESTApiClient:
 
     def _is_not_found(self, resp: requests.Response) -> bool:
         # Forgive me, I don't know how to map the error returned by gorm to juju/errors
-        return resp.status_code == 400 and "record not found" in resp.text
+        return (
+            resp.status_code == 404
+            or resp.status_code == 400
+            and "record not found" in resp.text
+        )
 
     def _check_resp(self, resp: requests.Response) -> None:
         if resp.status_code != 200:
@@ -414,9 +418,9 @@ class YataiRESTApiClient:
         self._check_resp(resp)
         return schema_from_json(resp.text, ModelWithRepositoryListSchema)
 
-    def get_deployment_list(self, cluster_name: str) -> DeploymentListSchema | None:
+    def get_deployment_list(self, cluster_name: str, params: dict[str,str|int] | None = None) -> DeploymentListSchema | None:
         url = urljoin(self.endpoint, f"/api/v1/clusters/{cluster_name}/deployments")
-        resp = self.session.get(url)
+        resp = self.session.get(url, params=params)
         if self._is_not_found(resp):
             return None
         self._check_resp(resp)
--- a/src/bentoml/_internal/cloud/config.py
+++ b/src/bentoml/_internal/cloud/config.py
@@ -18,8 +18,6 @@ from ..configuration.containers import BentoMLContainer
 logger = logging.getLogger(__name__)
 
 default_context_name = "default"
-default_cluster_name = "default"
-default_kube_namespace = "yatai"
 
 
 @inject
--- a/src/bentoml/_internal/cloud/deployment.py
+++ b/src/bentoml/_internal/cloud/deployment.py
@@ -46,7 +46,13 @@ class Resource:
 
     @classmethod
     def for_runner(cls, **kwargs) -> DeploymentTargetRunnerConfig:
-        return bentoml_cattr.structure(kwargs, DeploymentTargetRunnerConfig)
+        exclusive_api_server_key = {
+            v for v in kwargs if v not in attr.fields_dict(DeploymentTargetRunnerConfig)
+        }
+        return bentoml_cattr.structure(
+            {k: v for k, v in kwargs.items() if k not in exclusive_api_server_key},
+            DeploymentTargetRunnerConfig,
+        )
 
     @classmethod
     def for_api_server(cls, **kwargs) -> DeploymentTargetConfig:
@@ -234,7 +240,7 @@ class Deployment:
             ],
         }
 
-        cls._update_deployment(
+        return cls._update_deployment(
             deployment_name,
             kube_namespace,
             bentoml_cattr.structure(dct_update, UpdateDeploymentSchema),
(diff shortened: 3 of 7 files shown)


src/bentoml/_internal/cloud/client.py
```diff
@@ -49,7 +49,11 @@ def __init__(self, endpoint: str, api_token: str) -> None:
 
     def _is_not_found(self, resp: requests.Response) -> bool:
         # Forgive me, I don't know how to map the error returned by gorm to juju/errors
-        return resp.status_code == 400 and "record not found" in resp.text
+        return (
+            resp.status_code == 404
+            or resp.status_code == 400
```
aarnphm, line 54: Why are we return of it is 404?


Ah I was always meant to submit a PR for error improvement on this end. Eh maybe when we refactor cloud_rest_client

src/bentoml/_internal/cloud/client.py
```diff
@@ -414,9 +418,9 @@ def get_models_list(self) -> ModelWithRepositoryListSchema | None:
         self._check_resp(resp)
         return schema_from_json(resp.text, ModelWithRepositoryListSchema)
 
-    def get_deployment_list(self, cluster_name: str) -> DeploymentListSchema | None:
+    def get_deployment_list(self, cluster_name: str, params: dict[str,str|int] | None = None) -> DeploymentListSchema | None:
```
aarnphm, line 421: make to run `pre-commit run --all`

src/bentoml/_internal/cloud/deployment.py
```diff
@@ -45,13 +46,7 @@ def for_hpa_conf(cls, **kwargs) -> DeploymentTargetHPAConf:
 
     @classmethod
     def for_runner(cls, **kwargs) -> DeploymentTargetRunnerConfig:
-        exclusive_api_server_key = {
-            v for v in kwargs if v not in attr.fields_dict(DeploymentTargetRunnerConfig)
-        }
-        return bentoml_cattr.structure(
-            {k: v for k, v in kwargs.items() if k not in exclusive_api_server_key},
-            DeploymentTargetRunnerConfig,
-        )
```
aarnphm, line 54: Why are we removing this?

src/bentoml/_internal/cloud/deployment.py
```diff
@@ -149,7 +144,15 @@ def update(
...
-    ):
+    ) -> DeploymentSchema:
+        if mode is None:
+            mode = DeploymentMode.Function
+        if type is None:
+            type = DeploymentTargetType.STABLE
+        if kube_namespace is None:
+            kube_namespace = default_kube_namespace
+        if cluster_name is None:
+            cluster_name = default_cluster_name
```
aarnphm, line 155: use first_not_none(value, default=default)

src/bentoml_cli/deployment.py
```diff
@@ -0,0 +1,476 @@
...
+        if "," in v:
+            parsed += tuple(v.split(","))
+        else:
+            parsed += tuple(v.split())
+        parsedLs.append(
+            dict(map(lambda x: x.strip()[1:-1].split("="), filter(lambda x: x, parsed)))
+        )
+        # Remove trailing space then single quotation mark in front and end, split by = to make a key,val pair
+        # --runner a b c d
+        # ("'a'", "'b'")
```
aarnphm, line 88:
```suggestion
```

src/bentoml_cli/deployment.py
```diff
@@ -0,0 +1,476 @@
...
+    return parsedLs
+
+
+def add_deployment_command(cli: click.Group) -> None:
+    from bentoml_cli.utils import BentoMLCommandGroup
+    from bentoml._internal.configuration.containers import BentoMLContainer
+
+    client = BentoMLContainer.bentocloud_client.get()
```
aarnphm, line 96:
```suggestion
    client = BentoMLContainer.bentocloud_client.get()
   
   	output = click.option("-o", '--output', help='display the output of this command', type=click.Choice(['json', 'porcelain']))
```

aarnphm: Then you can use this decorator for all of the command to process it.

src/bentoml/_internal/cloud/client.py
Haivilo: I found that now when the model or deployment doesn't exist, it returns 404.

frostming: This is changed by me, we have fixed the error handling of BentoCloud that correctly returns 404 for not found.

src/bentoml_cli/deployment.py
```diff
@@ -0,0 +1,476 @@
...
+        "--runner",
+        cls=NargsOptions,
+        multiple=True,
+        callback=parse_runners_callback,
+        default=None,
+        help="Runners.",
+    )
+    def create(  # type: ignore
+        deployment_name: str,
+        bento: str,
```
aarnphm, line 177:
```suggestion
        bento: str,
        output: t.Literal['default', 'json']
```

src/bentoml/_internal/cloud/deployment.py
aarnphm: I guess we set forbid_extra_keys to False
we should just set forbid_extra_keys to true

src/bentoml_cli/deployment.py
```diff
@@ -83,41 +85,67 @@ def parse_runners_callback(
         # ("'a'", "'b'")
     return parsedLs
 
+
 class Mutex(click.Option):
```
aarnphm, line 89: It is not really Mutex option anymore is it? I think this is RequiredAllMutexOption

src/bentoml_cli/deployment.py
```diff
@@ -83,41 +85,67 @@ def parse_runners_callback(
...
+        options = [click.argument(
+        "deployment_name",
+        type=click.STRING,
+        required=True,
+    ),
+    click.option(
+        "--context", type=click.STRING, default=None, help="Yatai context name."
+    ),
+    click.option(
+        "--cluster-name", type=click.STRING, default=None, help="Name of the cluster."
```
aarnphm, line 136: run formatter please 😄

src/bentoml_cli/deployment.py
```diff
@@ -127,23 +155,35 @@ def deployment_cli():
...
+            "resource_instance",
+            "min_replicas",
+            "max_replicas",
+            "type",
+            "mode",
+            "runner",
+            "force",
+            "threads",
+            "push"
+        ],
```
aarnphm, line 178: I still think having to implement this as a eager feature works the best. This is not ideal imho

src/bentoml_cli/deployment.py
Haivilo: It differs a bit imo, here `[--deployment-name, --bento]` and `[--file]` are required & mutually exclusive, but other options except `--output` are forbidden too when using `--file`. but they are optional

src/bentoml_cli/deployment.py
Haivilo: The issue with eager is that it is unable to check if `--file` and `--deployment_name` are provided together (we only want one to be provided). I will check if we can add that check somehow with eager.

src/bentoml_cli/deployment.py
aarnphm: so if users pass in --file it should be executed immediately right?

aarnphm: in terms of setting context and the other variable for create from file we can tell user with the --file workflow to use environment variable to set for that variable specifically

src/bentoml_cli/deployment.py
aarnphm: if --file is passed then it ignore everything else if it is an eager option right? we confirmed that behaviour the other day?

Haivilo: Yes that's right, but eager is processed first so it cannot disallow commands such as `bentoml deployment create -f create.json --deployment-name test-1`. However `--deployment-name` here is omitted. Not sure if we want to forbid extra commands when `-f/--file` is present? Currently it'll give a warning of `Error: Illegal usage: 'file' is mutually exclusive with 'deployment_name'` in that situation.

src/bentoml_cli/deployment.py
Haivilo: `--output` and `--context` are valid options when using `--file`. I agree `context` could be stored as an environment variable, but `--output` might be better to be passed in?...

src/bentoml_cli/deployment.py
aarnphm: What I'm saying is in the case of `--file` user need to put into the environment variable for deployment-name and cluster-name

aarnphm:
```bash
BENTOCLOUD_DEPLOYMENT_NAME=... BENTOCLOUD_CLUSTER_NAME=... bentoml deploy -f create.json
```

src/bentoml_cli/deployment.py
```diff
@@ -0,0 +1,637 @@
...
+        type: DeploymentTargetType,
+        mode: DeploymentMode,
+        runner: dict[str, t.Any],
+        output: t.Literal["json", "default"],
+        force: bool,
+        threads: int,
+        push: bool,
+    ) -> DeploymentSchema:
+        """Create a deployment."""
+        if file is not None:
```
aarnphm, line 308: add a _echo here just sayiing that --file is passed, then only context and cluster_name will be get from environment variable

src/bentoml_cli/deployment.py
```diff
@@ -0,0 +1,637 @@
...
+        type: DeploymentTargetType,
+        mode: DeploymentMode,
+        runner: dict[str, t.Any],
+        output: t.Literal["json", "default"],
+    ) -> DeploymentSchema:
+        """Update a deployment"""
+        if file is not None:
+            return client.deployment.update_from_file(
+                path=file, context=context, cluster_name=cluster_name
+            )
```
aarnphm, line 467:
```suggestion
            return client.deployment.update_from_file(path_or_stream=file)
```

src/bentoml_cli/deployment.py
```diff
@@ -0,0 +1,637 @@
...
+
+    def shared_decorator(f: t.Callable[..., t.Any]):
+        options = [
+            click.argument(
+                "deployment_name",
+                type=click.STRING,
+                required=True,
+            ),
+            click.option(
+                "--context", type=click.STRING, default=None, help="Yatai context name."
```
aarnphm, line 150:
```suggestion
                "--context", type=click.STRING, default=None, help="Yatai context name.", envvar=...
```

src/bentoml/_internal/cloud/config.py
```diff
@@ -18,8 +18,6 @@
 logger = logging.getLogger(__name__)
 
 default_context_name = "default"
-default_cluster_name = "default"
-default_kube_namespace = "yatai"
```
aarnphm, line 22: why are we removintg this?

src/bentoml_cli/deployment.py
```diff
@@ -0,0 +1,248 @@
...
+        ]
+        for opt in reversed(options):
+            f = opt(f)
+        return f
+
+    @cli.group(name="deployment", cls=BentoMLCommandGroup)
+    def deployment_cli():
+        """Deployment Subcommands Groups"""
+
+    @deployment_cli.command(context_settings={"ignore_unknown_options": True})
```
aarnphm, line 61:
```suggestion
    @deployment_cli.command()
```

src/bentoml_cli/deployment.py
```diff
@@ -0,0 +1,248 @@
...
+        elif output == "json":
+            click.echo(json.dumps(bentoml_cattr.unstructure(res), indent=2))
+        return res
+
+    @deployment_cli.command()
+    @click.option(
+        "-f",
+        "--file",
+        type=click.File(),
+        required=True,
```
aarnphm, line 91:
```suggestion
```

aarnphm (approved):

PR #4002, opened 2023-06-28
chore: deprecate 3.7 EOL

aarnphm:

## What does this PR address?

See https://devguide.python.org/versions/

Fixes #(issue)

## Before submitting:


sauyon:
If you're going to have real changes along with format changes can you split them into different commits for quicker review?

src/bentoml/_internal/server/http_app.py
```diff
@@ -387,9 +387,6 @@ async def api_func(request: Request) -> Response:
                         )
                     else:
                         response = JSONResponse("", status_code=status)
-                except asyncio.CancelledError:
-                    # Special handling for Python 3.7 compatibility
-                    raise
```
sauyon, line 392: Hm, I don't remember but this feels like it's necessary even now, @frostming do you remember why this is 3.7 only?

frostming: No, we can remove all `except CancelledError` now. Because on Python 3.7 CancelledError subclasses `Exception`, which will unexpected caught by the general `except Exception` block. It is changed to subclass `BaseException` on Python 3.8, so we don't need to bother now.

frostming: There are some occurrences in other files, @aarnphm remove them as well.

sauyon: I see, yeah, we should remove them all @aarnphm.

aarnphm:
cc @frostming @sauyon updated now. PTAL at the latest commit to see if it is correct

frostming (approved):

sauyon (approved):
👍

PR #3993, opened 2023-06-26
fix(scheduling): expose Strategy class

bojiang:

PR #4003, opened 2023-06-28
ci: Sort imports with isort

frostming:

## What does this PR address?

Introduce isort as the formatter to fix the issue that ruff doesn't sort imports correctly(#3993)


.pre-commit-config.yaml
```diff
@@ -2,11 +2,15 @@ ci:
   autoupdate_schedule: monthly
 exclude: '.*\.(css|js|svg)$'
 repos:
-  - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: 'v0.0.274'
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.0.275
```
aarnphm, line 6: @frostming I think there is something wrong with ruff, because I have the same config for OpenLLM and ruff is correctly respect the single line imports when running within hatch and pre-commit

frostming: Agree, it should have something to do with the parser. Ruff's parser is borrowed from RustPython, which may have different behaviors from CPython's.

When the parser fails, it quits silently

aarnphm: hmm. Not to strongly on this, we can add isort back for now.

aarnphm:
@frostming we need to do this https://beta.ruff.rs/docs/settings/#select

```toml

[tool.ruff]
extend-select = ["I"]
```

frostming:
> @frostming we need to do this https://beta.ruff.rs/docs/settings/#select
> 
> ```toml
> [tool.ruff]
> extend-select = ["I"]
> ```
> 
> See the list of linter with `ruff linter`

Yeah, that works! See #4032

PR #4022, opened 2023-07-06
feat(resource): respecting CUDA_VISIBLE_DEVICES

aarnphm:
Signed-off-by: aarnphm-ec2-dev <29749331+aarnphm@users.noreply.github.com>

## What does this PR address?

Fixes #(issue)

## Before submitting:

### DIFF
--- a/src/bentoml/_internal/resource.py
+++ b/src/bentoml/_internal/resource.py
@@ -211,7 +211,7 @@ class NvidiaGpuResource(Resource[t.List[str]], resource_id="nvidia.com/gpu"):
                     if spec.startswith("GPU"):
                         return [spec]
                     if "," in spec:
-                        return [i.strip() for i in spec.split(",")]
+                        return cls.from_spec([int(i.strip()) for i in spec.split(",")])
                     raise ValueError
             else:
                 return [str(x) for x in spec]
--- a/src/bentoml/_internal/resource.py
+++ b/src/bentoml/_internal/resource.py
@@ -205,10 +205,13 @@ class NvidiaGpuResource(Resource[t.List[str]], resource_id="nvidia.com/gpu"):
                 return [str(i) for i in range(spec)]
             elif isinstance(spec, str):
                 try:
-                    return cls.from_spec(int(spec))
+                    resolved = int(spec)
+                    return cls.from_spec(resolved)
                 except ValueError:
                     if spec.startswith("GPU"):
                         return [spec]
+                    if "," in spec:
+                        return [i.strip() for i in spec.split(",")]
                     raise ValueError
             else:
                 return [str(x) for x in spec]
--- a/src/bentoml/_internal/resource.py
+++ b/src/bentoml/_internal/resource.py
@@ -12,7 +12,6 @@ from abc import abstractmethod
 import psutil
 
 from ..exceptions import BentoMLConfigException
-from .types import LazyType
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


src/bentoml/_internal/resource.py
```diff
@@ -244,6 +213,13 @@ def from_system(cls) -> list[int]:
         """
         import pynvml  # type: ignore
 
+        cuda_visible_devices = os.environ.get("CUDA_VISIBLE_DEVICES", None)
+        if cuda_visible_devices is not None:
+            if "," in cuda_visible_devices:
+                return [int(i) for i in cuda_visible_devices.split(",")]
+            else:
```
sauyon, line 220: Seems like the spec says comma separated (https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#env-vars); why are we supporting space?

tests/unit/_internal/test_resource.py
```diff
@@ -63,6 +63,14 @@ def test_CpuResource():
         CpuResource.from_spec((1, 2, 3))
 
 
+@pytest.mark.parametrize("env", ["0, 1", "0,1", "0 1"])
```
sauyon, line 66: Maybe add a test for an empty value?

sauyon (commented):
Seems like you haven't implemented `-1` (see https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#env-vars), we should support this properly.

Also, int casting is probably incorrect, we should allow strings, as per the spec.

tests/unit/_internal/test_resource.py
aarnphm: sg

src/bentoml/_internal/runner/strategy.py
```diff
@@ -150,7 +151,12 @@ def get_worker_env(
                 ]
                 dev = ",".join(map(str, assigned_gpu))
             else:
-                dev = str(nvidia_gpus[worker_index // workers_per_resource])
+                idx = worker_index // workers_per_resource
+                if len(nvidia_gpus) == idx:
```
frostming, line 155:
```suggestion
                if idx >= len(nvidia_gpus):
```

src/bentoml/_internal/resource.py
```diff
@@ -183,120 +188,92 @@ def query_os_cpu_count() -> int:
...
+                if spec == -1:
+                    return []
                 if spec < -1:
                     raise ValueError
-                return list(range(spec))
+                return [str(i) for i in range(spec)]
             elif isinstance(spec, str):
-                return cls.from_spec(int(spec))
+                try:
+                    return cls.from_spec(int(spec))
```
frostming, line 208: Better to convert str before evaluation to avoid recursive call.

aarnphm: Now thinking a bit more, I don't think it wouldn't recurse unless the spec is a valid int right?

aarnphm: Actually this would fail with `0,1,2`

src/bentoml/_internal/resource.py
```diff
@@ -183,120 +188,95 @@ def query_os_cpu_count() -> int:
...
-    def from_spec(cls, spec: int | str | list[int | str]) -> list[int]:
+    def from_spec(cls, spec: int | str | list[int | str]) -> list[str]:
         if not isinstance(spec, (int, str, list)):
             raise TypeError(
                 "NVidia GPU device IDs must be int, str or a list specifing the exact GPUs to use."
             )
 
         try:
             if isinstance(spec, int):
+                if spec == -1:
```
sauyon, line 201: Actually, does `pynvml` not have support for parsing `CUDA_VISIBLE_DEVICES`? Maybe we can just use that?

src/bentoml/_internal/resource.py
```diff
@@ -183,120 +188,95 @@ def query_os_cpu_count() -> int:
...
                     raise ValueError
-                return list(range(spec))
+                return [str(i) for i in range(spec)]
             elif isinstance(spec, str):
-                return cls.from_spec(int(spec))
+                try:
+                    resolved = int(spec)
+                    return cls.from_spec(resolved)
+                except ValueError:
+                    if spec.startswith("GPU"):
```
sauyon, line 211: I believe we should also support `MIG` strings. Let's maybe limit parsing to as little as possible so we don't have to keep up with changes in the Nvidia libraries.

src/bentoml/_internal/resource.py
```diff
@@ -183,120 +188,95 @@ def query_os_cpu_count() -> int:
...
+                        return [spec]
+                    if "," in spec:
+                        return [i.strip() for i in spec.split(",")]
+                    raise ValueError
```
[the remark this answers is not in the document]
aarnphm: We are catching it below right? I guess that was what we are doing b4. I can raise more specific error for this then.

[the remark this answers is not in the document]
sauyon: Oh, no need. Can just reraise in this case, I think it should make full traces look better anyway.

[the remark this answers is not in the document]
sauyon:
```suggestion
                    raise
```

src/bentoml/_internal/resource.py
bojiang, line 1: @aarnphm Shall we add some unit tests for this CUDA_VISIBLE_DEVICES resolver?
I guess you already know this doc,
https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#cuda-environment-variables
maybe we should add test cases in this doc

GrigoryEvko:
Guys how this wasn't implemented? How do I run exactly one service per one gpu? bentoml doesn't respect env variables, I can't configure it inside bentoml service? so how??

PR #4028, opened 2023-07-08
docs: Save models via `import_model.py`

hi-rustin:
## What does this PR address?

To add a better description to the import model section, I am a user who is not familiar with Python and machine learning, so I don't actually know how that code would be executed. I found some instructions in your other documentation. So I thought it would be better to add it here.
I also made a few minor tweaks:
1. state the Python version description, which I also found in your other documentation.
2. we should not use $ for bash blocks that have no output. You can check the [MD014/commands-show-ouput](https://github.com/updownpress/markdown-lint/blob/master/rules/014-commands-show-output.md) rule. Also, some of your places actually obey this rule, so I think it would be better to be uniform.

## Before submitting:

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ packaging, and production deployment.</p>
 
 # Documentation
 
-- Installation: `pip install bentoml`
+- Installation: `pip install bentoml` (Requires Python 3.8+)
 - Full Documentation: [docs.bentoml.com](https://docs.bentoml.com/en/latest/)
 - Tutorial: [Intro to BentoML](https://docs.bentoml.com/en/latest/tutorial.html)
 
@@ -92,7 +92,7 @@ packaging, and production deployment.</p>
 
 # Getting Started
 
-Save or import models in BentoML local model store:
+Save or import models in BentoML local model via `import_model.py`:
 
 ```python
 import bentoml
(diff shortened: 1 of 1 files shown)


aarnphm (changes requested):
This is not technically a file, you can python this into a python repl and it should work. You don't need a file here.

Also this section in the README.md are more for demonstration.

aarnphm:
I'm going to close this because it is not relevant for README.md 😃

hi-rustin:
> This is not technically a file, you can python this into a python repl and it should work. You don't need a file here.

Thank you for the explanation. Should we put it in the documentation? As a golang and rust developer I even forgot that python has a repl feature. 😆 


> I'm going to close this because it is not relevant for README.md 😃

It also fixes a markdown warning and adds a python version requirement.

> Also this section in the README.md are more for demonstration.

But it may be the first step for some users. That's why I think it's important.

Thanks for your review!  💚 💙 💜 💛 ❤️

aarnphm:
Let's open a different PR for it

hi-rustin:
> Let's open a different PR for it

Why do we need to reopen a new pull request? I can change it in this PR.

hi-rustin:
Actually, use the `import_model.py` instruction that comes from your [document](https://docs.bentoml.org/en/latest/frameworks/diffusers.html#importing-a-pre-trained-model).

PR #4032, opened 2023-07-10
ci: add isort to the ruff lint rules

frostming:

## What does this PR address?

Adding `extend-select` actually makes it work!
But since `ruff` doesn't support sort imports by length, changes are expected.

Supersedes #4003 

@aarnphm


aarnphm (approved):
Can you add the changes for diffusers here as well?

aarnphm (approved):

PR #4038, opened 2023-07-12
fix: correct the logic of delete model

frostming:

## What does this PR address?

Previously, when the model version is missing, model_store.delete() will delete the **whole** `$model_name` directory. And if the version is `latest`, it will raise an error since `$model_name/latest` isn't a directory.

Fix this to resolve to the actual version to get the path correctly.

## Before submitting:

### DIFF
--- a/src/bentoml/_internal/cloud/yatai.py
+++ b/src/bentoml/_internal/cloud/yatai.py
@@ -873,19 +873,21 @@ class YataiClient(CloudClient):
         context: str | None = None,
         query: str | None = None,
     ) -> Model:
-        try:
-            model = model_store.get(tag)
-            if not force:
-                self.log_progress.add_task(
-                    f'[bold blue]Model "{tag}" already exists locally, skipping'
-                )
-                return model
-            else:
-                model_store.delete(tag)
-        except NotFound:
-            pass
-        yatai_rest_client = get_rest_api_client(context)
         _tag = Tag.from_taglike(tag)
+        try:
+            model = model_store.get(_tag)
+        except NotFound:
+            model = None
+        else:
+            if _tag.version not in (None, "latest"):
+                if not force:
+                    self.log_progress.add_task(
+                        f'[bold blue]Model "{tag}" already exists locally, skipping'
+                    )
+                    return model
+                else:
+                    model_store.delete(tag)
+        yatai_rest_client = get_rest_api_client(context)
         name = _tag.name
         version = _tag.version
         if version in (None, "latest"):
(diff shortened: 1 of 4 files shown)


src/bentoml_cli/models.py
```diff
@@ -254,7 +254,6 @@ def import_from(model_path: str) -> None:  # type: ignore (not accessed)
         "--context", type=click.STRING, default=None, help="Yatai context name."
     )
     @click.option(
-        "-f",
```
frostming, line 257: Conflicting short option name with `--force`

sauyon: Let's just rename this to `-F`?

src/bentoml/_internal/store.py
```diff
@@ -182,6 +182,13 @@ def register(self, tag: t.Union[str, Tag]):
...
         _tag = Tag.from_taglike(tag)
 
+        if _tag.version in (None, "latest"):
+            try:
+                _tag.version = self._fs.readtext(_tag.latest_path())
+            except fs.errors.ResourceNotFound:
+                # if latest path doesn't exist, we don't need to delete anything
+                return
+
         if not self._fs.exists(_tag.path()):
```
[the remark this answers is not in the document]
frostming: I thought it a bit further, we don't need to delete the latest version, but allow to update the latest version from the remote store.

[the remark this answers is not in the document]
aarnphm: I see. that makes sense as well.

src/bentoml/_internal/cloud/yatai.py
```diff
@@ -873,19 +873,21 @@ def _do_pull_model(
...
+            model = None
+        else:
+            if _tag.version not in (None, "latest"):
+                if not force:
+                    self.log_progress.add_task(
+                        f'[bold blue]Model "{tag}" already exists locally, skipping'
+                    )
+                    return model
+                else:
+                    model_store.delete(tag)
```
aarnphm, line 889: You will need to also add this to `_internal/cloud/bentocloud.py` as well 😄

frostming: Oh yes, may be we could make it DRYer in the future.

aarnphm: I think the idea for yatai.py and bentocloud.py is to have different implementation, but currently the bentocloud push for both model and bento are the same with yatai

aarnphm (approved):

aarnphm:
I Think we should have an option with `bentoml delete --all` or `bentoml prune` to just remove all the bentos and models available locally

PR #4078, opened 2023-07-27
feat(model): skip creating tempfs when saving

aarnphm:

## What does this PR address?

This PR aims to solve a few pain points:
- When saving large model, such as any of the LLM usecase, it is probably best to just save it directly to the model store, and skip copying from temp to the actual store itself.
- When users send a KeyboardInterrupt, the store should cleanly remove all items within the folder to ensure `validate` works, otherwise it will leave with dangling incomplete `bentoml.Model` object.

## Before submitting:

### DIFF
--- a/src/bentoml/_internal/bento/bento.py
+++ b/src/bentoml/_internal/bento/bento.py
@@ -119,42 +119,36 @@ This is a Machine Learning Service created with BentoML."""
     return doc
 
 
-@attr.define(repr=False, auto_attribs=False)
+@attr.define(repr=False, init=False)
 class Bento(StoreItem):
-    _tag: Tag = attr.field()
-    __fs: FS = attr.field()
-
+    _tag: Tag
+    _fs: FS
     _info: BentoInfo
+    _doc: str | None = attr.field(default=None)
+    _model_store: ModelStore | None = attr.field(default=None)
 
-    _model_store: ModelStore | None = None
-    _doc: str | None = None
+    if t.TYPE_CHECKING:
+
+        def __attrs_init__(
+            self, tag: Tag, bento_fs: FS, info: BentoInfo, doc: str | None
+        ) -> None:
+            ...
+
+    def __init__(self, tag: Tag, bento_fs: FS, info: BentoInfo, doc: str | None = None):
+        self.__attrs_init__(tag, bento_fs, info, doc)
+        # old behaviour
+        try:
+            self._model_store = ModelStore(bento_fs.opendir("models"))
+        except fs.errors.ResourceNotFound:
+            pass
 
     @staticmethod
     def _export_ext() -> str:
         return "bento"
 
-    @__fs.validator  # type:ignore # attrs validators not supported by pyright
-    def check_fs(self, _attr: t.Any, new_fs: FS):
-        try:
-            models = new_fs.opendir("models")
-        except fs.errors.ResourceNotFound:
-            return
-        else:
-            self._model_store = ModelStore(models)
-
-    def __init__(self, tag: Tag, bento_fs: "FS", info: "BentoInfo"):
-        self._tag = tag
-        self.__fs = bento_fs
-        self.check_fs(None, bento_fs)
-        self._info = info
-
     @property
     def tag(self) -> Tag:
         return self._tag
-
-    @property
-    def _fs(self) -> FS:
-        return self.__fs
 
     @property
     def info(self) -> BentoInfo:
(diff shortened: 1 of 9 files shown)


src/bentoml/_internal/models/model.py
```diff
@@ -178,7 +199,17 @@ def create(
...
+            model_fs = fs.open_fs(f"temp://bentoml_model_{tag.name}")
+        else:
+            if _model_store.exists(tag):
+                raise ItemAlreadyExists(
+                    f"Item '{tag}' already exists in the store {_model_store}"
+                )
+            model_fs = fs.open_fs(
+                _model_store.path_of(tag.path()), create=True, writeable=True
+            )
+            _model_store.update_latest(tag)
```
frostming, line 212: This can be moved to `save()`?

src/bentoml/_internal/store.py
```diff
@@ -159,25 +160,40 @@ def get(self, tag: t.Union[Tag, str]) -> Item:
...
         self._fs.makedirs(item_path)
         try:
-            yield self._fs.getsyspath(item_path)
-        finally:
+            yield self.path_of(item_path)
+        except KeyboardInterrupt:
+            # During save, if user hit Ctrl+C, we will remove the item path cleanly
+            # so that validate() will work correctly.
+            self._fs.removetree(item_path)
+            raise SystemExit(1)
```
frostming, line 178: Simply reraise here.

src/bentoml/models.py
```diff
@@ -255,12 +256,14 @@ def create(
...
         context=context,
+        use_tempfs=use_tempfs,
     )
     external_modules = [] if external_modules is None else external_modules
-    imported_modules = []
+    imported_modules: t.List[ModuleType] = []
     try:
         res.enter_cloudpickle_context(external_modules, imported_modules)
         yield res
+        res.exit_cloudpickle_context(imported_modules)
```
frostming, line 266: Why move it here from `finally`?

aarnphm: it is the same as finally right?

aarnphm:
@frostming PTAL

src/bentoml/_internal/store.py
```diff
@@ -159,25 +160,40 @@ def get(self, tag: t.Union[Tag, str]) -> Item:
...
             )
         self._fs.makedirs(item_path)
         try:
-            yield self._fs.getsyspath(item_path)
-        finally:
+            yield self.path_of(item_path)
+        except KeyboardInterrupt:
+            # During save, if user hit Ctrl+C, we will remove the item path cleanly
+            # so that validate() will work correctly.
+            self._fs.removetree(item_path)
```
frostming, line 177: If I don't miss anything the clear up is only required for `use_temp_fs=False`, but the `register()` isn't on the path of `use_temp_fs=False` right?

It should be moved here instead: https://github.com/bentoml/BentoML/blob/c794024eef895405fbfe53dea5e9b129ed60b5b3/src/bentoml/models.py#L312-L313

frostming: Fine, then moving to bentoml.models handles both.

aarnphm: Oh you are correct, let me do that.

src/bentoml/_internal/bento/bento.py
```diff
@@ -330,7 +324,7 @@ def export(
         subpath: t.Optional[str] = None,
     ) -> str:
         # copy the models to fs
-        models_dir = self._fs.makedir("models", recreate=True)
+        models_dir = self.fs.makedir("models", recreate=True)
```
sauyon, line 327: I'm not sure we should expose `fs` itself. Maybe we can write some abstractions?

src/bentoml/_internal/models/model.py
```diff
@@ -135,18 +145,21 @@ def __hash__(self) -> int:
...
         module: str,
         api_version: str,
         signatures: ModelSignaturesType,
+        context: ModelContext,
         labels: dict[str, str] | None = None,
         options: ModelOptions | None = None,
         custom_objects: dict[str, t.Any] | None = None,
         metadata: dict[str, t.Any] | None = None,
-        context: ModelContext,
+        use_tempfs: bool = True,
```
sauyon, line 161: Honestly, instead of using the tempfs can we do this in a non-memory tmp location? The original reason we use a move at the end was to make the process atomic, and really we should never have used a memory fs for this.

We also don't have to provide this option then.

src/bentoml/_internal/bento/bento.py
aarnphm: I think fs is already itself an abstraction? I think we should expose this fs object.

src/bentoml/_internal/models/model.py
aarnphm: maybe a NamedTemporaryDir?

src/bentoml/_internal/bento/bento.py
sauyon: I'd prefer not to expose external APIs we have no control over if we have a choice, is why. It's also something that would need to be replaced in the potentially soon:tm: rust rewrite.

aarnphm: i see. makes sense then

Issue #3743, opened 2023-04-05
feature: Server-Sent Events (SSE) support in API


ssheng:
### Feature request

SSE is well-suited for delivering a large volume of data that changes frequently, making it an ideal protocol for real-time applications such as chatbots, live monitoring systems, and financial trading platforms. With the growing popularity of LLM, use of SSE allows streaming data in real-time on a long-lived connection on an active session for contexts.

Upvote if this is a must-have feature for you.

PR #4083, opened 2023-07-31
feat: SSE (Experimental)

jianshen92:
## What does this PR address?
This features adds support of SSE into BentoML. This implementation requires user to write a custom Runnable to allow SSE.

Usage:
```python
import bentoml
import asyncio

class SSERunnable(bentoml.Runnable):
    SUPPORTED_RESOURCES = ("cpu",)
    SUPPORTS_CPU_MULTI_THREADING = True

    @bentoml.Runnable.method(stream=True)
    async def count(self, input_text:str):
        for i in range(10):
            await asyncio.sleep(1)
            yield f"{input_text} {i} \n"

sse_runner = bentoml.Runner(SSERunnable)
svc = bentoml.Service("sse", runners=[sse_runner])

@svc.api(stream=True, input=bentoml.io.Text(), output=bentoml.io.TextStream())
async def count(input_text:str):
    ret = sse_runner.count.async_stream(input_text)
    return ret
```
Running
```bash
curl "http://localhost:3000/count" -d "Hello World"
```
outputs
```
Hello World 0
Hello World 1
Hello World 2
Hello World 3
Hello World 4
Hello World 5
Hello World 6
Hello World 7
Hello World 8
Hello World 9
```
Changes:
~1. Add a boolean `stream` argument to `svc.api`~
~2. Add a boolean `stream` argument to `Runnable.method`~
3. Add `async_stream` for Runner method, this returns an Async Generator 
~4. Add `TextStream` for streaming io, currently only supports output~

**UPDATE**: 
Discussion within the team results to
1. Removal of the need of `stream` argument both in `svc.api` and `Runnable.method`
2. Removal the need of a new `TextStream` io

Everything will be done by inspecting whether the methods is a normal function or a generator. BentoML will treat it as a stream if the runner method is a generator.

https://github.com/bentoml/BentoML/issues/3743

---------------------


jianshen92:
@aarnphm @frostming @sauyon @larme PTAL! 
I think there are parts which is not being covered properly yet, i.e. error handling. Please comment rigourously!

src/bentoml/_internal/io_descriptors/text_stream.py
```diff
@@ -0,0 +1,163 @@
...
+    _mime_type = MIME_TYPE
+
+    def __init__(self, *args: t.Any, **kwargs: t.Any):
+        if args or kwargs:
+            raise BentoMLException(
+                f"'{self.__class__.__name__}' is not designed to take any args or kwargs during initialization."
+            ) from None
+
+    def _from_sample(self, sample: str | bytes):
+        raise NotImplementedError("TextStream does not support input type for now")
```
aarnphm, line 98: This can be the same as text. Should be easy to implement.

src/bentoml/_internal/io_descriptors/text_stream.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def input_type(self) -> t.Type[str]:
+        raise NotImplementedError("TextStream does not support input type for now")
```
aarnphm, line 101: We need this for client

```suggestion
    def input_type(self) -> t.Type[str]:
        return str
```

src/bentoml/_internal/io_descriptors/text_stream.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def to_spec(self) -> dict[str, t.Any]:
+        return {"id": self.descriptor_id}
```
aarnphm, line 104: I think there is something worth to put in here to indicate this is streaming, since the client doesn't now whether the endpoint is streaming or not,

src/bentoml/_internal/marshal/dispatcher.py
```diff
@@ -401,3 +401,46 @@ async def outbound_call(self, inputs_info: tuple[Job, ...]):
...
+
+        @functools.wraps(callback)
+        async def _func(data: t.Any) -> t.Any:
+            try:
+                r = await self.inbound_call(data)
+            except asyncio.CancelledError:
+                return None
+            if isinstance(r, Exception):
+                raise r
+            return r
```
aarnphm, line 444:
```suggestion
            try:
                return await self.inbound_call(data)
            except Exception as err:
            	logger.error("Exception caught while sending request:\n%s", err, exc_info=err)
            	raise err from None
```

src/bentoml/_internal/runner/runnable.py
```diff
@@ -101,6 +103,7 @@ def method(
     def method(
         meth: t.Callable[t.Concatenate[T, P], R] | None = None,
         *,
+        stream: bool = False,
```
aarnphm, line 106: fwiw I think we can infer whether the function is streaming or not via asyncgen and generator, since there is no need for this `stream` value.

src/bentoml/_internal/runner/runnable.py
```diff
@@ -153,3 +157,4 @@ class RunnableMethodConfig:
     batch_dim: tuple[int, int]
     input_spec: AnyType | tuple[AnyType, ...] | None = None
     output_spec: AnyType | None = None
+    is_stream: bool = False
```
aarnphm, line 160: same here, this boolean can become a property and can be calculated based on whether this runnablemethod is an asyncgen, gen or not.

src/bentoml/_internal/runner/runner_handle/__init__.py
```diff
@@ -5,6 +5,7 @@
 from abc import ABC
 from abc import abstractmethod
 from typing import TYPE_CHECKING
+from typing import AsyncGenerator
```
aarnphm, line 8: s/AsyncGenerator/t.AsyncGenerator

src/bentoml/_internal/runner/runner_handle/remote.py
```diff
@@ -274,6 +274,71 @@ async def async_run_method(
...
+
+        async def stream_response() -> t.AsyncGenerator[R, None]:
+            """
+            Async generator that yields results from the remote runner.
+            """
+            try:
+                async with self._client.post(
+                    f"{self._addr}/{path}",
+                    data=data,
+                    headers=headers,
```
aarnphm, line 331: iirc the client has `stream=True`, and we probably don't need to iterate through the line manually.

Correct me if I'm wrong, but `stream=True` then the context manager returns an iterator over the payload anw.

src/bentoml/_internal/server/runner_app.py
```diff
@@ -68,11 +69,14 @@ def fallback():
...
-                fallback=fallback,
-            )
+            if method.config.is_stream:
+                self.dispatchers[method.name] = NoOpDispatcher()
+            else:
+                self.dispatchers[method.name] = CorkDispatcher(
+                    max_latency_in_ms=method.max_latency_ms,
+                    max_batch_size=max_batch_size,
+                    fallback=fallback,
+                )
```
aarnphm, line 79: let's also make this a bit more clean. Can we make it similar to how we dispatch the class for runner_handle?

src/bentoml/_internal/service/service.py
```diff
@@ -288,7 +288,10 @@ def is_service_importable(self) -> bool:
     # case 1: function is not defined, but input and output are
     @t.overload
     def api(self, input: IODescriptor[IOType], output: IODescriptor[IOType]) ->  _inference_api_wrapper[IOType]: ...
-    # case 2: the decorator itself with custom routes
+    # case 2: streaming is enabled
+    @t.overload
+    def api(self, input: IODescriptor[IOType], output: IODescriptor[IOType], *, stream: bool) ->  _inference_api_wrapper[IOType]: ...
```
aarnphm, line 293: For correctness
```suggestion
    def api(self, input: IODescriptor[IOType], output: IODescriptor[IOType], *, stream: t.Literal[True]) ->  _inference_api_wrapper[IOType]: ...
```

src/bentoml/io.py
```diff
@@ -10,6 +10,7 @@
 from ._internal.io_descriptors.pandas import PandasDataFrame
 from ._internal.io_descriptors.pandas import PandasSeries
 from ._internal.io_descriptors.text import Text
+from ._internal.io_descriptors.text_stream import TextStream
```
aarnphm, line 13: also let's expose this in `src/bentoml/_internal/io_descriptor/__init__.py` as well for correctness

src/bentoml/_internal/io_descriptors/text_stream.py
jianshen92: Isn't this for input? We don't have streaming input for now

aarnphm: This can be example for output as well right?

src/bentoml/_internal/marshal/dispatcher.py
sauyon: Hm, did `inbound_call` return an exception before? What line 442 for originally?

src/bentoml/_internal/marshal/dispatcher.py
```diff
@@ -401,3 +401,46 @@ async def outbound_call(self, inputs_info: tuple[Job, ...]):
                     if not fut.done():
                         fut.cancel()
             self._sema.release()
+
+
+class NoOpDispatcher:
```
sauyon, line 406: I don't know if this was addressed already, but was there a reason you couldn't use the normal dispatcher with batching disabled?

src/bentoml/_internal/server/runner_app.py
```diff
@@ -244,18 +248,37 @@ async def infer_batch(
 
             infer = self.dispatchers[runner_method.name](infer_batch)
         else:
+            if runner_method.config.is_stream:
```
sauyon, line 251: `elif`?

src/bentoml/_internal/io_descriptors/text_stream.py
sauyon: If we end up doing this, this should be an async generator, right?

aarnphm: oh right this should be async generator

src/bentoml/_internal/runner/runnable.py
jianshen92: Do you think we should remove both the `stream` value for `svc.api` and `runnable.method` ? and infer from the function type?

src/bentoml/_internal/server/runner_app.py
jianshen92: thought that nesting this would make sense, i.e the first part is for batching, second part is for non batch.

src/bentoml/_internal/runner/runnable.py
aarnphm: yes, because it is a generator or async generator then we can assume that it has to be stream. We can tell users that if it is a generator, then it should be using within streaming cases. Ofc you can write generator within sync, but that will be anti-pattern

jianshen92: I agree with this from the user POV, makes the path to streaming as smooth as possible.

src/bentoml/_internal/server/runner_app.py
sauyon: Ah, I see. I think nested ifs can be a little hard to read, but if you feel like this distinction is important, I also feel like it makes more sense for the top-level difference to be streamed / not streamed first, then batched / unbatched.

src/bentoml/_internal/server/runner_app.py
```diff
@@ -244,18 +248,37 @@ async def infer_batch(
...
+                async def infer_stream(
+                    paramss: t.Sequence[Params[t.Any]],
+                ) -> t.AsyncGenerator[Payload, None]:
+                    assert len(paramss) == 1
+
+                    params = paramss[0].map(AutoContainer.from_payload)
+
+                    ret = await runner_method.async_run(*params.args, **params.kwargs)
+
+                    if runner_method.config.is_stream:
```
sauyon, line 262: Is this check necessary?

src/bentoml/_internal/io_descriptors/text_stream.py
```diff
@@ -0,0 +1,163 @@
...
+            res = StreamingResponse(
+                obj,
+                media_type=self._mime_type,
+                headers=ctx.response.metadata,  # type: ignore (bad starlette types)
+                status_code=ctx.response.status_code,
+            )
+            set_cookies(res, ctx.response.cookies)
+            return res
+        else:
+            return StreamingResponse(obj, media_type=self._mime_type)
```
frostming, line 157: For SSE, we don't want a simple TextStream, but indeed an [EventStream](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#event_stream_format), it would be most appropriate to wrap it here. I've spotted this point on the meeting, but it seems we still need to align, cc @aarnphm @sauyon @ssheng

src/bentoml/_internal/runner/runnable.py
frostming: Agree to remove it from `Runnable.method()` and `svc.api()`, but having such field in the config is still valuable, to cache and prevent redundant calculation, compared to a property.

src/bentoml/_internal/io_descriptors/text_stream.py
frostming: I can't execute the request on the swagger page so I don't know it is implemented correctly or not. But theoretically, an valid event stream should be generated with mime type `text/event-stream`

jianshen92: Just making sure that we are not speaking pass each other, you are suggesting that the returned object conform to the SSE data format, i.e. with `event` `data` `id` `retry` fields right? because the mime type is already set to `text/event-stream`

src/bentoml/_internal/marshal/dispatcher.py
jianshen92: I was copying from https://github.com/bentoml/BentoML/blob/main/src/bentoml/_internal/marshal/dispatcher.py#L181

src/bentoml/_internal/marshal/dispatcher.py
jianshen92: It was mainly for convenience without conflating the current implementation, I could give it a try with the `CorkDispatcher`

src/bentoml/_internal/io_descriptors/text_stream.py
sauyon: Is there an issue with simply streaming text? I knowing we want to support "true" SSE we want to events, but for the simple text stream this should be fine, as long as we're not setting the mime incorrectly, right?

(If we are setting the mime to event-stream we should indeed be doing as frost says, though.)

aarnphm: We are setting the mime to event-stream in this case? Even for text, I agree with frost here.

src/bentoml/_internal/runner/runnable.py
aarnphm: +1 on caching this for later use, if necessary.

src/bentoml/_internal/runner/runner.py
```diff
@@ -54,6 +54,9 @@ def run(self, *args: P.args, **kwargs: P.kwargs) -> R:
     async def async_run(self, *args: P.args, **kwargs: P.kwargs) -> R:
         return await self.runner._runner_handle.async_run_method(self, *args, **kwargs)
 
+    def async_stream(self, *args: P.args, **kwargs: P.kwargs) -> R:
```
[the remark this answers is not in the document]
aarnphm: This should be `t.AsynGenerator[R, None]`

src/bentoml/_internal/io_descriptors/text_stream.py
sauyon: I'm not really sure what you mean, I'm saying we can just change the mime type to text/plain and everything is fine as-is.

frostming: > Just making sure that we are not speaking pass each other, you are suggesting that the returned text to conform to the SSE data format, i.e. with `event` `data` `id` `retry` fields right? because the mime type is already set to `text/event-stream`

Yes.

> I'm not really sure what you mean, I'm saying we can just change the mime type to text/plain and everything is fine as-is.

A text stream could be convenient only when user is interacting with the API. However, event stream is the standard protocol of SSE that is supported by main stream browsers, and event stream can carry more information in the event data. A possible solution is to add another type or an option to switch between plain text stream and event stream.

frostming: One more thing to add, for building an event, it should have the flexibility to define the protocol. For example, Open AI API marks the end of stream by a `data: [DONE]` message, others could have different protocol.

jianshen92: Would asking user to define the whole event be pushing too much complexity down to the user? would asking user to produce a generator of string and we wrap it with
`data: {string} \n\n` at the boundary works?

frostming: More references,

[Claude's stream API](https://docs.anthropic.com/claude/reference/streaming#example) marks the end of stream with a non-empty `stop_reason` field
While [Open AI chat completion](finish_reason) does this with a nested `choices.finish_reason` field.

The protocols differ a lot. What I mean is to allow users to mark the end of stream with flexibility.

src/bentoml/_internal/runner/runner_handle/remote.py
jianshen92: `stream=True` is available in `requests` to read response in chunk. aiohttp doesn't have this and doesn't need it because the response is already behave like a stream. 

We will need to new generator here because we need to some unpickling and converting to a data container

src/bentoml/_internal/marshal/dispatcher.py
jianshen92: The biggest difference here is that for streaming, the `callback` returns a async generator, and it is handled differently for a normal response. in particular this part of the code.
```python
            outputs = await self.callback(
                tuple(t.cast(t.Any, input_info.data) for input_info in inputs_info)
            )
            assert len(outputs) == len(inputs_info)
            for input_info, out in zip(inputs_info, outputs):
                fut = input_info.future
                if not fut.done():
                    fut.set_result(out)
            _done = True
            self.optimizer.log_outbound(
                n=len(inputs_info),
                wait=_time_start - inputs_info[-1].enqueue_time,
                duration=time.time() - _time_start,
            )
```

Not sure if it's worth conflating the implementation here to use a single dispatcher?

jianshen92: And speaking of this, I think we might need to give it a better name? Since its also a dispatcher to handle generators?

sauyon: I think we should update the old dispatcher to support streaming in every case, no real reason not to.

src/bentoml/_internal/io_descriptors/text_stream.py
sauyon: > A text stream could be convenient only when user is interacting with the API. However, event stream is the standard protocol of SSE that is supported by main stream browsers, and event stream can carry more information in the event data. A possible solution is to add another type or an option to switch between plain text stream and event stream. After all the title "SSE" refers to an event stream, or it is called simply "streamed/chunked texts"

I think streamed responses predate SSE and support is widely available in JS for both; but yes, I agree that the title SSE is incorrect for this PR.

src/bentoml/_internal/marshal/dispatcher.py
jianshen92: Just trying to understand, what are the benefits by using the current dispatcher?

sauyon: I believe that all the cancellation logic still applies in a streaming context, and besides, I think every response should be streamed anyway, which means that having two doesn't really make sense.

sauyon: Yep!

jianshen92:
`Runnable.method()` and `svc.api()` no longer requires `stream=True` to specify streaming method. we will check whether the function is a generator by using `inspect` module

jianshen92:
Hey guys @aarnphm @sauyon @frostming, as discussed, the changes:
1. remove `stream` arguments in the decorators
2. remove `TextStream` in favour just using `Text` 
3. Every output for `io.Text` uses `StreamResponse`, even if just 1 output. 
4. User can pass in an optional `content_type` to `io.Text` should they want to use a different mime, e.g `text/event-stream` if they want browser's SSE support
5. remove the need of an additional dispatcher

src/bentoml/_internal/runner/runnable.py
```diff
@@ -112,9 +112,12 @@ def method(
         def method_decorator(
             meth: t.Callable[t.Concatenate[T, P], R]
         ) -> RunnableMethod[T, P, R]:
+            import inspect
+
             return RunnableMethod(
                 meth,
                 RunnableMethodConfig(
+                    is_stream=inspect.isasyncgenfunction(meth),
```
frostming, line 120:
```suggestion
                    is_stream=inspect.isasyncgenfunction(meth) or inspect.isgeneratorfunction(meth),
```
The user defined function can be either sync or async

src/bentoml/_internal/runner/runner_handle/local.py
```diff
@@ -61,3 +62,11 @@ async def async_run_method(
...
             limiter=self._limiter,
         )
+
+    def async_stream_method(
+        self,
+        __bentoml_method: RunnerMethod[t.Any, P, R],
+        *args: P.args,
+        **kwargs: P.kwargs,
+    ) -> AsyncGenerator[R, None]:
+        return getattr(self._runnable, __bentoml_method.name)(*args, **kwargs)
```
frostming, line 72: Need to convert the sync generator to async one here. A possible approach:

```python
async def async_stream_method(...) -> AsyncGenerator[R, None]:
    method = getattr(self._runnable, __bentoml_method.name)
    if inspect.isasyncgenfunction(method):
        async for chunk in method(*args, **kwargs):
            yield chunk
    else:  # sync generator
        for chunk in method(*args, **kwargs):
            yield chunk
```

src/bentoml/_internal/runner/runnable.py
jianshen92: Do you think we should enforce user to write an async function?

frostming: No, since it isn't a big deal handling both on our side, we can have more flexibility.

src/bentoml/_internal/server/runner_app.py
```diff
@@ -196,66 +196,84 @@ def _mk_request_handler(
...
-
-                batched_params, indices = payload_paramss_to_batch_params(
-                    params_list, input_batch_dim
-                )
+        if runner_method.config.is_stream:
+            # Streaming does not have batching implemented yet
+            async def infer_stream(
+                paramss: t.Sequence[Params[t.Any]],
+            ) -> t.Sequence[t.AsyncGenerator[Payload, None]]:
+                async def _():
```
aarnphm, line 204:
```suggestion
                async def inner():
```

lets avoid using _

src/bentoml/_internal/server/runner_app.py
```diff
@@ -196,66 +196,84 @@ def _mk_request_handler(
...
+                    # This is a workaround to allow infer stream to return a iterable of
+                    # async generator, to align with how non stream inference works
+                    params = paramss[0].map(AutoContainer.from_payload)
+                    ret = runner_method.async_stream(*params.args, **params.kwargs)
+                    async for data in ret:
+                        payload = AutoContainer.to_payload(data, 0)
+                        yield payload
+
+                return (_(),)
```
aarnphm, line 213:
```suggestion
                return (inner(),)
```

tests/e2e/bento_server_http/service.py
```diff
@@ -68,6 +68,7 @@ def ensure_metrics_are_registered(data: str) -> str:  # pylint: disable=unused-a
         if m.type == "counter"
     ]
     assert "test_metrics" in counters
+    return "ok"
```
aarnphm, line 71: Can we quickly add a stream endpoint?

src/bentoml/_internal/runner/runner_handle/local.py
frostming: LGTM

src/bentoml/_internal/runner/runner_handle/local.py
```diff
@@ -61,3 +63,19 @@ async def async_run_method(
...
+    ) -> AsyncGenerator[R, None]:
+        generator = getattr(self._runnable, __bentoml_method.name)(*args, **kwargs)
+
+        # This allows user to define both sync and async generator
+        if inspect.isasyncgen(generator):
+            async for chunk in generator:
+                yield chunk
+        else:
+            for chunk in generator:
+                yield chunk
```
sauyon, line 81: This concerns me a bit, doesn't this mean that we're blocking the async thread?

jianshen92: yeah, it's not really converting sync -> async for generators as we were doing for the `async_run` and `run`

jianshen92:
e2e test for streaming endpoint is not working for distributed setting atm

src/bentoml/_internal/runner/runner_handle/remote.py
```diff
@@ -274,6 +274,71 @@ async def async_run_method(
...
+                ) as resp:
+                    async for line in resp.content.iter_any():
```
sauyon, line 333: Are we sure we want to use newline as a separator? Can pickled content not contain a newline?

jianshen92: I don't think we are explicitly using newline as separator, starlette handles the chunking delimiter when you pass a generator to the `StreamingResponse`. on the client here we are essentially iterating the chunks 
(i could be wrong)

should probably use `chunk` instead of `line` here

jianshen92: actually you might be right

sauyon: I think chunking should be transparent to us, yeah.

jianshen92: what would you suggest here?

sauyon: We need to have our own separator if we want separate messages in a stream.

src/bentoml/_internal/runner/runner_handle/local.py
sauyon: Let's do that?

src/bentoml/_internal/runner/runner_handle/remote.py
sauyon: Actually, given pickle probably contains arbitrary binary data, we'd probably have to do something like
`<content length> <pickle content>`
and use that to get each object off the stream.

jianshen92: actually after looking up a little, what I initially said was probably right. `<content length> <pickle content>` is actually being abstracted away by `StreamingResponse` and `iter_any`. Actually most ASGI clients will handle the chunking for us in this form https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding#chunked_encoding

I think we can keep this at it is until we see the need of a lower level control?

sauyon: As long as we can ensure both the server and client are sending chunk boundaries and not truncating chunks at a max size, that works, though I think you want `iter_chunks`.

jianshen92: My test with `iter_any` is that it will dynamically read the chunks without a fixed bytes length, though not sure why its a much lesser documented API

sauyon: My understanding of the [aiohttp documentation](https://docs.aiohttp.org/en/stable/streams.html#asynchronous-iteration-support) that `iter_any` will give you partial chunks if the chunks are very large.

Have you verified that very large chunks on the server side are not split into smaller chunks?

src/bentoml/_internal/runner/runner_handle/local.py
frostming: [PEP 492 has an example](https://peps.python.org/pep-0492/#example-2) for getting async generator from a generator. But it looks the same as this approach?

sauyon: I suspect this problem is why the example says "While this is not a very useful thing to do..."

It may be that we want to follow django's example and simply disallow async generators; we may want to measure perf overhead here.

jianshen92: if that is the case, then we should just disallow sync generators no? django

I'm thinking to maybe further improve this in another PR as there's additional intricacies to cast sync generators to a async via a threadpool, and like you said, we may want to do some benchmarking.

sauyon: Let's disallow this for now then.

src/bentoml/_internal/runner/runner_handle/remote.py
frostming: To iterate by line, simply do `async for line in resp.content`
https://docs.aiohttp.org/en/stable/streams.html#asynchronous-iteration-support

sauyon: We don't want to iterate by line as pickle content may include newlines; we must either use chunk boundaries or implement our own chunking, essentially.

jianshen92: Thanks @sauyon for pointing out this. Tested on a big chunk of data with a single `yield` on the server side, indeed it will be split into multiple chunks. I added logic to handle this

src/bentoml/_internal/runner/runner_handle/local.py
```diff
@@ -61,3 +64,16 @@ async def async_run_method(
...
+        __bentoml_method: RunnerMethod[t.Any, P, R],
+        *args: P.args,
+        **kwargs: P.kwargs,
+    ) -> AsyncGenerator[R, None]:
+        generator = getattr(self._runnable, __bentoml_method.name)(*args, **kwargs)
+        # This allows user to define both sync and async generator
+        if inspect.isasyncgen(generator):
+            return generator
+        else:
+            return iterate_in_threadpool(generator)
```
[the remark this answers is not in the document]
aarnphm: Hmm, what is the reason behind this?

src/bentoml/_internal/runner/runner_handle/local.py
jianshen92: @aarnphm ^

src/bentoml/_internal/runner/utils.py
```diff
@@ -152,3 +154,28 @@ def payload_paramss_to_batch_params(
...
+        raise _StopIteration
+
+
+async def iterate_in_threadpool(
+    iterator: t.Iterator[T],
+) -> t.AsyncIterator[T]:
+    while True:
+        try:
+            yield await anyio.to_thread.run_sync(_next, iterator)
+        except _StopIteration:
```
frostming, line 180: Do you know why it is using a custom error instead of `StopIteration`?

src/bentoml/_internal/runner/runner_handle/local.py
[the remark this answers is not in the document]
frostming: This LGTM

src/bentoml/_internal/runner/runner_handle/remote.py
```diff
@@ -274,6 +274,80 @@ async def async_run_method(
...
+                            payload = pickle.loads(buffer)
+                            yield AutoContainer.from_payload(payload)
+
+                            # Clearing the buffer for the next data
+                            buffer = bytearray()
+
+            except aiohttp.ClientOSError as e:
+                raise RemoteException("Failed to connect to runner server.") from e
+
+        return stream_response()
```
aarnphm, line 349: If we just call the function then why do we need to put this into a function in the first place?

jianshen92: Because we need to return an async generator, so it's not just executing the function itself.

src/bentoml/_internal/runner/utils.py
jianshen92: From what I gathered, [PEP 479](https://peps.python.org/pep-0479/) will bubble `RuntimeError` up the context when `StopIteration` is raised within a generator. Hence, to make things clearer, we used a custom error here.

src/bentoml/_internal/runner/runner_handle/remote.py
aarnphm: But you call the function immediately right? Wouldn't that be the same as
```python
            try:
                async with self._client.post(
                    f"{self._addr}/{path}",
                    data=data,
                    headers=headers,
jianshen92 marked this conversation as resolved.
                ) as resp:
                    buffer = bytearray()
                    async for b, end_of_http_chunk in resp.content.iter_chunks():
                        buffer.extend(b)

                        # This is to handle large payload that is split into multiple chunks
                        if end_of_http_chunk and len(buffer) > 0:
                            # TODO: To remove pickling so that we can stream data as it is
                            payload = pickle.loads(buffer)
                            yield AutoContainer.from_payload(payload)

                            # Clearing the buffer for the next data
                            buffer = bytearray()

            except aiohttp.ClientOSError as e:
                raise RemoteException("Failed to connect to runner server.") from e
```

Because this itself already returns an async generator?

jianshen92: great spot. just updated

src/bentoml/_internal/runner/runnable.py
```diff
@@ -112,9 +112,13 @@ def method(
         def method_decorator(
             meth: t.Callable[t.Concatenate[T, P], R]
         ) -> RunnableMethod[T, P, R]:
+            import inspect
+
```
aarnphm, line 116: Let's move import top level for this

src/bentoml/_internal/server/runner_app.py
```diff
@@ -311,6 +330,27 @@ async def _request_handler(request: Request) -> Response:
...
+            from starlette.responses import StreamingResponse
+
+            if runner_method.config.is_stream:
+
+                async def streamer() -> t.AsyncGenerator[bytes, None]:
+                    """
+                    Extract Data from a AsyncGenerator[Payload, None]
+                    """
+                    async for p in payload:
+                        yield pickle.dumps(p)
```
frostming, line 342: I have a few concerns about this approach:

1. Wouldn't this cause double pickling which will affect the performance?
2. Can the aiohttp client split the chunks correctly? How does it know the boundaries?

jianshen92: 1. Yes there's double pickling, it is similar to how we are handling `multiple_outputs`. I was doing this to align with how we currently pass payloads across the wire. 

2. I've tested with 1-100MB worth of strings, the aiohttp server will chunk them and on the client side, it is handled with this [logic]. (https://github.com/bentoml/BentoML/pull/4083/commits/7f568dc276911de38e8ecbd404bc8a3302929ed5) The boundaries is handled by library following this [protocol](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding#chunked_encoding)

frostming: > * Would we only allow user to yield data of one type, or multiple?

> * How would the IO descriptor looks like for streaming of multiple data type

I assume we only support text stream, but this approach seems to allow `yield picklable_object` in the runnable method as well.

> 2\. The boundaries is handled by library following this [protocol](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding#chunked_encoding)

TIL, so it can handle newlines in the serialized form correctly?

src/bentoml/testing/utils.py
```diff
@@ -91,60 +87,3 @@ async def async_request(
             f"Headers assertion failed: {resp.headers!r}",
         )
     return resp.status, Headers(resp.headers), body
-
-
-def assert_distributed_header(headers: multidict.CIMultiDict[str]) -> None:
```
bojiang, line 96: I know it. This is just for ensure headers that Yatai needed.

src/bentoml/_internal/server/runner_app.py
jianshen92: > I assume we only support text stream, but this approach seems to allow yield picklable_object in the runnable method as well.

Actually user can yield anything, image, nd.array, json, and this approach allows us to generalize that as long as we use the right descriptor and implement streaming for them.

> TIL, so it can handle newlines in the serialized form correctly?
yeap, i can confirm that!

aarnphm: fwiw we might want to do a/b testing on this one day

aarnphm (approved):
On my end LGTM in general. My concern is mainly what is the RPS would look like, but I think this might not be a top priority atm.

src/bentoml/_internal/server/runner_app.py
jianshen92: you mean a/b test for the additional payload and see the overhead?

frostming: Anyway, the internal exchange format won't affect the user interface, so LGTM and we can improve the performance(if it has issues) in the future.

jianshen92:
Upon merging I will raise another PR for the docs.

PR #4092, opened 2023-08-02
feat(store): update annotations

aarnphm:
working with store

We also should flush the yaml before enter the context, for validate to work

### DIFF
--- a/src/bentoml/_internal/store.py
+++ b/src/bentoml/_internal/store.py
@@ -173,14 +173,13 @@ class Store(ABC, t.Generic[Item]):
         except Exception:
             self._fs.removetree(item_path)
             raise
-        finally:
-            # item generation is most likely successful, link latest path
-            if (
-                not self._fs.exists(_tag.latest_path())
-                or self.get(_tag).creation_time >= self.get(_tag.name).creation_time
-            ):
-                with self._fs.open(_tag.latest_path(), "w") as latest_file:
-                    latest_file.write(_tag.version)
+        # item generation is most likely successful, link latest path
+        if (
+            not self._fs.exists(_tag.latest_path())
+            or self.get(_tag).creation_time >= self.get(_tag.name).creation_time
+        ):
+            with self._fs.open(_tag.latest_path(), "w") as latest_file:
+                latest_file.write(_tag.version)
 
     def delete(self, tag: t.Union[str, Tag]) -> None:
         _tag = Tag.from_taglike(tag)
--- a/src/bentoml/_internal/store.py
+++ b/src/bentoml/_internal/store.py
@@ -159,7 +159,7 @@ class Store(ABC, t.Generic[Item]):
             )
 
     @contextmanager
-    def register(self, tag: t.Union[str, Tag]):
+    def register(self, tag: str | Tag) -> t.Generator[str, None, None]:
         _tag = Tag.from_taglike(tag)
 
         item_path = _tag.path()
@@ -170,14 +170,26 @@ class Store(ABC, t.Generic[Item]):
         self._fs.makedirs(item_path)
         try:
             yield self._fs.getsyspath(item_path)
-        finally:
+        except Exception:
+            raise
+        else:
             # item generation is most likely successful, link latest path
-            if (
-                not self._fs.exists(_tag.latest_path())
-                or self.get(_tag).creation_time >= self.get(_tag.name).creation_time
-            ):
-                with self._fs.open(_tag.latest_path(), "w") as latest_file:
-                    latest_file.write(_tag.version)
+            self.update_latest(_tag)
+
+    def exists(self, tag: str | Tag) -> bool:
+        return self._fs.exists(Tag.from_taglike(tag).path())
+
+    def path_of(self, path: str) -> str:
+        return self._fs.getsyspath(path)
+
+    def update_latest(self, tag: str | Tag) -> None:
+        _tag = Tag.from_taglike(tag)
+        if (
+            not self._fs.exists(_tag.latest_path())
+            or self.get(_tag).creation_time >= self.get(_tag.name).creation_time
+        ):
+            with self._fs.open(_tag.latest_path(), "w") as latest_file:
+                latest_file.write(_tag.version)
 
     def delete(self, tag: t.Union[str, Tag]) -> None:
         _tag = Tag.from_taglike(tag)
--- a/src/bentoml/_internal/store.py
+++ b/src/bentoml/_internal/store.py
@@ -173,7 +173,7 @@ class Store(ABC, t.Generic[Item]):
         except Exception:
             self._fs.removetree(item_path)
             raise
-        else:
+        finally:
             # item generation is most likely successful, link latest path
             if (
                 not self._fs.exists(_tag.latest_path())
--- a/src/bentoml/_internal/store.py
+++ b/src/bentoml/_internal/store.py
@@ -170,24 +170,17 @@ class Store(ABC, t.Generic[Item]):
         self._fs.makedirs(item_path)
         try:
             yield self._fs.getsyspath(item_path)
+        except Exception:
+            self._fs.removetree(item_path)
+            raise
         else:
             # item generation is most likely successful, link latest path
-            self.update_latest(_tag)
-
-    def exists(self, tag: str | Tag) -> bool:
-        return self._fs.exists(Tag.from_taglike(tag).path())
-
-    def path_of(self, path: str) -> str:
-        return self._fs.getsyspath(path)
-
-    def update_latest(self, tag: str | Tag) -> None:
-        _tag = Tag.from_taglike(tag)
-        if (
-            not self._fs.exists(_tag.latest_path())
-            or self.get(_tag).creation_time >= self.get(_tag.name).creation_time
-        ):
-            with self._fs.open(_tag.latest_path(), "w") as latest_file:
-                latest_file.write(_tag.version)
+            if (
+                not self._fs.exists(_tag.latest_path())
+                or self.get(_tag).creation_time >= self.get(_tag.name).creation_time
+            ):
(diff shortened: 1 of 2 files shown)


src/bentoml/_internal/store.py
```diff
@@ -170,14 +170,26 @@ def register(self, tag: t.Union[str, Tag]):
         self._fs.makedirs(item_path)
         try:
             yield self._fs.getsyspath(item_path)
-        finally:
+        except Exception:
+            raise
```
sauyon, line 174: What's this for?

aarnphm: If any exceptions where that happens during the context manager, we should raise it

src/bentoml/_internal/store.py
```diff
@@ -170,14 +170,26 @@ def register(self, tag: t.Union[str, Tag]):
...
+        return self._fs.getsyspath(path)
+
+    def update_latest(self, tag: str | Tag) -> None:
+        _tag = Tag.from_taglike(tag)
+        if (
+            not self._fs.exists(_tag.latest_path())
+            or self.get(_tag).creation_time >= self.get(_tag.name).creation_time
+        ):
+            with self._fs.open(_tag.latest_path(), "w") as latest_file:
+                latest_file.write(_tag.version)
```
sauyon, line 192: Why did you factor this out?

src/bentoml/_internal/store.py
```diff
@@ -170,14 +170,26 @@ def register(self, tag: t.Union[str, Tag]):
...
-            ):
-                with self._fs.open(_tag.latest_path(), "w") as latest_file:
-                    latest_file.write(_tag.version)
+            self.update_latest(_tag)
+
+    def exists(self, tag: str | Tag) -> bool:
+        return self._fs.exists(Tag.from_taglike(tag).path())
+
+    def path_of(self, path: str) -> str:
+        return self._fs.getsyspath(path)
```
sauyon, line 183: Why are you putting these here and not using the `Bento`'s `path_of`?

src/bentoml/models.py
```diff
@@ -257,14 +301,19 @@ def create(
...
+        res.flush()
         res.enter_cloudpickle_context(external_modules, imported_modules)
         yield res
     except Exception as e:
         raise e
+    except KeyboardInterrupt:
+        # During save, if user hit Ctrl+C, we will remove the item path cleanly
+        # so that validate() will work correctly.
+        res._fs.removetree(res.path_of(res.tag.path()))
+        raise
```
sauyon, line 315: This shouldn't have mattered; do you know why this is necessary?

src/bentoml/models.py
```diff
@@ -257,14 +301,19 @@ def create(
...
     )
     external_modules = [] if external_modules is None else external_modules
-    imported_modules = []
+    imported_modules: t.List[ModuleType] = []
     try:
+        res.flush()
         res.enter_cloudpickle_context(external_modules, imported_modules)
         yield res
     except Exception as e:
         raise e
```
sauyon, line 310: I'm not sure why these are necessary either, isn't this identical to not having a try block?

src/bentoml/_internal/store.py
aarnphm: This is the store fs, not the bento. Bento and Model has its own path_of

src/bentoml/models.py
aarnphm: My suggestion is during the move, if users press Ctrl+C, it is not cleanup correctly.

src/bentoml/_internal/store.py
frostming: This can be removed and it works the same

src/bentoml/_internal/store.py
sauyon: Right, but when do you think that this will be useful? I think we should enforce that people should use `get` and then use the Bento's FS.

aarnphm: Do we want to expose the store API in the future? I think it is useful in general

src/bentoml/_internal/store.py
aarnphm, line 174:
```suggestion
```

src/bentoml/_internal/store.py
sauyon: No, I don't think we should.

src/bentoml/_internal/store.py
sauyon: (I don't think we should support this as a separate operation)

src/bentoml/models.py
sauyon: Are you going to remove this?

src/bentoml/models.py
sauyon: ^

src/bentoml/models.py
```diff
@@ -227,6 +227,50 @@ def pull(
...
+    custom_objects: dict[str, t.Any] | None = ...,
+    external_modules: t.List[ModuleType] | None = ...,
+    metadata: dict[str, t.Any] | None = ...,
+    context: ModelContext,
+    use_tempfs: bool = ...,
+    _model_store: ModelStore = ...,
+) -> t.Generator[Model, None, None]:
+    ...
+
+
```
larme, line 273: I'm a little bit lost here but what's purpose of writing definitions in this way?

src/bentoml/models.py
aarnphm: done

src/bentoml/models.py
aarnphm: Oh this is to support create with unpack
```python
kwargs = dict(context=..., not_a_valid_type="something")
with bentoml.models.create("some_name", **kwargs): # <- should throw an error here saying kwargs doesn't have a valid argument
```

src/bentoml/models.py
```diff
@@ -227,6 +227,48 @@ def pull(
...
+        context: t.Required[ModelContext]
+        _model_store: t.NotRequired[ModelStore]
+
+
+@t.overload
+@contextmanager
+def create(
+    name: Tag | str, /, **attrs: t.Unpack[CreateKwargs]
+) -> t.Generator[Model, None, None]:
+    ...
```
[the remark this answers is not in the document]
aarnphm: oh it supports `**kwargs`:
```python
attrs = dict(module="bentoml.transformers", context=ModelContext())

with bentoml.models.create("name", **attrs): # type-safe
  ...
```

where this will fail:
```python
attrs = dict(module="bentoml.transformers", not_a_valid_arg="asdf")
with bentoml.models.create("name, **attrs): # <- should show error with type check
  ...
```

src/bentoml/models.py
```diff
@@ -257,14 +299,14 @@ def create(
         context=context,
     )
     external_modules = [] if external_modules is None else external_modules
-    imported_modules = []
+    imported_modules: t.List[ModuleType] = []
```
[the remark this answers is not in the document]
aarnphm: because we have a list above right? 😄

src/bentoml/models.py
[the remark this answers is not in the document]
sauyon: Ah, nice! I'm surprised that it doesn't work normally...

src/bentoml/models.py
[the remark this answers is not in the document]
sauyon: Oh, right, this is `models`, ugh.

src/bentoml/models.py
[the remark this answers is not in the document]
aarnphm: Oh this is a new feature type-safe https://docs.python.org/3.13/library/typing.html#typing.Unpack

src/bentoml/_internal/store.py
```diff
@@ -170,7 +170,10 @@ def register(self, tag: t.Union[str, Tag]):
         self._fs.makedirs(item_path)
         try:
             yield self._fs.getsyspath(item_path)
-        finally:
+        except Exception:
+            self._fs.removetree(item_path)
+            raise
+        else:
```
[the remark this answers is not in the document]
aarnphm: this else will still get hit if there is no exception right?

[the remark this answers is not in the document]
sauyon: Yes, but it's also identical to not having the else.

src/bentoml/models.py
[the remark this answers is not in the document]
sauyon: No, I mean that it's strange that `Unpack` typing is required at all for `**attrs` to work.

src/bentoml/_internal/store.py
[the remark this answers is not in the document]
aarnphm: finally will run regardless of whether we catch the exception or not right?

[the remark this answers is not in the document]
sauyon: This isn't a finally, it's an else.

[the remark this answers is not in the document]
aarnphm: yes you need except to use else. you can `try..else`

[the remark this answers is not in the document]
sauyon: I'm saying that this else does nothing and can be removed.

aarnphm, line 176:
```suggestion
        finally:
```

src/bentoml/_internal/store.py
```diff
@@ -170,6 +170,9 @@ def register(self, tag: t.Union[str, Tag]):
         self._fs.makedirs(item_path)
         try:
             yield self._fs.getsyspath(item_path)
+        except Exception:
+            self._fs.removetree(item_path)
+            raise
         finally:
```
sauyon, line 176: I didn't want you to change it back to a finally, I agree this should not execute when there's an exception. Let's just dedent the block and remove the finally?

sauyon (approved):
Alright let's get this one in!

PR #4100, opened 2023-08-08
chore(deps): open Starlette version, to allow latest

alexeyshockov:
## What does this PR address?

Open the version constraints for Starlette, to allow the latest version (it's a consumer decision, as BentoML is used as a library is many cases).

This also allows to use the latest FastAPI (a popular combination AFAIK), as it usually requires latest version of Starlette. 

## Before submitting:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,7 +59,7 @@ dependencies = [
     "rich>=11.2.0",
     "schema",
     "simple-di>=0.1.4",
-    "starlette>=0.13.5,<0.29",
+    "starlette>=0.13.5",
     "uvicorn",
     "watchfiles>=0.15.0",
 ]


sauyon:
We limited this because at some point Starlette made breaking changes, but we should really keep up with updates better. @aarnphm did we disable dependabot?

aarnphm:
yes we did, because starlette introduces breaking change, hence upper bound limit.

sauyon:
Wait, we disabled dependabot for Starlette? We shouldn't do that, we should be doing the work to ensure compatibility, right?

alexeyshockov:
I think it was one time BC break that is already addressed, right? So opening the constraints is not an issue now, you just want to be safe in the future?

In this case I would argue that all the other deps have no upper bound, so any of them can introduce a BC break with a new version. I don't see any reason why Starlette should not be special.

Also a point against dependabot approach: we are still sitting on BentoML 1.0.20, which constraints Starlette to a "really old" version (yes-yes, going to upgrade in the next week).

IMO it should be a consistent approach on dependency constraints.

sauyon:
> In this case I would argue that all the other deps have no upper bound, so any of them can introduce a BC break with a new version. I don't see any reason why Starlette should be special.

I think the approach is leave dependencies unlocked until they break on us and we no longer trust that backwards compatibility will work as expected.

> Also a point against dependabot approach: we are still sitting on BentoML 1.0.20, which constraints Starlette to a "really old" version (yes-yes, going to upgrade in the next week).

We do try to ensure backwards compatibility as much as we can, if we break something you're using in an update that's a bug and should be reported.

judahrand:
It is really important to us to be able to manage dependencies such as Starlette for example in the case of security vulnerabilities. It doesn't feel like the right solution for BentoML to pin this dependency. Downstream users should really be pinning dependencies in their environment.

aarnphm:
I disagree with the case of Starlette. Surely if there are any security vulnerabilities, we are happy to bump the upper bound for Starlette. In the past, they have introduced breaking change, which breaks BentoML, so we decided that we will upper limit the version that we know works with bentoml.

sauyon:
At this point this PR has been open long enough that I think we should just merge it.

aarnphm (approved):

Issue #4115, opened 2023-08-11
bug: connects are created for every request


judahrand:
### Describe the bug

In both the GRPC and HTTP clients channels/sessions are created for every single request. Both GRPC and AIOHTTP recommend reusing channels/sessions for the lifecycle of the application. 

- https://grpc.io/docs/guides/performance/#general
- https://docs.aiohttp.org/en/stable/client_reference.html#client-session

In both cases the channel/session encapsulates a connection pool. If a new channel/session is created for every request this connection pool must also be recreated. This leads to performance which is poorer than it could be. 

Additionally, to really fix this problem each client needs to have its own implementation of `_sync_call` which is distinct from `_call`. Using `asyncio.run` in the background causes problems. It might be simpler and clearer to separate the clients into `AsyncHTTPClient`, `HTTPClient`, `AsyncGrpcClient`, and `GrpcClient`. Each of which could handle its simpler case.

### Expected behavior

 Reuse of channels/sessions

### Environment

N/A

sauyon:
HTTP resolved in #4116; we should probably consider gRPC at some point.

PR #4116, opened 2023-08-11
feat: separate and optimize async and sync clients

judahrand:
## What does this PR address?

A draft attempt at addressing #4115 

Currently, this introduces a lot of duplicate code. However, I think it should be possible to reduce this significantly. 

I believe by having separate sync and async implementations optimal performance can be achieved for both.

I'd be interested to get thoughts and feedback on this PR as well as advice on how to benchmark it against `main` as I believe it should see some performance benefit across multiple requests on the same client.

Fixes #(issue)

## Before submitting:


sauyon (changes requested):
For backwards compatibility, the client implementations will have to retain the sync/async methods; perhaps the default client could instantiate a sync / async client?

judahrand:
> For backwards compatibility, the client implementations will have to retain the sync/async methods; perhaps the default client could instantiate a sync / async client?

The latest changes should maintain backwards compatibility. Is this sort of a solution acceptable? If so I'll push forwards with getting the tests passing and perhaps also look at splitting the tests up by sync and async implementation.

src/bentoml/_internal/client/__init__.py
```diff
@@ -30,50 +33,61 @@ def __init__(self, svc: Service, server_url: str):
...
 
-            HTTPClient.wait_until_server_ready(host, port, timeout, **kwargs)
+            SyncHTTPClient.wait_until_server_ready(host, port, timeout, **kwargs)
         except BadStatusLine:
             # when address is a RPC
-            from .grpc import GrpcClient
+            from .grpc import SyncGrpcClient
 
-            GrpcClient.wait_until_server_ready(host, port, timeout, **kwargs)
+            SyncGrpcClient.wait_until_server_ready(host, port, timeout, **kwargs)
```
aarnphm, line 90: nit: For sync client naming would be `GrpcClient`,
and Async would be `AsyncGrpcClient` (similar to what aiohttp does)

One more thing is we might want to fix the sync client `async function` behaviour within a running loop

judahrand: > nit: For sync client naming would be GrpcClient,
> and Async would be AsyncGrpcClient (similar to what aiohttp does)

I chose not to do that since `GrpcClient` already exists and its API is maintained for backwards compatibility.

sauyon: I think most of these should in theory be mostly memory bound and therefore valid usecases for async functions, but we are intending to properly rework IO descriptors (that's pretty much next on my plate, and should hopefully have some preliminary work soon).

Re: making sync functions run in an async environment: I'm not sure that this is a real issue. In an async environment you should really never be calling the sync functions, and it's actively harmful to do so. We should really just create a wrapper around `asyncio.run` that tells users to use the async methods if they're already in a running loop.

judahrand:
I think that this PR has gotten a bit muddled. This is something which I am keen to address as at the moment I think the client implementations are quite poor. However, I may take another stab at this later understanding more about all the various moving parts.

There seems to be a lot which needs changes to:

- make the synchronous clients work reliably and efficiently
- make the async clients work efficiently
- separate the client implementation from the core BentoML package to avoid carrying huge numbers of irrelevant dependencies around in client side software.

src/bentoml/_internal/client/__init__.py
judahrand: > Re: making sync functions run in an async environment: I'm not sure that this is a real issue. In an async environment you should really never be calling the sync functions, and it's actively harmful to do so.

I'm not sure this is true in Python, is it? If you write an `async def` function which never calls `await` inside it it will never yield to the event loop, will it? `to_proto` and `from_proto` will never call `await` mostly because Protobuf doesn't have any `async` implementations and doesn't intend to. Making `to/from_proto` `async def` doesn't magically yield to the event loop.

If you don't want to block the event loop I think you'd be way better off making `to/from_proto` synchronous functions and then calling them in `async` functions like this:

```python
executor = concurrent.futures.ThreadPoolExecutor(max_workers=3)
loop = asyncio.get_event_loop()
result = await loop.run_in_executor(executor, self.to_proto, obj)
```

sauyon: Yeah, I'm not entirely sure why the from/to proto functions are async now. Probably just so they matched everything else, I think that's probably just a problem we didn't think about very hard.

sauyon: > I'm not sure this is true in Python, is it? If you write an `async def` function which never calls `await` inside it it will never yield to the event loop, will it?

You're correct, I'm in fact  not sure that a single one of our implementations are actually correctly taking advantage of async, this is somewhat of a mess, and something we need to improve.

src/bentoml/_internal/client/__init__.py
```diff
@@ -17,63 +16,92 @@
...
+    from .http import SyncHTTPClient
 
 
 class Client(ABC):
     server_url: str
     _svc: Service
     endpoints: list[str]
 
+    _sync_client: SyncClient
+    _async_client: AsyncClient
```
sauyon, line 33: Where are these initialized?

src/bentoml/_internal/io_descriptors/base.py
```diff
@@ -175,18 +175,29 @@ def input_type(self) -> InputType:
...
         self, obj: IOType, ctx: Context | None = None
     ) -> Response:
-        raise NotImplementedError
+        return self.sync_to_http_response(obj, ctx)
 
     @abstractmethod
-    async def from_proto(self, field: t.Any) -> IOType:
+    def sync_to_http_response(
+        self, obj: IOType, ctx: Context | None = None
+    ) -> Response:
```
sauyon, line 186: I think we unfortunately can't add new abstract methods or change this base class as it's public and we support defining your own IO descriptors; sync / async methods will be included in the next incarnation, I think.

sauyon (commented):
A few things that I think need changing. Haven't reviewed the async client code entirely carefully yet.

@aarnphm can you help review the grpc client?

src/bentoml/_internal/io_descriptors/base.py
sauyon: (This is also what's causing the test failure, FWIW)

sauyon:
> I think that this PR has gotten a bit muddled. This is something which I am keen to address as at the moment I think the client implementations are quite poor. However, I may take another stab at this later understanding more about all the various moving parts.

I think apart from the IO descriptor change we should get this in, then probably circle back after we rework those.

> There seems to be a lot which needs changes to:
> 
>     * make the synchronous clients work reliably and efficiently
>     * make the async clients work efficiently

Yep.

Somewhat related, there was at one point an issue where connections were being dropped and we were running into strange errors when we reused a connection which we haven't gotten to the bottom of, which I probably could have added a comment for, but we really need to figure that out.

>     * separate the client implementation from the core BentoML package to avoid carrying huge numbers of irrelevant dependencies around in client side software.

We probably need to think more carefully about this one in general; a lot of our dependencies are only required for servers / only required for clients, etc. @parano @bojiang something to think about, I think.

aarnphm:
FWIW I think we can separate out a `bentoml-client` package, and by default `bentoml` will include `bentoml-client`

judahrand:
> FWIW I think we can separate out a `bentoml-client` package, and by default `bentoml` will include `bentoml-client`

Agreed. I also wonder if the serialization/deserialization logic should live in a separate package (given that both the client and server will want to depend on it). This is something a lot of Rust projects do very very well (see `arrow-rs`). They split out all their subcomponents into sub crates.

I believe PDM might even be able to cope with this setup in Python? https://pdm.fming.dev/latest/usage/advanced/#use-pdm-to-manage-a-monorepo

aarnphm:
> > FWIW I think we can separate out a `bentoml-client` package, and by default `bentoml` will include `bentoml-client`
> 
> Agreed. I also wonder if the serialization/deserialization logic should live in a separate package (given that both the client and server will want to depend on it). This is something a lot of Rust projects do very very well (see `arrow-rs`). They split out all their subcomponents into sub crates.
> 
> I believe PDM might even be able to cope with this setup in Python? [pdm.fming.dev/latest/usage/advanced/#use-pdm-to-manage-a-monorepo](https://pdm.fming.dev/latest/usage/advanced/#use-pdm-to-manage-a-monorepo)

this will be worked into bentoml-core 🎉

sauyon:
I don't know that it needs to be a monorepo.

sauyon:
Would it be ok if I took this one over so we can get it in while we work on the rework?

judahrand:
> Would it be ok if I took this one over so we can get it in while we work on the rework?

100% - sorry for letting this one lag. Various demands on time etc etc etc - same as everyone else 🤣

src/bentoml/_internal/client/__init__.py
```diff
@@ -127,35 +130,263 @@ def from_url(
...
+
+    def __init__(self, svc: Service, server_url: str):
+        self._svc = svc
+        self.server_url = server_url
+
+        if len(svc.apis) == 0:
+            raise BentoMLException("No APIs were found when constructing client.")
+
+        self.endpoints = []
+        self.endpoints = []
```
frostming, line 172: Duplicate lines?

src/bentoml/_internal/client/http.py
```diff
@@ -8,25 +8,193 @@
...
+                async with httpx.AsyncClient(base_url=f"{host}:{port}") as session:
+                    resp = await session.get("/readyz")
+                    if resp.status_code == 200:
+                        break
+                    else:
+                        await asyncio.sleep(check_interval)
+            except (
+                ConnectionError,
+                urllib.error.URLError,
+                socket.timeout,
```
frostming: Can these catch the errors raised by httpx?

frostming (commented):
I am mostly fine with it.

sauyon:
@aarnphm @frostming should be ready for another look!

aarnphm (approved):

aarnphm (approved):

PR #4135, opened 2023-08-20
fix: UnboundLocalError `wait` in dispatcher  

### LINKED ISSUE #4128: bug:  UnboundLocalError in the dispatcher
### Describe the bug

```
2023-08-16T18:02:15+0800 [ERROR] [runner:baichuan_runner:1] Traceback (most recent call last):
  File "/root/miniconda3/envs/chen/lib/python3.10/site-packages/bentoml/_internal/marshal/dispatcher.py", line 221, in train_optimizer
    if n < batch_size and (batch_size * a + b) + w0 <= wait:
UnboundLocalError: local variable 'wait' referenced before assignment
```

Image

This happens sporadically but the application will still run.

### To reproduce

_No response_

### Expected behavior

_No response_

### Environment

-

KimSoungRyoul:
## What does this PR address?

Fixes #4128 

this issue is so wired

this fix won't help us figure out that why `wait` does not exist under line `if batch_size > 1: ...`
I monkey-patched bentoml internal code due to this issue.

### DIFF
--- a/src/bentoml/_internal/marshal/dispatcher.py
+++ b/src/bentoml/_internal/marshal/dispatcher.py
@@ -193,12 +193,6 @@ class CorkDispatcher:
         if self.max_batch_size < batch_size:
             batch_size = self.max_batch_size
 
-        if batch_size > 1:
-            wait = min(
-                self.max_latency * 0.95,
-                (batch_size * 2 + 1) * (self.optimizer.o_a + self.optimizer.o_b),
-            )
-
         req_count = 0
         try:
             while req_count < num_reqs_to_train:
@@ -217,6 +211,10 @@ class CorkDispatcher:
                 if batch_size > 1:  # only wait if batch_size
                     a = self.optimizer.o_a
                     b = self.optimizer.o_b
+                    wait = min(
+                        self.max_latency * 0.95,
+                        (batch_size * 2 + 1) * (self.a + self.optimizer.o_b),
+                    )
 
                     if n < batch_size and (batch_size * a + b) + w0 <= wait:
                         await asyncio.sleep(self.tick_interval)
--- a/src/bentoml/_internal/marshal/dispatcher.py
+++ b/src/bentoml/_internal/marshal/dispatcher.py
@@ -213,7 +213,7 @@ class CorkDispatcher:
                     b = self.optimizer.o_b
                     wait = min(
                         self.max_latency * 0.95,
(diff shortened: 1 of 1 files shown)


src/bentoml/_internal/marshal/dispatcher.py
```diff
@@ -217,6 +211,10 @@ async def train_optimizer(
                 if batch_size > 1:  # only wait if batch_size
                     a = self.optimizer.o_a
                     b = self.optimizer.o_b
+                    wait = min(
+                        self.max_latency * 0.95,
+                        (batch_size * 2 + 1) * (self.a + self.optimizer.o_b),
```
frostming, line 216: `self.a` is a typo?

src/bentoml/_internal/marshal/dispatcher.py
```diff
@@ -217,6 +211,10 @@ async def train_optimizer(
                 if batch_size > 1:  # only wait if batch_size
                     a = self.optimizer.o_a
                     b = self.optimizer.o_b
+                    wait = min(
+                        self.max_latency * 0.95,
+                        (batch_size * 2 + 1) * (a + b),
+                    )
```
aarnphm, line 217: I don't think moving wait here solves the issue

Since wait should already be defined as before

aarnphm (changes requested):
Don't think this is correct.
Also putting wait here will increase the bytecode since it is inside the while loop, as this is also a performance-sensitive code.

src/bentoml/_internal/marshal/dispatcher.py
frostming: It does since `wait` is only used by L219 below. But it seems to change the behavior since `a`, `b` and `self.max_latency` all change during the loop while `wait` doesn't change before.

KimSoungRyoul: I agree this fix does not good solution . it's just temporary fix

It is a temporary fix Pull Request aimed at hotfix.

currently we can not use `workers_per_resource Option` in BentoML `v1.1.0 ~v1.1.1`

aarnphm: @sauyon does this look ok with you?

KimSoungRyoul:
This hotfix feels rushed and sloppy. 
Even if there is a bug right now, it might be better to analyze the exact cause and then fix it.

thanks, I close this PR myself

PR #4161, opened 2023-08-31
fix: use closer-integrated callbackwrapper

sauyon:
This fixes a few weird issues I was seeing with the httpx `content=`, as previously some functions were not behaving correctly.

### DIFF
--- a/src/bentoml/_internal/cloud/base.py
+++ b/src/bentoml/_internal/cloud/base.py
@@ -1,10 +1,10 @@
 from __future__ import annotations
 
+import io
 import typing as t
 from abc import ABC
 from abc import abstractmethod
 from contextlib import contextmanager
-from functools import wraps
 
 from rich.console import Group
 from rich.panel import Panel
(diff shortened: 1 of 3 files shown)


aarnphm (approved):

src/bentoml/_internal/cloud/base.py
```diff
@@ -26,61 +26,36 @@
...
+        if size is not None:
+            res = super().read(size)
         else:
-            raise KeyError("Can only wrap read/write methods")
+            res = super().read()
+        if self.read_cb is not None:
+            self.read_cb(len(res))
+        return res
+
+    def write(self, data: bytes):  # type: ignore  # python buffer types are too new and seem to not support something like Buffer+Sized as of now
```
aarnphm, line 53:
```suggestion
    def write(self, data: bytes) -> t.Any:  # type: ignore  # python buffer types are too new and seem to not support something like Buffer+Sized as of now
```

aarnphm (changes requested):
one smol comment

aarnphm (approved):

PR #4174, opened 2023-09-07
fix: apply env map for distributed runner workers

bojiang:


bojiang:
@sauyon  Here's a feature we will need to adopt to Rust in the future.

aarnphm (approved):

sauyon:
Where is `scheduled_worker_env_map` set?

aarnphm:
> Where is `scheduled_worker_env_map` set?

runners.py#L348

sauyon:
Oh, it's from our scheduler. This seems like it should go away with our new architecture anyway, right? I feel like we should just remove all the resource stuff, anyway.

ssheng (approved):

PR #4185, opened 2023-09-14
chore: refactor to a monorepo including core

sauyon:
@aarnphm @frostming I think this was somewhat overdue; I'll continue to work on the other PRs but really rebasing shouldn't be too much of a chore.


core/__pypackages__/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
```
aarnphm, line 1: I don't think we should include this folder here

Is there a specific reason for this?

aarnphm (commented):
tiny comment, but still need @frostming to take another look on __pypackages__ for pdm

aarnphm:
Also we might need to regenerate pdm.lock

sauyon:
PDM is somewhat broken; I still need to fix the CI.

core/__pypackages__/.gitignore
sauyon: Nope, that's a mistake.

sauyon:
Hm, I suspect something is wrong, since `pdm lock -d` has so far taken 33 minutes; I'll probably go to sleep for today and look at this again tomorrow...

sauyon:
I've pushed what I have in case you wanted to take a look @frostming

sauyon:
Oh, this was expected, I guess, since pdm doesn't pick up the config from bentoml about our dependency overrides.

sauyon:
@ssheng @parano not sure who controls readthedocs, I think it'll have to be updated to render the correct directory, though we could possibly just leave docs at root.

sauyon:
Hm, I don't think PDM monorepo support is quite there yet; it probably makes more sense to just treat the two projects as separate and have CI just operate in each directory separately.

PR #4200, opened 2023-09-20
docs: Add best practices doc for cost optimization

Sherlock113:
## What does this PR address?

Fixes #(issue)

## Before submitting:

### DIFF
--- a/docs/source/bentocloud/best-practices/cost-optimization.rst
+++ b/docs/source/bentocloud/best-practices/cost-optimization.rst
@@ -1,0 +1,65 @@
+=================
+Cost Optimization
+=================
+
+This document contains a list of best practices for optimizing costs on BentoCloud. You can use the information on this page as
+a reference when trying to reduce or rightsize your overall cloud spending. This helps you maximize the value of your cloud resources.
+
+.. note::
+    
+    If you're new to BentoCloud, this document may not be the best place to start. We recommend you begin with this :doc:`/bentocloud/getting-started/quickstart`
+    to understand the basics of BentoCloud.
+
+Basic
+-----
+
+For those with a basic understanding of BentoML and BentoCloud, consider the following practices:
+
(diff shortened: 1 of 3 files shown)


jianshen92:
I think these (this one and the CI/CD doc) could fall under "topic guides" instead of creating a "best practices" category. Topic guide are designed for more open ended articles like this.

Thoughts on calling it "Best Practices" (purely opinion)
- I felt it is a very offensive stance of telling user that you "should" do this, while it's just a recommended way.
- What best for one might not be best for each other, hence i think calling something "best" is on the extreme opinioniated side, unless it is backed with facts

maybe we could just call it :
- "Cost Saving Measures"
- "CICD with BentoCloud"
?

Sherlock113:
> I think these (this one and the CI/CD doc) could fall under "topic guides" instead of creating a "best practices" category. Topic guide are designed for more open ended articles like this.
> 
> Thoughts on calling it "Best Practices" (purely opinion)
> 
> * I felt it is a very offensive stance of telling user that you "should" do this, while it's just a recommended way.
> * What best for one might not be best for each other, hence i think calling something "best" is on the extreme opinioniated side, unless it is backed with facts
> 
> maybe we could just call it :
> 
> * "Cost Saving Measures"
> * "CICD with BentoCloud"
>   ?

Well, the "best practices" wording is fine and very common in docs actually. For example, here is the SageMaker docs: https://docs.aws.amazon.com/sagemaker/latest/dg/best-practices.html. And I can give you more examples like this in Google Cloud, Azure docs, or other IT & AI product docs. So for Best Practices as titles or sections, I think it is ok.

But you are correct. In terms of tone, "should" needs to be avoided in best practices docs. This is why I did not use this word, and used "can", "recommend", and "consider" instead in this doc. Obviously, these are just suggestions and users can choose to do or not, depending on their use cases/needs.

Sherlock113:
Merging...

PR #4209, opened 2023-09-26
fix: subprocess not using same python as main process causing `bentoml.bentos.build` to crash

nickolasrm:
## What does this PR address?

When using `bentoml.bentos.build` from the python api, my project crashes because I am using an in-project virtual env Thus the global `bentoml` command is not installed or it might be a call to another `bentoml` installation, which can cause conflicts.

This change forces subprocess to use the same python binary as the caller process, fixing the problem (tested locally)

## Before submitting:

### DIFF
--- a/src/bentoml/bentos.py
+++ b/src/bentoml/bentos.py
@@ -442,7 +442,7 @@ def build_bentofile(
     copied.setdefault("BENTOML_HOME", BentoMLContainer.bentoml_home.get())
     try:
         return get(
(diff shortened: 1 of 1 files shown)


frostming:
LGTM, just haven't checked if there is any other such usage?

nickolasrm:
> LGTM, just haven't checked if there is any other such usage?

I don't think so, I usually put `sys.executable -m` before my subprocess calls due to conflicts when using multiple python installations. Poetry sometimes give me headaches for the same reason haha

sauyon (approved):

PR #4227, opened 2023-10-10
fix: use httpx for testing

sauyon:


tests/unit/_internal/utils/test_analytics.py
```diff
@@ -67,7 +67,7 @@ def test_get_payload(event_properties: analytics.schemas.ModelSaveEvent):
     assert SCHEMA.validate(payload)
 
 
-@patch("bentoml._internal.utils.analytics.usage_stats.requests.post")
+@patch("bentoml._internal.utils.analytics.usage_stats.httx.post")
```
natsukium, line 70:
```suggestion
@patch("bentoml._internal.utils.analytics.usage_stats.httpx.post")
```
Here is a typo I found while packaging with nixpkgs (https://github.com/NixOS/nixpkgs/pull/262687).

PR #4228, opened 2023-10-10
fix: use httpx for usage tracking

sauyon:


pyproject.toml
```diff
@@ -34,6 +34,7 @@ dependencies = [
     "cloudpickle>=2.0.0",
     "deepmerge",
     "fs",
+    "httpx",
```
aarnphm, line 37: Let's also remove requests from the deps?

sauyon: We can't in any of these PRs, it's used by everything else.

aarnphm: so are you going to remove this once all of the PRs are merged?

sauyon: Yep, I think the tests PR is not working, so will probably remove it with that one.

aarnphm (approved):

sauyon:
I think this PR broke CI @aarnphm going to revert.

PR #4247, opened 2023-10-23
feat: import model for transformers framework

MingLiangDai:
## What does this PR address?

Add import model method to transformers framework allowing users to directly download models from huggingface to bentoml model store. Add additional functionality to `bentoml models import` to import from huggingface. (eg.  `bentoml models import hf:t5-base`)

## Before submitting:


aarnphm:
I think `bentoml models import` should accept the HF URL instead of the specific format `hf:t5-base`. 
`bentoml models import https://huggingface.co/meta-llama/Llama-2-7b-chat-hf` vs `bentoml models import hf:meta-llama/Llama-2-7b-chat-hf`

src/bentoml_cli/models.py
```diff
@@ -233,13 +233,20 @@ def export(model_tag: str, out_path: str) -> None:  # type: ignore (not accessed
...
+        type=click.STRING,
+        default=None,
+        help="The name to give to the model in the BentoML store",
+    )
+    def import_from(model_path: str, name: str | None) -> None:  # type: ignore (not accessed)
         """Import a previously exported Model archive file
 
         bentoml models import ./my_model.bentomodel
         bentoml models import s3://mybucket/models/my_model.bentomodel
+        bentoml models import hf:t5-base --name myModelName
```
aarnphm, line 247: I don't think we should have a custom name support here.

What stops people from doing the following

```bash

bentoml models import s3://mybucket/models/my_model.bentomodel --name ADiffName
```

Importing HF models should just follow a specific format
```prolog
repo--model-name
```

We can save the model format (whether it is PyTorch, Flax, TF) in either labels or context.

src/bentoml/models.py
```diff
@@ -102,19 +103,49 @@ def import_model(
...
+
+        _, repo_name = path.split(":")
+        name = name if name is not None else repo_name.replace("/", "--")
+        try:
+            from huggingface_hub import repo_info
+
+            res = repo_info(repo_name)
+            for tag in res.tags:
+                if tag in FRAMEWORK_MAPPING:
+                    return FRAMEWORK_MAPPING[tag](name, repo_name)
```
aarnphm, line 142: Ugh do we know for sure that all models from `repo_info` will always return tags that contains `transformers` and `diffusers`?

See https://huggingface.co/docs/huggingface_hub/v0.18.0.rc0/en/package_reference/hf_api#huggingface_hub.hf_api.ModelInfo tags can be optional. So this is won't work for some cases.

Which is ok? probably we will need a warning after the iteration saying that it might not be a diffusers/transformers model

```python
for tag in res.tags:
  if tag in {'transformers', 'diffusers'}: return FRAMEWORK_AMPPING[tag](name, repo_name)
else:
  logger.warning("BentoML cannot automatically determine which framework to save the model (transformers, diffusers). Make sure to set the environment BENTOML_IMPORT_FRAMEWORK='transformers|diffusers' to explicitly set loading framework.
```

@larme what do you think about the envvar `BENTOML_IMPORT_FRAMEWORK`?

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -169,6 +172,42 @@ def _autoclass_converter(
...
+
+def _try_import_accelerate():
+    try:
+        import accelerate  # noqa: F401
+    except ImportError:  # pragma: no cover
+        raise MissingDependencyException(
+            "'accelerate' is required in order to download pretrained transformer models, install with 'pip install accelerate'.",
+        )
+
+
```
aarnphm, line 210: lets refactor this under `bentoml/_internal/frameworks/utils/transformers.py` to use for both diffusers and transformers.

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -605,6 +643,337 @@ def make_default_signatures(pretrained: t.Any) -> ModelSignaturesType:
...
+    proxies: dict[str, str] | None = None,
+    revision: str = "main",
+    variant: str | None = None,
+    sync_with_hub_version: bool = False,
+    signatures: ModelSignaturesType | None = None,
+    labels: dict[str, str] | None = None,
+    custom_objects: dict[str, t.Any] | None = None,
+    external_modules: t.List[ModuleType] | None = None,
+    metadata: dict[str, t.Any] | None = None,
+    # ...
```
aarnphm, line 661:
```suggestion
```

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -605,6 +643,337 @@ def make_default_signatures(pretrained: t.Any) -> ModelSignaturesType:
...
 
+def import_model(
+    name: Tag | str,
+    model_name_or_path: str | os.PathLike[str],
+    *,
+    task_name: str | None = None,
+    task_definition: dict[str, t.Any] | TaskDefinition | None = None,
+    proxies: dict[str, str] | None = None,
+    revision: str = "main",
+    variant: str | None = None,
```
aarnphm, line 654: You should handle huggingface_hub kwargs separately? This I assume is the kwargs for passing into huggingface_hub?

src/bentoml/_internal/frameworks/transformers.py
aarnphm: And you can make it `is_accelerate_available` and `is_huggingface_hub_available`?

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -605,6 +643,337 @@ def make_default_signatures(pretrained: t.Any) -> ModelSignaturesType:
...
+                "jax": get_pkg_version("jax"),
+                "jaxlib": get_pkg_version("jaxlib"),
+            }
+        )
+    context = ModelContext(
+        framework_name="transformers", framework_versions=framework_versions
+    )
+
+    if signatures is None:
+        signatures = {
```
aarnphm, line 798: make sure to use the `make_signatures` function to correctly generate the signature for this model

You should also change that function to just check for the model architecture fwiw.

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -605,6 +643,337 @@ def make_default_signatures(pretrained: t.Any) -> ModelSignaturesType:
...
+            raise BentoMLException(
+                "Cannot sync version with huggingface hub when importing a local model"
+            )
+    else:
+        try:
+            _try_import_accelerate()
+            from accelerate import init_empty_weights
+            from transformers.utils import cached_file
+
+            with init_empty_weights():
```
aarnphm, line 828: > [!NOTE]
> Check for offline support @aarnphm

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -605,6 +643,337 @@ def make_default_signatures(pretrained: t.Any) -> ModelSignaturesType:
...
+    if os.path.isdir(model_name_or_path):
+        src_dir = model_name_or_path
+        if sync_with_hub_version:
+            raise BentoMLException(
+                "Cannot sync version with huggingface hub when importing a local model"
+            )
+    else:
+        try:
+            _try_import_accelerate()
+            from accelerate import init_empty_weights
```
aarnphm, line 825: hmm this logics depends on accelerate is kinda weird.

Since it will probably end up going for snapshot_download nonetheless, since we can assume most people won't have accelerate install?

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -605,6 +643,337 @@ def make_default_signatures(pretrained: t.Any) -> ModelSignaturesType:
...
+                options_args = t.cast(
+                    "tuple[str, TaskDefinition]",
+                    check_task(task_name)[:2],
+                )
+                if task_name.startswith("translation"):
+                    options_args = (task_name, options_args[1])
+            except Exception:
+                raise BentoMLException(
+                    f"Task '{task_name}' is not a valid task for pipeline (available: {get_supported_tasks()}."
+                )
```
aarnphm, line 903: Ye I'm still not sure why we need to check for pipeline in transformers here?

can you give me an example where a hf repo save with pipeline support?

Is it a new feature?

aarnphm (changes requested):
cc @larme for another look

src/bentoml/_internal/frameworks/transformers.py
MingLiangDai: I don't think hf repo saves with pipeline support but from what I'm aware of, our `save_model` method supports pipelines. So I assumed we wanted a similar behavior for `import_model` as well. Below is a comparison:

```python
# saving a pipeline
pipeline = transformers.pipeline("summarization", "t5-base")
bentoml.transformers.save_model("t5-base-summarization-pipeline", pipeline)

# importing to save as a pipeline
bentoml.transformers.import_model('t5-base-summarization-pipeline', 't5-base', task_name='summarization')

# service.py
bentoml.transformers.get("t5-base-summarization-pipeline").to_runner()
```

If this feature is not necessary, I can remove this 😃

aarnphm: I think we have discussed this. `import_model` shouldn't load the model at all. using pipeline requires users to have the pipeline explicitly.

If users want to save pipeline, they can still use `save_model`.

src/bentoml/_internal/frameworks/transformers.py
larme: agree

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -605,6 +643,337 @@ def make_default_signatures(pretrained: t.Any) -> ModelSignaturesType:
...
+def import_model(
+    name: Tag | str,
+    model_name_or_path: str | os.PathLike[str],
+    *,
+    task_name: str | None = None,
+    task_definition: dict[str, t.Any] | TaskDefinition | None = None,
+    proxies: dict[str, str] | None = None,
+    revision: str = "main",
+    variant: str | None = None,
+    sync_with_hub_version: bool = False,
```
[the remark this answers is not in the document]
larme: I'm not sure if this should always be True. Maybe it can be True by default, but in some situation, user may want to have different tag version (e.g. user want more memorable version name)

src/bentoml/_internal/frameworks/transformers.py
larme: @aarnphm can we just ask user to install accelerate if it's not available? I think accelerate is helpful for loading LLM models?

src/bentoml/models.py
```diff
@@ -65,6 +65,7 @@ def import_model(
     params: t.Optional[t.Dict[str, str]] = None,
     subpath: t.Optional[str] = None,
     _model_store: "ModelStore" = Provide[BentoMLContainer.model_store],
+    name: t.Optional[str] = None,
```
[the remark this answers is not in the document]
larme: I think user can provide their own model name if they don't want the name generated by us?

src/bentoml/models.py
larme: @aarnphm Why not add the framework variable to the function's arguments list? Then let cli interface convert the env var to the function argument?

aarnphm: yes that works as well.

src/bentoml/models.py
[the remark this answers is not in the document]
aarnphm: but this should already be passed in the first argument right?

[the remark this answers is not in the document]
aarnphm: or persay, `import_model` should be `import_bentomodel`

src/bentoml/_internal/frameworks/transformers.py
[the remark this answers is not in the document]
aarnphm: > I'm not sure if this should always be True. Maybe it can be True by default, but in some situation, user may want to have different tag version (e.g. user want more memorable version name)

The argument for more memorable version name should be set on the CLI. I don't think that we even have support for custom version on `save_model`? It is all generated by us right?

src/bentoml/_internal/frameworks/transformers.py
aarnphm: afaik accelerate is mostly well known for ppl who train model with transformers. usually people won't like installing additional dependencies unless they have to.

I mean we can recommend them, sure, but we can't always rely that accelerate is always available unless it becomes `bentoml[transformers]` or `bentoml[huggingface]`

src/bentoml/_internal/utils/transformers.py
```diff
@@ -0,0 +1,43 @@
+from __future__ import annotations
```
aarnphm, line 1: Oh we already have an utils for framework under `bentoml/_internal/frameworks/utils`. You can move these function there.

src/bentoml/models.py
```diff
@@ -131,20 +136,36 @@ def import_model(
...
-        _, repo_name = path.split(":")
-        name = name if name is not None else repo_name.replace("/", "--")
+        model_id = parsedURL.path.strip("/")
+
+        if model_id == "":
+            raise BentoMLException(
+                "Invalid HuggingFace model URL, please check your URL"
+            )
+
+        bento_name = inflection.dasherize(model_id.replace("/", "--"))
```
aarnphm, line 146: not sure if we depends on inflection. check `pyproject.dependencies` for this.

src/bentoml_cli/models.py
```diff
@@ -240,13 +241,14 @@ def export(model_tag: str, out_path: str) -> None:  # type: ignore (not accessed
...
+        """Import a previously exported Model archive file or a model from remote model store
 
         bentoml models import ./my_model.bentomodel
         bentoml models import s3://mybucket/models/my_model.bentomodel
-        bentoml models import hf:t5-base --name myModelName
+        bentoml models import https://huggingface.co/t5-base --name myModelName
         """
-        bentomodel = import_model(model_path, name=name)
+        BENTOML_IMPORT_FRAMEWORK = os.environ.get("BENTOML_IMPORT_FRAMEWORK", None)
+        bentomodel = import_model(model_path, framework=BENTOML_IMPORT_FRAMEWORK)  # type: ignore
```
aarnphm, line 251: Why the `# type: ignore` here?

src/bentoml/_internal/frameworks/transformers.py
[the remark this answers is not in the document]
larme: I'm not sure. In diffusers we allow user to specify the model name, but use something like `diffusers_simple` to provide a way to auto generate the model name.

I think we should make `benotml.diffusers` and `bentoml.transformers` more generic and have more flexible apis for more general usage, and provide `benotml.diffusers_simple` and `bentoml.transformers_simple` to have a simplified api with some specified model in mind.

src/bentoml/_internal/frameworks/transformers.py
larme: @MingLiangDai how much work needed to fallback to your original implementation if accelerate is not available?

src/bentoml/models.py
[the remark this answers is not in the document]
MingLiangDai: Sorry I'm not too sure how we can get the name from the bentomodel? I believe the first argument is going to be the hf url in our case. Maybe the name could be retrieved from the bentomodel if the path is local or s3, but not hf?

src/bentoml/models.py
MingLiangDai: inflection is included in the dependencies list. so should be good 👍

src/bentoml_cli/models.py
MingLiangDai: Ah there was a mismatch for the framework param since BENTOML_IMPORT_FRAMEWORK could be any str and framework is only 'transformers|diffusers'. Think I just used the quick fix tool suggested by vscode. Shouldn't be a big deal. I will remove the `# type: ignore`

src/bentoml/_internal/frameworks/transformers.py
MingLiangDai: I did a bit of digging into `init_empty_weights`. It seems I can easily extract that code out of the library for our use case (without any additional dependencies that we don't already have). So I will put `init_empty_weights` into our code base in `utils/transformers.py` and reference that instead.

src/bentoml_cli/models.py
MingLiangDai: > We can save the model format (whether it is PyTorch, Flax, TF) in either labels or context.

Can I have a bit more detail on what you mean by this?

src/bentoml/models.py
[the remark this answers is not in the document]
larme: just synced with Sean, I think we can left cli part out of this pr. I will discuss with you later about cli ui design @MingLiangDai

larme (changes requested):

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -605,6 +607,262 @@ def make_default_signatures(pretrained: t.Any) -> ModelSignaturesType:
...
+        if not downloaded_model_successfully:
+            # clone entire repository as fallback
+            is_huggingface_hub_available()
+            from huggingface_hub import snapshot_download
+
+            src_dir = snapshot_download(
+                model_name_or_path,
+                proxies=proxies,
+                revision=revision,
+            )
```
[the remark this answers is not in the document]
MingLiangDai: I will give the user an option (`clone_repository`) to download the entire repository if that's their intention. This is added as an argument to the function.

src/bentoml_cli/models.py
aarnphm: you can make the framework literal type

src/bentoml_cli/models.py
aarnphm: `pt-meta-llama--Llama-2-13b-chat-hf`
or `flax-google--vit-transformers-384...`

larme (dismissed):
LGTM

larme:
@aarnphm could you take a look of the revised codes?

Our status is:
1. delay cli implementation, waiting to sync with BentoProject design
2. explicitly expose some huggingface from_pretrained arguments to user (because they are download/import related), but also allow user to pass extra hf arguments
3. tests and documentations will be in another pr by ming

src/bentoml/_internal/frameworks/utils/transformers.py
```diff
@@ -46,3 +54,87 @@ def is_tf_available():
...
+    return commit_hash if regex_commit_hash.match(commit_hash) else None
+
+
+def is_huggingface_hub_available():
+    try:
+        import huggingface_hub  # noqa: F401
+    except ImportError:  # pragma: no cover
+        raise MissingDependencyException(
+            "'huggingface_hub' is required in order to download pretrained models, install with 'pip install huggingface-hub'. For more information, refer to https://huggingface.co/docs/huggingface_hub/quick-start",
+        )
```
frostming, line 83:
```suggestion
        ) from None
```

This can make the traceback a bit cleaner

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -605,6 +607,266 @@ def make_default_signatures(pretrained: t.Any) -> ModelSignaturesType:
...
+    model = None
+    if os.path.isdir(model_name_or_path):
+        src_dir = model_name_or_path
+        if sync_with_hub_version:
+            raise BentoMLException(
+                "Cannot sync version with huggingface hub when importing a local model"
+            )
+    else:
+        try:
+            is_huggingface_hub_available()
```
frostming, line 754: What is this line used for?

src/bentoml/_internal/frameworks/utils/transformers.py
```diff
@@ -46,3 +54,87 @@ def is_tf_available():
...
+    search = re.search(r"snapshots/([^/]+)/", resolved_dir)
+
+    if search is None:
+        return None
+
+    commit_hash = search.groups()[0]
+    return commit_hash if regex_commit_hash.match(commit_hash) else None
+
+
+def is_huggingface_hub_available():
```
aarnphm, line 77: Lets synchronize with with `is_torch_available` implementation

```python
def is_huggingface_hub_available() -> bool:return importlib.util.find_spec("huggingface_hub") is not None
```

src/bentoml/_internal/frameworks/diffusers.py
```diff
@@ -525,7 +498,7 @@ def import_model(
             )
 
     elif pipeline_class:
-        _try_import_huggingface_hub()
+        is_huggingface_hub_available()
```
aarnphm, line 501: Why do we need to check for huggingface_hub available in this branch here?

src/bentoml/_internal/frameworks/diffusers.py
```diff
@@ -534,14 +507,14 @@ def import_model(
...
-            version = _extract_commit_hash(src_dir, REGEX_COMMIT_HASH)
+            version = extract_commit_hash(src_dir, REGEX_COMMIT_HASH)
             if version is not None:
                 if variant is not None:
                     version = version + "-" + variant
                 tag.version = version
 
     else:
-        _try_import_huggingface_hub()
+        is_huggingface_hub_available()
```
aarnphm, line 517: Lets' then handle the exception for missing imports here.

src/bentoml/_internal/frameworks/utils/transformers.py
```diff
@@ -46,3 +54,87 @@ def is_tf_available():
...
+    try:
+        import huggingface_hub  # noqa: F401
+    except ImportError:  # pragma: no cover
+        raise MissingDependencyException(
+            "'huggingface_hub' is required in order to download pretrained models, install with 'pip install huggingface-hub'. For more information, refer to https://huggingface.co/docs/huggingface_hub/quick-start",
+        ) from None
+
+
+@contextmanager
+def init_empty_weights(include_buffers: bool = False):
```
aarnphm, line 87: Make sure to state why we vendor from accelerate

src/bentoml/_internal/frameworks/transformers.py
aarnphm, line 754:
```suggestion
```

aarnphm (changes requested):
final comments

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -605,6 +607,266 @@ def make_default_signatures(pretrained: t.Any) -> ModelSignaturesType:
...
+                    resume_download=resume_download,
+                    **extra_hf_hub_kwargs,
+                )
+            else:
+                from transformers.utils import cached_file
+
+                from .utils.transformers import init_empty_weights
+
+                with init_empty_weights():
+                    model = transformers.AutoModel.from_pretrained(
```
aarnphm, line 772: Btw, this will still download the model to cache. So on BentoCloud, once the pod once initialised, it will redownload again at runtime. So if users use `bentoml.transformers.import_model` in the service code this might happen.

So tbh I'm not sure. Maybe we should have a check before on L700 if the tag is already exists then we return the Model ref immediately.

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -605,6 +607,266 @@ def make_default_signatures(pretrained: t.Any) -> ModelSignaturesType:
...
+                    path_to_config = cached_file(
+                        model_name_or_path,
+                        CONFIG_JSON_FILE_NAME,
+                        proxies=proxies,
+                        revision=revision,
+                        force_download=force_download,
+                        resume_download=resume_download,
+                        **extra_hf_hub_kwargs,
+                    )
+                    src_dir = os.path.dirname(path_to_config)
```
aarnphm, line 789: This is actually incorrect, since the path_to_config is a symlink.

src/bentoml/_internal/frameworks/transformers.py
MingLiangDai: afaik, from_pretrained should be smart enough to detect if a certain version of a certain model has already been downloaded. It will download if not available otherwise, it grabs directly from the cache. 

I'm not too knowledgeable about the workflow on BentoCloud, can you give me a bit more info on what you mean? What do you mean by 'redownload' once the pod has been initialised?

src/bentoml/_internal/frameworks/diffusers.py
MingLiangDai: This is diffusers which is something Larme worked on so I'm not entirely sure. But a couple lines down, `from huggingface_hub.file_download import REGEX_COMMIT_HASH` seems to use huggingface hub.

src/bentoml/_internal/frameworks/diffusers.py
MingLiangDai: Got it 👍. I assume this is a follow up on the changes you suggested for `is_huggingface_hub_available`

src/bentoml/_internal/frameworks/transformers.py
MingLiangDai: The updated code now raises MissingDependencyException using is_huggingface_hub_available

src/bentoml/_internal/frameworks/transformers.py
aarnphm: I will always download the models if there is no cache. On bentocloud, we won't have the HUGGINGFACE_HUB_CACHE since the model is saved into bentoml models. So it will always download. Can you test it on bentocloud?

src/bentoml/_internal/frameworks/diffusers.py
```diff
@@ -525,7 +498,10 @@ def import_model(
             )
 
     elif pipeline_class:
-        _try_import_huggingface_hub()
+        if not is_huggingface_hub_available():
+            raise MissingDependencyException(
+                "'huggingface_hub' is required in order to download pretrained models, install with 'pip install huggingface-hub'. For more information, refer to https://huggingface.co/docs/huggingface_hub/quick-start",
+            ) from None
```
aarnphm, line 504: still not sure why we need to check this here. Since we don't use huggingface_hub here?

src/bentoml/_internal/frameworks/utils/transformers.py
```diff
@@ -46,3 +52,84 @@ def is_tf_available():
...
+    commit_hash = search.groups()[0]
+    return commit_hash if regex_commit_hash.match(commit_hash) else None
+
+
+def is_huggingface_hub_available() -> bool:
+    return importlib.util.find_spec("huggingface_hub") is not None
+
+
+# init_empty_weights is vendored from accelerate. This is a useful function as it allows us to call transformers.AutoModel.from_pretrained without loading
+# the actual weights of the model. This grants access to the model class and its attributes without using a significant amount of memory.
```
aarnphm, line 80: I was meant to put this into the docstring 😄

src/bentoml/_internal/frameworks/transformers.py
MingLiangDai: After playing around with BentoCloud for a while, I have a few questions:

1. Would users call `bentoml.transformers.import_model` during runtime code? To my understanding, the normal workflow would be to save models to bento model store using `import_model` or `save_model`. Then we create a bento and push to bento cloud. I don't see the likelihood of users calling `import_model` during runtime?
2. Is this the change you have in mind:
```
tag = Tag.from_taglike(name)
try: 
    model = bentoml.models.get(tag)
    return model
except bentoml.exceptions.NotFound:
    pass
```

src/bentoml/_internal/frameworks/diffusers.py
MingLiangDai: `pipeline_class.download()` uses `snapshot_download` and `hf_hub_download` under the hood

src/bentoml/_internal/frameworks/transformers.py
aarnphm: An example service file
```python

runner = bentoml.transformers.import_model("llama-7b", "meta-llama/Llama-2-7b-chat-hf").to_runner()

svc = bentoml.Service('llama-chat', runners=[runner])

```

larme: I guess here `bentoml.transformers.import_model` is for importing models before packaging the bento and in service.py user should use `bentoml.transformers.get` instead of `bentoml.transformers.import_model` to get the bento Model. Does this work flow make sense for you?

src/bentoml/_internal/frameworks/diffusers.py
aarnphm: Then it should be their responsibility to check whether huggingface_hub is available or not right?

src/bentoml/_internal/frameworks/transformers.py
aarnphm: ye I guess that works. I guess this would be more structured once we have BentoProject. Since I feel like having just one file that has one line to imports the models is excessive. Since `bentoml.transformers.import_model` returns a `bentoml.Model` users might do this

I think for correctness I suggest `import_model` for both transformers and diffusers not to return anything. cc @larme ?

src/bentoml/_internal/frameworks/diffusers.py
MingLiangDai: hmm, I don't see that they check for `huggingface_hub` availability in their code. I believe adding this check on our end is not too costly performance-wise and also doesn't take away code legibility.

src/bentoml/_internal/frameworks/transformers.py
MingLiangDai: > For correctness I suggest `import_model` for both transformers and diffusers not to return anything

Doesn't this break the backwards compatibility of users who already use diffusers import_model. 

e.g. if we make the above change, users who have done `diffusers.import_model().to_runner()` would now receive errors.

MingLiangDai: I do agree with Aaron on the above point regarding some users may want their setup code and runtime code together in one service.py file (e.g doing `bentoml.transformers.import_model().to_runner()` during runtime). Though this is most likely a bad practice. I can still add a check on the tag and call bentoml.models.get on the tag at the beginning of the function.

aarnphm: I think since we never document that behaviour, it is fine. @larme are you ok with breaking this?

Tho I do agree `import_model().to_runner()` is a very very bad practice.

src/bentoml/_internal/frameworks/diffusers.py
aarnphm: Actually you don't have to check this at all, since `huggingface-hub` is a core dependency of diffusers.

See https://github.com/huggingface/diffusers/blob/dd9a5caf61f04d11c0fa9f3947b69ab0010c9a0f/setup.py#L237

src/bentoml/_internal/frameworks/transformers.py
larme: I guess `import_model` that return a bento model is helpful in notebook examples. For example:

```python
bento_model = import_model(...)

# then after several cells
runner = bento_model.with_options(...).to_runner()
```

I think just don't emphasize it in our documentation should be OK. So user don't have an understanding of the relationship bewteeen bento model and runner won't do that?

src/bentoml/_internal/frameworks/diffusers.py
MingLiangDai: I see, I will remove the check. It seems for transformers, huggingface-hub is also a core dependency so I will remove the check in transformers as well.

src/bentoml/_internal/frameworks/transformers.py
aarnphm: Hmm, if we document in jupyter notebook then ppl will def put it into their service file.

aarnphm: But I'm fine with this as well.

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -605,6 +606,279 @@ def make_default_signatures(pretrained: t.Any) -> ModelSignaturesType:
...
+                        tag.version = version
+        except Exception:
+            logger.info(
+                "Failed to download model file for (%s).",
+                model_name_or_path,
+            )
+            raise
+
+    if model is None:
+        from .utils.transformers import init_empty_weights
```
aarnphm, line 817: Why there are two different import on L817 and L781 here? Shouldn't this just be on the top of the files?

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -605,6 +606,279 @@ def make_default_signatures(pretrained: t.Any) -> ModelSignaturesType:
...
+                src_dir = snapshot_download(
+                    model_name_or_path,
+                    proxies=proxies,
+                    revision=revision,
+                    force_download=force_download,
+                    resume_download=resume_download,
+                    **extra_hf_hub_kwargs,
+                )
+            else:
+                from transformers.utils import cached_file
```
aarnphm, line 779: THis can move below L106. We already eager import transformers

aarnphm (changes requested):
Final comments

src/bentoml/_internal/frameworks/transformers.py
MingLiangDai: oops, I think this was here because we had a check for accelerate availability before. Since we no longer depend on accelerate, I will move this up to the top.

aarnphm (approved):
lets get this in

PR #4255, opened 2023-11-02
fix(client): prepend http if necessary in wait

sauyon:


horw:
Hi, what about this method?
https://github.com/bentoml/BentoML/blob/34e0d4d4cb03020db6d574360c124e484720e101/src/bentoml/_internal/client/http.py#L192

horw (changes requested):
Hi, what about this method?
https://github.com/bentoml/BentoML/blob/34e0d4d4cb03020db6d574360c124e484720e101/src/bentoml/_internal/client/http.py#L192

aarnphm (approved):

sauyon:
Er, I'll create a new PR for the sync client I guess.

Issue #4217, opened 2023-10-02
bug: BentoML unexpected behavior when interacting with Editable Pip Modules


tokotchd:
### Describe the bug

If you happen to be using bentoml as a build server involving modular code that you're installing via `pip install -e <path_to_folder>`, you're going to encounter very strange/bad behavior:

1. Incorrect models, services, and runners being loaded
2. Inconsistent behavior between `bentoml build` and `bentoml build --containerize`
3. Possible dependency import issues
etc.

### To reproduce

Simplest possible reproduction of the bug:

Start with the following folder structure (describable as a bento project overtop of an installable python module):
```
- module_1
-- module1
--- __init__.py
-- bento_packer.py
-- bento_service.py
-- bentofile.yaml
-- setup.py
- module_2
-- module2
--- __init__.py
-- bento_packer.py
-- bento_service.py
-- bentofile.yaml
-- setup.py
```
Then, install both python modules:
`pip install -e module_1`
`pip install -e module_2`
Finally, attempt to `bentoml build` in `module_2`
> Error: [bentoml-cli] `build` failed: no Models with name 'module_1' exist in BentoML store <osfs '/home/user/bentoml/models'>

If we look at the --verbose flag, we're seeing that while the build process starts with the correct bentofile, it manages to lose it during the builds/imports process and ends up using another's instead due to the symlinks created by `pip install -e`
```
Importing service "bento_service:svc" from working dir: "/home/user/bentoml_bug/module_2"
...
 File "/home/user/bentoml_bug/module_1/bento_service.py", line 3, in <module>
    runner = bentoml.picklable_model.get("module_1:latest").to_runner()
...
bentoml.exceptions.NotFound: no Models with name 'module_1' exist in BentoML store <osfs '/home/user/bentoml/models'>
```
[bentoml_bug.zip](https://github.com/bentoml/BentoML/files/12785928/bentoml_bug.zip)

### Expected behavior

The ideal behavior would be that coincidentally having a `bentofile.yaml` anywhere on the system that happens to be in a python module installed with `pip install -e` will not break the `bentoml build` CLI.

I'm guessing the behavior I'm seeing is more due to path/symlink changes caused by pip than bentoML itself.
If the fix is outside of the scope of bentoML than I'm happy to at least have documented the issue here with the ask of adding something to BentoML documentation that the `bentoML` cli does not play well with editable python modules.

### Environment

bentoml==1.1.6
Python 3.9.15
Ubuntu 20.04

[...]

PR #4256, opened 2023-11-02
fix: Path issues with pip modules

tokotchd:
Always prepend current bentofile directory to system path to avoid unwanted behavior when other bentofiles are on the system PATH
This is especially evident when trying to use bentoml cli on python folders that are also pip modules see https://github.com/bentoml/BentoML/issues/4217

### DIFF
--- a/src/bentoml/_internal/service/loader.py
+++ b/src/bentoml/_internal/service/loader.py
@@ -71,15 +71,15 @@ def import_service(
         os.chdir(prev_cwd)
         BentoMLContainer.model_store.set(global_model_store)
 
-    try:
-        if working_dir is not None:
-            working_dir = os.path.realpath(os.path.expanduser(working_dir))
-            # Set cwd(current working directory) to the Bento's project directory,
-            # which allows user code to read files using relative path
-            os.chdir(working_dir)
-        else:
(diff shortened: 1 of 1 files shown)


sauyon (approved):
LGTM, thank you!

src/bentoml/_internal/service/loader.py
```diff
@@ -172,7 +169,7 @@ def recover_standalone_env_change():
         object.__setattr__(instance, "_import_str", f"{module_name}:{attrs_str}")
         return instance
     except ImportServiceError:
-        if sys_path_modified and working_dir:
+        if working_dir:
```
parano, line 172: we probably don't need this check neither now?

sauyon: I think this check is ensuring that if the assignment fails it doesn't try to remove `None`.

parano: Make sense, I'd suggest we move the working_dir assignment L75-81 out of the `try` block and remove this `if working_dir` check as well

parano: @sauyon made a quick patch, could you help take a look?

PR #4271, opened 2023-11-09
fix(transformers): support `trust_remote_code` and added unit tests

MingLiangDai:
## What does this PR address?

Add unit tests to transformers import model

## Before submitting:

### DIFF
--- a/src/bentoml/_internal/frameworks/transformers.py
+++ b/src/bentoml/_internal/frameworks/transformers.py
@@ -844,8 +844,8 @@ def import_model(
     if model is None:
         with init_empty_weights():
             if is_auto_class:
-                # using .from_config with init_empty_weights() is a sure-fire way to get an instance of a model class without loading the model weights
-                # for a detailed discussion between the use of .from_pretrained and .from_config with init_empty_weights(), check out https://github.com/huggingface/accelerate/issues/2163
+                # NOTE: Under `init_empty_weights`, `.from_config` won't load the model weights. Whereas for `.from_pretrained`, transformers needs to find out what layers to load with the architecture, thereby loading the models into memory, then unload afterwards.
+                # See https://github.com/huggingface/accelerate/issues/2163 for more information.
                 model = pretrained_model_class.from_config(
                     trust_remote_code=trust_remote_code,
                     config=config,
--- a/src/bentoml/_internal/frameworks/transformers.py
+++ b/src/bentoml/_internal/frameworks/transformers.py
@@ -766,11 +766,8 @@ def import_model(
     if getattr(config, "architectures", None):
         try:
             pretrained_model_class = getattr(transformers, config.architectures[0])
-            logger.info(
-                f"pretrained_model_class is not provided, bentoml will create a model with the following pretrained model class {config.architectures[0]}. Available pretrained classes for this model: {config.architectures}."
-            )
         except AttributeError:
-            logger.info("Config architecture does not exist in transformers module.")
+            pass
     if getattr(config, "auto_map", None):
         for auto_class in config.auto_map:
             if "AutoModel" in auto_class:
--- a/src/bentoml/_internal/frameworks/transformers.py
+++ b/src/bentoml/_internal/frameworks/transformers.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import inspect
 import logging
 import os
 import platform
@@ -750,10 +751,29 @@ def import_model(
         resume_download=resume_download,
         **extra_hf_hub_kwargs,
     )
+
+    is_auto_class = False
     if pretrained_model_class is None:
-        pretrained_model_class = getattr(transformers, config.architectures[0])
-        logger.info(
-            f"pretrained_model_class is not provided, bentoml will create a model with the following pretrained model class {config.architectures[0]}. Available pretrained classes for this model: {config.architectures}."
+        if getattr(config, "architectures", None):
+            try:
+                pretrained_model_class = getattr(transformers, config.architectures[0])
+                logger.info(
+                    f"pretrained_model_class is not provided, bentoml will create a model with the following pretrained model class {config.architectures[0]}. Available pretrained classes for this model: {config.architectures}."
+                )
+            except AttributeError:
+                logger.info(
+                    "Config architecture does not exist in transformers module."
+                )
+        if getattr(config, "auto_map", None):
+            for auto_class in config.auto_map:
+                if auto_class.startswith("AutoModel"):
+                    pretrained_model_class = getattr(transformers, auto_class)
+                    is_auto_class = True
+                    break
+
+    if pretrained_model_class is None:
+        raise BentoMLException(
+            "BentoML cannot automatically determine the pretrained model class/architecture for the given model. Please explicitly set the pretrained_model_class argument."
         )
 
     model = None
@@ -768,13 +788,19 @@ def import_model(
             if clone_repository:
                 from huggingface_hub import snapshot_download
 
+                # filter out kwargs as snapshot_download my not accept some kwargs
+                params = inspect.signature(snapshot_download).parameters.values()
+                param_names = {param.name for param in params}
+                input_params = dict(
+                    filter(lambda x: x[0] in param_names, extra_hf_hub_kwargs.items())
+                )
                 src_dir = snapshot_download(
                     model_name_or_path,
                     proxies=proxies,
                     revision=revision,
                     force_download=force_download,
                     resume_download=resume_download,
-                    **extra_hf_hub_kwargs,
+                    **input_params,
                 )
             else:
                 with init_empty_weights():
@@ -812,7 +838,12 @@ def import_model(
 
     if model is None:
         with init_empty_weights():
-            model = pretrained_model_class(config=config)
+            if is_auto_class:
+                model = pretrained_model_class.from_config(
+                    config=config, **extra_hf_hub_kwargs
+                )
+            else:
+                model = pretrained_model_class(config=config)
 
     pretrained = t.cast("PreTrainedProtocol", model)
     assert all(
--- a/src/bentoml/_internal/frameworks/transformers.py
+++ b/src/bentoml/_internal/frameworks/transformers.py
@@ -844,6 +844,8 @@ def import_model(
     if model is None:
         with init_empty_weights():
             if is_auto_class:
+                # using .from_config with init_empty_weights() is a sure-fire way to get an instance of a model class without loading the model weights
+                # for a detailed discussion between the use of .from_pretrained and .from_config with init_empty_weights(), check out https://github.com/huggingface/accelerate/issues/2163
                 model = pretrained_model_class.from_config(
                     trust_remote_code=trust_remote_code,
                     config=config,
--- a/src/bentoml/_internal/frameworks/transformers.py
+++ b/src/bentoml/_internal/frameworks/transformers.py
@@ -777,7 +777,7 @@ def import_model(
 
     if pretrained_model_class is None:
         raise BentoMLException(
-            "BentoML cannot automatically determine the pretrained model class/architecture for the given model. Please explicitly set the pretrained_model_class argument."
+            "BentoML cannot automatically determine the pretrained model class/architecture for the given model."
         )
 
     model = None
--- a/src/bentoml/_internal/frameworks/transformers.py
+++ b/src/bentoml/_internal/frameworks/transformers.py
@@ -113,7 +113,6 @@ MODULE_NAME = "bentoml.transformers"
 API_VERSION = "v2"
 PIPELINE_PICKLE_NAME = f"pipeline.{API_VERSION}.pkl"
 PRETRAINED_PROTOCOL_NAME = f"pretrained.{API_VERSION}.pkl"
-CONFIG_JSON_FILE_NAME = "config.json"
 
 logger = logging.getLogger(__name__)
 
@@ -529,7 +528,7 @@ def load_model(bento_model: str | Tag | Model, *args: t.Any, **kwargs: t.Any) ->
                 raise
 
 
-def make_default_signatures(pretrained: t.Any) -> ModelSignaturesType:
+def make_default_signatures(pretrained_cls: t.Any) -> ModelSignaturesType:
     default_config = ModelSignature(batchable=False)
     infer_fn = ("__call__",)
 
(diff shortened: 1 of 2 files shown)


tests/integration/frameworks/test_transformers_unit.py
```diff
@@ -291,3 +291,83 @@ def test_custom_pipeline(pair_classification_pipeline: PairClassificationPipelin
...
+    assert os.path.exists(bento_model.path_of("pytorch_model.bin"))
+    assert os.path.exists(bento_model.path_of("pretrained.v2.pkl"))
+    assert os.path.exists(bento_model.path_of("model.yaml"))
+    assert bento_model.info.labels == labels
+    bentoml.models.delete("tiny-bert:3956d303d3cddf0708ff20660c1ea5f6ec30e434")
+
+
+def test_import_model_with_pretrained_class():
+    revision = "3956d303d3cddf0708ff20660c1ea5f6ec30e434"
+    pretrained_class = transformers.BertForMaskedLM
```
aarnphm, line 351: I assume this also work for a remote model? Or trust_remote_code=True?

aarnphm (commented):
Actually how to do we import a trust_remote_code=True model? Sth like chatglm or baichuan

tests/integration/frameworks/test_transformers_unit.py
MingLiangDai: Yup using trust_remote_code=True does work. But testing for this functionality actually brought up an issue that I was not aware of before. Tested with Baichuan and chatglm, it seems that they have their own predefined architecture (not one of the pretrained ones from transformers) which doesn't seem to be compatible with `init_empty_weights` from `accelerate`. 

The only work around I found was to snapshot download the repo as that does not use `.from_pretrained`, thus not loading the model weights. This is not the ideal behavior but the transformers library seems to be lacking in terms of what we want to accomplish 😢
I have been following this thread: https://discuss.huggingface.co/t/downloading-a-model-from-the-hub-without-loading-it/48377

With the new changes I added, the below should work:
`bentoml.transformers.import_model('baichuan-test', 'baichuan-inc/Baichuan-7B', trust_remote_code=True, clone_repository=True)`

aarnphm: The from_pretrained should work with `trust_remote_code` path right? it is a trust_remote_code path in the meta class AutoModel.

But do we need the class itself for loading? we can just pickled the class afterwards right?

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -750,10 +751,29 @@ def import_model(
...
     )
+
+    is_auto_class = False
     if pretrained_model_class is None:
-        pretrained_model_class = getattr(transformers, config.architectures[0])
-        logger.info(
-            f"pretrained_model_class is not provided, bentoml will create a model with the following pretrained model class {config.architectures[0]}. Available pretrained classes for this model: {config.architectures}."
+        if getattr(config, "architectures", None):
+            try:
+                pretrained_model_class = getattr(transformers, config.architectures[0])
```
aarnphm, line 759: You shouldn't do this eagerly, I think having the string is fine. Transformers has a mapping maintaining all of the supported architecture (Either CONFIG_MAPPING or AUTO_MODEL_MAPPING or sth)

Then it won't break with trust_remote_code. We can't really rely on the models on the Hub to always contain the `auto_map`

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -750,10 +751,29 @@ def import_model(
         resume_download=resume_download,
         **extra_hf_hub_kwargs,
     )
+
+    is_auto_class = False
```
aarnphm, line 755: I don't think we should depend on this AutoModel class behaviour

tests/integration/frameworks/test_transformers_unit.py
MingLiangDai: from_pretrained does work with the `trust_remote_code` kwarg. 

However, the problem comes when `config.architectures` specifies an architecture not supported by transformers. When this happens, the repository comes with its own `modeling_xyz.py`, which defines its own pretrained model class. `init_empty_weights` is used to instantiate the model class without actually loading the weights, but it doesn't seem to be compatible with custom pretrained classes (so those that are not supported by transformers by default).

Therefore, the problem right now is that custom pretrained model classes will actually be loaded into memory. The only way to avoid it is to set the `clone_repository` kwarg. This will call snapshot download and just downloads the entire repository instead of being selective of what files we want (which from_pretrained was primarily used for).

MingLiangDai: > But do we need the class itself for loading? we can just pickled the class afterwards right?

An instance of the class is needed for generating default signatures.

src/bentoml/_internal/frameworks/transformers.py
MingLiangDai: I can only see two ways for us to automatically determine the model class for the user (config.architectures or config.auto_map). So if a repository does not have `architectures` AND `auto_map`, then `import_model` can't determine the model class.

tests/integration/frameworks/test_transformers_unit.py
aarnphm: This is a intended behaviour of trust_remote_code afaik. You don't really have any other way other than loading the model into memory right?

You have to load the model into memory with trust_remote_code to verify so basically I think it is fine.

src/bentoml/_internal/frameworks/transformers.py
aarnphm: What happens if `architectures=[]`?This will fail

MingLiangDai: Yeah if architectures=[], then will try to look at auto_map. If both empty, then there's nothing more we could do 😞

MingLiangDai: > You shouldn't do this eagerly, I think having the string is fine. Transformers has a mapping maintaining all of the supported architecture (Either CONFIG_MAPPING or AUTO_MODEL_MAPPING or sth)

I assume what you wanted was for me to use the `MODEL_MAPPING` and check if `config.architectures[0]` was in the map. But after playing around with `MODEL_MAPPING` and `CONFIG_MAPPING`, it appears that they are quite expensive to use. They are lazy loading maps that, when initialized, will reference a lot of classes in transformers. When I referenced the map, it took about a minute for the map to initialize. 

Below is a sample of the python output of `MODEL_MAPPING`: 
```
[(<class 'transformers.models.albert.configuration_albert.AlbertConfig'>, <class 'transformers.models.albert.modeling_albert.AlbertModel'>),
(<class 'transformers.models.align.configuration_align.AlignConfig'>, <class 'transformers.models.align.modeling_align.AlignModel'>),
```

src/bentoml/_internal/frameworks/transformers.py
MingLiangDai: This AutoClass flag is used for when we fall back to using `config.auto_map` when `config.architectures` is unavailable. 

`config.auto_map` gives us something like `transformers.AutoModel...xyz`, where as `config.architectures` will give us something like `transformers.T5ForConditionalGeneration` (the exact model class). 

This flag is only used for local imports or snapshot download imports in the following way:
```
if is_auto_class:
    # when we don't have the exact model class, so we call from_config
    model = pretrained_model_class.from_config(
        config=config, **extra_hf_hub_kwargs
    )
else:
    # we have the exact model class, so we initalize the class using config
    model = pretrained_model_class(config=config)
```

src/bentoml/_internal/frameworks/transformers.py
aarnphm: They have the name variant, we don't need to use the lazy map, so `MODEL_MAPPING_NAMES` and `CONFIG_MAPPING_NAMES`

MingLiangDai: I've also taken a look into the name variants, but they don't seem to work for our use case. 
`MODEL_NAMES_MAPPING` has something like this:
```
("albert", "ALBERT"),
        ("align", "ALIGN"),
        ("altclip", "AltCLIP"),
```
But `config.architectures` contains a string that is the name of a model class (eg. T5ForConditionalGeneration). I've found another map that has something similar to what we need but it has the names of the base classes only (eg. T5Model). 

Overall, grabbing the attribute from `transformers` inside of a try block might be the most robust approach.

aarnphm: We want to use `CONFIG_MAPPING_NAMES` to get the config class, then use `from_config` right?

MingLiangDai: calling `from_config` will still require the model class, not the config class. The model class is required for calling both `from_pretrained` and `from_config`.

aarnphm: Doesn't AutoModel has a from_config option?

MingLiangDai: Yes, AutoModel and equivalents (eg. AutoModelForCasualLM, AutoModelForSeq2Seq, etc) all have a from_config option. But I don't think CONFIG_MAPPING_NAMES and MODEL_NAMES_MAPPING can assist us in this case.

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -750,10 +751,29 @@ def import_model(
         resume_download=resume_download,
         **extra_hf_hub_kwargs,
```
aarnphm, line 752: shouldn't trust_remote_code will also have to pass it here as well?

src/bentoml/_internal/frameworks/transformers.py
aarnphm: I think you can get MODEL_MAPPING[config_class], since __getitem__ is lazy see https://github.com/huggingface/transformers/blob/9beb2737d758160e845b66742a0c01201e38007f/src/transformers/models/auto/auto_factory.py#L734

src/bentoml/_internal/frameworks/transformers.py
MingLiangDai: trust_remote_code is packaged into extra_hf_hub_kwargs here. Alternatively, if you think trust_remote_code should be explicitly declared as a keyword arg, I can make that change as well.

aarnphm: it should

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -812,7 +838,12 @@ def import_model(
 
     if model is None:
         with init_empty_weights():
-            model = pretrained_model_class(config=config)
+            if is_auto_class:
+                model = pretrained_model_class.from_config(
+                    config=config, **extra_hf_hub_kwargs
+                )
+            else:
+                model = pretrained_model_class(config=config)
```
aarnphm, line 846:
```suggestion
model = pretrained_model_class.from_pretrained(model_id_or_path, config=config, **extra_hf_hub_kwargs)
```

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -768,13 +788,19 @@ def import_model(
...
+                    filter(lambda x: x[0] in param_names, extra_hf_hub_kwargs.items())
+                )
                 src_dir = snapshot_download(
                     model_name_or_path,
                     proxies=proxies,
                     revision=revision,
                     force_download=force_download,
                     resume_download=resume_download,
-                    **extra_hf_hub_kwargs,
+                    **input_params,
```
aarnphm, line 803: config = transformers.AutoConfig.from_pretrained(src_dir)
model = getattr(transformers, config['architectures'][0])

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -743,33 +746,37 @@ def import_model(
...
-            for auto_class in config.auto_map:
-                if auto_class.startswith("AutoModel"):
-                    pretrained_model_class = getattr(transformers, auto_class)
-                    is_auto_class = True
-                    break
+    pretrained_model_class = None
+    if getattr(config, "architectures", None):
+        try:
+            pretrained_model_class = getattr(transformers, config.architectures[0])
+            logger.info(
```
aarnphm, line 769: This line is not relevant anymore innit?

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -743,33 +746,37 @@ def import_model(
...
+                f"pretrained_model_class is not provided, bentoml will create a model with the following pretrained model class {config.architectures[0]}. Available pretrained classes for this model: {config.architectures}."
+            )
+        except AttributeError:
+            logger.info("Config architecture does not exist in transformers module.")
```
aarnphm, line 773: no need to info log about this. just pass. we can raise an example if both config.auto_map and architectures is not found

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -784,11 +814,12 @@ def import_model(
                         revision=revision,
                         force_download=force_download,
                         resume_download=resume_download,
+                        trust_remote_code=trust_remote_code,
                         **extra_hf_hub_kwargs,
                     )
                     path_to_config = transformers.utils.cached_file(
                         model_name_or_path,
-                        CONFIG_JSON_FILE_NAME,
+                        transformers.CONFIG_NAME,
```
aarnphm, line 822: Do you know that this is stable?

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -812,7 +843,14 @@ def import_model(
 
     if model is None:
         with init_empty_weights():
-            model = pretrained_model_class(config=config)
+            if is_auto_class:
+                model = pretrained_model_class.from_config(
```
aarnphm, line 847: probably add a note, and link to marc response on accelerate https://github.com/huggingface/accelerate/issues/2163#event-11018561032 explaining why we do this.

aarnphm (changes requested):
Final comments, then GTG

src/bentoml/_internal/frameworks/transformers.py
MingLiangDai: This is referenced by multiple files in the transformers library. I think it is unlikely they will just change this constant.

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -742,18 +746,38 @@ def import_model(
...
+                pretrained_model_class = getattr(transformers, auto_class)
+                is_auto_class = True
+                break
+
     if pretrained_model_class is None:
-        pretrained_model_class = getattr(transformers, config.architectures[0])
-        logger.info(
-            f"pretrained_model_class is not provided, bentoml will create a model with the following pretrained model class {config.architectures[0]}. Available pretrained classes for this model: {config.architectures}."
+        raise BentoMLException(
+            "BentoML cannot automatically determine the pretrained model class/architecture for the given model. Please explicitly set the pretrained_model_class argument."
```
aarnphm, line 780: There is no `pretrained_model_class` anymore right?

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -812,7 +843,16 @@ def import_model(
 
     if model is None:
         with init_empty_weights():
-            model = pretrained_model_class(config=config)
+            if is_auto_class:
+                # using .from_config with init_empty_weights() is a sure-fire way to get an instance of a model class without loading the model weights
+                # for a detailed discussion between the use of .from_pretrained and .from_config with init_empty_weights(), check out https://github.com/huggingface/accelerate/issues/2163
```
aarnphm, line 848:
```suggestion
                # NOTE: Under `init_empty_weights`, `.from_config` won't load the model weights. Whereas for `.from_pretrained`, transformers needs to find out what layers to load with the architecture, thereby loading the models into memory, then unload afterwards.
                # See https://github.com/huggingface/accelerate/issues/2163 for more information.
```

aarnphm (approved):

PR #4282, opened 2023-11-15
docs: Add docs for new transformers model import API

Sherlock113:
## What does this PR address?

Add docs for https://github.com/bentoml/BentoML/pull/4247.

## Before submitting:

### DIFF
--- a/docs/source/frameworks/diffusers.rst
+++ b/docs/source/frameworks/diffusers.rst
@@ -19,6 +19,10 @@ You can import a pretrained diffusion model from huggingface hub or local direct
 	"stabilityai/stable-diffusion-2-1",  # huggingface model name
     )
 
+.. note::
+
+    ``bentoml.diffusers.save_model`` can also be used to import diffusion models from Hugging Face, but it requires models to be loaded into memory first,
+    which is resource-intensive for large models. By contrast, ``bentoml.diffusers.import_model`` imports diffusion models directly without loading them into memory.
 
 If you plan to use the model with a custom pipeline that has method other than :code:`__call__` (e.g. a :code:`StableDiffusionMegaPipeline`), you need to explicitly list them like this:
 
--- a/docs/source/frameworks/diffusers.rst
+++ b/docs/source/frameworks/diffusers.rst
@@ -22,7 +22,7 @@ You can import a pretrained diffusion model from huggingface hub or local direct
 .. note::
 
     ``bentoml.diffusers.save_model`` can also be used to import diffusion models from Hugging Face, but it requires models to be loaded into memory first,
-    which is resource-intensive for large models. By contrast, ``bentoml.diffusers.import_model`` imports diffusion models directly without loading them into memory.
+    which is resource-intensive for large models. By contrast, ``bentoml.diffusers.import_model`` tries to import diffusion models directly without loading them into memory.
 
 If you plan to use the model with a custom pipeline that has method other than :code:`__call__` (e.g. a :code:`StableDiffusionMegaPipeline`), you need to explicitly list them like this:
 
--- a/docs/source/frameworks/transformers.rst
+++ b/docs/source/frameworks/transformers.rst
@@ -110,7 +110,7 @@ The ``bentoml.transformers.import_model`` function has two required parameters:
 * ``name``: The name of the model in the BentoML Model Store.
 * ``model_name_or_path``: This can be a string, a Hugging Face repository identifier (repo_id), or a directory path containing weights saved using ``transformers.AutoModel.save_pretrained`` (for example, ``./my_pretrained_directory/``).
 
-When importing models from repositories that require the keyword argument ``trust_remote_code=True`` for custom pipeline execution, BentoML will load the model into memory by default.
+When importing models from repositories that require the keyword argument ``trust_remote_code=True`` for custom defined model classes, BentoML will load the model into memory by default.
 In such cases, to avoid loading the model into memory, add the keyword argument ``clone_repository=True``. Note that since this downloads all the files in the repository instead of selectively picking certain model files,
 it results in greater storage requirements. Here is how you can invoke this:
 
--- a/docs/source/frameworks/transformers.rst
+++ b/docs/source/frameworks/transformers.rst
@@ -87,6 +87,29 @@ such as tokenizers, preprocessors, and feature extractors, can also be saved as 
 
 To load the pre-trained instances for testing and debugging, use :code:`bentoml.transformers.load_model` with the same tags.
 
+Starting from BentoML version 1.1.9, importing pre-trained Transformer models from Hugging Face has been further streamlined.
+You can choose to use the new ``bentoml.transformers.import_model`` function to import models directly into the BentoML Model Store without the overhead of loading them into memory.
+By contrast, ``bentoml.transformers.save_model`` necessitates prior model loading and can be resource-intensive for models with large weights.
+Here is an example of using the new function:
+
+.. code-block:: python
+    :caption: `download_model.py`
+
+    import bentoml
+    from transformers import AutoTokenizer
+
+    # Save the tokenizer with minimal memory overhead
+    tokenizer = AutoTokenizer.from_pretrained("t5-small")
+    bentoml.transformers.save_model('t5-small-tokenizer', tokenizer)
+
+    # Import the model without loading into memory, conserving memory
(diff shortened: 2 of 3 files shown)


docs/source/frameworks/transformers.rst
```diff
@@ -87,6 +87,29 @@ such as tokenizers, preprocessors, and feature extractors, can also be saved as
...
+    bentoml.transformers.save_model('t5-small-tokenizer', tokenizer)
+
+    # Import the model without loading into memory, conserving memory
+    bentoml.transformers.import_model("t5-small-model", "t5-small")
+
+The ``bentoml.transformers.import_model`` function has two required parameters:
+
+* ``name``: The name of the model in the BentoML Model Store.
+* ``model_name_or_path``: This can be a string, a Hugging Face repository identifier (repo_id), or a directory path containing weights saved using ``transformers.AutoModel.save_pretrained`` (for example, ``./my_pretrained_directory/``).
+
```
MingLiangDai, line 112: Hey Sherlock, everything looks good to me so far. I would like to mention that there is a small note that should be added:

At the moment, repositories that require the argument `trust_remote_code=True` to be set will actually fail to download. I am currently working on a change to fix this issue. 

The new changes should be merged this week as well. With the new changes, users will unfortunately need to load the model into memory by default for repositories that require `trust_remote_code=True`. To avoid loading into memory, users will have to set the `clone_repository=True` keyword argument.

MingLiangDai: This change can probably summarized as follows:

NOTE: Importing from model repositories that use the keyword argument `trust_remote_code=True` will require loading the model into memory by default. To completely avoid loading the model into memory, we recommend using the keyword argument `clone_repository=True`. This downloads all the files in the repository instead of selectively picking certain model files by using the method `.from_pretrained` under the hood, avoiding any potential memory usage of the model weights.

Sherlock113: Thank you for the clarification. So `clone_repository=True` should be added like this?

```python
import bentoml

model = "sshleifer/distilbart-cnn-12-6"
task = "summarization"

# Import the model without loading into memory by cloning the entire repository
bentoml.transformers.import_model(
   name=task, 
   model_name_or_path=model,
   clone_repository=True,  # This will avoid loading the model into memory
   metadata=dict(model_name=model)
)

```

aarnphm: `import_model` won't load the model into memory regardless `clone_repository` is set or not.

The model only load into memory if the models require to use `trust_remote_code=True` (for examples, any Yi AI models, Baichuan, ChatGLM models). In that case, setting `clone_repository=True` won't load such models into memory

docs/source/frameworks/diffusers.rst
```diff
@@ -19,6 +19,10 @@ You can import a pretrained diffusion model from huggingface hub or local direct
 	"stabilityai/stable-diffusion-2-1",  # huggingface model name
     )
 
+.. note::
+
+    ``bentoml.diffusers.save_model`` can also be used to import diffusion models from Hugging Face, but it requires models to be loaded into memory first,
+    which is resource-intensive for large models. By contrast, ``bentoml.diffusers.import_model`` imports diffusion models directly without loading them into memory.
```
larme, line 25: maybe "tries to import diffusion models..." is better

docs/source/frameworks/transformers.rst
```diff
@@ -87,6 +87,29 @@ such as tokenizers, preprocessors, and feature extractors, can also be saved as
 
 To load the pre-trained instances for testing and debugging, use :code:`bentoml.transformers.load_model` with the same tags.
 
+Starting from BentoML version 1.1.9, importing pre-trained Transformer models from Hugging Face has been further streamlined.
+You can choose to use the new ``bentoml.transformers.import_model`` function to import models directly into the BentoML Model Store without the overhead of loading them into memory.
```
larme, line 91: I think in some edge cases we still need to load it into memory? @MingLiangDai

docs/source/frameworks/transformers.rst
larme: however `clone_repository=True` may increase the size of the downloaded model. Should we mention this?

larme (commented):
LGTM for most part.

docs/source/frameworks/transformers.rst
MingLiangDai: Both Aaron and Larme said are correct. For most cases, model will not load into memory. But some repositories require the use of `trust_remote_code=True` keyword and will load model to memory. To avoid this, we can set `clone_repository=True`, which will download all the repo files (hence the larger storage size requirement) but will not load model to memory. 

```
# Import trust_remote_code=True model without loading into memory by cloning the entire repository

bentoml.transformers.import_model(
   name=task, 
   model_name_or_path=model,
   trust_remote_code=True,
   clone_repository=True,  # This will avoid loading the model into memory
   metadata=dict(model_name=model)
)
```

Sherlock113: Thank you all. I just updated the docs based on your comments.

docs/source/frameworks/transformers.rst
```diff
@@ -87,6 +87,50 @@ such as tokenizers, preprocessors, and feature extractors, can also be saved as
...
+    bentoml.transformers.import_model("t5-small-model", "t5-small")
+
+The ``bentoml.transformers.import_model`` function has two required parameters:
+
+* ``name``: The name of the model in the BentoML Model Store.
+* ``model_name_or_path``: This can be a string, a Hugging Face repository identifier (repo_id), or a directory path containing weights saved using ``transformers.AutoModel.save_pretrained`` (for example, ``./my_pretrained_directory/``).
+
+When importing models from repositories that require the keyword argument ``trust_remote_code=True`` for custom pipeline execution, BentoML will load the model into memory by default.
+In such cases, to avoid loading the model into memory, add the keyword argument ``clone_repository=True``. This downloads all the files in the repository instead of selectively picking certain model files,
+resulting in greater storage requirements; this approach uses the ``.from_pretrained`` method under the hood, avoiding any potential memory usage of the model weights. Here is how you can invoke this:
```
MingLiangDai, line 115: oops, sorry I might have flipped the things I said before. 

Using `clone_repository=True` actually avoids the use of `from_pretrained`, that's why we can avoid the memory usage. `clone_repository=False` will use `from_pretrained` by default and this will normally not load into memory except for when `trust_remote_code=True`. 

Sorry about my wordy explanation beforehand.

docs/source/frameworks/transformers.rst
MingLiangDai: Yup, so far I think the only identified issue is `trust_remote_code=True`. Mentioned in another comment.

docs/source/frameworks/transformers.rst
Sherlock113: OK. I think we can just removed the `from_pretrained` explanation part.

docs/source/frameworks/transformers.rst
```diff
@@ -87,6 +87,50 @@ such as tokenizers, preprocessors, and feature extractors, can also be saved as
...
+
+    # Import the model without loading into memory, conserving memory
+    bentoml.transformers.import_model("t5-small-model", "t5-small")
+
+The ``bentoml.transformers.import_model`` function has two required parameters:
+
+* ``name``: The name of the model in the BentoML Model Store.
+* ``model_name_or_path``: This can be a string, a Hugging Face repository identifier (repo_id), or a directory path containing weights saved using ``transformers.AutoModel.save_pretrained`` (for example, ``./my_pretrained_directory/``).
+
+When importing models from repositories that require the keyword argument ``trust_remote_code=True`` for custom pipeline execution, BentoML will load the model into memory by default.
```
MingLiangDai, line 113:
```suggestion
When importing models from repositories that require the keyword argument ``trust_remote_code=True`` for custom defined model classes, BentoML will load the model into memory by default.
```

MingLiangDai: I think this is the last comment I have. Thanks Sherlock!

docs/source/frameworks/transformers.rst
aarnphm: > however `clone_repository=True` may increase the size of the downloaded model. Should we mention this?

Worth noting is we probably want to have an options for ppl to save safetensors vs legacy bin weights

Sherlock113:
Docs were updated based on comments. Merging this...

PR #4307, opened 2023-11-29
feat: models.save api and tests

MingLiangDai:
## What does this PR address?

Add bentoml.models.save and tests coverage for changes

## Before submitting:


src/bentoml/models.py
```diff
@@ -318,6 +318,35 @@ def create(
...
+        api_version=api_version,
+        signatures={},
+        metadata=metadata,
+        context=ModelContext(framework_name="", framework_versions={}),
+    )
+
+    try:
+        yield res
+    except Exception:
+        raise
```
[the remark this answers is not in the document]
MingLiangDai: Leaving an open exception like this is done on purpose. The user code will actually be injected into the try block when calling `with bentoml.models.save():`. So if there is any exception raised by the user's code, we shouldn't handle it differently and just raise it. I believe that is why they did the same thing in `bentoml.models.create`

src/bentoml/models.py
```diff
@@ -318,6 +318,35 @@ def create(
         res.exit_cloudpickle_context(imported_modules)
 
 
+@inject
+@contextmanager
+def save(
```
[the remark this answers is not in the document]
aarnphm: seems to me this is more of a syntactic sugar?

src/bentoml/models.py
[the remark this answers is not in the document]
aarnphm: No I don't think this is correct. All exceptions should either wrapped within BentoMLException and raise from it.

src/bentoml/_internal/models/model.py
```diff
@@ -606,8 +606,9 @@ def with_options(self, **kwargs: t.Any) -> ModelInfo:
     def imported_module(self) -> ModuleType:
         if self._cached_module is None:
             try:
+                module_to_import = self.module or "bentoml.dummy_module"
                 object.__setattr__(
-                    self, "_cached_module", importlib.import_module(self.module)
+                    self, "_cached_module", importlib.import_module(module_to_import)
```
aarnphm, line 611: I don't think we should do this.

src/bentoml/dummy_module.py
```diff
@@ -0,0 +1,18 @@
+import logging
```
aarnphm, line 1: Let's not do this as a top level.

src/bentoml/_internal/models/model.py
larme: @aarnphm  there's some codes in other places relies on this `self.imported_module` has certain functions. We have a `dummy_module` to implement a dummy version of these functions so that we are not distinguishing our framework codes and users' codes.

aarnphm: Then I think we should fix this. I think `_cached_module` is a hack.

MingLiangDai: I'm confused by what you mean by 'fix this'. Do we not want `_cached_module` or `imported_module`? I think those are needed for other functionalities.

MingLiangDai: Another way to avoid this empty module error (right now, `self.module` is "" due to the nature of the `modules.save` method) is to have `self.imported_module` return None if `self_module == ""`. Then in all other methods that use `self.imported_module`, we add a check for the value of `self.imported_module`.

The current approach we are taking is to add a dummy_module that imitates all the functionalities that an actual module will have. This way, other methods that use the `imported_module` only need to worry about how they interact with modules and what the modules are.

src/bentoml/models.py
```diff
@@ -318,6 +318,37 @@ def create(
...
+    name: Tag | str,
+    *,
+    module: str = "",
+    api_version: str | None = None,
```
parano, line 327: why expose module and api_version here?

src/bentoml/models.py
```diff
@@ -318,6 +318,37 @@ def create(
...
+        module=module,
+        api_version=api_version,
+        labels=labels,
+        signatures={},
+        metadata=metadata,
+        context=ModelContext(framework_name="", framework_versions={}),
+    )
+
+    try:
+        yield res
```
parano, line 344: should we just yield the path here instead of Model?

src/bentoml/models.py
MingLiangDai: I was following the design here (at the bottom of https://www.notion.so/bentoml/Model-API-f020bba69482465198c7dcea8b23cc55). I suppose these two would not be too useful to users. In your opinion, should these be removed?

src/bentoml/models.py
MingLiangDai: I think leaving it as Model is okay. Users can access values that we have auto-generated through the Model (eg. Model.tag, etc). Model also provides some useful properties (eg. Model.file_size). Though not too sure helpful they'd be to the users 😅 

imo, it may be more intuitive to have the `model_ref` rather than the `model_file_path` when running the following code:
```
with bentoml.models.save as model_ref:
    ...
# vs
with bentoml.models.save as model_file_path:
    ...
```

src/bentoml/models.py
MingLiangDai: removed these two parameters

src/bentoml/_internal/models/skeleton_module.py
```diff
@@ -0,0 +1,18 @@
+import logging
```
larme, line 1: @MingLiangDai we can move `import logging` and `logger = ...` into `_log_warning_msg`. Because these functions are not supposed to be called, we do not need to import the logging module

larme:
@MingLiangDai could you fix the unitest. I think an extra "module" argument is passed to unit test

src/bentoml/models.py
larme: @parano An example is that if user does not provide a version in the model name, the model_ref will have an auto-generated version, which may be useful for users.

src/bentoml/models.py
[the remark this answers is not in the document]
sauyon: Ming is correct, this is so that `flush` and `save` aren't run if user code raises an exception. We don't want to mess with the actual exception in this case.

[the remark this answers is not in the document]
frostming: catching a general `Exception` and reraise it is just like we don't catch. It isn't necessary at all.

The following examples are equivalent to each other:

```python
try:
    model.save_pretrained(res.path)
    tokenizer.save_pretrained(res.path)
except Exception:
    raise
```

```python
model.save_pretrained(res.path)
tokenizer.save_pretrained(res.path)
```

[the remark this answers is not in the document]
sauyon: I'm actually not sure this try block should exist, I thought that the context manager would ignore the exception but that isn't true.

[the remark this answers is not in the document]
sauyon: I'm not really sure how `yield res` could fail, anyway?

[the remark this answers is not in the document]
sauyon: We definitely don't want to wrap user exceptions in a `BentoMLException`.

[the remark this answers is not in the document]
frostming: > I'm not really sure how `yield res` could fail, anyway?

Oh, definitely:
```pytb
In [1]: from contextlib import contextmanager

In [2]: @contextmanager
   ...: def catching():
   ...:     try:
   ...:         yield
   ...:     except Exception as e:
   ...:         print(f"The exception is {type(e)}: {e}")
   ...:         raise
   ...:     finally:
   ...:         print('finally called')
   ...: 

In [3]: with catching():
   ...:     1/0
   ...: 
The exception is <class 'ZeroDivisionError'>: division by zero
finally called
```

src/bentoml/_internal/models/model.py
sauyon: I think we can just raise an error here if this method is called at all for a model saved without a module.

src/bentoml/models.py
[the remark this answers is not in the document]
sauyon: Let's just change `create` to do the things `save` does? Doesn't seem like we need both of these APIs to me.

src/bentoml/models.py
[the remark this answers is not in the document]
aarnphm: I think we might want to wrap the users exception in a exceptiongroup stack. https://peps.python.org/pep-0654/

src/bentoml/_internal/models/model.py
larme: @sauyon This `import_module` is somehow called when bentoml push a model to BentoCloud so we need to avoid raising the error here.

src/bentoml/models.py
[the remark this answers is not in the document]
larme: @aarnphm could you explain the benefits of using an exception group. And is this feature backward compatible with python 3.8 etc.?

src/bentoml/_internal/models/model.py
sauyon: We should probably just remove the use in push?

MingLiangDai: Following the error traceback of calling `bentoml.models.push`, it appears that the error is triggered when we call the `options` property of the `ModelInfo` class. This line: https://github.com/bentoml/BentoML/blob/main/src/bentoml/_internal/cloud/bentocloud.py#L639C44-L639C44

Accessing the `options` will call `imported_module` under the hood. I don't think we can easily avoid this error in push.

sauyon: I think it just seems like more effort than necessary when we can just patch `options`?

sauyon: I think we should error for both of those. Honestly I would want to error for `options` as well, but we'd have to figure something else out for push and that seems more annoying.

MingLiangDai:
Cool! Made all the necessary changes. Please have one final look! 🙏

src/bentoml/models.py
```diff
@@ -285,6 +285,10 @@ def create(
     context: ModelContext,
     _model_store: ModelStore = Provide[BentoMLContainer.model_store],
 ) -> t.Generator[Model, None, None]:
+    """
+    `bentoml.models.create` is deprecated and replaced with `bentoml.models.save`. Moving forward, we encourage users to use `bentoml.models.save`.
+    """
```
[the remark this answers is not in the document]
MingLiangDai: Will resolve due to outdated code.

src/bentoml/_internal/frameworks/catboost.py
```diff
@@ -205,7 +205,7 @@ def save_model(
         training_class_name=model.__class__.__name__,
     )
 
-    with bentoml.models.create(
+    with bentoml.models._create(
```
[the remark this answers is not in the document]
MingLiangDai: I believe this is what the team has decided on. But you are correct, it is throwing a warning about private access.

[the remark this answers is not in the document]
sauyon: If it's a big deal we can always just add a flag to silence warnings, maybe users want that too.

src/bentoml/_internal/frameworks/catboost.py
```diff
@@ -205,7 +205,7 @@ def save_model(
         training_class_name=model.__class__.__name__,
     )
 
-    with bentoml.models.create(
+    with bentoml.models._create(  # pyright: ignore [reportPrivateUsage]
```
aarnphm, line 208: /s/pyright: ignore [reportPrivateUsage]/type: ignore

So that both mypy and pyright understand this.

src/bentoml/_internal/frameworks/catboost.py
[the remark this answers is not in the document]
frostming: This is supposed to be a temporary solution right? It seems a bit strange to me to call a private member

src/bentoml/_internal/frameworks/catboost.py
MingLiangDai: Unfortunately, VS code, the IDE I'm using, does not use mypy so I have no way of testing if the ignore suppresses the warning. I also tried copying your suggestion above but it did not seem to work. 

What I can do is `# type: ignore` to ignore the line entirely instead of specifically targeting private method use. Lmk if you would like that instead.

src/bentoml/_internal/frameworks/catboost.py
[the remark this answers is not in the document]
aarnphm: Doesn't seem like a tmp sol.

PR #4326, opened 2023-12-07
add with_tokenizer parameter to transformers.import_model

MingLiangDai:
## What does this PR address?

feat: add with_tokenizer parameter to transformers.import_model

## Before submitting:

### DIFF
--- a/src/bentoml/_internal/frameworks/transformers.py
+++ b/src/bentoml/_internal/frameworks/transformers.py
@@ -613,6 +613,7 @@ def import_model(
     name: Tag | str,
     model_name_or_path: str | os.PathLike[str],
     *,
+    with_tokenizer: bool = False,
     proxies: dict[str, str] | None = None,
     revision: str = "main",
     force_download: bool = False,
@@ -911,6 +912,12 @@ def import_model(
         )
         with open(bento_model.path_of(PRETRAINED_PROTOCOL_NAME), "wb") as f:
             cloudpickle.dump(pretrained.__class__, f)
+
+        if with_tokenizer:
+            transformers.AutoTokenizer.from_pretrained(
+                model_name_or_path
+            ).save_pretrained(bento_model.path)
+
         return bento_model
 
 
--- a/src/bentoml/_internal/frameworks/transformers.py
+++ b/src/bentoml/_internal/frameworks/transformers.py
@@ -613,7 +613,6 @@ def import_model(
     name: Tag | str,
     model_name_or_path: str | os.PathLike[str],
     *,
-    with_tokenizer: bool = False,
     proxies: dict[str, str] | None = None,
     revision: str = "main",
     force_download: bool = False,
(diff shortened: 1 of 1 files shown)


MingLiangDai:
Slack thread: https://bentoml-team.slack.com/archives/C02QLC8RB5W/p1701943951358159?thread_ts=1701412487.168309&cid=C02QLC8RB5W

src/bentoml/_internal/frameworks/transformers.py
```diff
@@ -613,6 +613,7 @@ def import_model(
     name: Tag | str,
     model_name_or_path: str | os.PathLike[str],
     *,
+    with_tokenizer: bool = False,
```
aarnphm, line 616: why do we need this? can't we just always save the tokenizer with the model once we do import_model?

I think that is actually more intuitive right, since `import_model` users coming from HF will think that it will actually import the repository from HF.

MingLiangDai: The current work flow (in our documentations) is that we save the tokenizer and the model separately. We want this param to allow users to have the choice of saving the tokenizer into the same directory as the model weights

current workflow: 
```
import bentoml
from transformers import AutoTokenizer

hf_model_name = "sentence-transformers/all-MiniLM-L6-v2"
tokenizer = AutoTokenizer.from_pretrained(hf_model_name)

bentoml.transformers.import_model("all-MiniLM-L6-v2", hf_model_name)
bentoml.transformers.save_model("all-MiniLM-L6-v2-tokenizer", tokenizer)
```

aarnphm: Ye I don't think that makes much sense. Since we are moving to the new save methods in 1.2 and, probably just need to save the tokenizer together with the model, then users can do 
```python

tokenizer = transformers.AutoTokenizer.from_pretrained(bentomodel.path)
```

larme: @aarnphm why originally we didn't save tokenizer with model? If there is some usage for this case, we can set `with_tokenizer=True` and keep the option of not saving the tokenizer. If there's no usage, then we can remove `with_tokenizer` and always save tokenizer with model

aarnphm: I think we should always save tokenizer with model. Makes this similar to hugging face `from_pretrained` without actually loading the model into memory.

ssheng: I think we should continue to keep models and tokenizers separate. If users want to save them in the same model, they can use the `bentoml.models.create` API.

ssheng: Please do not merge this before syncing up with me.

MingLiangDai:
Closed due to the fact that users should be encouraged to use 1.2 to save model and tokenizer in same directory.

PR #4330, opened 2023-12-11
fix(stream): streaming enable to work with proxy

jianshen92:
## What does this PR address?
Relates to #4310 
1) streaming now just stream raw text instead of pickled data container
2) provides `to_sse` and `from_sse` to be used at boundaries (api server and runner) if user wants to stream text in SSE format

Fixes #(issue)

## Before submitting:


src/bentoml/_internal/server/runner_app.py
```diff
@@ -212,8 +212,8 @@ async def inner():
                         traceback.print_exc()
                         raise
                     async for data in ret:
-                        payload = AutoContainer.to_payload(data, 0)
-                        yield payload
+                        # payload = AutoContainer.to_payload(data, 0)
+                        yield data
```
aarnphm, line 216:
```suggestion
                        yield data
```

src/bentoml/_internal/io_descriptors/utils.py
```diff
@@ -0,0 +1,25 @@
...
+        sse_data += f"id: {id}\n"
+    if event is not None:
+        sse_data += f"event: {event}\n"
+    if retry is not None:
+        sse_data += f"retry: {retry}\n"
+    sse_data += f"data: {data}\n\n"
+    return sse_data
+
+
+async def from_sse(async_iterator: AsyncIterator[str]) -> AsyncGenerator[str, None]:
```
aarnphm, line 19: This is not really from sse right? rather from iterator that yields sse format.

We can do something like this 

https://github.com/bentoml/OpenLLM/blob/main/openllm-client/src/openllm_client/_stream.py

jianshen92: Yeah the naming of the function is probably inaccurate. It's just a wrapper iterator that will combine chunks(if they are chunked) into a complete SSE event. so that we can just do `for event in from_sse(runner.async_stream.xx()`

Wasn't exactly thinking about implementing a new data structure here but mere simple functions. 
since OpenLLM would be mainly using these utils, maybe you could state your preferences

aarnphm: I wonder when users will actually using this

```python
bentoml.io.from_see(runner.generate.async_stream())
```
?

This means `async_stream` should either yield a sse object, or a stream with sse format

Preferably to use the object.

jianshen92: The decision from @ssheng and @bojiang discussion the other day was to keep the `async_stream` as pure text stream and if we expect an SSE format text stream, then we should wrap it with this `from_sse` method, or whatever its called, otherwise we don't have guarantee that the event will not be chunked. 

CMIIW

aarnphm: If async_stream yield pure text_stream then `from_sse` expects the generator is a text-stream with SSE format and then should yield a SSE object (similar to what I did with `openllm-client`)

aarnphm: I don't really see the use of these utilities to be frank.

aarnphm:
@jianshen92 I think might be worth merging this to 1.2

src/bentoml/_internal/io_descriptors/utils.py
jianshen92: okay, if you as the user wouldn't use it we can exclude it from this PR. User will then have the responsibility to patch the text chunks if they are being chunked

jianshen92:
> @jianshen92 I think might be worth merging this to 1.2

@bojiang should I be merging to 1.2 instead?

bojiang:
‌‌‌@aarnphm This is a bug fix. It's affecting the deployment of openllm under certain circumstances. There's still some time for 1.2 to release. Let's merge to main first?

bojiang:
‌‌‌‌By the way, I think what Aaron meant might not be that we don't offer these utils, but rather that we provide utils which then return SSE objects. @jianshen92 
I wrote some unit test for the utils. Let me push them here.

aarnphm:
@bojiang @jianshen92 what is the status of this PR?

bojiang:
@aarnphm Jianshen is still working on the utils.

jianshen92:
Usage:

runner
```
class Runner(bentoml.Runnable):
    ...
    
    async def raw_text_streaming():
        yield "this is valid"
    
    async def sse_streaming():
        yield SSE(data="my data", event="my event", id="a123cbx", retry=123)
```

api-server
```
# if we expect raw text, nothing changes. we dont care if they get chunked 
# if we expect a SSE text stream, we can use the `from_chunks` method to reconstruct the chunked data
async for event in SSE.from_chunks(runner_iterator):
    yield event.to_str()
    yield event.data
```

src/bentoml/_internal/io_descriptors/utils.py
```diff
@@ -0,0 +1,55 @@
...
+        data_lines = self.data.split("\n")
+        for line in data_lines:
+            sse_data += f"data: {line}\n"
+
+        # Add an extra newline to mark the end of the message
+        sse_data += "\n"
+        return sse_data
+
+    @staticmethod
+    async def from_chunks(
```
aarnphm, line 36:
```suggestion
    async def from_iterator(
```

src/bentoml/_internal/io_descriptors/utils.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+@attr.s(auto_attribs=True)
+class SSE:
+    data: str
+    event: Optional[str] = None
+    id: Optional[str] = None
+    retry: Optional[int] = None
+
+    def to_str(self) -> str:
```
aarnphm, line 17:
```suggestion
    def marshal(self) -> str:
```

bojiang (approved):

PR #4335, opened 2023-12-13
feat(bentocloud): deployment v2 api client + cli

Haivilo:
## What does this PR address?
```
bentoml.deployment.create(...)
bentoml.deployment.update(...)
bentoml.deployment.apply(...)
bentoml.deployment.get(...)
bentoml.deployment.list(...)
bentoml.deployment.terminate(...)
bentoml.deployment.delete(...)
```

```
bentoml deploy .
bentoml deployment update
bentoml deployment get
bentoml deployment list
bentoml deployment terminate
bentoml deployment delete
```

## Before submitting:


src/bentoml/_internal/cloud/deployment.py
```diff
@@ -560,3 +565,98 @@ def terminate(
...
         return res
+
+    @classmethod
+    def _create_deploymentV2(
+        cls,
+        create_deployment_schema: CreateDeploymentSchemaV2,
+        context: str | None = None,
+    ) -> DeploymentSchema:
+        cloud_rest_client = get_rest_api_client(context)
+        res = cloud_rest_client.create_deployment_v2(create_deployment_schema)
```
aarnphm, line 576: nit: should we move this to v2 namespace?

```suggestion
        res = cloud_rest_client.v2.create_deployment(create_deployment_schema)
```

aarnphm: I mean the rest_client is never a public API, so breaking is ok?

src/bentoml/deployment.py
```diff
@@ -0,0 +1,48 @@
+"""
+User facing python APIs for deployment
+"""
+
+
```
aarnphm, line 5:
```suggestion
from __future__ import annotations
```

frostming (commented):
What about making `get_rest_api_client` to accept a version parameter instead of accessing via `client.v1`?

```python
def get_rest_api_client(version = "v2"):
    ...
    return getattr(client, version)
```

src/bentoml/__init__.py
```diff
@@ -154,7 +155,7 @@
     exceptions = _LazyLoader("bentoml.exceptions", globals(), "bentoml.exceptions")
     monitoring = _LazyLoader("bentoml.monitoring", globals(), "bentoml.monitoring")
     cloud = _LazyLoader("bentoml.cloud", globals(), "bentoml.cloud")
```
aarnphm, line 157: Should probably deprecate this module.

Haivilo: We also have push/pull bentos and models methods within the cloud client, I think we might need to continue supporting them? We can probably just remove `deployment` client from our `BentoCloudClient`?

aarnphm: we should deprecate the deployment from cloud then.

Haivilo:
> What about making `get_rest_api_client` accept a version parameter instead of accessing via `client.v1`?
> 
> ```python
> def get_rest_api_client(version = "v2"):
>     ...
>     return getattr(client, version)
> ```

I am not sure if we should do this, because some methods use both `v1` and `v2`. If we were to do it this way, then we need to instantiate it in 2 lines instead of 1. Is there any draw back to return the whole client?
```python
client_v1 = get_rest_api_client(version = "v1")
client_v2 = get_rest_api_client(version = "v2")
...

**Comment by aarnphm** [comment]
> > What about making `get_rest_api_client` accept a version parameter instead of accessing via `client.v1`?
> > ```python
> > def get_rest_api_client(version = "v2"):
> >     ...
> >     return getattr(client, version)
> > ```
> 
> I am not sure if we should do this, because some methods use both `v1` and `v2`. If we were to do it this way, then we need to instantiate it in 2 lines instead of 1. Is there any draw back to return the whole client?
> 
> ```python
> client_v1 = get_rest_api_client(version = "v1")
> client_v2 = get_rest_api_client(version = "v2")
> ...
> ```


This `get_rest_api_client` is internal right?

Why do we need v1 still?

**Comment by Haivilo** [comment]
> > > What about making `get_rest_api_client` accept a version parameter instead of accessing via `client.v1`?
> > > ```python
> > > def get_rest_api_client(version = "v2"):
> > >     ...
> > >     return getattr(client, version)
> > > ```
> > 
> > 
> > I am not sure if we should do this, because some methods use both `v1` and `v2`. If we were to do it this way, then we need to instantiate it in 2 lines instead of 1. Is there any draw back to return the whole client?
> > ```python
> > client_v1 = get_rest_api_client(version = "v1")
> > client_v2 = get_rest_api_client(version = "v2")
> > ...
> > ```
> 
> This `get_rest_api_client` is internal right?
> 
> Why do we need v1 still?

Only update and create deployment has v2 api. We will be using v1 api for all the rest of the endpoints:
https://github.com/bentoml/BentoML/blob/af2a79411bd9a823f9c63a3ceb1992997955cbb3/src/bentoml/_internal/cloud/client.py#L539-L555

**Comment by aarnphm** [comment]
> > > > What about making `get_rest_api_client` accept a version parameter instead of accessing via `client.v1`?
> > > > ```python
> > > > def get_rest_api_client(version = "v2"):
> > > >     ...
> > > >     return getattr(client, version)
> > > > ```
> > > 
> > > 
> > > I am not sure if we should do this, because some methods use both `v1` and `v2`. If we were to do it this way, then we need to instantiate it in 2 lines instead of 1. Is there any draw back to return the whole client?
> > > ```python
> > > client_v1 = get_rest_api_client(version = "v1")
> > > client_v2 = get_rest_api_client(version = "v2")
> > > ...
> > > ```
> > 
> > 
> > This `get_rest_api_client` is internal right?
> > Why do we need v1 still?
> 
> Only update and create deployment has v2 api. We will be using v1 api for all the rest of the endpoints:
> 
> https://github.com/bentoml/BentoML/blob/af2a79411bd9a823f9c63a3ceb1992997955cbb3/src/bentoml/_internal/cloud/client.py#L539-L555

Hmm I propose this: for v2 client it would just extends v1. So only `update` and `create` deployments is different. Otherwise it will still be the same.

**Comment by Haivilo** [comment]
Discussed with @aarnphm and @frostming , let's keep using `client.v1.xxx` and `client.v2.xxx` since it's internal.

**Review by frostming — APPROVED** [review]

File: src/bentoml/_internal/cloud/__init__.py [hunk 4]
```diff
@@ -1,4 +1,3 @@
 from .base import CloudClient as CloudClient
 from .bentocloud import BentoCloudClient as BentoCloudClient
-from .deployment import Resource as Resource
```
aarnphm, line 3: Technically this is a breaking change.

Probably should throw deprecation warning either way a getattr

src/bentoml/_internal/cloud/client.py
```diff
@@ -8,47 +8,42 @@
...
+from .schemas.utils import schema_from_object
+from .schemas.utils import schema_to_json
 
 logger = logging.getLogger(__name__)
 
 
-class RestApiClient:
-    def __init__(self, endpoint: str, api_token: str) -> None:
+class BaseRestApiClient:
+    def __init__(self, endpoint: str, session: requests.Session) -> None:
```
aarnphm, line 44: Nit: can we update this with httpx instead? We r removing requests soon

aarnphm (commented):
A few final comments. Then we can get this in

src/bentoml/_internal/cloud/client.py
Haivilo: Um, should we merge these after this PR? 
https://github.com/bentoml/BentoML/pull/4224
https://github.com/bentoml/BentoML/pull/4227

src/bentoml/_internal/cloud/schemas/schemasv2.py
```diff
@@ -0,0 +1,55 @@
...
+
+import attr
+
+from bentoml._internal.cloud.schemas.modelschemas import AccessControl
+from bentoml._internal.cloud.schemas.modelschemas import ApiServerBentoFunctionOverrides
+from bentoml._internal.cloud.schemas.modelschemas import BentoConfigServiceSchema
+from bentoml._internal.cloud.schemas.modelschemas import BentoRequestOverrides
+from bentoml._internal.cloud.schemas.modelschemas import DeploymentStrategy
+from bentoml._internal.cloud.schemas.modelschemas import DeploymentTargetConfigV2
+from bentoml._internal.cloud.schemas.modelschemas import DeploymentTargetHPAConf
```
FogDong, line 13: We can delete `cpu`, `memory`, `gpu`, `qps` in `DeploymentTargetHPAConf` right now.

src/bentoml/deployment.py
```diff
@@ -0,0 +1,444 @@
...
+        config_dct=config_dct,
+        config_file=config_file,
+    )
+
+
+def get(
+    name: str,
+    context: str | None = None,
+    cluster_name: str | None = None,
+    kube_namespace: str | None = None,
```
FogDong, line 398: I think we can delete `kube_namespace` parameter and only use default namespace to call the API.

src/bentoml/_internal/cloud/client.py
aarnphm: I don't see why not to do it now. Probably better to do it now.

Haivilo: Rebased https://github.com/bentoml/BentoML/pull/4224 here, httpx should be good to go!

src/bentoml/_internal/cloud/__init__.py
Haivilo: Actually this is a class so I don't think we can make getattr warning 🤔

src/bentoml_cli/deployment.py
```diff
@@ -20,17 +23,145 @@ def add_deployment_command(cli: click.Group) -> None:
...
+        multiple=True,
+    )
+    @click.option(
+        "--config-file",
+        type=click.File(),
+        help="Configuration file path",
+        default=None,
+    )
+    @click.option(
+        "--no_wait",
```
aarnphm, line 90:
```suggestion
        "--wait/--no-wait", default=False
```

src/bentoml_cli/deployment.py
```diff
@@ -42,26 +173,16 @@ def shared_decorator(
         def decorate(f: t.Callable[..., t.Any]) -> t.Callable[..., t.Any]:
             options = [
                 click.argument(
-                    "deployment-name",
+                    "name",
```
aarnphm, line 176: This is breaking change.

src/bentoml_cli/deployment.py
```diff
@@ -42,26 +173,16 @@ def shared_decorator(
...
                     type=click.STRING,
                     required=required_deployment_name,
                 ),
-                cog.optgroup.group(
-                    cls=cog.AllOptionGroup, name="cluster and kube namespace options"
-                ),
-                cog.optgroup.option(
-                    "--cluster-name",
+                click.option(
+                    "--cluster",
```
aarnphm, line 181: also these.

src/bentoml_cli/deployment.py
```diff
@@ -42,26 +173,16 @@ def shared_decorator(
...
                     type=click.STRING,
                     default=None,
                     help="Name of the cluster.",
                 ),
-                cog.optgroup.option(
```
aarnphm, line 58: let's deprecate this.

aarnphm (changes requested):
Few things.

src/bentoml_cli/deployment.py
Haivilo: Since we are having a pretty big refactor and most things behave differently, I think it's fine to make these breaking changes here and not provide `deployment-name`, `cluster-name`, `kube-namespace` anymore?

src/bentoml/_internal/cloud/__init__.py
Haivilo: Thanks, this is fixed

src/bentoml_cli/deployment.py
aarnphm: ok sg.

src/bentoml/cloud.py
```diff
@@ -1,3 +1,12 @@
 from ._internal.cloud import BentoCloudClient as BentoCloudClient
-from ._internal.cloud import Resource as Resource
 from ._internal.cloud import YataiClient as YataiClient
+
+deprecated_names = ["Resource"]
+
+
+def __getattr__(name: str):
+    if name in deprecated_names:
```
aarnphm, line 8: This defeats the whole purpose of deprecation right? this is still breaking.

Haivilo: Oh, I misunderstood. Just realized you said throwing warning not exception, but I believe we are not going to provide backward compatibility for this either, just like CLI?

FogDong (approved):

src/bentoml/_internal/cloud/deployment.py
```diff
@@ -37,526 +43,632 @@
...
+                and config_struct.scaling != DeploymentTargetHPAConf()
+            ):
+                raise BentoMLException(
+                    "The 'scaling' field is not allowed for distributed deployments. Please specify it per service in the services field."
+                )
+            if config_struct.deployment_strategy is not None:
+                raise BentoMLException(
+                    "The 'deployment_strategy' field is not allowed for distributed deployments. Please specify it per service in the services field."
+                )
+            if config_struct.extras is not None:
```
FogDong, line 148: This field is allowed in distributed mode.

aarnphm:
LGTM

aarnphm (approved):
LGTM

PR #4356, opened 2023-12-22
chore(sse): refactor sse utils with cursor

bojiang:
## What does this PR address?

Fixes #(issue)

## Before submitting:

### DIFF
--- a/src/bentoml/_internal/io_descriptors/utils.py
+++ b/src/bentoml/_internal/io_descriptors/utils.py
@@ -21,9 +21,7 @@ class SSE:
     id: str | None = None
     retry: int | None = None
 
-    def marshal(
-        self,
-    ) -> str:
+    def marshal(self) -> str:
         with io.StringIO() as buffer:
             if self.event:
                 if "\n" in self.event:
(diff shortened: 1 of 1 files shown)


src/bentoml/_internal/io_descriptors/utils.py
```diff
@@ -1,68 +1,83 @@
...
+    id: str | None = None
+    retry: int | None = None
 
-        # Handle multi-line data
-        data_lines = self.data.split("\n")
-        for line in data_lines:
-            sse_data += f"data: {line}\n"
+    def marshal(
+        self,
+    ) -> str:
```
aarnphm, line 26:
```suggestion
    def marshal(self) -> str:
```

src/bentoml/_internal/io_descriptors/utils.py
```diff
@@ -1,68 +1,97 @@
...
+            read_cursor = 0
+            last_chunk = ""
+            async for chunk in async_iterator:
+                buffer.write(chunk)
+                if (
+                    "\n\n" in chunk
+                    or chunk
+                    and chunk[0] == "\n"
+                    and last_chunk
+                    and last_chunk[-1] == "\n"
```
jianshen92, line 92: Is this part needed? Could we always just append the buffer until we see `\n\n` and then update the cursor?

bojiang: Chunking may happens right in the meddle of `\n\n`

jianshen92: Yeah, if that happens maybe we just simply wait for the next chunk to arrive?

bojiang: what do you mean?

bojiang: I don't think so. You can take a closer look at the process here.

jianshen92: Right, i understood what you wrote. I think we can just have `while '\n\n' in chunks` condition, and that will solve two problem at once:

1. The scenario you mentioned, where its chunked between two `\n` character, it will simply wait for the next chunk to enter the buffer then we will have the full delimiter again.
2. Multiple events in the same chunk, it will then yield multiple times before it receives the next chunk into buffer

bojiang:
open from another branch

PR #4377, opened 2023-12-29
chore(sdk): able to specify service name

bojiang:


aarnphm (approved):

src/bentoml/_internal/configuration/v2/__init__.py
```diff
@@ -63,6 +63,7 @@
     },
 }
 _SERVICE_CONFIG = {
+    s.Optional("name", default=None): s.Or(str, None),
```
aarnphm, line 66: hmm this means within the config this might happen

```yaml
services:
  llm-phi-service:
    name: llm-phi-2
    batching: 
      enabled: true
```

I think we should ignore or respect name in this case?

aarnphm: cc @ssheng thoughts on this as well.

bojiang: The configs are only in code in bentoml 1.2

frostming: I don't think we need to add this into the config, or is there any use case to specify it in config file?
Make it an argument of `bentoml.service()` is enough.

PR #4386, opened 2024-01-05
fix(http-server): limit threads per service

bojiang:


src/_bentoml_impl/server/app.py
```diff
@@ -8,6 +8,7 @@
 import typing as t
 from pathlib import Path
 
+import anyio
```
aarnphm, line 11: Why do we decide to only limit to one thread per API function?

src/_bentoml_impl/server/app.py
```diff
@@ -36,6 +37,31 @@
 R = t.TypeVar("R")
 
 
+LIMITER_MAP: dict[str, t.Any] = {}
```
frostming, line 40: It's suboptimal to use a global map, instead I would prefer to store this map inside the app factory.

src/_bentoml_impl/server/app.py
bojiang: answered offline

src/_bentoml_impl/server/app.py
bojiang: makes more sense since we need service level

src/_bentoml_sdk/service/config.py
```diff
@@ -194,6 +194,7 @@ class ServiceConfig(TypedDict, total=False):
     max_runner_connections: Posint
     resources: ResourceSchema
     workers: WorkerSchema
+    threads: Posint
```
frostming, line 197: Also need to update the schema in `configurations/v2`

aarnphm (approved):

PR #4391, opened 2024-01-09
fix(cli): fill out missing `apply` CLI args

aarnphm:
There are some click CLI args that are missing for `bentoml deployment apply`


aarnphm:
Closed due to duplication of #4390

PR #4395, opened 2024-01-12
chore: use case sensitive service name

bojiang:


frostming:
In what case users want define two services named `Iris` and `iRiS`?

bojiang:
‌‌‌@frostming We could think in reverse. Automatically converting everything to lowercase might be a kind of magic that users find hard to understand. Previously, we thought 'cloud' was case-insensitive, but it seems that's not the case.

frostming (approved):

PR #4415, opened 2024-01-18
fix: loading service from an absolute path

frostming:

## What does this PR address?

This is the case when running `Service.serve_http()`

Fixes #(issue)

## Before submitting:

### DIFF
--- a/src/_bentoml_sdk/service/factory.py
+++ b/src/_bentoml_sdk/service/factory.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import inspect
 import logging
+import pathlib
 import math
 import os
 import sys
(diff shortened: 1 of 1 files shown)


src/_bentoml_impl/loader.py
```diff
@@ -70,20 +71,23 @@ def normalize_identifier(
...
-            return normalize_package(service_identifier), pathlib.Path(
-                working_dir if working_dir is not None else "."
-            )
+            module_spec, _, service_path = service_identifier.partition(":")
+            module_path = pathlib.Path(module_spec)
+            if not module_path.suffix:
+                return service_identifier, pathlib.Path(working_dir or ".")
+            if not module_path.is_absolute():
+                module_path = pathlib.Path(working_dir or ".").joinpath(module_path)
+            return f"{module_path.stem}:{service_path}", module_path.parent
```
bojiang, line 88: Shall we return CWD as the bento path?

frostming: bento_path here is to tell `import_service` where to import the service, so it makes sense to return the parent dir of it.

This is to handle the case where `bento_identifier` is a path.

bojiang: bento_path is used to be the working directory of sub-processes

frostming: So `normalize_identifier` is to return two values, import string and bento_path. For a bento identifier as absolute path. What shall it return? `normalize_identifier("/path/to/package.py:Inference")`

bojiang: @frostming

bojiang: You just said this case happens when user call `service.serve_http()`, thus I think we should inherit the working directory, which is '.'
Will that work?
@frostming

frostming: but the first return value will be directly feed to `import_service`. An absolute path isn't a valid import string right?

PR #4498, opened 2024-02-10
feat: Dynamic Service creation 

### LINKED ISSUE #4460: feature: Dynamically add endpoints to Service
### Feature request

Depending on available models, support the possibility to add endpoints to a service.

Main Idea (only a draft, does not work like this in python)

```
import numpy as np
import bentoml
from bentoml.io import NumpyNdarray

# assuming modeltags holds all available models on device
runner_list = []
for model_tag in modeltags:
    runner_list.add[bentoml.sklearn.get("model_tag").to_runner()]

svc = bentoml.Service("detection", runners=runner_list)


for idx in range(len(runner_list)):
	@svc.api(input=NumpyNdarray(), output=NumpyNdarray(), route=f"runner_{idx}")
	def classify(input_series: np.ndarray) -> np.ndarray:
		result = iris_clf_runner[i].predict.run(input_series)
		return result
```

### Motivation

This feature would solve the problem of not having the possibility to dynamically add endpoints to a service

### Other

_No response_

holzweber:
## What does this PR address?

Fixes #4460 

## Before submitting:

- [] Did your changes require updates to the documentation? Have you updated


examples/dynamic_service/pipeline/service.py
```diff
@@ -0,0 +1,80 @@
+from typing import Any
+
+import bentoml
+from bentoml import Runner
```
bojiang, line 4: Hi holzweber. We are deprecating Runner API from 1.2. But I see the values of the dynamic servive example. Would you like update this once we finished the 1.2 DOC?

bojiang: Thus may I mark this as merge-hold?

holzweber: @bojiang i just checked the new documentation. but i do not know if my idea will work any longer, as i can not access the service object with the new service-annotation idea. i am scared that we really need to generate a pyhton file for dynamic services.. which is kind of an ugly solution...

any other idea how to get this working with 1.2?

aarnphm: Dynamic service in 1.2 will still work, as OpenLLM has to do something similar like this.

You can probably hijack directly into the service object, since `bentoml.Service` will return the new service object, which treats all runnable as a normal python class.

I think one huge difference here is that the lifecycle is just a class, so probably a lot simpler.

aarnphm: Will send more example once I finish openllm revamp 😄

holzweber: @aarnphm ... Any more updates on this?
I was trying to do something like this, but it does not work....

```
def wrap_service_methods(model: bentoml.Model, targets: Any):

    async def predict(input_doc: str):
        predictions = await model.predict.async_run([input_doc])
        return {"result": targets[predictions[0]]}

    async def predict_proba(input_doc: str):
        predictions = await model.predict_proba.async_run([input_doc])
        return predictions[0]

    return predict, predict_proba
model_lst = []
api_dict = {}
for idx, available_model in enumerate(bentoml.models.list()):
    if "twenty_news_group" in available_model.tag.name:
        print(f"Creating Endpoint {idx}")
        model_lst.append(available_model)
        bento_model = bentoml.sklearn.get(f"{available_model.tag.name}:latest")
        target_names = bento_model.custom_objects["target_names"]
        path_predict = f"predict_model_{idx}"
        path_predict_proba = f"predict_proba_model_{idx}"
        predict, predict_proba = wrap_service_methods(bento_model, target_names)
        api_dict[path_predict_proba] = bentoml.api(
            func=predict_proba,
            input_spec=Text(),
            output_spec=JSON(),
            name=path_predict_proba,
            route=path_predict_proba
        )
        api_dict[path_predict] = bentoml.api(
            func=predict,
            input_spec=Text(),
            output_spec=JSON(),
            name=path_predict,
            route=path_predict
        )
@bentoml.service(
workers=1, resources={"cpu": "1"}, apis=api_dict, models=model_lst
)
class DynamicService:
    def __init__(self):
        pass
```

holzweber: Edit: i added the methods via the locals() function

aarnphm: you can probably use types.new_class here, or even type() to construct subclass.

holzweber: true, i tried it with type() and it seems to work. updated my latest push.. so it is visible now :)

frostming:
The legacy Service class is deprecated now and we are not going to add features to it. Closing it now.

PR #4524, opened 2024-02-23
feat: add back service hooks

frostming:

## What does this PR address?

```python
import bentoml

@bentoml.service
class MyService:
    def __init__(self) -> None:
        # do something on service startup, per replica
        ...

    @bentoml.on_shutdown   
    def on_shutdown(self) -> None:
        # do something on service shutdown
        ...
```


bojiang:
Any use case for on_service_deployment? @frostming

bojiang (changes requested):
* not add on_deployment now
* not add ctx for on_shutdown

bojiang (approved):

PR #4534, opened 2024-02-28
chore(io): allow implicit pydantic type conversion

aarnphm:


src/_bentoml_sdk/api.py
```diff
@@ -28,13 +29,19 @@ def _only_include(data: dict[str, t.Any], fields: t.Container[str]) -> dict[str,
     return {k: v for k, v in data.items() if k in fields}
 
 
+def _io_descriptor_converter(it: t.Any) -> type[IODescriptor]:
+    if isinstance(it, type) and issubclass(it, IODescriptor):
+        return it
+    return ensure_io_descriptor(it)
```
frostming, line 35: Let's restrict the `it` to subclass of `BaseModel` only, as we are not supporting plain types for now.

frostming: Also can you fix the class name returned by `ensure_io_descriptor` since it has been only used by output types before.

aarnphm: done. PTAL.

frostming (approved):

frostming (approved):

PR #4544, opened 2024-03-01
docs: Add how to get deployment info doc

Sherlock113:
## What does this PR address?

Fixes #(issue)

## Before submitting:


docs/source/bentocloud/how-tos/manage-deployments.rst
```diff
@@ -40,41 +40,58 @@ To retrieve details about a specific Deployment:
...
+      # Output: DeploymentInfo(name='deploy-1', admin_console='https://test.cloud.bentoml.com/deployments/deploy-1/access?cluster=aws-ca-1&namespace=test--aws-ca-1', created_at='2024-03-01 05:00:19', created_by='bentoml-user', cluster='aws-ca-1')
+
+    To check the Deployment's status:
+
+    .. code-block:: python
+
+      import bentoml
+
+      dep = bentoml.deployment.get(name="deploy-1")
+      status = dep.get_status()
```
FogDong, line 91: Should we explicitly show all the valid method and attributes that `dep` have?

Also,  get_status has one parameter, refetch, like: `dep.get_status(refetch=True)`. By default, refetch is True, but user can also set it to False

Sherlock113: I suggest we only list the common ones, but depending on their purpose, they may be covered in different docs. So in addition to `get_status()`, what are other methods or attributes we can use for `dep`?

Sherlock113: I added some paramaters and methods. Please take a look again. Thanks.

docs/source/bentocloud/how-tos/manage-deployments.rst
```diff
@@ -40,41 +42,83 @@ To retrieve details about a specific Deployment:
...
-            traffic:
-              timeout: 10
-    status:
-      status: running
-      created_at: '2024-02-20 09:27:52'
-      updated_at: '2024-02-21 05:46:18'
+      import bentoml
+
+      dep = bentoml.deployment.get(name="deploy-1")
+      print(dep)
```
FogDong, line 84: I think we should introduce `dep.to_dict()` here, since this method will include everything like config and endpoint_urls in deployment.

docs/source/bentocloud/how-tos/manage-deployments.rst
```diff
@@ -40,41 +42,83 @@ To retrieve details about a specific Deployment:
...
+      import bentoml
+
+      dep = bentoml.deployment.get(name="deploy-1")
+      status = dep.get_status()
+      print(status.to_dict()) # Show the current status of the Deployment
+      # Output: {'status': 'running', 'created_at': '2024-03-01 05:00:19', 'updated_at': '2024-03-06 03:55:17'}
+
+    .. note::
+
+       ``get_status()`` has a parameter ``refetch`` to automatically refresh the status, which defaults to ``True``. You can use ``dep.get_status(refetch=False)`` to disable it.
```
FogDong, line 106: do we need to include `get_bento()` and `get_endpoint_urls()` as well?

docs/source/bentocloud/how-tos/manage-deployments.rst
Sherlock113: Added.

docs/source/bentocloud/how-tos/manage-deployments.rst
Sherlock113: Added `get_bento()`, but `get_endpoint_urls()` does not seem to be working. I used the following and I got an AttributeError: `'DeploymentInfo' object has no attribute 'get_endpoint_urls'`

```python
import bentoml
dep = bentoml.deployment.get(name="deploy-1")
urls = dep.get_endpoint_urls()
print(urls)
```

But I think this is fine as it was already documented here.
Screenshot 2024-03-06 at 14 57 33

FogDong (approved):

PR #4612, opened 2024-03-26
feat: download git+ssh dependencies to wheels/ folder

frostming:

## What does this PR address?

Fixes #(issue)

## Before submitting:

### DIFF
--- a/src/_bentoml_sdk/method.py
+++ b/src/_bentoml_sdk/method.py
@@ -34,7 +34,7 @@ def _only_include(data: dict[str, t.Any], fields: t.Container[str]) -> dict[str,
 def _io_descriptor_converter(it: t.Any) -> type[IODescriptor]:
     if not inspect.isclass(it):
         raise ValueError(f"{it} must be a class type")
-    if not issubclass(it, pydantic.BaseModel):
+    if not issubclass(it, (pydantic.BaseModel, dict)):
         raise ValueError(f"{it} is not a valid IODescriptor accepted type.")
     if issubclass(it, IOMixin):
         return it
--- a/src/bentoml/_internal/utils/__init__.py
+++ b/src/bentoml/_internal/utils/__init__.py
@@ -273,6 +273,39 @@ def copy_file_to_fs_folder(
     fs.copy.copy_file(src_fs, file_name, dst_fs, dst_path)
 
 
+def download_and_zip_git_repo(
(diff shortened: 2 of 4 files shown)


src/_bentoml_sdk/method.py
```diff
@@ -34,7 +34,7 @@ def _only_include(data: dict[str, t.Any], fields: t.Container[str]) -> dict[str,
 def _io_descriptor_converter(it: t.Any) -> type[IODescriptor]:
     if not inspect.isclass(it):
         raise ValueError(f"{it} must be a class type")
-    if not issubclass(it, pydantic.BaseModel):
+    if not issubclass(it, (pydantic.BaseModel, dict)):
```
aarnphm, line 37: should we separate this out?

seems like not related to this PR?

src/bentoml/_internal/utils/__init__.py
```diff
@@ -273,6 +273,39 @@ def copy_file_to_fs_folder(
...
+            subprocess.check_call(git_command, env=env)
+            if ref:
+                subprocess.check_call(["git", "fetch", "-q", url, ref], cwd=dest_dir)
+                subprocess.check_call(["git", "checkout", "FETCH_HEAD"], cwd=dest_dir)
+        except subprocess.CalledProcessError as e:
+            raise BentoMLException(
+                f"Failed to clone git repository {url}: {e.stderr}"
+            ) from e
+        zipball = os.path.join(dst_path, subdirectory or name)
+        shutil.rmtree(os.path.join(dest_dir, ".git"), ignore_errors=True)
```
aarnphm, line 303: should we just do a shadow clone? or even a sparse checkout instead?

frostming: This piece of code is being used in both pip and unearth.

I think the reason is we do not know whether user has specified the `@ref` part, in that case, we have to clone the full history and checkout to it later. the `ref` is an arbitrary legal git reference so we can't pass it in `--branch` option with `git clone`.

We could do a shallow clone as a shorcut when `ref` isn't given.

aarnphm: I see. Is there a current use case for this?

frostming: The following git url is allowed:
```
git+ssh://git@github.com/pallets/flask.git@dev
```

aarnphm (approved):

PR #4680, opened 2024-04-19
fix(cloud client): various type error

bojiang:
fix bentoml push failed and more

### DIFF
--- a/src/bentoml/_internal/cloud/deployment.py
+++ b/src/bentoml/_internal/cloud/deployment.py
@@ -308,13 +308,12 @@ def get_bento_info(
         except NotFound:
             bento_schema = None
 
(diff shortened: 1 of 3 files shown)


src/bentoml/_internal/cloud/deployment.py
```diff
@@ -379,8 +379,8 @@ def to_dict(self) -> dict[str, t.Any]:
             "created_at": self.created_at,
             "created_by": self.created_by,
             "config": (
-                self.get_config(refetch=False).to_dict(with_meta=False)
-                if self.get_config(refetch=False) is not None
+                config.to_dict(with_meta=False)
```
FogDong, line 382: Do we need this line?

FogDong (approved):

PR #4699, opened 2024-05-02
fix: cloud login now also updates context

jianshen92:
## What does this PR address?
currently, using `bentoml cloud login` will not change the context, so it is not exactly "logging in".

Fixes #(issue)

## Before submitting:

### DIFF
--- a/src/bentoml/_internal/cloud/config.py
+++ b/src/bentoml/_internal/cloud/config.py
@@ -41,7 +41,7 @@ class CloudClientContext:
                     "Unable to get current user from yatai server"
                 )
             self.email = user.email
-            add_context(self, ignore_warning=True)
+            _ = add_context(self, ignore_warning=True)
         return self.email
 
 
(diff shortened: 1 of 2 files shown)


src/bentoml/_internal/cloud/config.py
```diff
@@ -41,7 +41,7 @@ def get_email(self) -> str:
                     "Unable to get current user from yatai server"
                 )
             self.email = user.email
-            add_context(self, ignore_warning=True)
+            _ = add_context(self, ignore_warning=True)
```
frostming, line 44: What is the purpose of using a underscore to keep the result?

src/bentoml_cli/cloud.py
```diff
@@ -61,7 +61,8 @@ def login(shared_options: SharedOptions, endpoint: str, api_token: str) -> None:
         email=user.email,
     )
 
-    add_context(ctx)
+    new_config = add_context(ctx)
+    _ = new_config.set_current_context(ctx.name)
```
frostming, line 65: Why do we update the current context? This would be a behavior change.

I often do `bentoml cloud login ... --context dev` and then `bentoml push --context dev`, because I want to use the dev environment for testing but don't want to update the current context.

jianshen92: I see. Because if we have multiple context:

doing `bentoml cloud login ...` as copied from the UI doesn't actually log you in the context. As a user I would expect when i login my context will change to the one i just logged in.

frostming: > doing `bentoml cloud login ...` as copied from the UI doesn't actually log you in the context.

Okay, but the approach to fixing it is wrong. We should instead, login using the **current active context** when no `--context` is given.

aarnphm: I agree with @frostming here, we shouldn't updating the context automatically like this, which is confusing.

Probably better to be explicit at this case.

PR #4723, opened 2024-05-09
feat(build): expose `pip_preheat_packages`

aarnphm:
With the addition of #4690, added a field under `docker.pip_preheat_packages` under bentofile.yaml allowing user to specifying a list of dependencies to be preheated within the cache layers for improvement of build time

```yaml
docker:
  pip_preheat_packages:
    - vllm==0.4.2
    - lmdeploy
```

Would be useful for openllm as openllm will lock specific vllm version.

### DIFF
--- a/src/bentoml/_internal/bento/build_config.py
+++ b/src/bentoml/_internal/bento/build_config.py
@@ -174,6 +174,7 @@ class DockerOptions:
     setup_script: t.Optional[str] = None
     base_image: t.Optional[str] = None
     dockerfile_template: t.Optional[str] = None
+    pip_preheat_packages: t.Optional[t.List[str]] = None
 
     def __attrs_post_init__(self):
         if self.base_image is not None:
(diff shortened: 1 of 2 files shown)


aarnphm:
@bojiang any comments on this?

eledhwen:
This would be a great addition for us, another issue we have with pre-heating is that it does not play nice with pip markers, eg.

```
torch==2.3.0 ; platform_machine!='x86_64'
torch==2.3.0 --index-url https://download.pytorch.org/whl/cpu ; platform_machine=='x86_64'
```

..or with a specific wheel for a given target platform, the syntax throws the direct `pip install {} || true` off.

Would it be a possibility to generate an intermediary requirements file with the exact same definition found in the original, and `pip install -r` that instead?

bojiang:
@eledhwen  The issue should be fixed here: https://github.com/bentoml/BentoML/pull/4737

bojiang:
What are the packages you wanna cover if we supported custom pip_preheat_packages ?

aarnphm:
@bojiang this is different right? #4737 doesn't support custom packages, rather just finding torch or vllm from our predefined packages right?

I guess we do parse the python packages for vllm and torch. But then it will be pretty hard for us to manage additional packages in the future. Should we let users to have control over this?

PR #4729, opened 2024-05-14
feat: add a switch to opt-out packing git packages

frostming:

## What does this PR address?

Fixes #(issue)

## Before submitting:

### DIFF
--- a/src/bentoml/_internal/bento/build_config.py
+++ b/src/bentoml/_internal/bento/build_config.py
@@ -463,7 +463,11 @@ class PythonOptions:
         default=None,
         validator=attr.validators.optional(attr.validators.instance_of(ListStr)),
     )
-    lock_packages: t.Optional[bool] = attr.field(
+    lock_packages: bool = attr.field(
+        default=True,
+        validator=attr.validators.optional(attr.validators.instance_of(bool)),
+    )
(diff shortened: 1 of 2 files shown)


src/bentoml/_internal/bento/build_config.py
```diff
@@ -463,7 +463,11 @@ class PythonOptions:
         default=None,
         validator=attr.validators.optional(attr.validators.instance_of(ListStr)),
     )
-    lock_packages: t.Optional[bool] = attr.field(
+    lock_packages: bool = attr.field(
```
aarnphm, line 466: Should we just don't pack git packages when `lock_packages=False`? I don't think having a separate option makes sense here. 

If user don't want to pack git packages, then they shouldn't lock other non-git packages as well, and vice versa. Probably more easier to understand

frostming: `pack_git_packages` solves multiple problems. One is `pip-tools` doesn't lock the git commit hash when locking packages, the other is to remove the need of ssh keys during `containerize`.

There are chances that users don't want to lock packages, while still don't want to inject ssh keys into containerize.

bojiang (approved):

PR #4755, opened 2024-05-26
fix(build): make sure to install .tar.gz wheels

aarnphm:
this will fix installing bentoml dev wheels.

### DIFF
--- a/src/bentoml/_internal/bento/install.sh.j2
+++ b/src/bentoml/_internal/bento/install.sh.j2
@@ -37,6 +37,14 @@ if [ ${#wheels[@]} -gt 0 ]; then
 fi
 {% endraw %}
 
+targzs=($WHEELS_DIR/*.tar.gz)
+{% raw %}
+if [ ${#targzs[@]} -gt 0 ]; then
+    echo "Installing tar.gz packaged in Bento.."
+    pip3 install "${PIP_ARGS[@]}" "${targzs[@]}"
+fi
+{% endraw %}
+
 # Install the BentoML from PyPI if it's not already installed
 if python3 -c "import bentoml" &> /dev/null; then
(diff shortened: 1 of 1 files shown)


frostming:
Why do we need to install tar.gz separately? the dev sdist will be added to the `requirements.txt`:

https://github.com/bentoml/BentoML/blob/a882bfbc25dd029d2a90bfbbf49b02200e773ca8/src/bentoml/_internal/bento/build_config.py#L591-L592

aarnphm:
> Why do we need to install tar.gz separately? the dev sdist will be added to the `requirements.txt`:
> 
> https://github.com/bentoml/BentoML/blob/a882bfbc25dd029d2a90bfbbf49b02200e773ca8/src/bentoml/_internal/bento/build_config.py#L591-L592

currently we only check for ending with *.whl. this means .tar.gz won't be installed.

frostming:
> currently we only check for ending with *.whl. this means .tar.gz won't be installed.

I mean we add them into the requirements.txt and it will be picked up by `pip install -r requirements.txt`

aarnphm:
oh what happened to custom .tar.gz that can be installed with pip?

I guess I run into this issue with openllm, but then wrote a dockerfile template to install the custom .tar.gz separately.

edit: I think for this case users should write custom dockerfile template then, since installing arbitrary tar seems like a bad idea regardless.

frostming:
> oh what happened to custom .tar.gz that can be installed with pip?

How to add these tarballs? we only support `python.wheels` in build config

aarnphm:
hmm I guess `wheels=[wheel_fs.getsyspath(f"/{i.split('/')[-1]}") for i in built_wheels]`.

but probably out of scope. I don't think this is needed.

PR #4801, opened 2024-06-13
Use `orjson` for JSON serialization

judahrand:
## What does this PR address?

`orjson` seems to perform ~10x better than the stdlib `json` for roundtrip serialization:

```python
In [1]: import json

In [2]: import orjson

In [3]: import random

In [4]: py_list = [random.random() for x in range(1000)]

In [5]: %timeit orjson.loads(orjson.dumps(py_list))
40.9 μs ± 178 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [6]: %timeit json.loads(json.dumps(py_list))
522 μs ± 8.43 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [7]: %timeit orjson.loads(orjson.dumps('string'))
93.4 ns ± 0.882 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [8]: %timeit json.loads(json.dumps('string'))
824 ns ± 3.81 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

Fixes #(issue)

## Before submitting:

### DIFF
--- a/pdm.lock
+++ b/pdm.lock
@@ -5,7 +5,7 @@
 groups = ["default", "all", "testing", "io", "grpc-channelz", "tracing-otlp", "monitor-otlp", "io-image", "aws", "docs", "tracing-zipkin", "grpc-reflection", "grpc", "tracing", "tooling", "tracing-jaeger", "io-pandas", "triton"]
 strategy = ["cross_platform", "inherit_metadata"]
 lock_version = "4.4.1"
-content_hash = "sha256:5cf1da64b609193bb96742f04b8701869268af8c9d029fc9be7d2c375a21c222"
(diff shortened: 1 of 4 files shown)


frostming:
This should better be an optional dependency and expose a uniform API from a wrapper module.

judahrand:
> This should better be an optional dependency and expose a uniform API from a wrapper module.

Why do you think that is necessary? `pydantic` is already a dependency which (if wheels are not available) requires a Rust toolchain. It doesn't feel like adding `orjson` is much of a stretch especially since it builds wheels for basically every platform. In addition it is only ~500kB.

PR #4807, opened 2024-06-14
fix(monitoring): don't override user set environment variables

judahrand:
## What does this PR address?

This PR is stacked on top of #4808 

It shocked me that this bit of code sets environment variables! Does this not make it possible for `monitoring` and `tracing` configuration to conflict?

https://github.com/bentoml/BentoML/pull/3257#discussion_r1639956769

## Before submitting:

### DIFF
--- a/src/bentoml/_internal/monitoring/otlp.py
+++ b/src/bentoml/_internal/monitoring/otlp.py
@@ -27,7 +27,12 @@ from ..context import trace_context
 from .base import MonitorBase
 
 try:
-    from opentelemetry.exporter.otlp.proto.http._log_exporter import OTLPLogExporter
+    from opentelemetry.exporter.otlp.proto.grpc._log_exporter import (
+        OTLPLogExporter as OTLPGrpcLogExporter,
+    )
+    from opentelemetry.exporter.otlp.proto.http._log_exporter import (
+        OTLPLogExporter as OTLPHttpLogExporter,
+    )
 except ImportError:
     raise MissingDependencyException(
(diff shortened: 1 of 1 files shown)


aarnphm (changes requested):
no, we need to set via environment for controlling on kubernetes.

I'm find with having hierarchical to check for env -> variable

gPRC exporter seems to be a separate PR.

judahrand:
> no, we need to set via environment for controlling on kubernetes.

Why? Is that specific to BentoCloud because the user doesn't actually have control over the Pod's spec? Otherwise, the user should still be responsible for setting their environment variables of their own Pods?

> I'm find with having hierarchical to check for env -> variable

That still strikes me as leading so some slightly odd behaviour. For example, imagine you want to send traces and logs to two different OTLP Collectors (as I do). I set the environment variables match the endpoint I want to send the traces to and I configure the logs via code on in the `bentoml.service` decorator. Where do my traces go? 🤷‍♂️ I genuinely don't know the answer but it seems very confusing and delicate and could change unexpectedly given that the code is sometimes manually overriding my environment variables but sometimes not.

judahrand:
> gPRC exporter seems to be a separate PR.

True... but also some of the arguments and environment variables being set literally do nothing 😆 `insecure` for example seems to only be relevant for the GRPC exporter. Happy to split it out but surprised it made it through code review.

judahrand:
> no, we need to set via environment for controlling on kubernetes.

Alternatively, should we be setting only the `OTEL_EXPORTER_OTLP_LOGS_ENDPOINT` variable for the endpoint? Still not sure setting these at all is great. But it'd be better.

aarnphm:
we can revisit this if you have bandwidth

PR #4845, opened 2024-07-09
cli: Fix deployment CLI not working with `deploy . -f config.yaml`

jianshen92:
## What does this PR address?
`bentoml deploy . -f config.yaml` ignores the config file passed in.

Fixes #(issue)

## Before submitting:

### DIFF
--- a/src/bentoml/_internal/cloud/deployment.py
+++ b/src/bentoml/_internal/cloud/deployment.py
@@ -73,17 +73,19 @@ class DeploymentConfigParameters:
     def verify(
         self,
     ):
-        deploy_by_param = (
-            self.name
-            or self.bento
-            or self.cluster
-            or self.access_authorization
-            or self.scaling_min
-            or self.scaling_max
-            or self.instance_type
-            or self.strategy
-            or self.envs
-            or self.extras
+        deploy_by_param = any(
+            param is not None and not "" and len(param) > 0
+            for param in [
+                self.name,
+                self.cluster,
+                self.access_authorization,
+                self.scaling_min,
+                self.scaling_max,
+                self.instance_type,
+                self.strategy,
+                self.envs,
+                self.extras,
+            ]
         )
 
         if (
@@ -151,6 +153,11 @@ class DeploymentConfigParameters:
 
         if self.cfg_dict is None:
             self.cfg_dict = {}
+
+        # This is a special case to allow
+        # bentoml deploy . -f config.yaml to work
+        if self.bento is not None:
+            self.cfg_dict["bento"] = self.bento
 
         bento_name = self.cfg_dict.get("bento")
         # determine if bento is a path or a name
--- a/src/bentoml/_internal/cloud/deployment.py
+++ b/src/bentoml/_internal/cloud/deployment.py
@@ -89,12 +89,9 @@ class DeploymentConfigParameters:
         )
 
         if (
-            self.config_dict
-            and self.config_file
-            or self.config_dict
-            and deploy_by_param
-            or self.config_file
-            and deploy_by_param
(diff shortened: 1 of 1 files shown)


src/bentoml/_internal/cloud/deployment.py
```diff
@@ -73,26 +73,25 @@ class DeploymentConfigParameters:
     def verify(
         self,
     ):
-        deploy_by_param = (
-            self.name
-            or self.bento
```
jianshen92, line 78: Removing self.name here

Haivilo: Are we removing `self.name` or `self.bento`?

src/bentoml/_internal/cloud/deployment.py
```diff
@@ -73,26 +73,25 @@ class DeploymentConfigParameters:
...
-            or self.cluster
-            or self.access_authorization
-            or self.scaling_min
-            or self.scaling_max
-            or self.instance_type
-            or self.strategy
-            or self.envs
-            or self.extras
+        deploy_by_param = any(
+            param is not None and not "" and len(param) > 0
```
frostming, line 77: What does `not ""` do here?

src/bentoml/_internal/cloud/deployment.py
jianshen92: Sorry, I meant removing self.bento

src/bentoml/_internal/cloud/deployment.py
jianshen92: Wanted to check empty string. An oversight. Made a fix

src/bentoml/_internal/cloud/deployment.py
```diff
@@ -73,26 +73,25 @@ class DeploymentConfigParameters:
...
-            or self.cluster
-            or self.access_authorization
-            or self.scaling_min
-            or self.scaling_max
-            or self.instance_type
-            or self.strategy
-            or self.envs
-            or self.extras
+        deploy_by_param = any(
+            param is not None and len(param) > 0
```
frostming, line 77: I am afraid you can't judge with this, given there are integer values (scaling_min and scaling_max)
Can we just filter None, meaning it is specified by the users

src/bentoml/_internal/cloud/deployment.py
FogDong: Why we need to remove `self.bento` here?

frostming: Otherwise it would be rejected by the check:

```python
if (
            (self.config_dict and self.config_file)
            or (self.config_dict and deploy_by_param)
            or (self.config_file and deploy_by_param)
        ):
            raise BentoMLException(
                "Configure a deployment can only use one of the following: config_dict, config_file, or the other parameters"
```

jianshen92: Because `bentoml deploy .` will consider `.` as the bento and this check evaluates to true and then it will skip the file argument

src/bentoml/_internal/cloud/deployment.py
jianshen92: Yeah initially that was the case but it then the default for envs is a list so it will always evaluate to True if we dont check for it

jianshen92:
Decided to throw error if user use both inline argument and file.
User should specify the `name` in the bentofile, i.e `name: '.'`

CC: @Sherlock113 to update documentation

jianshen92:
Decided to throw error if user use both inline argument and file.
User should specify the name in the bentofile, i.e name: '.'

CC: @Sherlock113 to update documentation

PR #4860, opened 2024-07-15
fix: nvcc not found

serkankavak:
## What does this PR address?
This PR fixes the installation process of packages that are dependent on CUDA. 

One of my Python packages is `deepspeed`, and it requires CUDA while installation. However, I was getting error as `nvcc: command not found`. After searching for a while, I came across with [this issue](https://github.com/NVIDIA/nvidia-container-toolkit/issues/161). If you use `devel`, instead of `runtime` version of CUDA, nvcc commands become available. 
Otherwise, for `runtime` version, `nvcc` and other CUDA related commands are available only if you use `docker run --gpus all` but since you get the error while building container, you can't even come to that part.

I tested this fix only on Ubuntu 20.04. UBI 8 distro is also using `runtime` instead of `devel` at line 84 in the same file. If anyone tests that distro it would be helpful for the future improvements as well.

### DIFF
--- a/src/bentoml/_internal/container/frontend/dockerfile/__init__.py
+++ b/src/bentoml/_internal/container/frontend/dockerfile/__init__.py
@@ -93,7 +93,7 @@ CONTAINER_METADATA: dict[str, dict[str, t.Any]] = {
             "supported_architectures": SUPPORTED_ARCHITECTURES,
         },
         "cuda": {
-            "image": "nvidia/cuda:{spec_version}-cudnn8-runtime-ubuntu20.04",
+            "image": "nvidia/cuda:{spec_version}-cudnn8-devel-ubuntu20.04",
             "supported_architectures": ["amd64", "arm64"],
         },
         "miniconda": {


frostming:
@bojiang Please comment if this fix is acceptable, would this make the base image much larger?

aarnphm:
i don't think this make sense, since the base image will be significantly larger. you can already change the base image under docker of bentofile's `base_image`: https://docs.bentoml.com/en/latest/guides/build-options.html#docker-options-table

serkankavak:
> i don't think this make sense, since the base image will be significantly larger. you can already change the base image under docker of bentofile's `base_image`: https://docs.bentoml.com/en/latest/guides/build-options.html#docker-options-table

I tried this approach as well. However, `base_image` overrides other options so you can't install Python and other dependencies automatically. How about adding a new option (`cuda_type` maybe?) to the docker field under bentofile.yaml such as:
```yaml
docker:
  distro: debian
  python_version: "3.10"
  cuda_version: "12.1.1"
  cuda_type: "devel"  # Default value: runtime
```

aarnphm:
you can also customise the base dockerfile with `dockerfile_template` to include python installation if you want to. I don't think that customising cuda_type at this granular level is beneficial as it would add additional complexity to this field.

Feel free to ask if you have anymore questions wrt this.

PR #4894, opened 2024-07-31
feature: add custom runner support for time series models like ARIMA using BentoML

jsamantaucd:
Feature request
I have implemented and tested a feature that trains a time series model - ARIMA and serves it using BentoML's custom runner to make predictions.

Motivation
I believe it would be a nice guide to someone exploring time series models with custom runner implementation for models like ARIMA.

Others:
This feature was tested on tag v1.0.20


sebasmos:
oh that's a great feature

xianml:
It would be Great if it is based on bentoml >= 1.2

jsamantaucd:
Sure @xianml, I will create this example with bentoml >= 1.2, thanks for suggesting.

PR #4935, opened 2024-08-23
fix: uploading large files saving to disk instead of memory

frostming:

## What does this PR address?

Fixes #(issue)

## Before submitting:

### DIFF
--- a/src/bentoml/_internal/cloud/bentocloud.py
+++ b/src/bentoml/_internal/cloud/bentocloud.py
@@ -1,8 +1,8 @@
 from __future__ import annotations
 
+import math
 import tarfile
 import tempfile
-import threading
 import typing as t
 import warnings
 from concurrent.futures import ThreadPoolExecutor
@@ -84,6 +84,7 @@ class BentoCloudClient(CloudClient):
         threads: int = 10,
         rest_client: RestApiClient = Provide[BentoMLContainer.rest_api_client],
         model_store: ModelStore = Provide[BentoMLContainer.model_store],
+        bentoml_tmp_dir: str = Provide[BentoMLContainer.tmp_bento_store_dir],
     ):
         name = bento.tag.name
         version = bento.tag.version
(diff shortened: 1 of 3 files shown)


src/bentoml/_internal/cloud/bentocloud.py
```diff
@@ -663,38 +663,37 @@ def _do_push_model(
...
 
-        io_mutex = threading.Lock()
-
         def io_cb(x: int):
-            with io_mutex:
-                self.spinner.transmission_progress.update(upload_task_id, advance=x)
+            self.spinner.transmission_progress.update(upload_task_id, advance=x)
 
-        with CallbackIOWrapper(read_cb=io_cb) as tar_io:
+        with NamedTemporaryFile(
```
sauyon, line 669: Should we use `SpooledTemporaryFile` instead?

sauyon:
I think there is a separate issue with building large Bentos on Linux (I think we do actually use temp directory for building?), so I can't test this, but LGTM otherwise.

src/bentoml/_internal/cloud/bentocloud.py
frostming: NamedTemporyFile isn't a in-memory approach when `dir` points to a custom location. But `SpooledTemporaryFile` keeps file in memory with a max size, I don't see much benefit because it must be saved to disk in the later multipart uploading step.

frostming:
> I think there is a separate issue with building large Bentos on Linux (I think we do actually use temp directory for building?), so I can't test this, but LGTM otherwise.

Yes, let's improve it later.

src/bentoml/_internal/cloud/bentocloud.py
sauyon: Why does it have to be saved?

frostming: Because we need to split it into several chunks to be read from multi-thread. The idea is we don't want to load the whole chunk into memory, so we save the file and read it with separate file pointers.

sauyon: Seems like we could use the buffer like we did before if it's still in memory, but maybe that's too much work for a temporary fix.

frostming: That will be an implementation detail hidden behind `SpooledTemporaryFile` and we don't need to change other code.

frostming: Yes, making such distinctions will make the code verbose and ugly.

sauyon (approved):

bojiang:
Tested with 15G models. The time is acceptable.

bojiang (approved):

PR #4949, opened 2024-08-30
feat(framework): unsloth

aarnphm:
follow the https://colab.research.google.com/drive/1Ys44kVvmeZtnICzWz0xgpRnrIOjZAuxp?usp=sharing#scrollTo=QmUBVEnvCDJv

Then one can save with `bentoml.unsloth.build_bento`

```python
bentoml.unsloth.build_bento(model, tokenizer, chat_template="<chat_template_that_trained_with>", quantization_method="bitsandbytes")
```

If model is continued from a fine-tune checkpoint, then model_name should be passed:

```python
bentoml.unsloth.build_bento(model, tokenizer, model_name="llama-3-continued-from-checkpoint", chat_template="<chat_template_that_trained_with>", quantization_method="bitsandbytes")
```

supports llama3, gemma, gemma2, mistral and most chat templates that is supported with fine-tune. Based on BentoVLLM and OpenLLM logics for template.


src/_bentoml_impl/frameworks/unsloth/template/.gitattributes
```diff
@@ -0,0 +1,2 @@
+* text=auto eol=lf
+ui/** linguist-vendored=true
```
aarnphm, line 2:
```suggestion
**/ui/** linguist-vendored=true
```

src/_bentoml_impl/frameworks/unsloth/mapping.py
```diff
@@ -0,0 +1,49 @@
+RUNTIME_MAPPING = {
+    "llama": {
+        "service_config": {
+            "traffic": {"timeout": 300},
+            "resources": {"gpu": 1, "gpu_type": "nvidia-l4"},
```
bojiang, line 5: llama 8b and llama 70b will need different resources

src/_bentoml_impl/frameworks/unsloth/__init__.py
```diff
@@ -0,0 +1,187 @@
...
+
+    try:
+        bentomodel = bentoml.models.get(model_name)
+        if overwrite:
+            raise bentoml.exceptions.NotFound("Overwriting new models.")
+    except bentoml.exceptions.NotFound:
+        with bentoml.models.create(model_name) as bentomodel:
+            model.save_pretrained_merged(
+                bentomodel.path, tokenizer, save_method=save_method
+            )
```
ssheng, line 119: It's OK to just save another copy. No need to add the complicated logic here.

src/_bentoml_impl/frameworks/unsloth/__init__.py
```diff
@@ -0,0 +1,187 @@
...
+    with tempfile.TemporaryDirectory() as tempdir:
+        tempdir = pathlib.Path(tempdir)
+        shutil.copytree(
+            pathlib.Path(__file__).parent / "template", tempdir, dirs_exist_ok=True
+        )
+        with (tempdir / "service_config.yaml").open("w") as f:
+            f.write(
+                yaml.safe_dump(
+                    dict(
+                        chat_template=chat_template,
```
ssheng, line 148: Why should BentoML be managing the chat template?

src/_bentoml_impl/frameworks/unsloth/template/service.py
```diff
@@ -0,0 +1,200 @@
...
+import uuid
+from typing import Any
+from typing import AsyncGenerator
+from typing import Literal
+from typing import Optional
+
+import fastapi
+import fastapi.staticfiles
+import pydantic
+import vllm.entrypoints.openai.api_server as vllm_api_server
```
ssheng, line 17: This is not needed.

src/_bentoml_impl/frameworks/unsloth/mapping.py
aarnphm: think that will be included in the example to change the resources for `build_bento`

src/_bentoml_impl/frameworks/unsloth/__init__.py
aarnphm: Otherwise, the prompt won't be correct here. Let say the model is fine-tuned with a specific chat template, then the service should also use that chat_template as override.

src/_bentoml_impl/frameworks/unsloth/mapping.py
aarnphm: Since users can continue fine tune from a saved checkpoint, and we can only get model_type from config

src/_bentoml_impl/frameworks/unsloth/template/service.py
aarnphm: Are you suggesting we shouldn't use vllm?

src/_bentoml_impl/frameworks/unsloth/__init__.py
ssheng: Chat templates can get updates. It shouldn't be BentoML's responsibility to keep the chat templates update-to-date. Where did we get these chat templates?

src/_bentoml_impl/frameworks/unsloth/template/service.py
ssheng: On a second thought, using vLLM is fine. Sorry for the confusion.

src/_bentoml_impl/frameworks/unsloth/__init__.py
aarnphm: I believe we got it from https://github.com/chujiezheng/chat_templates

ssheng: When a model is saved, isn't `tokenizer_config.json` saved with the model as well, which includes the bos/eos tokens and chat template? Having BentoML managing the chat template, eos/bos tokens, and system prompts doesn't seem right.

ssheng:
Could you please also create a documentation similar to this section of Unsloth? https://docs.unsloth.ai/tutorials/how-to-finetune-llama-3-and-export-to-ollama#id-13.-exporting-to-ollama

aarnphm:
> Could you please also create a documentation similar to this section of Unsloth? https://docs.unsloth.ai/tutorials/how-to-finetune-llama-3-and-export-to-ollama#id-13.-exporting-to-ollama

Should we submit this to unsloth?

ssheng:
> Should we submit this to unsloth?

We should but let's create a draft to review first.

src/_bentoml_impl/frameworks/unsloth/__init__.py
aarnphm: Not actually, save_pretrained won't apply the chat templates that user fine-tune the model with. Unless users override the default chat template.

The chat template comes with `tokenizer_config.json` might not be the one that users use to fine tune with.

aarnphm: They will have to do something like this: https://huggingface.co/docs/transformers/en/chat_templating#what-template-should-i-use

src/_bentoml_impl/frameworks/unsloth/train.py
```diff
@@ -0,0 +1,67 @@
...
+    formatting_prompts_func,
+    batched=True,
+  )
+  return dataset
+
+
+def main(max_seq_length: int = 8196) -> int:
+  import unsloth, bentoml, trl, transformers
+
+  model, tokenizer = unsloth.FastLanguageModel.from_pretrained("unsloth/Meta-Llama-3.1-8B-bnb-4bit", max_seq_length=max_seq_length, load_in_4bit=True)
```
aarnphm, line 41: tokenizer.chat_templates = "<alpaca_prompt>"

src/_bentoml_impl/frameworks/unsloth/template/service.py
```diff
@@ -0,0 +1,86 @@
...
+    ("/chat/completions", vllm_api_server.create_chat_completion, ["POST"]),
+    ("/completions", vllm_api_server.create_completion, ["POST"]),
+    ("/models", vllm_api_server.show_available_models, ["GET"]),
+]:
+    openai_api_app.add_api_route(
+        path=route, endpoint=endpoint, methods=methods, include_in_schema=True
+    )
+
+# special handling for prometheus_client of bentoml
+if "prometheus_client" in sys.modules:
```
frostming, line 46: Do we still need this? I think we already get rid of this hacking.

src/_bentoml_impl/frameworks/unsloth/__init__.py
```diff
@@ -0,0 +1,151 @@
...
+    if engine_config is None:
+        engine_config = {}
+
+    service_config.update({**MAPPINGS[model_type]["service_config"]})
+
+    engine_config.update(MAPPINGS[model.config.model_type]["engine_config"])
+    if quantization_method is not None:
+        engine_config.update(
+            {"quantization": quantization_method, "load_format": quantization_method}
+        )
```
ssheng, line 89: Why do we take `quantization_method` out as a separate parameter instead of relying on `engine_config`?

aarnphm: because for users who use bnb they also have to set load_format.

src/_bentoml_impl/frameworks/unsloth/train.py
```diff
@@ -0,0 +1,69 @@
+# ruff: ignore
```
ssheng, line 1: Does this need to be merged?

src/_bentoml_impl/frameworks/unsloth/__init__.py
ssheng: I think they are better down through documentation. This abstraction is not clear about its override priority. If `quantization` and `load_format` must be the same, we can add engine_config validation logic.

src/_bentoml_impl/frameworks/unsloth/train.py
aarnphm: this doesn't. I was just using this for testing.

PR #4960, opened 2024-09-06
fix: support pydantic 2.9

frostming:

## What does this PR address?

The new approach doesn't monkey patch globally and doesn't rely on internal members of pydantic.


pyproject.toml
```diff
@@ -54,7 +54,7 @@ dependencies = [
     "pip-requirements-parser>=31.2.0",
     "prometheus-client>=0.10.0",
     "psutil",
-    "pydantic<3",
```
aarnphm, line 57: did they have a breaking change in 2.9?

frostming: Yes, and I've updated the implementation to work for pydantic 2.9, this limitation was removed.

frostming:
The framework testing failures can be ignored as they are handled in dev/1.4 branch.

aarnphm (approved):

PR #4972, opened 2024-09-12
refactor(1.4): Make the cloud client APIs more structured and aligned

frostming:

## What does this PR address?

Make the cloud client APIs more structured and aligned

```python
import bentoml
import os

# Use default client
deployment = bentoml.deployment.get('mydeployment')
# Use from cloud client
client = bentoml.BentoCloudClient("token")
deployment = client.deployment.get("mydeployment")
# Env vars are supported
os.environ['BENTO_CLOUD_API_KEY'] = 'mytoken'
os.environ['BENTO_CLOUD_API_ENDPOINT'] = 'https://my.cloud.bentoml.org'
client = bentoml.BentoCloudClient()
deployment = client.deployment.get("mydeployment")
```

### DIFF
--- a/src/bentoml/__init__.py
+++ b/src/bentoml/__init__.py
@@ -29,7 +29,7 @@ from pydantic import Field
 
 # BentoML built-in types
 from ._internal.bento import Bento
-from ._internal.cloud import YataiClient
+from ._internal.cloud import BentoCloudClient
 from ._internal.context import ServiceContext as Context
 from ._internal.context import server_context
 from ._internal.models import Model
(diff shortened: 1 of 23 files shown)


src/bentoml/__init__.py
```diff
@@ -29,7 +29,7 @@
 
 # BentoML built-in types
 from ._internal.bento import Bento
-from ._internal.cloud import YataiClient
```
aarnphm, line 32: hmm what about OSS yatai users?

frostming: good question, in fact the yatai client has been broken for a long time. you can't run it on main nor recent releases. i think we intentionally don't put effort in it. so i didn't touch it. everything remains as it is for yatai

aarnphm: fair enough.

PR #5008, opened 2024-10-06
feat: gradio integration

parano:
## What does this PR address?

This PR simplifies how users can serve Gradio UI for their BentoML service. Previously user will need to manually do multiple settings and callbacks, e.g.:

```python
import bentoml
import gradio as gr

def summarize_text(text: str) -> str:
    svc_instance = bentoml.get_current_service()
    return svc_instance.summarize([text])[0]

io = gr.Interface(
    fn=summarize_text,
    inputs=[gr.Textbox(lines=5, label="Enter Text", value=EXAMPLE_INPUT)],
    outputs=[gr.Textbox(label="Summary Text")],
    title="Summarization",
    description="Enter text to get summarized text."
)
gradio_app = gr.routes.App.create_app(io)

io.dev_mode = False
io.show_error = True
io.validate_queue_settings()

@bentoml.mount_asgi_app(gradio_app, path="/ui")
@bentoml.service(resources={"cpu": "4"})
class Summarization:

    def __init__(self) -> None:
        import torch
        from transformers import pipeline

        # gradio_app.get_blocks().queue()
        gradio_app.get_blocks().startup_events()
        # gradio_app.get_blocks().root_path="/ui"

        ...
```

With this PR, this can be simplified as:

```python
def summarize_text(text: str) -> str:
    svc_instance = bentoml.get_current_service()
    return svc_instance.summarize([text])[0]

io = gr.Interface(
    fn=summarize_text,
    inputs=[gr.Textbox(lines=5, label="Enter Text", value=EXAMPLE_INPUT)],
    outputs=[gr.Textbox(label="Summary Text")],
    title="Summarization",
    description="Enter text to get summarized text."
)

@bentoml.service(resources={"cpu": "4"})
@bentoml.gradio.mount_gradio_app(io, path="/ui")
class Summarization:
      ...
```

## Before submitting:

### DIFF
--- a/examples/gradio/service.py
+++ b/examples/gradio/service.py
@@ -3,8 +3,8 @@ from __future__ import annotations  # I001
 import bentoml
 
 with bentoml.importing():
-    import torch
     import gradio as gr
+    import torch
     from transformers import pipeline
 
 EXAMPLE_INPUT = "Breaking News: In an astonishing turn of events, the small \
--- a/examples/gradio/service.py
+++ b/examples/gradio/service.py
@@ -1,11 +1,10 @@
 from __future__ import annotations  # I001
 
-import gradio as gr
-
 import bentoml
 
 with bentoml.importing():
     import torch
+    import gradio as gr
     from transformers import pipeline
 
 EXAMPLE_INPUT = "Breaking News: In an astonishing turn of events, the small \
--- a/examples/gradio/service.py
+++ b/examples/gradio/service.py
@@ -1,7 +1,8 @@
 from __future__ import annotations  # I001
 
-import bentoml
 import gradio as gr
+
+import bentoml
 
 with bentoml.importing():
     import torch
@@ -16,23 +17,24 @@ place in Thompson's backyard, is now being investigated by scientists for potent
 breaches in the laws of physics. Local authorities are considering a town festival \
 to celebrate what is being hailed as 'The Leap of the Century."
 
+
 def summarize_text(text: str) -> str:
     svc_instance = bentoml.get_current_service()
     return svc_instance.summarize([text])[0]
+
 
 io = gr.Interface(
     fn=summarize_text,
     inputs=[gr.Textbox(lines=5, label="Enter Text", value=EXAMPLE_INPUT)],
     outputs=[gr.Textbox(label="Summary Text")],
     title="Summarization",
-    description="Enter text to get summarized text."
+    description="Enter text to get summarized text.",
 )
 
 
 @bentoml.service(resources={"cpu": "4"})
 @bentoml.gradio.mount_gradio_app(io, path="/ui")
 class Summarization:
-
     def __init__(self) -> None:
         device = "cuda" if torch.cuda.is_available() else "cpu"
         self.pipeline = pipeline("summarization", device=device)
--- a/examples/gradio/service.py
+++ b/examples/gradio/service.py
@@ -1,0 +1,43 @@
+from __future__ import annotations  # I001
+
+import bentoml
+import gradio as gr
+
+with bentoml.importing():
+    import torch
+    from transformers import pipeline
+
+EXAMPLE_INPUT = "Breaking News: In an astonishing turn of events, the small \
+town of Willow Creek has been taken by storm as local resident Jerry Thompson's cat, \
+Whiskers, performed what witnesses are calling a 'miraculous and gravity-defying leap.' \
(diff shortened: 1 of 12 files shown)


src/_bentoml_sdk/service/factory.py
```diff
@@ -266,6 +271,14 @@ def add_asgi_middleware(
     ) -> None:
         self.middlewares.append((middleware_cls, options))
 
+    def gradio_app_startup_hook(self, max_concurrency: int):
+        gradio_apps = getattr(self.inner, "__bentoml_gradio_apps__", [])
+        if gradio_apps:
```
aarnphm, line 276:
```suggestion
        if len(gradio_apps) > 0:
```

src/bentoml/_internal/gradio.py
```diff
@@ -0,0 +1,71 @@
...
+    )
+
+
+def mount_gradio_app(
+    blocks: Blocks,
+    path: str,
+    name: str = "gradio_ui",
+):
+    from _bentoml_sdk.service import Service
+    from bentoml._internal import server
```
aarnphm, line 25: we don't call the server from here anymore right? it should be `_bentoml_impl.server` iirc.

aarnphm (commented):
probably want frost opinion on this:

For any integration going forward, should it be under `bentoml._internal` or `_bentoml_sdk`?

I think it should be the second option?

src/bentoml/_internal/gradio.py
frostming: +1

examples/gradio/service.py
```diff
@@ -0,0 +1,45 @@
...
+    return svc_instance.summarize([text])[0]
+
+
+io = gr.Interface(
+    fn=summarize_text,
+    inputs=[gr.Textbox(lines=5, label="Enter Text", value=EXAMPLE_INPUT)],
+    outputs=[gr.Textbox(label="Summary Text")],
+    title="Summarization",
+    description="Enter text to get summarized text.",
+)
```
frostming, line 32: I wonder can we automatically generate this from the schema?
Or do we want to give more flexibility to users?

src/bentoml/_internal/gradio.py
```diff
@@ -0,0 +1,71 @@
...
+        if isinstance(obj, Service):
+            # For scenario:
+            #
+            # @bentoml.gradio.mount_gradio_app(..)
+            # @bentoml.service(..)
+            # class MyService: ...
+            #
+            # If the Service instance is already created, mount the ASGI app immediately
+            target = obj.inner
+            obj.mount_asgi_app(gradio_app, path=path, name=name)
```
frostming, line 52: It seems not necessary to mount it here, otherwise it will be mounted twice?

examples/gradio/service.py
aarnphm: Isn't you need to have this interface to mount it to a bentoml service?

edit: I just realized that the fn called the service instance below :)

parano: we should give more flexibility to users. the gradio UI can be quite dynamic and includes UI specific logic, no need to couple it with the model inference / pipeline logic defined in BentoML APIs.

src/bentoml/_internal/gradio.py
parano: it has to do with the Service wrapper class creation time, similar to the mount_asgi_app method here: https://github.com/bentoml/BentoML/blob/main/src/_bentoml_sdk/decorators.py#L124

src/_bentoml_sdk/service/factory.py
```diff
@@ -276,8 +276,14 @@ def gradio_app_startup_hook(self, max_concurrency: int):
         if gradio_apps:
             for gradio_app, path, name in gradio_apps:
                 logger.info(f"Initializing gradio app at: {path or '/'}")
-                gradio_app.get_blocks().queue(default_concurrency_limit=max_concurrency)
-                gradio_app.get_blocks().startup_events()
+                blocks = gradio_app.get_blocks()
+                blocks.queue(default_concurrency_limit=max_concurrency)
+                if hasattr(blocks, "startup_events"):
```
aarnphm, line 281: should we support gradio 5.0 only?

for folks who need to write a new BentoML service they prob should also upgrade their gradio interface to new version?

ig for ppl who alr have gradio 4.0 then 🤔

frostming (approved):

PR #5039, opened 2024-10-23
feat: Opt-in bento image support

frostming:

### DIFF
--- a/src/_bentoml_sdk/images.py
+++ b/src/_bentoml_sdk/images.py
@@ -153,9 +153,11 @@ class PythonImage(Image):
                 raise BentoMLConfigException(
                     f"Unsupported distro: {self.distro}, expected one of {CONTAINER_SUPPORTED_DISTROS}"
                 )
-
+            python_version = self.python_version
+            if self.distro in ("ubi8",):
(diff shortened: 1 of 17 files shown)


src/bentoml/_internal/container/frontend/dockerfile/__init__.py
```diff
@@ -21,7 +21,7 @@
 
 
 # Python supported versions
-SUPPORTED_PYTHON_VERSIONS = ["3.8", "3.9", "3.10", "3.11", "3.12"]
+SUPPORTED_PYTHON_VERSIONS = ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
```
aarnphm, line 24: probably good to also drop 3.8

aarnphm (approved):
tiny

src/_bentoml_sdk/images.py
```diff
@@ -0,0 +1,239 @@
+from __future__ import annotations
```
aarnphm, line 1: Oh final thing, we should also expose this in `bentoml/__init__.py` so that `bentoml.Image` works

aarnphm (approved):

PR #5054, opened 2024-10-31
doc: suggest to use pip for CUDA first

bojiang:


docs/source/guides/gpu-inference.rst
```diff
@@ -98,6 +79,25 @@ This means when building your Bento, you DO NOT need to specify ``cuda_version``
...
+    service: "service:GPUSVC"
+    labels:
+      owner: bentoml-team
+      stage: demo
+    include:
+    - "*.py"
+    python:
+      requirements_txt: "./requirements.txt"
+    docker:
+      python_version: "3.11.7"
```
Sherlock113, line 98: No config here is about GPU specifically, right? My understanding is that we don't want to use `cuda_version`. Maybe we can just remove these lines and let people [see the doc](https://docs.bentoml.com/en/latest/guides/build-options.html#docker-options-table) here.

docs/source/guides/gpu-inference.rst
```diff
@@ -98,6 +79,25 @@ This means when building your Bento, you DO NOT need to specify ``cuda_version``
         - torch
         - tensorflow[and-cuda]
 
+If you know what you are doing and want to customize the installation of CUDA driver and libraries via ``system_packages``, ``setup_script``, or ``base_image`` options under the :ref:`docker-configuration` field.
```
Sherlock113, line 82:
```suggestion
If you want to customize the installation of CUDA driver and libraries, use ``system_packages``, ``setup_script``, or ``base_image`` options under the :ref:`docker-configuration` field.
```

docs/source/guides/gpu-inference.rst
bojiang: sure

Sherlock113:
```suggestion
```

docs/source/guides/gpu-inference.rst
```diff
@@ -98,6 +79,11 @@ This means when building your Bento, you DO NOT need to specify ``cuda_version``
         - torch
         - tensorflow[and-cuda]
 
+If you want to customize the installation of CUDA driver and libraries, use ``system_packages``, ``setup_script``, or ``base_image`` options under the :ref:`docker-configuration` field.
+
+
+
+
```
Sherlock113, line 86:
```suggestion
```

Sherlock113 (approved):

Issue #5058, opened 2024-11-01
feature: Stop and resume tasks


Leon-Sander:
### Feature request

I would like to be able to manage stopping and resuming of tasks.

### Motivation

If I have a long running task like large batch processing and the server needs a reboot for example, then I would want to save the intermediate results and resume later on.

frostming:
A task can be interrupted only if it can respond to interrupt signals and save the state of the scene for resumption.

This requires considerable efforts to implement and hard to make it generic. You can implement for specific frameworks but not for all.

Leon-Sander:
Yes absolutely, I agree. 

I thought about it and here is the idea I had. Making a generic pause and resuming, which can be activated by choice if one wants to implement the logic for his framework.

So @betoml.task(is_pausable=True).

It would just be necessary to make the status of the task_id accessible. 
is_pausable=True would make the api endpoints of pause and resume available. If the user does not want to implement it, is_pausable default is False and everything else stays the same.

frostming:
Hi @Leon-Sander  After our discussion, we decided not to do this because `bentoml.task` is not positioned as a mechanism with complete task management.

Leon-Sander:
@frostming Thanks for discussing it with the team! I’m planning to implement this on my end as a way to extend BentoML's functionality for my use case. If I complete it, would a pull request for this feature still be open for consideration, or would it likely be declined?

PR #5059, opened 2024-11-03
feat: Add task_id to request inference context

Leon-Sander:
## What does this PR address?

**feat: Add `task_id` tracking to request inference context**

### Description
This PR introduces `task_id` tracking in `Inference context request state`, making `task_id` accessible within API functions with the task decorator by accessing `ctx.request.state.task_id`. Based on the starlette documentation the request.state attribute is there to add additional information. This provides access to `task_id` across task functions, setting up support for features like state management and task control (e.g., pausing/resuming #5058).

Key changes:
- Adds `task_id` to `ctx.request.state`, available during task execution.
- Enables task functions to retrieve `task_id` directly, improving task observability.

Edit:
Initially I put the task_id in the response context, but after the comment from @frostming I modified it to be contained within the request context since the response context is mainly for modifying the properties of the response to be returned and task_id is a fixed value and therefore fits better in the request inference context.

---

## Before submitting:

- [ ] Does the Pull Request follow [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/#summary) naming? 
- [ ] Does the code follow BentoML's code style, `pre-commit run -a` script has passed?
- [ ] Did your changes require updates to the documentation? Documentation has been updated accordingly.
- [ ] Added tests to cover the new feature’s behavior.

### DIFF
--- a/src/_bentoml_impl/server/app.py
+++ b/src/_bentoml_impl/server/app.py
@@ -440,6 +440,7 @@ class ServiceAppFactory(BaseAppFactory):
 
     async def _run_task(self, task_id: str, name: str, request: Request) -> None:
         try:
+            self.service.context.response.task_id = task_id
             resp = await self.api_endpoint_wrapper(name, request)
             await self._result_store.set_result(
                 task_id,
--- a/src/_bentoml_impl/server/app.py
+++ b/src/_bentoml_impl/server/app.py
@@ -440,7 +440,7 @@ class ServiceAppFactory(BaseAppFactory):
 
     async def _run_task(self, task_id: str, name: str, request: Request) -> None:
         try:
-            self.service.context.response.task_id = task_id
+            self.service.context.request.state.task_id = task_id
             resp = await self.api_endpoint_wrapper(name, request)
             await self._result_store.set_result(
                 task_id,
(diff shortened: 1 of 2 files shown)


frostming:
One question, why setting it in response context instead of request, which seems a more proper place to store this information.

Leon-Sander:
@frostming thats a good point and when I saw that the response context is mainly for modifying the properties of the response to be returned I also questioned that.

My initial thought was that when I call the submit_task endpoint, then I get the task_id in the response. And from the documentation I understood that the context allows you to access information about the incoming request and it's response. And since the task_id came with the response to my request, I naturally first looked into the response context in my inference function, but didn't find it there and therefore thought about adding it there.

frostming:
Also `_run_task` is called **after** the response is sent, so I wonder how you can access this properties.

Another point I didn't mention is the whole task implementation is local only, it works when you serve the bento without your own infras. And for bento cloud we have another implementation for tasks.

Leon-Sander:
> Also `_run_task` is called **after** the response is sent, so as background task, so I wonder how you can access this property.

Well I guess then the actual api function which gets the context is called after the response was sent.
Within _run_task the function api_endpoint_wrapper is called which calls the function api_endpoint. Within those two functions the other context variables are set and I guess the actual api function is called. So by accessing it in _run_task I am accessing it even earlier than the main code.

frostming (approved):

PR #5139, opened 2024-12-19
fix: raise error when certfile is not given but ssl enabled and fix local dep

frostming:

## What does this PR address?

Fixes #(issue)

## Before submitting:

### DIFF
--- a/src/bentoml/_internal/utils/pkg.py
+++ b/src/bentoml/_internal/utils/pkg.py
@@ -2,8 +2,10 @@ from __future__ import annotations
 
 import importlib.metadata
 import importlib.util
+import subprocess
 from functools import lru_cache
 from importlib.metadata import PackageNotFoundError
+from pathlib import Path
 from types import ModuleType
 from typing import cast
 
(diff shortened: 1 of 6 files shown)


src/bentoml/_internal/utils/pkg.py
```diff
@@ -35,3 +37,45 @@ def source_locations(pkg: str) -> str | None:
...
     (module_path,) = module.submodule_search_locations  # type: ignore (unfinished typed)
     return cast(str, module_path)
+
+
+DEFAULT_BENTOML_GIT_URL = "https://github.com/bentoml/bentoml.git"
+
+
+@lru_cache(maxsize=1)
+def get_local_bentoml_dependency() -> str:
+    import logging
```
aarnphm, line 47: why do we have to create a separate logger here?

Is it because we are running this in a subprocess?

aarnphm: hmm how does this work with our workflow, where we have a local bentoml installation without checking it into git yet?

we will still build the wheels, but the version for that case/deps would fail the build process no (given that the rev and url does not exist, yet)

aarnphm (approved):

src/bentoml/_internal/utils/pkg.py
frostming: >why do we have to create a separate logger here?

It's best practice to create logger with `getLogger(__name__)` for each module, because it will propagate to the parent till `getLogger('bentoml')` if configured properly

>hmm how does this work with our workflow, where we have a local bentoml installation without checking it into git yet?

This is only used by image spec, where we couldn't build wheel for the editable project. This just provide a way to replicate the current bentoml as close as possible.

PR #5155, opened 2024-12-26
fix: use kantoku instead of circus

bojiang:
Have to fork for Python3.13


xianml (approved):

adamliter:
It looks like part of your codebase still depends on `circus`. With the `1.3.19` release, I'm getting this error when trying to do `bentoml serve .` for one of my projects:

```
ModuleNotFoundError: No module named 'circus'
```

The error originates at [this line](https://github.com/bentoml/BentoML/blob/main/src/bentoml/_internal/utils/circus/__init__.py#L10).

PR #5186, opened 2025-01-17
feat(metrics): init zero for request count of apis

Haivilo:
This initializes endpoints metrics with 0 request count. This way, doing `rate()` in PromQL will correctly capture the rate of change. Currently,  if it's the first time an endpoint returns error, metric is incremented to 1, but 0 value is not recorded. `rate()` will show 0 for error counts. Also imo we don't have to enumerate all http code, which spams the metrics, but we can. 
Screenshot 2025-01-16 at 6 48 04 PM

### DIFF
--- a/src/_bentoml_impl/worker/service.py
+++ b/src/_bentoml_impl/worker/service.py
@@ -186,6 +186,7 @@ def main(
     BentoMLContainer.development_mode.set(development_mode)
 
     server_context.service_name = service.name
+    server_context.service_apis = [api for api in service.apis.keys()]
     if service.bento is None:
         server_context.bento_name = service.name
         server_context.bento_version = "not available"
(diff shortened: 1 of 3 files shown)


src/_bentoml_impl/worker/service.py
```diff
@@ -186,6 +186,7 @@ def main(
     BentoMLContainer.development_mode.set(development_mode)
 
     server_context.service_name = service.name
+    server_context.service_apis = [api for api in service.apis.keys()]
```
frostming, line 189:
```suggestion
    server_context.service_apis = [api.route for api in service.apis.values()]
```
Make it consistent since the request_in_total metric is collected with the url path(route) not the url name

Haivilo: Ah yes. Fixed, and changed the name to `service_routes`.

frostming (approved):

PR #5216, opened 2025-02-17
feat(build): simplier tqdm log viewer

aarnphm:
A simplier tqdm for downloading models from huggingface to reduce the amount of polution on Grafana log viewer


frostming:
This would not take effect when model seeder has downloaded the model properly?

aarnphm:
Yeah, and I realized that tqdm_class doesn't get passed to each of the file_download jobs regardless, so this is not useful.

PR #5244, opened 2025-02-22
Fix RunPod unhashable model error by converting svc.models to hashable ids

rivaon:
✅ Issue: TypeError: unhashable type: 'list'
✅ Cause: svc.models returned a list, which set() cannot handle.
✅ Fix: Convert each model to id(model), making it hashable.
✅ Branch Name: fix-runpod-unhashable-model-error
✅ Commit Message: "Fix RunPod unhashable model error by converting svc.models to hashable IDs"

### DIFF
--- a/src/bentoml/_internal/utils/analytics/usage_stats.py
+++ b/src/bentoml/_internal/utils/analytics/usage_stats.py
@@ -175,7 +175,8 @@ def _track_serve_init(
                 if svc.name in seen:
                     return set()
                 seen.add(svc.name)
-                models = set(svc.models)
+                # Fix: Ensure models are hashable by using id()
+                models = {id(model) for model in svc.models} if isinstance(svc.models, list) else {id(svc.models)}
                 for dep in svc.dependencies.values():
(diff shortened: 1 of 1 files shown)


src/bentoml/_internal/utils/analytics/usage_stats.py
```diff
@@ -175,7 +175,12 @@ def _get_models(
                 if svc.name in seen:
                     return set()
                 seen.add(svc.name)
-                models = set(svc.models)
+                # Fix: Ensure models are hashable by using id()
+                models = (
+                    {id(model) for model in svc.models}
+                    if isinstance(svc.models, list)
+                    else {id(svc.models)}
```
frostming, line 182: Is there any case `svc.models` isn't a list?

frostming (commented):
what models are you using? this helps understand why they are unhashable

frostming (approved):

rivaon:
I am using llama3.2 3B and 7B with ubuntu imgae over runpod

PR #5250, opened 2025-02-26
refactor: drop deepmerge dependency

frostming:

## What does this PR address?

Supersedes #5239


aarnphm (approved):
LGTM

PR #5262, opened 2025-03-07
docs: Update private repo notes

Sherlock113:
## What does this PR address?

Fixes #(issue)

## Before submitting:

### DIFF
--- a/docs/source/build-with-bentoml/runtime-environment.rst
+++ b/docs/source/build-with-bentoml/runtime-environment.rst
@@ -113,7 +113,7 @@ If your project depends on a private GitHub repository, you can include the Pyth
     image = bentoml.images.PythonImage(python_version='3.11') \
         .python_packages("git+ssh://git@github.com/username/repository.git@branch_name")
 
-To configure PyPI indexes and other pip options:
+To configure PyPI indexes and other pip options (e.g. custom package sources and private repositories):
 
 .. code-block:: python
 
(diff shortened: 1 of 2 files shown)


docs/source/build-with-bentoml/runtime-environment.rst
```diff
@@ -128,15 +128,39 @@ To configure PyPI indexes and other pip options:
...
         )
 
-    # Multiple pip options
+    # Configuring multiple PyPI options, including a private repository
     image = bentoml.images.PythonImage(python_version='3.11') \
         .python_packages(
-            "--index-url https://pypi.org/simple",
-            "--extra-index-url https://my.private.pypi/simple",
-            "--trusted-host my.private.pypi",
+            "--index-url https://pypi.org/simple", # Default PyPI index
```
aarnphm, line 134: @frostming I don't think this works with lock_packages=True

or uv don't support this use case yet.

aarnphm: I have to do this https://github.com/bentoml/BentoVLLM/blob/c32bea67e2f15a960b9f45f193b65cc4292245ee/deepseek-r1-distill-llama3.1-8b-tool-calling/service.py#L33 (otherwise building on bentocloud would fail)

frostming: Well yes. Why not use env var `UV_INDEX_URL`

aarnphm: hmm because in case user want to specify `--extra-index-url` or `--find-links`, it can't seem to be resolved to a requirements.txt

fwiw it is also find to separate it out to a separate run step, to cache on image build.

frostming: I mean you can use:

```python
python PythonImage().run("export UV_INDEX_URL=...").requirements_file("requirements.txt")
```

The commands are run in sequential in the same script so `export` works here, a small trick.

aarnphm: lock won't work with `--index-url` in requirements.txt last time I tested with this.

frostming: so it neither reads or writes that option in requirements.txt? well it's embarrassing

aarnphm: or is this `--find-links`? iirc one of these options doesn't work with `lock` on bentocloud (the image builder failed at env-layer)

frostming: Oh, the image builder doesn't lock but consumes the result of `lock` which is performed at `bentoml build`

The problem is it doesn't write that option into the result so it is missing when building image

Sherlock113: @frostming @aarnphm Just to follow up on this. Any change needed for the doc? Like using an env var (`UV_INDEX_URL`) during image building to ensure the index URL is properly set at install time:

```python
image = bentoml.images.PythonImage(python_version='3.11') \
          .run("export UV_INDEX_URL=https://my.private.pypi/simple") \
          .requirements_file("requirements.txt")
```

One thing I am thinking is that it may confuse users. Most people should just expect that specifying `--index-url` in `.python_packages()` would just work

frostming: @Sherlock113 you can keep it as-is now. The latest bentoml version supports global options in `requirements.txt`. now

aarnphm (approved):

PR #5263, opened 2025-03-08
feat(serve): add `--env` and inject service-defined envs to serving environment

aarnphm:
## What does this PR address?

Inject env defined under `@bentoml.service` similar to behaviour as `bentoml deploy`

this PR also remove `env_manager` (given that we barely have any data pointers on people actually using this, or we might have to think of a better way to run serve in an isolated environment)

env_manager only used in conda, so idk, good to also remove this.

### DIFF
--- a/src/bentoml_cli/serve.py
+++ b/src/bentoml_cli/serve.py
@@ -8,8 +8,6 @@ import typing as t
 import click
 import rich
 
-from .deployment import convert_env_to_dict
-
 if t.TYPE_CHECKING:
     P = t.ParamSpec("P")
     F = t.Callable[P, t.Any]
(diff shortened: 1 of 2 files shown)


src/bentoml_cli/serve.py
```diff
@@ -192,7 +193,12 @@ def cli():
         show_default=True,
         hidden=True,
     )
-    @env_manager
+    @click.option(
+        "--env",
+        type=click.STRING,
+        help="List of environment variables pass by --env key[=value] --env ...",
```
frostming, line 199: It's less useful proving such option in serve command, users are always able to achieve this by setting env in their terminal. Can we just leave this out for this PR?

frostming (approved):

PR #5280, opened 2025-03-19
fix: replace namespace import to `python_multipart`

yxtay:
## What does this PR address?

replace `multipart` imports with `python_multipart`

Apparently, there is another python library called `multipart`, which is different from `python_multipart`
- https://github.com/Kludex/python-multipart
- https://github.com/defnull/multipart

By default, `python_multipart` is also available as `multipart`. However, issues arise when `multipart` is also installed. This can be easily fixed by referencing `python_multipart` instead, which is also recommended based on the documentation.

I have searched the repo and replaced 2 instance of `multipart` I found.

Below is run in `python:slim` container image.
```python
root@d7f23af8a56e:/# pip install bentoml multipart
root@d7f23af8a56e:/# python
Python 3.13.2 (main, Mar 18 2025, 01:50:09) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import bentoml.io
/usr/local/lib/python3.13/site-packages/bentoml/io.py:7: BentoMLDeprecationWarning: `bentoml.io` is deprecated since BentoML v1.4 and will be removed in a future version. Please upgrade to new style IO types instead.
  warn_deprecated(
  File "<python-input-0>", line 1, in <module>
    import bentoml.io
  File "/usr/local/lib/python3.13/site-packages/bentoml/io.py", line 12, in <module>
    from ._internal.io_descriptors import from_spec
  File "/usr/local/lib/python3.13/site-packages/bentoml/_internal/io_descriptors/__init__.py", line 6, in <module>
    from .file import File
  File "/usr/local/lib/python3.13/site-packages/bentoml/_internal/io_descriptors/file.py", line 9, in <module>
    from multipart.multipart import parse_options_header
ModuleNotFoundError: No module named 'multipart.multipart'; 'multipart' is not a package
>>> from python_multipart.multipart import parse_options_header
```

## Before submitting:

### DIFF
--- a/src/_bentoml_impl/client/http.py
+++ b/src/_bentoml_impl/client/http.py
@@ -592,8 +592,8 @@ class SyncHTTPClient(HTTPClient[httpx.Client]):
     def _parse_file_response(
         self, endpoint: ClientEndpoint, resp: httpx.Response
     ) -> pathlib.Path | Image.Image:
-        from python_multipart.multipart import parse_options_header
         from PIL import Image
+        from python_multipart.multipart import parse_options_header
 
         content_disposition = resp.headers.get("content-disposition")
         content_type = resp.headers.get("content-type", "")
@@ -767,8 +767,8 @@ class AsyncHTTPClient(HTTPClient[httpx.AsyncClient]):
     async def _parse_file_response(
         self, endpoint: ClientEndpoint, resp: httpx.Response
     ) -> pathlib.Path | Image.Image:
-        from python_multipart.multipart import parse_options_header
         from PIL import Image
+        from python_multipart.multipart import parse_options_header
 
         content_disposition = resp.headers.get("content-disposition")
         content_type = resp.headers.get("content-type", "")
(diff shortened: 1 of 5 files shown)


aarnphm (changes requested):
This code is on the hotpath, and should be considered and benchmark accordingly.

We have ran into problems with multipart in the past, I'm not entirely sure with in any case where you install a different multipart library.

edit: i misread this, it seems like they moved to the new namespace as python-multipart.

@frostming can we take this opportunity to test out the `multipart` library?

yxtay:
To clarify, `bentoml` currently depends on `python-multipart`. I am not suggesting any change to that. I'm only suggesting to change the import statements to use `python_multipart` instead, in case the user is using some library that installs the other `multipart` library that takes the `multipart` namespace.

Hope this is clear. There is no change in the code that is actually used.

frostming:
We don't depend on `multipart` now and @yxtay we are just using the alias provided by `python-multipart`, so this PR LGTM.

frostming (commented):
Can you fix all occurrences of `multipart` imports?

yxtay:
> Can you fix all occurrences of `multipart` imports?

Done. Did not realise there were other occurences.

frostming (approved):

PR #5307, opened 2025-04-08
chore: locking transformers <4.51 for quickstart tests

aarnphm:
https://github.com/huggingface/transformers/issues/37326

Given that we don't need newer version of transformers for quickstart, probably ok to lock the version of transformers

### DIFF
--- a/tests/e2e/fixtures/quickstart/requirements.txt
+++ b/tests/e2e/fixtures/quickstart/requirements.txt
@@ -1,3 +1,3 @@
 bentoml
 torch
-transformers
+transformers<=4.51.0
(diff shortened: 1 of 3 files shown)


tests/e2e/fixtures/quickstart/requirements.txt
```diff
@@ -1,3 +1,3 @@
 bentoml
 torch
-transformers
+transformers<=4.51.0
```
frostming, line 3:
```suggestion
transformers<4.51.0
```

no equal

frostming:
Oh, the correct requirements.txt file should be `tests/e2e/bento_new_sdk/requirements.txt`

PR #5353, opened 2025-05-11
chore: fix link in readme

imba-tjd:
Wrong relative link

image


frostming (approved):

frostming:
This fix is not correct. [Any encoding other than UTF-8 should be considered invalid TOML](https://arc.net/l/quote/dnflblcs)

PR #5363, opened 2025-05-15
fix: allow route name without leading slash

jianshen92:
## What does this PR address?

Fixes #(issue)

## Before submitting:

### DIFF
--- a/src/_bentoml_sdk/decorators.py
+++ b/src/_bentoml_sdk/decorators.py
@@ -97,13 +97,7 @@ def api(
             "max_latency_ms": max_latency_ms,
         }
         if route is not None:
-
-            def parse_route(route: str) -> str:
-                if route.startswith("/"):
-                    return route
-                return f"/{route}"
-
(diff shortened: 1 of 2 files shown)


frostming:
Do you think it's better if user specifies `name=$name`, also sets `route` to `/$name` automatically?

An other concern is, which one is better between auto fixing and erroring out?

aarnphm:
> Do you think it's better if user specifies `name=$name`, also sets `route` to `/$name` automatically?
> 
> An other concern is, which one is better between auto fixing and erroring out?

+1 for error out instead of auto behaviour

frostming:
> +1 for error out instead of auto behaviour

me too. Then a `validator=` in `attr.field()` would make more sense

jianshen92:
> > +1 for error out instead of auto behaviour
> 
> me too. Then a `validator=` in `attr.field()` would make more sense

+1 for erroring out.

aarnphm (approved):

PR #5470, opened 2025-09-15
feat: Allow bento description to be passed to service decorator

sean-hickey-wf:
## What does this PR address?

While migrating to the python sdk released in `v1.3.20` I noticed that there was a regression in the ability to set Bento Descriptions which feed directly into the OpenAPI docs. 

Previously in the `bentofile.yaml` the description could be set either via a file or a str/markdown like below:

```
service: "service:svc"
description: |
    ## Description For My Bento 🍱

    Use **any markdown syntax** here!

    > BentoML is awesome!
```

```
service: "service:svc"
description: "file: ./README.md"
```

Using the pure python sdk:
- Doesn't allow you to define where the README.md file is 
- Will use the README in the same directory as the service.py file if it exists

Both are behaviour that is sub-optimal

This PR:
- Adds an optional `description` to the service class and decorator
- If the description has been populated, as part of the bento build it will copy it over to the `build_config` (Missing step currently) which then allows the description creation process to continue

The optional description parameter supports both:
- str/markdown 
- via a file path "file: <<PATH_TO_MY_FILE>>/README.md"

## Before submitting:

### DIFF
--- a/src/_bentoml_sdk/service/factory.py
+++ b/src/_bentoml_sdk/service/factory.py
@@ -85,6 +85,7 @@ class Service(t.Generic[T]):
     config: Config = attrs.field(factory=Config)
     inner: type[T] = _DummyService
     image: t.Optional[Image] = None
+    description: t.Optional[str] = None
     envs: t.List[BentoEnvSchema] = attrs.field(factory=list, converter=convert_envs)
     labels: t.Dict[str, str] = attrs.field(factory=dict)
     models: list[Model[t.Any]] = attrs.field(factory=list)
@@ -541,6 +542,7 @@ def service(
     *,
     name: str | None = None,
     image: Image | None = None,
+    description: str | None = None,
     envs: list[dict[str, str]] | None = None,
     labels: dict[str, str] | None = None,
     cmd: list[str] | None = None,
@@ -567,6 +569,7 @@ def service(
             config=config,
             inner=inner,
             image=image,
+            description=description,
             envs=envs or [],
             labels=labels or {},
             cmd=cmd,
--- a/src/bentoml/_internal/bento/bento.py
+++ b/src/bentoml/_internal/bento/bento.py
@@ -264,6 +264,8 @@ class Bento(StoreItem):
             )
             build_config.envs.extend(svc.envs)
             build_config.labels.update(svc.labels)
+            if build_config.description is None and svc.description is not None:
+                object.__setattr__(build_config, "description", svc.description)
             if svc.image is not None:
                 image = svc.image
         if not disable_image:


frostming:
You could however just read it from `Service.inner.__doc__` for that purpose.

Specifically, you can modify `doc` property of Service class: https://github.com/bentoml/BentoML/blob/30e2e81571df3e50fe30ddb88e05f2abaaaa0238/src/_bentoml_sdk/service/factory.py#L249-L256 and modify the related part in `Bento.create` method.

sean-hickey-wf:
> You could however just read it from `Service.inner.__doc__` for that purpose.
> 
> Specifically, you can modify `doc` property of Service class:
> 
> https://github.com/bentoml/BentoML/blob/30e2e81571df3e50fe30ddb88e05f2abaaaa0238/src/_bentoml_sdk/service/factory.py#L249-L256
> 
> and modify the related part in `Bento.create` method.

Hey @frostming !

Do you mean something like this?
```python
@property
def doc(self) -> str:
    from bentoml._internal.bento.bento import get_default_svc_readme

    if self.bento is not None:
        return self.bento.doc
    
    if self.inner.__doc__:
        return self.inner.__doc__
    
    return get_default_svc_readme(self) 

```

If this is what you mean then I can see pros and cons to both options. 
- Your way means no new decorator parameter and very minimal code change 🎉 
- The proposed way means adding an additional parameter and arguably more code change

I will advocate for my proposal though 😅 While it introduces a new parameter is keeps the functionality consistent between the `Bentofile.yaml` and the new python sdk. If a user was using a Bentofile they can drop it and just put `description = "file: ./README.md"` in the service decorator and it'll all just work so it's less work on the user side.

Also if we make a change to the `doc` property I think it might have some unintended consequences because if someone has a top level `README.md` file but also have a Docstring on their service class, making this change will swap their OpenAPI docs to the docstring. In the current proposal, leaving the new `description` parameter as None will still default to the top level README.

Either way I'm happy so let me know what you prefer and I can make the necessary changes. 
Also, if I've completely misunderstood what you were saying please let me know!

Thanks for the comments 😄

frostming:
Hi, @sean-hickey-wf your argument makes sense to me. Do you often need to provide the description? because you know the README.md will be used automatically if no other instructions. From my point of view the description is usually long, and not convenient to provide either in the decorator or via CLI argument.

sean-hickey-wf:
> Hi, @sean-hickey-wf your argument makes sense to me. Do you often need to provide the description? because you know the README.md will be used automatically if no other instructions. From my point of view the description is usually long, and not convenient to provide either in the decorator or via CLI argument.

@frostming thanks for the reply!
I find that Repos, especially private repos have more than one README and what was doing with the `bentofile.yaml` was passing the location of the README that contained all of the information that I wanted in the OpenAPI docs which is usually not the same as the root README of the repo. so the functionality that I have with the bentofile that I don't currently have without it is doing something like 

```yaml
description: "file: wf/bentoml_poc_txn_tagger/README.md"
```
^^ In the above case I can point to a sepcific yaml file and this PR allows you to do the same via the additional description parameter

```python 

@bentoml.service(
    image=my_image,
    description="file: src/README.md",
    resources={"cpu": "2"},
    traffic={"timeout": 30},
)
class Summarization:

```

Alternatively you could do something like:
```python 
README_PATH = "src/README.md"

with open(README_PATH, "r") as f:
    description = f.read()


@bentoml.service(
    image=my_image,
    description=description,
    resources={"cpu": "2"},
    traffic={"timeout": 30},
)
class Summarization:
```

Both work with this implementation

frostming (approved):

PR #5541, opened 2026-01-28
fix(client): recreate connector on session refresh to prevent closed session errors

Wirg:
## What does this PR address?

Fixes session refresh causing RuntimeError: Session is closed in the async client.

### Problem
SessionManager creates the connector once in __init__ via a lambda, then reuses it across session refreshes. When session.close() is called during refresh, aiohttp also closes the shared connector (default connector_owner=True). Subsequent sessions then fail with "Session is closed" because they're created with an already-closed connector.

### Solution
Refactor _make_client from a lambda to a method that creates a fresh connector for each session. This ensures each session is self-contained and session.close() properly cleans up without affecting future sessions.
The performance impact is negligible since session refresh only happens every 100 requests or 5 minutes by derfault.

[x] Does the Pull Request follow Conventional Commits specification naming?
[x] Does the code follow BentoML's code style, pre-commit run -a script has passed?
[x] Did you read through contribution guidelines and follow development guidelines?
[x] Did your changes require updates to the documentation?
[ ] Did you write tests to cover your changes?

### DIFF
--- a/src/_bentoml_impl/client/proxy2.py
+++ b/src/_bentoml_impl/client/proxy2.py
@@ -75,24 +75,38 @@ class SessionManager:
         self._created_at = 0.0
         self._request_count = 0
 
-        parsed = urlparse(url)
+        # Store config for _make_session
+        self._app = app
+        self._headers = headers
+        self._timeout = aiohttp.ClientTimeout(total=timeout)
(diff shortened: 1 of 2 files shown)


frostming (approved):

frostming:
Thanks, the fix LGTM

PR #5555, opened 2026-02-25
feat: native support for src-layout projects

### LINKED ISSUE #5537: feature: Native support for src-layout projects
### Feature request

Add native support for Python projects using the src-layout pattern, where the package code resides in a src/ subdirectory rather than at the project root. BentoML should automatically detect projects with pyproject.toml that specify [tool.setuptools.packages.find] where = ["src"] and install them as editable packages during the Docker build process.

### Motivation

The src-layout is a [widely recommended best practice] in the Python packaging community because it:

- Prevents accidental imports from the development directory
- Ensures tests run against the installed package, not local files
- Provides better isolation between development and installed code

Currently, BentoML requires workarounds to support src-layout projects:

**Problem**: When using src-layout, the service module path is src/package_name/service.py, but BentoML includes it as src/src/package_name/**/*.py in the bento. At runtime, Python cannot import package_name because:

1. The package isn't installed in site-packages
2. Setting PYTHONPATH=/home/bentoml/bento/src/src conflicts with OpenTelemetry auto-instrumentation and other tools that modify PYTHONPATH

Current Workarounds:

- Custom Dockerfile templates with manual uv pip install -e commands
- PYTHONPATH manipulation (breaks with OTEL and other instrumentation)
- Restructuring projects to use flat layout (abandons best practices)

### Other

build-ctx doesn't work as often you have some files to include outside of the src/ folder like .env or some artifacts.

VedantMadane:
## Description
Adds native support for Python projects using the src-layout pattern. During `bentoml build`, when a project's source code resides in `src/package_name/`, the `src/` prefix is stripped so the package is importable at runtime without editable installs.

## Changes
- **`bento.py`**: Strip `src/` prefix from file paths during the copy loop when `is_src_layout` is enabled
- **`build_config.py`**: Add `is_src_layout` option to `PythonOptions`; auto-detect by checking if `src/` directory exists alongside `pyproject.toml`
- **`test_src_layout.py`**: Tests for auto-discovery, prefix stripping, and requirements generation

## How it works
1. **Auto-detection**: When loading config from `pyproject.toml`, checks if a `src/` directory exists in the project root
2. **Manual override**: Set `is_src_layout: true` in `bentofile.yaml` under `python:`
3. **Build behavior**: Files under `src/` are copied with the prefix stripped (`src/mypackage/foo.py` → `mypackage/foo.py`)

No packaging or editable installs are involved — this is purely a file-copy path adjustment, consistent with how `bentoml build` works as a regular import mechanism.

Fixes #5537

### DIFF
--- a/src/bentoml/_internal/bento/build_config.py
+++ b/src/bentoml/_internal/bento/build_config.py
@@ -477,6 +477,7 @@ class PythonOptions:
         default=None,
         validator=attr.validators.optional(attr.validators.instance_of(ListStr)),
     )
+    is_src_layout: t.Optional[bool] = None
 
     def __attrs_post_init__(self):
         if self.requirements_txt and self.packages:
@@ -606,7 +607,6 @@ class PythonOptions:
                     f.write(f"{bentoml_req}\n")
                 elif sdist_name:
                     f.write(f"./wheels/{sdist_name}\n")
-
             is_empty = f.tell() == 0
 
         if self.lock_packages and not is_empty:
@@ -673,6 +673,8 @@ class PythonOptions:
             new_attrs["lock_packages"] = self.pack_git_packages is not False
         if self.pack_git_packages is None:
             new_attrs["pack_git_packages"] = True
+        if self.is_src_layout is None:
+            new_attrs["is_src_layout"] = False
         return attr.evolve(self, **new_attrs) if new_attrs else self
 
     @staticmethod
@@ -906,7 +908,9 @@ class BentoBuildConfig:
         return cls.load(yaml_content)
 
     @classmethod
-    def from_pyproject(cls, stream: t.BinaryIO) -> BentoBuildConfig:
+    def from_pyproject(
+        cls, stream: t.BinaryIO, base_dir: str | None = None
+    ) -> BentoBuildConfig:
         if sys.version_info >= (3, 11):
             import tomllib
         else:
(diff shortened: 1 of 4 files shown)


src/bentoml/_internal/bento/build_config.py
```diff
@@ -920,13 +930,47 @@ def from_pyproject(cls, stream: t.BinaryIO) -> BentoBuildConfig:
...
+                == "src"
+            ):
+                is_src_layout = True
+            # Poetry
+            elif any(
+                p.get("from") == "src"
+                for p in data.get("tool", {}).get("poetry", {}).get("packages", [])
+                if isinstance(p, dict)
+            ):
+                is_src_layout = True
```
frostming: It's best not to handle tool-specific logic.

frostming (changes requested):

src/bentoml/_internal/bento/build_config.py
```diff
@@ -611,7 +611,7 @@ def write_to_bento(
                     f.write(f"./wheels/{sdist_name}\n")
 
             if self.is_src_layout:
-                f.write("-e ../../src\n")
+                f.write("./src\n")
```
frostming: Still didn't explain why it needs to be added to requirements, don't rely on AI.

VedantMadane:
Thanks for the detailed feedback @frostming. I understand the approach needs to change fundamentally:

Instead of adding \src/\ to \
equirements.txt\, the build should strip the \src/\ prefix and copy the package files directly into the bento.

To confirm my understanding: during \entoml build\, when a src-layout project is detected (e.g., importable code lives under \src/package_name/\), the build should copy \src/package_name/\ into the bento as just \package_name/\ so imports work at runtime without editable installs. Is that the right direction?

I'll rework the PR once confirmed.

VedantMadane:
Reworked the PR based on your feedback @frostming:

1. **Stripping approach**: During the file copy loop in \bento.py\, files under \src/\ get their prefix stripped so they're importable at runtime. No editable installs, no changes to \
equirements.txt.

2. **No tool-specific logic**: Detection is a simple \os.path.isdir(os.path.join(base_dir, 'src'))\ check so no Setuptools/Hatch/PDM/Poetry parsing.

3. **Removed unrelated changes**: The previous commits included xgboost test changes that didn't belong here. This is now a single clean commit touching only the src-layout files.

4. **Manual override**: Users can also set \is_src_layout: true\ in \bentofile.yaml\ under \python:\ if auto-detection isn't appropriate for their project.

Happy to adjust further if this direction needs tweaking.

frostming (approved):

PR #5558, opened 2026-02-25
fix: resolve SQLite 'database is locked' errors under high concurrency

VedantMadane:
## Description
This PR fixes intermittent \database is locked\ errors when multiple BentoML workers concurrently access the SQLite task store.

## Changes
- **\Sqlite3Store.init_db\**: Now explicitly enables WAL (Write-Ahead Logging) mode and sets a 5000ms busy timeout during database initialization.
- **\Sqlite3Store.__aenter__\**: Removed the redundant \PRAGMA journal_mode=WAL\ call. Keeping it in \__aenter__\ caused multiple connections to attempt to upgrade the journal mode simultaneously during worker startup, leading to locking issues.

## Verification
- Verified that \iosqlite\ connection accepts the timeout parameter.
- Verified that WAL mode is correctly enabled after connection.
- Performed a stress test with 50 concurrent task submissions, which previously failed with \database is locked\ and now passes with 0 errors.

### DIFF
--- a/src/_bentoml_impl/tasks/result.py
+++ b/src/_bentoml_impl/tasks/result.py
@@ -111,7 +111,6 @@ class Sqlite3Store(ResultStore[Request, Response]):
 
     async def __aenter__(self) -> "t.Self":
         self._conn = await self._conn
-        await self._conn.execute("PRAGMA journal_mode=WAL")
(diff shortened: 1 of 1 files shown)


frostming:
Don't stack your changes on other PRs, keep your changeset clean in one PR.

PR #5572, opened 2026-03-16
fix: Allow dockerfile_template access during bentoml containerize

### LINKED ISSUE #5566: bug: `bentoml containerize` fails with `ValueError: Accessing file outside of current working directory` when using `dockerfile_template` (v1.4.36)
### Describe the bug

After upgrading from BentoML 1.4.27 to 1.4.36, `bentoml containerize` fails when the bento's `bentofile.yaml` specifies a `dockerfile_template`:
```
ValueError: Accessing file outside of current working directory is not allowed.
```
This is a regression introduced in PR #5548 (commit `4e0eb007`), which added a `secure=True` default to `resolve_user_filepath` in `bentoml/_internal/utils/filesystem.py`.

## Context: Two-Step Build Pattern

We use the standard two-step BentoML workflow in CI:

1. **`bentoml build`** — packages the bento archive (model + service + dependencies)
2. **`bentoml containerize`** — produces the final Docker image from the archive

This separation is common in production pipelines where you need to inject secrets or private registry credentials at containerize time (e.g., `--build-arg` for pip index URLs, artifactory tokens, etc.) that shouldn't be baked into the bento archive itself. The `dockerfile_template` feature enables this by letting users customize the Dockerfile with additional build steps — such as configuring private PyPI mirrors or installing system packages — that require secrets only available in the CI environment.

This two-step workflow is now broken in 1.4.36 for any bento that uses `dockerfile_template`.

## Root Cause

During `bentoml containerize`, BentoML extracts the bento archive to a temp directory under `/tmp/...`, then calls `resolve_user_filepath(user_templates, build_ctx)` in `generate.py:183` to locate the `Dockerfile.template`. Because `secure=True` checks that the resolved path is relative to `os.getcwd()`, and `/tmp` is not under CWD, the call raises a `ValueError`.

The `secure` check is appropriate for `bentoml build` (where user-supplied paths should be under the project directory), but during `bentoml containerize` the build context is a BentoML-controlled temp directory — not a user-supplied path — so the check is a false positive.

## Workaround

We are currently patching `generate.py` at install time via a sed command in our CI pipeline:

```bash
sed -i 's/resolve_user_filepath(user_templates, build_ctx)/resolve_user_filepath(user_templates, build_ctx, secure=False)/' \
  "$(python -c 'import bentoml; print(bentoml.__path__[0])')/_internal/container/generate.py"
```

### To reproduce

```bash
mkdir /tmp/bentoml-repro && cd /tmp/bentoml-repro
python -m venv .venv && source .venv/bin/activate
pip install bentoml==1.4.36

# Create minimal service
cat > service.py << 'EOF'
import bentoml

@bentoml.service(name="my_service")
class MyService:
    @bentoml.api
    def predict(self) -> str:
        return "hello"
EOF

# Create custom Dockerfile template (common pattern for injecting private registry auth)
cat > Dockerfile.template << 'EOF'
{% extends bento_base_template %}
{{ super() }}
RUN echo "custom dockerfile template"
{% endblock %}
EOF

# Create bentofile
cat > bentofile.yaml << 'EOF'
service: "service:MyService"
include:
  - "*.py"
docker:
  dockerfile_template: Dockerfile.template
EOF

# Step 1: Build succeeds
bentoml build

# Step 2: Containerize fails
bentoml containerize my_service:latest
```

## Error Output

```
  File ".venv/lib/python3.11/site-packages/bentoml/_internal/container/generate.py", line 183, in generate_containerfile
    dir_path = os.path.dirname(resolve_user_filepath(user_templates, build_ctx))
  ...
  File ".venv/lib/python3.11/site-packages/bentoml/_internal/utils/filesystem.py", line 155, in resolve_user_filepath
    raise ValueError("Accessing file outside of current working directory is not allowed.")
ValueError: Accessing file outside of current working directory is not allowed.
```

### Expected behavior

`bentoml containerize` should successfully resolve the `Dockerfile.template` from the extracted bento in `/tmp` and proceed with the Docker build, as it did in 1.4.27. The two-step `build` → `containerize` workflow with `dockerfile_template` should continue to work.

## Suggested Fix

In `generate.py:183`, pass `secure=False` when resolving the template path during containerize, since the build context is a BentoML-managed temp directory (not a user-supplied path):

```python
# Before (broken)
dir_path = os.path.dirname(resolve_user_filepath(user_templates, build_ctx))

# After (fixed)
dir_path = os.path.dirname(resolve_user_filepath(user_templates, build_ctx, secure=False))
```

### Environment

- **BentoML version:** 1.4.36
- **Python version:** 3.11
- **OS:** Ubuntu 22.04 (CI) / macOS (local)
- **Install method:** `pip install bentoml==1.4.36`

BillionClaw:
## What does this PR address?

Fixes #5566

PR #5548 introduced a security fix that added `secure=True` as the default for `resolve_user_filepath()`. This inadvertently broke `bentoml containerize` when using `dockerfile_template` or `setup_script` paths outside the current working directory.

## Root Cause

After the security fix in PR #5548, the `resolve_user_filepath()` function defaults to `secure=True`, which:
1. Rejects absolute paths
2. Rejects paths that resolve outside the current working directory
3. Rejects hidden files

The `dockerfile_template` and `setup_script` options in the bento configuration are user-provided paths that should legitimately be allowed to reference files outside the immediate build context (e.g., shared templates in a parent directory).

## Changes

This fix adds `secure=False` to the `resolve_user_filepath()` calls for:
- `dockerfile_template` in `build_config.py` and `generate.py`
- `setup_script` in `build_config.py`

## Testing

- Ran existing unit tests in `tests/unit/_internal/bento/test_bento.py` - all pass
- Verified the fix allows accessing templates outside the cwd

## Before submitting:

- [x] Did you read through [contribution guidelines](https://github.com/bentoml/BentoML/blob/main/CONTRIBUTING.md) and follow [development guidelines](https://github.com/bentoml/BentoML/blob/main/DEVELOPMENT.md)?

---

I have read the CLA Document and I hereby sign the CLA.

### DIFF
--- a/src/bentoml/_internal/bento/build_config.py
+++ b/src/bentoml/_internal/bento/build_config.py
@@ -255,7 +255,7 @@ class DockerOptions:
 
         if self.setup_script:
             try:
-                setup_script = resolve_user_filepath(self.setup_script, build_ctx)
+                setup_script = resolve_user_filepath(self.setup_script, build_ctx, secure=False)
             except FileNotFoundError as e:
                 raise InvalidArgument(f"Invalid setup_script file: {e}") from None
             if not os.access(setup_script, os.X_OK):
@@ -271,7 +271,7 @@ class DockerOptions:
         # This template then can be used later to containerize.
         if self.dockerfile_template is not None:
             shutil.copy2(
-                resolve_user_filepath(self.dockerfile_template, build_ctx),
+                resolve_user_filepath(self.dockerfile_template, build_ctx, secure=False),
                 docker_folder / "Dockerfile.template",
             )
 
--- a/src/bentoml/_internal/bento/build_config.py
+++ b/src/bentoml/_internal/bento/build_config.py
@@ -255,7 +255,9 @@ class DockerOptions:
 
         if self.setup_script:
             try:
-                setup_script = resolve_user_filepath(self.setup_script, build_ctx, secure=False)
+                setup_script = resolve_user_filepath(
+                    self.setup_script, build_ctx, secure=False
+                )
             except FileNotFoundError as e:
                 raise InvalidArgument(f"Invalid setup_script file: {e}") from None
             if not os.access(setup_script, os.X_OK):
@@ -271,7 +273,9 @@ class DockerOptions:
         # This template then can be used later to containerize.
         if self.dockerfile_template is not None:
             shutil.copy2(
-                resolve_user_filepath(self.dockerfile_template, build_ctx, secure=False),
+                resolve_user_filepath(
+                    self.dockerfile_template, build_ctx, secure=False
+                ),
                 docker_folder / "Dockerfile.template",
             )
 
--- a/src/bentoml/_internal/container/generate.py
+++ b/src/bentoml/_internal/container/generate.py
@@ -180,7 +180,7 @@ def generate_containerfile(
 
     user_templates = docker.dockerfile_template
     if user_templates is not None:
-        dir_path = os.path.dirname(resolve_user_filepath(user_templates, build_ctx))
+        dir_path = os.path.dirname(resolve_user_filepath(user_templates, build_ctx, secure=False))
         user_templates = os.path.basename(user_templates)
         TEMPLATES_PATH.append(dir_path)
         environment = ENVIRONMENT.overlay(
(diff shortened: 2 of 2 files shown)


frostming:
But this beats the purpose of original effort to avoid including arbitrary files by accident.

Doesn't changing directory to folder containing setup script and docker template work?

onthebed:
Friendly bump -- this PR has been open for a while. Happy to make any changes if needed, or close it if no longer relevant.

eledhwen:
Hello 👋 Any news on this front ? @frostming @onthebed?

PR #5631, opened 2026-06-04
fix: make_safe_connect crashes on Windows due to missing uvloop

stayliv3:
## Problem

`make_safe_connect()` in `src/bentoml/_internal/utils/uri.py` unconditionally imports `uvloop` (line 69), which is unavailable on Windows. This causes a `ModuleNotFoundError` crash on any BentoML service that accepts file/image URL inputs, leaving Windows deployments with **zero SSRF protection**.

## Impact

- **Affected platforms**: Windows (all versions)
- **Affected feature**: URL-based file input (multipart form with URL string instead of file upload)
- **Security impact**: Windows users have no SSRF protection — the function meant to block internal network requests crashes before it can do its job
- **Symptom**: HTTP 500 with `ModuleNotFoundError: No module named uvloop` in server logs

## Fix

Added a `try/except ImportError` around the uvloop import with a fallback implementation:
- When uvloop is unavailable, patches `socket.getaddrinfo` to check resolved IPs
- Blocks private, loopback, and link-local addresses (same security policy as the uvloop path)
- Properly restores original `getaddrinfo` in the `finally` block
- Respects proxy configuration (same as existing code)

## Testing

**Before fix (Windows):**
```
Status: 500
Log: ModuleNotFoundError: No module named uvloop
```

**After fix (Windows):**
```
Status: 400
Body: [{"error":"Connection blocked due to insecure input URL"}]
```

**Linux (unchanged):**
```
Status: 400
Body: [{"error":"Connection blocked due to insecure input URL"}]
```

Verified on:
- Windows 10/11 + Python 3.12 + BentoML 1.4.39
- Linux (Kali) + Python 3.13 + BentoML 1.4.39 + uvloop 0.22.1

### DIFF
--- a/src/bentoml/_internal/utils/uri.py
+++ b/src/bentoml/_internal/utils/uri.py
@@ -57,29 +57,72 @@ def is_http_url(url: str) -> bool:
 
 
 original_create_connection = None
+_original_getaddrinfo = None
 
 
 @contextlib.contextmanager
 def make_safe_connect():
-    """Patch loop.create_connection() method to reject unsafe URLs."""
+    """Patch network connection to reject requests to private/internal IP addresses.
+
+    On Linux/macOS with uvloop: patches uvloop.Loop.create_connection.
+    On Windows or without uvloop: patches socket.getaddrinfo as fallback.
+    """
 
     from urllib.request import getproxies
 
     import httpx
-    from uvloop import Loop
 
     from bentoml.exceptions import BadInput
 
-    global original_create_connection
+    try:
+        from uvloop import Loop
+    except ImportError:
+        Loop = None
 
-    if original_create_connection is None:
-        original_create_connection = Loop.create_connection
+    global original_create_connection
+    global _original_getaddrinfo
 
     # Do not check connections with proxy servers
     proxies = [
         (parsed.hostname, parsed.port)
         for parsed in map(urlparse, getproxies().values())
     ]
+
+    if Loop is None:
+        # Fallback for platforms without uvloop (e.g. Windows):
+        # Patch socket.getaddrinfo to check resolved IPs before connection.
+        if _original_getaddrinfo is None:
+            _original_getaddrinfo = socket.getaddrinfo
+
+        def safe_getaddrinfo(host, port, *args, **kwargs):
+            results = _original_getaddrinfo(host, port, *args, **kwargs)
+            if host is not None and (host, port) not in proxies:
+                for family, type_, proto, canonname, sockaddr in results:
+                    try:
+                        ip = ipaddress.ip_address(sockaddr[0])
+                    except ValueError:
+                        continue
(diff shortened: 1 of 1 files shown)


eeshsaxena (commented):
The core problem is real - `from uvloop import Loop` at module use time hard-fails on Windows (and anywhere uvloop isn't installed), so `make_safe_connect()` blows up before it can protect anything. Making uvloop optional and falling back to a `socket.getaddrinfo` guard is a reasonable shape, and checking the *resolved* IPs there is actually a nice property (it catches DNS-rebinding-style hosts that resolve to a private address). The e2e/unit coverage is good to see.

A few things I'd want addressed:

**1. The `getaddrinfo` restore isn't safe under concurrency / re-entrancy.** The pattern is:

```python
if _original_getaddrinfo is None:
    _original_getaddrinfo = socket.getaddrinfo
...
socket.getaddrinfo = safe_getaddrinfo
finally:
    socket.getaddrinfo = _original_getaddrinfo
```

`socket.getaddrinfo` is a *process-global*. If two `make_safe_connect()` contexts overlap (or it's entered while already active), the second capture reads the already-patched `safe_getaddrinfo` as the "original", and on exit the `finally` restores `safe_getaddrinfo` permanently - so the patch leaks past the context and every later lookup keeps doing the SSRF check (or worse, nests). The `_original_getaddrinfo is None` sentinel that guards the capture is never reset, so it also can't recover. The uvloop path has the same shape but is less exposed because it restores to a captured local. Safer would be to capture the previous value into a local at enter and restore *that* local in `finally`, without the module-global sentinel.

**2. This duplicates the block-decision logic, and #5646 is about to change it.** The condition `ip.is_private or ip.is_loopback or ip.is_link_local` now lives in two places (this fallback and the existing `safe_create_connection`). #5646 is currently extending that exact check to add the CGNAT range (`100.64.0.0/10`) and `is_reserved`. If both land, the Windows fallback silently won't get the CGNAT protection. Factoring the "is this IP unsafe?" decision into one helper (which is essentially what #5646 introduces as `is_unsafe_address`) and calling it from both paths would keep them from diverging - worth coordinating with that PR.

**3. Error-mapping asymmetry.** The uvloop path maps a blocked connection to `BadInput("Connection blocked due to insecure input URL")`, but the fallback raises `socket.gaierror(...)` from inside `getaddrinfo` and only converts `httpx.ConnectError` whose message contains `"All connection attempts failed"`. A `gaierror` may surface differently (e.g. `httpx.ConnectError` with a resolution message, or a different exception entirely depending on the client), so a blocked request on Windows might raise a raw resolver error instead of the friendly `BadInput`. Worth asserting the *exception type* users get in the Windows test, not just that the connection is blocked.

Nice direction overall - the fixes above are about making the fallback as robust and consistent as the uvloop path. Deferring to the maintainers.

PR #5632, opened 2026-06-05
feat: support configurable proxy host and user-provided aiohttp client

### LINKED ISSUE #5568: feat: support user-provided aiohttp.ClientSession and configurable proxy host in create_proxy_app
## Problem

In `src/_bentoml_impl/server/proxy.py` (line 67), there is a `# TODO: support aiohttp client` comment. When a user provides their own `aiohttp.ClientSession` on their service class via `self.client`, BentoML detects it but **cannot use it correctly**.

The root cause: BentoML builds request URLs as relative paths (e.g. `/predict`) and relies on `base_url` being set on the `ClientSession`. BentoML's own client sets `base_url=http://localhost:{proxy_port}`, so it works. But the user's client has no `base_url` — they cannot know BentoML's internal `proxy_url` at the time they create their client.

Additionally, the current proxy **hardcodes `localhost`** as the proxy host, which means it cannot forward requests to a **remote or external ML server** running on a different machine or network.

## Use Cases This Enables

1. **Custom auth/SSL**: User brings a pre-configured client with custom headers, SSL certificates, or retry policies.
2. **Remote ML server**: User's custom ML server runs on a different machine or container (e.g., `http://192.168.1.50:8000`), not on `localhost`.

## Current Behavior

```python
# proxy.py line 61-68
proxy_port = service.config.get("http", {}).get("proxy_port", 8000)
proxy_url = f"http://localhost:{proxy_port}"  # hardcoded localhost

if instance_client is not None and isinstance(instance_client, aiohttp.ClientSession):
    # TODO: support aiohttp client
    client = instance_client  # used as-is — breaks because no base_url
```

Later in `reverse_proxy`:
```python
url = yarl.URL.build(path=f"/{path}")  # relative path only — no host
client.request(url=url)                 # crashes for user-provided client
```

## Proposed Fix

**1. Allow configurable proxy host** in service config:
```python
proxy_host = service.config.get("http", {}).get("proxy_host", "localhost")
proxy_port = service.config.get("http", {}).get("proxy_port", 8000)
proxy_url = f"http://{proxy_host}:{proxy_port}"
```

**2. Always build the full absolute URL** in `reverse_proxy`, instead of relying on `base_url`:
```python
# In reverse_proxy — always construct full URL
full_url = f"{proxy_url.rstrip('/')}/{path}"
client.request(url=full_url, ...)
```

**3. Accept user's client but override its base_url awareness** by always passing full URLs.

## Files Affected

- `src/_bentoml_impl/server/proxy.py`

## Notes

- When `proxy_host` is a remote host, `should_start_process` must be `False` — BentoML cannot spawn a process on a remote machine.
- This allows users to bring pre-configured clients (auth, SSL, retries) while BentoML handles URL construction.

Incheonkirin:
## Summary

`create_proxy_app` hardcoded `localhost` as the proxy host and could not use a user-provided `aiohttp.ClientSession` (set via `self.client`). Closes #5568.

The user's session has no `base_url` — they cannot know BentoML's internal `proxy_port` at client-creation time — so the existing code (which relied on `base_url` and built relative URLs) broke for user-provided clients. The host was also hardcoded to `localhost`, so requests could not be forwarded to a remote/external upstream server.

## Changes

- **Configurable proxy host**: add optional `http.proxy_host` (default `localhost`) so requests can be forwarded to a remote/external ML server.
- **Remote-safe process management**: when `proxy_host` is remote, `should_start_process` is `False` — BentoML cannot spawn the upstream command on another machine.
- **Absolute URLs**: build full absolute URLs in `reverse_proxy` and the startup health check instead of relying on the session's `base_url`, so a user-provided client works as-is (custom auth/SSL/retries supported).
- **Config registration**: register `proxy_host` in the v2 config schema, the SDK `HTTPSchema`, and the default configuration.

## Testing

- `py_compile` passes for all changed Python files; default config YAML parses.
- Verified URL construction (local / remote / multi-segment path / query string) and the remote-host classification (`localhost`/`127.0.0.1`/`::1` → local; others → remote).

## Notes

Backward compatible: when `proxy_host` is unset it defaults to `localhost`, preserving current behavior. No new dependencies.

### DIFF
--- a/src/_bentoml_impl/server/proxy.py
+++ b/src/_bentoml_impl/server/proxy.py
@@ -52,24 +52,33 @@ def create_proxy_app(service: Service[t.Any]) -> Starlette:
     ) -> t.AsyncGenerator[dict[str, t.Any], None]:
         server_instance = get_current_service()
         assert server_instance is not None, "Current service is not initialized"
-        # Only start process for the first worker or when replicate_process is True
-        should_start_process = (
+        http_config = service.config.get("http", {})
+        proxy_host = http_config.get("proxy_host", "localhost")
+        proxy_port = http_config.get("proxy_port", 8000)
+        proxy_url = f"http://{proxy_host}:{proxy_port}"
+        health_url = f"{proxy_url}/{health_endpoint.lstrip('/')}"
+        # A remote proxy host points at an externally managed upstream server,
+        # so this process must not try to spawn the command locally.
+        is_remote_proxy = proxy_host not in ("localhost", "127.0.0.1", "::1")
+        # Only start the process for the first worker (or when replicate_process
+        # is True), and never when proxying to a remote host.
+        should_start_process = not is_remote_proxy and (
             service.config.get("replicate_process", False)
             or server_context.worker_index == 1
         )
         async with contextlib.AsyncExitStack() as stack:
-            proxy_port = service.config.get("http", {}).get("proxy_port", 8000)
-            proxy_url = f"http://localhost:{proxy_port}"
             proc: Process | None = None
             if (
                 instance_client := getattr(server_instance, "client", None)
             ) is not None and isinstance(instance_client, aiohttp.ClientSession):
-                # TODO: support aiohttp client
+                # Use the user-provided session as-is. We always build absolute
+                # URLs (see the health check and reverse_proxy below), so the
+                # user's client does not need a base_url pointing at BentoML's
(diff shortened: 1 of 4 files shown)


Incheonkirin:
Replacement PR is now #5659 from a proper fork. It contains the rebased change and current validation; no reopen is needed.

PR #5634, opened 2026-06-07
feat: add native pylock.toml support for container dependency resolution

### LINKED ISSUE #5466: feature: support for pylock.toml
### Feature request

I would like it if BentoML also supports using [pylock.toml](https://packaging.python.org/en/latest/specifications/pylock-toml/) to specify Python dependencies. 

### Motivation

The current options to specify Python dependencies have their limitations because it is impossible to easily configure per-dependency index urls. 

- Using `bentoml.images.Image().python_packages()`
    - will duplicate dependency specification (e.g. for the project in pyproject.toml, and for the bento the service). 
    - does not support `--extra-index-url`. In other words, GPU acceleration with torch does not work. 
 - Using `bentoml.images.Image().requirements_file()`
     - only supports `--extra-index-url` when specified in the file, but  [uv explicitly does not export these](https://github.com/astral-sh/uv/issues/13572#issuecomment-2897654562) becuase requirements.txt is not powerful enough. 
     - BentoML can add the `--extra-index-url` to the requirements.txt file on the fly, but this will apply to all packages.
     - BentoML can add `pip_args` to specify `--index-strategy unsafe-best-match` but these are added to the `uv pip install`. Maybe the documentation is outdated? 
 - Using `bentoml.images.Image().run()` or `run_script()`.
     - There is no way to copy artifacts (e.g. a requirements.txt, or pylock.toml) to the Docker before executing the run.

Hence, any situation where some package require conflicting `--extra-index-url` the current setup is not sufficient. I downgraded to `bentoml<1.3` to fix this. 

I think supporting `pylock.toml` will fix this, because it explicitly defines all the index urls. 

### Other

I'm creating this as a feature request, because I don't think there is a good solution based on requirements.txt. That will only work when specifying the index urls per package line, but uv is not going to do that.

TemidayoA:
## What does this PR address?

Fixes #5466

This PR introduces native support for `pylock.toml` in the Bento containerization build process, allowing users to strictly lock down and resolve Python dependencies using `uv` (bypassing current limitations with `requirements.txt` extra-index-urls).

**Implementation Details:**
1. **Configuration Schema (`build_config.py`)**: Added `pylock_toml` to the `PythonOptions` schema and implemented the file copy logic to stage the lockfile in the `env/python/` build context.
2. **Container Generation (`generate.py`)**: Injected a `_has_pylock` boolean flag into the Jinja templating context by evaluating the existence of the staged lockfile.
3. **Dockerfile Templates (`base.j2`)**: Updated the base Dockerfile template to conditionally execute `uv pip sync /home/bentoml/bento/env/python/pylock.toml` when `_has_pylock` is true, safely falling back to the standard `install.sh` pipeline if no lockfile is present.

This ensures that deployments requiring complex or isolated package indexes can rely entirely on standard lockfiles during the Docker build step without breaking existing backwards compatibility.

### DIFF
--- a/src/bentoml/_internal/bento/build_config.py
+++ b/src/bentoml/_internal/bento/build_config.py
@@ -443,6 +443,11 @@ class PythonOptions:
         default=None,
         validator=attr.validators.optional(attr.validators.instance_of(str)),
     )
+    pylock_toml: t.Optional[str] = attr.field(
+        default=None,
+        validator=attr.validators.optional(attr.validators.instance_of(str)),
+    )
+
     packages: t.Optional[t.List[str]] = attr.field(
         default=None,
         validator=attr.validators.optional(attr.validators.instance_of(ListStr)),
@@ -493,7 +498,7 @@ class PythonOptions:
             )
 
     def is_empty(self) -> bool:
-        return not self.requirements_txt and not self.packages
+        return not self.requirements_txt and not self.packages and not self.pylock_toml
 
     @property
     def _jinja_environment(self) -> jinja2.Environment:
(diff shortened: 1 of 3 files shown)


TemidayoA:
Hi @bojiang, just giving this feature a gentle bump.

It looks like the GitHub Actions workflows are paused pending maintainer approval. Could you hit the **"Approve and run"** button so the CI tests for the `pylock.toml` container generation can execute?

Happy to iterate on the templating logic if the runner catches anything, or if you have any feedback on the implementation!

PR #5637, opened 2026-06-17
Add adaptive batching dispatch metrics

RitwijParmar:
### Summary

This adds dispatcher-level metrics for adaptive batching so operators can see why batches are released and how much queue wait they are paying.

Today BentoML exposes batch size, but it is hard to tell whether an endpoint is filling batches, waiting on the optimizer window, or sitting with low item counts under load. This PR adds a lightweight observer on CorkDispatcher and wires it into both Service and Runner serving paths.

New Prometheus signals include dispatch reason, queued jobs at release time, item count after payload-size aware splitting, and oldest request queue delay. The existing batch-size metric is kept and moved to the same dispatch observer so it is recorded from one place.

### Validation

- `python3 -m ruff check src/bentoml/_internal/marshal/dispatcher.py src/bentoml/_internal/server/runner_app.py src/_bentoml_impl/server/app.py tests/unit/_internal/test_dispatcher_metrics.py`
- `python3 -m compileall -q src/bentoml/_internal/marshal/dispatcher.py src/bentoml/_internal/server/runner_app.py src/_bentoml_impl/server/app.py tests/unit/_internal/test_dispatcher_metrics.py`
- `git diff --check`
- `/Users/ritwij/.cache/codex-runtimes/codex-primary-runtime/dependencies/python/bin/python3 -m pytest tests/unit/_internal/test_dispatcher_metrics.py -q`

### DIFF
--- a/docs/source/get-started/adaptive-batching.rst
+++ b/docs/source/get-started/adaptive-batching.rst
@@ -102,6 +102,21 @@ When you specify ``max_batch_size`` and ``max_latency_ms`` parameters, BentoML e
 
     When using a synchronous endpoint in one Service to call a batchable endpoint in another Service, it sends only one request at a time and waits for a response before sending the next. This is due to the default concurrency of 1 for synchronous endpoints. To enable concurrent requests and allow batching, set the ``threads=N`` parameter in the ``@bentoml.service`` decorator.
 
+Observe adaptive batching
+-------------------------
+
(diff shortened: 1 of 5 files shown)


RitwijParmar:
Closing this PR because it has been inactive for several months and is no longer merge-ready. Thanks for the consideration.

PR #5643, opened 2026-06-28
fix(sdk): check generic args length on bare iterator return annotations to avoid IndexError

### LINKED ISSUE #5625: BUG: IndexError in IODescriptor.from_output() with bare (unparameterized) iterator return annotations
## What breaks

`IODescriptor.from_output()` crashes with an `IndexError` when a service method's return annotation is an unparameterized iterator type such as `t.Iterator`, `t.Generator`, `t.AsyncIterator`, or `t.AsyncGenerator` (and their `collections.abc` equivalents).

## How to trigger

```python
import typing as t
from _bentoml_sdk.io_models import IODescriptor

def my_stream_fn() -> t.Iterator:
    yield 1

IODescriptor.from_output(my_stream_fn)  # raises IndexError
```

## Traceback

```
File "src/_bentoml_sdk/io_models.py", line 435, in from_output
    return_annotation = get_args(return_annotation)[0]
                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: tuple index out of range
```

## Root cause

`is_iterator_type()` correctly returns `True` for bare generics, but `get_args()` returns an empty tuple `()` for unparameterized types. Line 435 unconditionally indexes position 0 without checking the length first. The surrounding `try/except` on lines 447–450 only catches `ValueError` and `TypeError`, so the `IndexError` escapes and crashes service startup.

Hasnaathussain:
IODescriptor.from_output() currently crashes with an IndexError when a service method's return annotation is an unparameterized iterator type (e.g. t.Iterator, t.Generator, etc.) because it unconditionally indexes get_args(return_annotation)[0].

This PR fixes it by checking if generic arguments exist first, and defaulting the annotation to t.Any if get_args() returns an empty tuple.

I have added regression unit tests to verify the fix works correctly for bare iterators and generators.

Closes #5625

### DIFF
--- a/src/_bentoml_sdk/io_models.py
+++ b/src/_bentoml_sdk/io_models.py
@@ -432,7 +432,8 @@ class IODescriptor(IOMixin, BaseModel):
             )
         media_type: str | None = None
         if is_iterator_type(return_annotation):
-            return_annotation = get_args(return_annotation)[0]
+            args = get_args(return_annotation)
+            return_annotation = args[0] if args else t.Any
         elif is_annotated(return_annotation):
             content_type = next(
                 (a for a in get_args(return_annotation) if isinstance(a, ContentType)),
--- a/tests/unit/_bentoml_sdk/test_io_models.py
+++ b/tests/unit/_bentoml_sdk/test_io_models.py
@@ -1,0 +1,19 @@
+from __future__ import annotations
+
+import typing as t
+from _bentoml_sdk.io_models import IODescriptor
+
+def test_from_output_with_bare_iterator():
+    def bare_iterator_fn() -> t.Iterator:
+        yield 1
+
+    descriptor = IODescriptor.from_output(bare_iterator_fn)
+    assert descriptor is not None
+
+
+def test_from_output_with_bare_generator():
+    def bare_generator_fn() -> t.Generator:
+        yield 1
+
+    descriptor = IODescriptor.from_output(bare_generator_fn)
+    assert descriptor is not None
--- a/tests/unit/_bentoml_sdk/test_io_models.py
+++ b/tests/unit/_bentoml_sdk/test_io_models.py
@@ -1,7 +1,9 @@
 from __future__ import annotations
 
 import typing as t
+
 from _bentoml_sdk.io_models import IODescriptor
+
 
 def test_from_output_with_bare_iterator():
     def bare_iterator_fn() -> t.Iterator:


Hasnaathussain:
@bojiang - This PR was requested for review 8 days ago and is ready for merge.

**Summary:**
- Fixes issue #5625: Prevents IndexError when service method return annotation is bare iterator
- Includes regression tests for both bare iterators and generators
- All tests passing

**Status:** ✅ Ready for review and merge

Could you please review when available? Or if unavailable, could another BentoML maintainer review?

eeshsaxena (commented):
The fix looks correct to me, and I was able to confirm the root cause locally.

`is_iterator_type` keys off `get_origin`, and `t.Iterator` / `t.Generator` are in `SYNC_ITERATOR_TYPES`, so a bare (unparameterized) iterator annotation passes the `is_iterator_type(return_annotation)` check. But `get_args(t.Iterator)` is `()` for the bare form, so the old `get_args(return_annotation)[0]` indexes into an empty tuple and raises `IndexError`:

```python
>>> from typing import get_args, get_origin
>>> import typing as t
>>> get_origin(t.Iterator)      # collections.abc.Iterator -> is_iterator_type True
>>> get_args(t.Iterator)        # () -> get_args(...)[0] raises IndexError
```

Falling back to `t.Any` when `get_args()` is empty is the right call and matches how an untyped iterator would otherwise be treated. The two regression tests cover both the `Iterator` and `Generator` bare cases.

One small optional thought: the tests assert `descriptor is not None`, which does guard against the `IndexError` regression (that's the important part). If you wanted to make them a touch stronger you could also assert the resolved element type falls back to `Any` (mirroring the parameterized case), but that's very much optional - the current tests already pin the crash fix.

Deferring to the maintainers for the final review.

Hasnaathussain:
I replaced the unsigned source-plus-pre-commit stack with one signed contributor commit at `1dbb554b`. The implementation and regression tests are unchanged; the bot-only formatting commit was folded into the source commit so the PR has a clean DCO history.

Hasnaathussain:
@eeshsaxena, the branch is now DCO-clean at signed source `1dbb554b`, with the expected pre-commit.ci formatting commit at `b6f8598490`; the implementation and regression test are otherwise unchanged since your root-cause confirmation. Could you take another look when you have time?

Hasnaathussain:
@eeshsaxena, a brief check-in on #5643. The branch remains DCO-clean and mergeable; your root-cause confirmation is reflected in the current implementation and no code changes have been made since. Please let me know if any follow-up test or adjustment is needed.

Hasnaathussain:
With #5725 now addressing #5625, I am withdrawing #5643 so the issue has one active implementation to review. This branch predates the newer PR and will not be kept as a competing review target.

PR #5646, opened 2026-07-01
fix(security): block CGNAT range in SSRF connection guard (#5644)

### LINKED ISSUE #5644: SSRF blocklist bypass via CGNAT range (100.64.0.0/10)
### Describe the bug

reported via email on 2 June 2026 - no responses since. 

I am reporting an incomplete SSRF fix in BentoML that allows unauthenticated network attackers to make the server issue outbound HTTP requests to CGNAT-range addresses (RFC 6598, [100.64.0.0/10](http://100.64.0.0/10)), which the existing blocklist in make_safe_connect() does not cover.

Background: CVE-2025-54381 (GHSA-mrmq-3q62-6cc8) patched SSRF in file upload processing by adding make_safe_connect() to reject connections to private, loopback, and link-local IPs. The patch is correct for those ranges but does not include the CGNAT range [100.64.0.0/10](http://100.64.0.0/10).

Root cause: Python's ipaddress module does not classify [100.64.0.0/10](http://100.64.0.0/10) as private, loopback, or link-local, so the check in uri.py lines 88-95 passes these addresses unchallenged.

Affected code:

src/bentoml/_internal/utils/uri.py -- make_safe_connect(), lines 88-95:
  if ip.is_private or ip.is_loopback or ip.is_link_local:
      raise socket.gaierror(f"Blocked private IP address {host}")

This condition evaluates to False for 100.64.1.1 (CGNAT), allowing the connection.

Trigger paths:
  src/_bentoml_impl/serde.py -- MultipartSerde.ensure_file(), lines 206-210
  src/_bentoml_impl/serde.py -- JSONSerde.parse_request(), lines 176-183

PoC (tested on commit 71f55b5e, bentoml 1.4.39, Linux/uvloop):

Service with PIL.Image.Image input:

  curl -X POST http://localhost:3000/classify -F "img=[http://100.64.1.1:80/internal](http://100.64.1.1/internal)"

Result: connection attempted (no "blocked" error), timeout after 5 seconds -- confirmed in server log as "httpx.ConnectTimeout" rather than the expected "Connection blocked due to insecure input URL".

For comparison, a private IP is correctly blocked:

  curl -X POST http://localhost:3000/classify -F "img=http://192.168.0.1/internal"
  Response: {"error":"Connection blocked due to insecure input URL"}

Suggested fix: Extend the blocked condition to include is_reserved and to explicitly check for CGNAT:

  BLOCKED_NETWORKS = [ipaddress.ip_network("[100.64.0.0/10](http://100.64.0.0/10)")]  # RFC 6598 CGNAT
  if ip.is_private or ip.is_loopback or ip.is_link_local or ip.is_reserved or any(ip in n for n in BLOCKED_NETWORKS):
      raise socket.gaierror(...)

Impact: Allows unauthenticated attackers to probe and potentially read from internal services reachable at CGNAT addresses. In cloud environments (AWS, GCP, ISP-managed infrastructure) where CGNAT IPs are used for internal load balancers or service endpoints, this enables meaningful SSRF exploitation.

Affected versions: all versions through 1.4.39 (latest).

### To reproduce

_No response_

### Expected behavior

_No response_

### Environment

1.4.39 on docker

Bharath-970:
Closes #5644.

## Problem

`make_safe_connect()` in `src/bentoml/_internal/utils/uri.py` blocks outbound connections to private, loopback, and link-local IPs (added for CVE-2025-54381 / GHSA-mrmq-3q62-6cc8). Python's `ipaddress` module does not classify the RFC 6598 Shared Address Space (CGNAT, `100.64.0.0/10`) as private, loopback, or link-local, so those addresses pass the check unblocked:

```pycon
>>> import ipaddress
>>> ip = ipaddress.ip_address("100.64.1.1")
>>> ip.is_private, ip.is_loopback, ip.is_link_local, ip.is_reserved
(False, False, False, False)
```

CGNAT addresses are used for carrier-grade NAT and, in several cloud environments, for internal load balancers and service endpoints reachable from the server — so this is a real SSRF gap for requests to user-supplied URLs (e.g. file/image inputs).

## Fix

- Extend the guard with `is_reserved` and an explicit check against the `100.64.0.0/10` network (guarded to IPv4 to avoid a version-mismatch error when comparing against an IPv6 address).
- Extract the predicate into a small `is_unsafe_address()` function so the security-critical logic is unit-testable directly, without needing to patch uvloop's event loop.

## Test

Added `test_is_unsafe_address_blocks_internal_ranges` / `test_is_unsafe_address_allows_public_addresses` in `tests/unit/_internal/utils/test_uri.py`, covering the CGNAT boundary (`100.63.255.255` allowed, `100.64.0.1`/`100.127.255.255` blocked, `100.128.0.1` allowed) plus existing private/loopback/link-local/reserved ranges and IPv6.

Verified as a true regression guard: reverting the fix fails the new CGNAT test cases; restoring it passes all 16 tests in the file.

```
tests/unit/_internal/utils/test_uri.py ................  16 passed
```

### DIFF
--- a/src/bentoml/_internal/utils/uri.py
+++ b/src/bentoml/_internal/utils/uri.py
@@ -71,6 +71,13 @@ def is_unsafe_address(
     (e.g. when fetching a user-supplied URL). Covers private, loopback, link-local,
     and reserved ranges, plus the IPv4 CGNAT range which `ipaddress` does not
     classify as any of the above."""
+    # Canonicalize IPv4-mapped IPv6 addresses (e.g. ::ffff:100.64.1.1) to their
+    # embedded IPv4 form before checking. `is_private`/`is_loopback`/etc already
+    # see through the mapping on CPython >= 3.10, but the CGNAT range below is a
+    # manual membership check that doesn't, so a mapped CGNAT address would
+    # otherwise slip past this guard.
+    if ip.version == 6 and ip.ipv4_mapped is not None:
+        ip = ip.ipv4_mapped
     return bool(
         ip.is_private
         or ip.is_loopback
--- a/src/bentoml/_internal/utils/uri.py
+++ b/src/bentoml/_internal/utils/uri.py
@@ -58,6 +58,27 @@ def is_http_url(url: str) -> bool:
 
 original_create_connection = None
 
+# RFC 6598 Shared Address Space (CGNAT), e.g. used for carrier-grade NAT and internal
+# cloud load balancers. Not classified as private/loopback/link-local/reserved by
+# ipaddress, so it must be checked explicitly (GHSA-mrmq-3q62-6cc8 follow-up).
+_CGNAT_NETWORK = ipaddress.ip_network("100.64.0.0/10")
+
+
+def is_unsafe_address(
+    ip: "ipaddress.IPv4Address | ipaddress.IPv6Address",
+) -> bool:
+    """Whether `ip` must be blocked from outbound connections initiated by BentoML
+    (e.g. when fetching a user-supplied URL). Covers private, loopback, link-local,
+    and reserved ranges, plus the IPv4 CGNAT range which `ipaddress` does not
+    classify as any of the above."""
+    return bool(
+        ip.is_private
+        or ip.is_loopback
+        or ip.is_link_local
+        or ip.is_reserved
+        or (ip.version == 4 and ip in _CGNAT_NETWORK)
+    )
+
 
 @contextlib.contextmanager
 def make_safe_connect():
@@ -91,7 +112,7 @@ def make_safe_connect():
             except ValueError:
                 raise socket.gaierror(f"Blocked invalid IP address {host}")
             else:
-                if ip.is_private or ip.is_loopback or ip.is_link_local:
+                if is_unsafe_address(ip):
                     raise socket.gaierror(f"Blocked private IP address {host}")
         return await original_create_connection(
             self, protocol_factory, host=host, port=port, **kwargs
--- a/tests/unit/_internal/utils/test_uri.py
+++ b/tests/unit/_internal/utils/test_uri.py
@@ -1,3 +1,4 @@
+import ipaddress
 import os
 import typing as t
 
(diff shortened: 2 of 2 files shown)


eeshsaxena (commented):
This is a solid fix for #5644 - factoring out `is_unsafe_address`, adding `is_reserved`, and covering `100.64.0.0/10` all look right, and I confirmed the core case: `is_unsafe_address(ip_address("100.64.1.1"))` is now `True` while `8.8.8.8` stays allowed.

One gap I think is worth closing while you're here: the CGNAT check is gated on `ip.version == 4`, so it misses the **IPv4-mapped IPv6** form of the same address. On CPython, the existing `is_private` / `is_loopback` / `is_link_local` checks *do* transparently see through a mapped address, so those ranges are covered in both forms:

```
::ffff:127.0.0.1    is_loopback=True     -> blocked (good)
::ffff:192.168.1.1  is_private=True      -> blocked (good)
::ffff:100.64.1.1   is_private=False, version=6  -> is_unsafe_address = False  (bypasses the new CGNAT block)
```

So the new CGNAT protection ends up weaker than the sibling checks specifically for the mapped form. Since the threat model here is an attacker-supplied URL, `http://[::ffff:100.64.1.1]/` would reach the CGNAT target that the direct `http://100.64.1.1/` form now blocks.

A small normalization fixes it - resolve the mapped address before the range check, e.g.:

```python
v4 = ip if ip.version == 4 else ip.ipv4_mapped
return bool(
    ip.is_private
    or ip.is_loopback
    or ip.is_link_local
    or ip.is_reserved
    or (v4 is not None and v4 in _CGNAT_NETWORK)
)
```

It'd also be worth adding `::ffff:100.64.1.1` to the blocked-address test cases alongside `100.64.1.1`, so the mapped form is pinned.

To be clear this is narrow (only CGNAT, only the mapped form - all the standard private ranges are already handled), so it doesn't detract from the PR; just flagging it so the new guard is as strong as the ones around it. Deferring to the maintainers, who know the resolution path better than I do.

zelinewang:
Nice fix — the CGNAT range is an easy one to miss. One completeness note in the same class, in case it's useful before this lands:

The CGNAT clause here is IPv4-gated (`ip.version == 4 and ip in _CGNAT_NETWORK`), so the same CGNAT host written in **IPv4-mapped IPv6** notation slips past it. `::ffff:100.64.1.1` is `version == 6`, so the clause is skipped, and on CPython ≥ 3.10 `ipaddress` delegates `is_private` / `is_reserved` / `is_loopback` / `is_link_local` to the embedded IPv4 — all `False` for a CGNAT address — so `is_unsafe_address()` returns `False` and the guard permits it.

Verified against this PR's code (`f9934b9`), no DB/network needed:

```
100.64.1.1               blocked=True    # bare CGNAT — this PR blocks it
::ffff:100.64.1.1        blocked=False   # mapped CGNAT — PERMITTED (bypass)
::ffff:100.127.255.255   blocked=False   # PERMITTED (bypass)
```

Reproduces on CPython ≥ 3.10 (3.9 happens to block it, but only incidentally — it marks all of `::ffff:0:0/96` reserved). The *other* mapped internal forms (`::ffff:127.0.0.1`, `::ffff:169.254.169.254`, …) are already caught by that `is_*` delegation on 3.10+ — CGNAT is the one that leaks, precisely because it rides on the manual `_CGNAT_NETWORK` list rather than on an `is_*` property.

Smallest fix I found is to canonicalize the mapped form before the existing checks, so `::ffff:<x>` is classified as `<x>` uniformly (and it stays robust if CPython's mapped-address classification shifts again):

```python
# in is_unsafe_address(), before the existing checks
if ip.version == 6 and ip.ipv4_mapped is not None:
    ip = ip.ipv4_mapped
```

I've got this together with regression tests (mapped CGNAT + mapped loopback/link-local blocked; `::ffff:8.8.8.8` still allowed) as a ~9-line change on top of this branch — `test_uri.py` passes 21/21, ruff-clean. Happy to send it as a PR onto this branch, or as a follow-up once this merges — whichever you'd prefer.

<sub>(The analysis here was AI-assisted; every claim above was verified against the imported guard and the repo's own test suite before posting. Noting it since I didn't see an AI-disclosure policy in CONTRIBUTING — happy to follow whatever you use.)</sub>

Bharath-970:
Good catch, both of you — verified the bypass directly (`::ffff:100.64.1.1` slipped through since `is_private`/etc see through the mapping on CPython >= 3.10 but the manual CGNAT range check didn't). Pushed a fix: canonicalize IPv4-mapped IPv6 to its embedded IPv4 form before all the checks, so mapped and bare forms are treated identically. Added mapped CGNAT/loopback/private/public cases to the existing parametrized tests — 21/21 pass.

@zelinewang appreciate the offer, went with the canonicalize-once approach so it stays robust the same way `is_private`/etc already are, rather than adding a second CGNAT-specific branch.

Issue #5246, opened 2025-02-24
bug: Bentoml mounted FastAPI app doesn't apply timeout value


jtong99:
### Describe the bug

Hi, I am running Bentoml and mount with FastAPI application.

```
app = FastAPI()

@bentoml.service(
    name="ai_service",
    resources={
        "cpu": CPU_THREADS,
        "memory": MEMORY,
    },
    traffic={"timeout": 3600}
)
@bentoml.asgi_app(app, path="/api")

@app.post('/v1/test')
async def ai_api(self, request: Request):
```

Even though I added timeout value to bentoml and run the service with timeout `bentoml serve service:svc  --timeout=3600`, the API `/v1/test` only has 60s timeout. Is there anyway to make the API route from FastAPI which is mounted to Bentoml can accept the timeout value?

Thanks

### Expected behavior

The mounted FastAPI service accepted timeout value from Bentoml

### Environment

Bentoml: latest

frostming:
I can't reproduce using the latest version, here is the testing code:

```python
import asyncio

import bentoml
import fastapi

app = fastapi.FastAPI()


@app.post("/greet")
async def greet():
    await asyncio.sleep(120)
    return "hello world"


@bentoml.service(traffic={"timeout": 360})
@bentoml.asgi_app(app)
class MyService:
    pass
```
Serve with `bentoml serve`
And the response returns after around 120s

jtong99:
Actually, it's about multiple APIs comes and it's timeout on processing after 60s, I thought it was about timeout variable but I checked the bentoml library and found it's because of this line at `/bentoml/_internal/server/http/traffic.py`:

> loop.call_later(self.timeout, self._set_timer_out, waiter)

After I commented out this line, the issue is gone. The `TimeoutMiddleware` did not read the environment variable `timeout`, correct? Is there anyway to control this? 

This is the error from bentoml regarding above error:

> {"error":"Not able to process the request in 60.0 seconds"}

ammmanism:
I can reproduce this. Root cause: in `HTTPAppFactory.__call__`
(src/bentoml/_internal/server/http_app.py) the `TimeoutMiddleware` /
`MaxConcurrencyMiddleware` are added to the base Starlette app, but mounted
ASGI apps are attached afterwards via `app.mount(...)`. Because Starlette
mounts create a sub-application, the mounted FastAPI app sits outside the
middleware chain and never receives the timeout.

[...]

PR #5654, opened 2026-07-10
fix: apply timeout/max_concurrency middleware to mounted FastAPI/ASGI…

ammmanism:
@bentoml/maintainers I've reproduced the issue reported in #5246 and opened a fix.

### Root cause

`TimeoutMiddleware` and `MaxConcurrencyMiddleware` were attached as Starlette middleware on the base application. However, applications mounted with `svc.mount_asgi_app(...)` are mounted as independent ASGI apps and bypass that middleware chain. As a result, `traffic.timeout` (and `max_concurrency`) were never enforced for mounted FastAPI/ASGI applications.

### Fix

The middleware is now applied as **outer ASGI wrappers after all mounted applications are attached**, ensuring that every incoming request—whether it targets BentoML routes or mounted FastAPI/ASGI apps—passes through the same timeout and concurrency enforcement.

The change covers both server implementations:

* **Legacy SDK:** `HTTPAppFactory`
* **New SDK:** `ServiceAppFactory`

### Validation

Added an end-to-end regression test that:

* mounts a FastAPI application,
* exposes an endpoint that intentionally exceeds a configured **1-second timeout**,
* verifies that the request correctly returns **HTTP 504**.

The test **fails on `main`** (returns `200 OK`) and **passes with this fix** (returns `504 Gateway Timeout`), confirming the issue and the solution.

Feedback is very welcome.


ammmanism:
Fixes #5246 — `traffic.timeout` and `max_concurrency` were not enforced for applications mounted via `svc.mount_asgi_app(...)`.

# Problem

Mounted FastAPI/ASGI applications could continue running indefinitely even when a global request timeout was configured.

Example configuration:

```yaml
traffic:
  timeout: 1
```

A mounted endpoint that sleeps for 2 seconds still returns **200 OK** after 2 seconds instead of timing out with **504 Gateway Timeout**.

---

# Root Cause

Both HTTP server factories registered `TimeoutMiddleware` and `MaxConcurrencyMiddleware` as ordinary Starlette middleware on the base application.

However, mounted ASGI applications are independent sub-applications and **do not inherit the parent middleware chain**.

```text
Incoming Request
        │
        ▼
Parent App
 ├── TimeoutMiddleware
 ├── MaxConcurrencyMiddleware
 └── BentoML Routes            ✅ Protected
        │
        └──────────────► Mounted FastAPI App
                           └── /mounted/...  ❌ Middleware bypassed
```

As a result, requests routed to mounted applications completely bypassed timeout and concurrency enforcement.

---

# Investigation

I traced the request flow through:

- `Service.asgi_app`
- `HTTPAppFactory`
- `ServiceAppFactory`
- middleware registration
- mount order

Then created a minimal reproduction:

- mount a FastAPI application
- add a `/delay` endpoint that sleeps for **2 seconds**
- configure:

```yaml
traffic:
  timeout: 1
```

### Before this fix

```text
GET /mounted/delay
→ waits ~2 seconds
→ 200 OK
```

### Expected

```text
GET /mounted/delay
→ ~1 second
→ 504 Gateway Timeout
```

This confirmed the mounted application was outside the timeout/concurrency middleware chain.

---

# Solution

Instead of registering these as ordinary Starlette middleware, the composed application is wrapped **after all mounted applications have been attached**.

This guarantees every request passes through the same middleware stack.

Applied to both implementations:

- **Legacy SDK:** `HTTPAppFactory`
- **New SDK:** `ServiceAppFactory`

Final request flow:

```text
Incoming Request
        │
        ▼
TimeoutMiddleware
        │
        ▼
MaxConcurrencyMiddleware
        │
        ▼
Entire ASGI Application
    ├── BentoML Routes
    └── Mounted FastAPI / ASGI Apps
```

`TimeoutMiddleware` remains the outermost wrapper so requests exceeding the configured timeout are terminated before continuing through the request pipeline.

---

# Tests

Added a new end-to-end regression test:

```
tests/e2e/bento_server_http/service_mount_timeout.py
```

The test:

- mounts a FastAPI application
- exposes `/mounted/delay`
- serves using:

```yaml
traffic:
  timeout: 1
```

Expected response:

```http
HTTP/1.1 504 Gateway Timeout

{
  "error": "Not able to process the request in 1 seconds"
}
```

---

# Verification

| Test | Before | After |
|------|--------|-------|
| `test_serve_mounted_app_timeout` | ❌ 200 OK | ✅ 504 Gateway Timeout |
| `test_serve_with_timeout` | ✅ Pass | ✅ Pass |

The new regression test fails on `main` and passes with this change, while existing timeout behavior remains unchanged.

eeshsaxena (commented):
The root cause is right, and I traced the wiring end to end - it holds together. For the record:

- `base_app.middlewares` only appends `TimeoutMiddleware`/`MaxConcurrencyMiddleware` when `self.timeout` / `self.max_concurrency` are truthy. In the new-style app (`_bentoml_impl/server/app.py`) those stay set, so `super().middlewares` still *builds* them and the new `continue` filters them back out; the wrapping in `__call__` then re-applies them outside the mounts. In `http_app.py` you instead pass `timeout=None`/`max_concurrency=None` to `super().__init__` (so the base adds nothing) and re-apply via the stored `self._timeout`/`self._max_concurrency` in that file's own `__call__`. Both paths end up applying each middleware exactly once, wrapping the fully-mounted app - which is the goal. Good, and nice that both an e2e mounted-timeout service and the existing timeout test are covered.

Two things worth a look:

**1. Ordering vs. the rest of the middleware stack.** Starlette applies `Middleware(...)` entries as the outermost layers of the app it builds, and this PR then wraps *that* app again with Timeout/MaxConcurrency. So the final onion is `Timeout -> MaxConcurrency -> [access-log, otel, CORS, user middleware, routes+mounts]`. Previously these two ran *inside* the Starlette middleware stack. The visible consequence: a timed-out request now short-circuits with the 504 *before* the access-log / OTel middleware see the response, so timeouts may stop showing up in access logs and traces. Depending on how you feel about that, you might want the timeout wrapper inside those observability layers rather than outside. Worth a conscious decision + a note.

**2. `skip_paths` no longer applies to `TimeoutMiddleware`.** `skip_paths` is passed only to `MaxConcurrencyMiddleware`; `TimeoutMiddleware(app, timeout=...)` has no skip list (its signature doesn't take one). Before, both lived in the same middleware chain where `livez`/`readyz` were excluded from concurrency limiting. Health checks were presumably never timeout-limited anyway, so this is likely fine - but since `readyz` can legitimately block while a service warms up, it's worth confirming the health endpoints shouldn't also be exempt from the timeout.

Minor: `skip_paths = [self.service.config.get("endpoints", {}).get("livez", "/health")]` will insert a literal `None` into `skip_paths` if `endpoints.livez` is explicitly configured as null; a small guard (or `or "/health"`) avoids a `None` entry in the list.

The core change looks correct to me. Deferring to the maintainers on the ordering/observability tradeoff, which is really a design call.

PR #5672, opened 2026-07-24
fix: prevent config pollution in Service.inject_config() (#5409)

### LINKED ISSUE #5409: bug: Bento Build -- Service.inject_config race condition on BentoMLContainer.config
### Describe the bug

Config updates and service access
/_bentoml_sdk/service/factory.py

def inject_config(self) -> None:
    # ... configuration processing ...
    existing = t.cast(t.Dict[str, t.Any], BentoMLContainer.config.get())
    deep_merge(existing, {"api_server": api_server_config, **rest_config})
    BentoMLContainer.config.set(existing)  # Updates global config

Looking at the code, there's a potential race condition and config pollution:
1, Global state modification: BentoMLContainer.config.set(existing) modifies global state
2, Service interference: Multiple services calling inject_config() can interfere with each other
3, No isolation: Changes from one service affect others

For example : 

# Service A calls inject_config()
service_a = Service("service_a")
service_a.inject_config()  # Updates global BentoMLContainer.config

# Service B calls inject_config() later
service_b = Service("service_b") 
service_b.inject_config()  # Gets the already-modified config from Service A!

### To reproduce

_No response_

### Expected behavior

_No response_

### Environment

`bentoml`: 1.4.17
`python`: 3.11.2

atharv0o:
## Summary

`Service.inject_config()` retrieves the shared config dict via
`BentoMLContainer.config.get()` and merges into it in place via
`deep_merge()`. Since `config.get()` returns a live reference (not a
copy), and `deep_merge()` mutates its first argument, calling
`inject_config()` on one Service can leak mutated config into
subsequent `inject_config()` calls.

Note: this reproduces **sequentially, with no concurrency involved** —
building Service A then Service B in the same process is enough to
see A's previously retrieved config change underneath it. This is a
config isolation issue, not a race condition.

## Fix

Deepcopy the config dict before merging, so mutation happens on an
isolated copy instead of the shared container state:

```python
existing = copy.deepcopy(t.cast(dict[str, t.Any], BentoMLContainer.config.get()))
deep_merge(existing, {"api_server": api_server_config, **rest_config})
BentoMLContainer.config.set(existing)
```

No public API changes, no new dependencies, no locking — the existing
config loading flow is unchanged, only the mutation target is isolated.

## Testing

Added a regression test in `tests/unit/_bentoml_sdk/test_service.py`
that mirrors the reported repro: builds Service A and calls
`inject_config()`, holds a live reference to its resulting config,
then builds and injects Service B, and asserts A's config reference is
unchanged. This test fails on current `main` and passes with this fix.

Ran the existing test suite for this module to confirm no other
behavior changed.

Fixes #5409

### DIFF
--- a/src/_bentoml_sdk/service/factory.py
+++ b/src/_bentoml_sdk/service/factory.py
@@ -454,7 +454,9 @@ class Service(t.Generic[T_co]):
         rest_config = {
             k: main_config[k] for k in main_config if k not in api_server_keys
         }
-        existing = copy.deepcopy(t.cast(dict[str, t.Any], BentoMLContainer.config.get()))
+        existing = copy.deepcopy(
+            t.cast(dict[str, t.Any], BentoMLContainer.config.get())
+        )
         deep_merge(existing, {"api_server": api_server_config, **rest_config})
         BentoMLContainer.config.set(existing)  # type: ignore
 
--- a/src/_bentoml_sdk/service/factory.py
+++ b/src/_bentoml_sdk/service/factory.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import copy
 import inspect
 import logging
 import math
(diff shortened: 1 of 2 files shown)


atharv0o:
Closing this PR in favor of #5679.

After further investigation, I found that the core issue was not a race condition but shared configuration mutation inside `Service.inject_config()`.

The newer PR takes a smaller and more focused approach:
- avoids mutating the shared `BentoMLContainer.config` object
- isolates configuration updates before applying service overrides
- preserves the existing configuration flow without introducing unnecessary complexity

Thanks for the discussion and review here.

PR #5681, opened 2026-08-03
docs: note OpenAI client base_url for multi-model gateways

seven7763:
## Summary

The BentoML vLLM example already shows the OpenAI-compatible client pattern.

This PR adds a short note that the same `OpenAI(base_url=...)` client pattern also works with OpenAI-compatible multi-model gateways when you are not self-hosting or using BentoCloud, using [DaoXE](https://daoxe.com) (`https://api.daoxe.com/v1`) as one concrete example.

Docs only — no runtime behavior changes.

## Test plan

- [ ] Docs still build
- [ ] Existing OpenAI client example unchanged
- [ ] Note is clearly optional / vendor-neutral

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/docs/source/examples/vllm.rst
+++ b/docs/source/examples/vllm.rst
@@ -200,6 +200,8 @@ BentoCloud provides fast and scalable infrastructure for building and scaling AI
 
             For more information, see the `OpenAI API reference documentation <https://platform.openai.com/docs/api-reference/introduction>`_.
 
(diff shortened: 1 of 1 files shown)


seven7763:
The change is a generic base_url note with nothing BentoML-specific in it, so there's no reason for it to sit in review. That was a waste of your queue and I'm sorry for it.

PR #5691, opened 2026-08-15
fix: send one multipart part per URL in list file fields

### LINKED ISSUE #5675: Remote clients drop preceding URL values for multipart list file inputs
### Summary

At current `main` (`73c4dbead99be6515fa25fcd91e348ac30f5c22e`), a schema-discovered BentoML client sending two URL-backed values for one multipart `list[Path]` field sends only the final URL value. I reproduced the same loss through the HTTPX sync/async clients and the separate aiohttp/RemoteProxy client.

A direct multipart request with two same-name parts reaches the service as two values, and local `Path` uploads through the client also preserve both.

### Service under test

```python
from pathlib import Path
import bentoml

@bentoml.service(metrics={"enabled": False})
class MultipartService:
    @bentoml.api
    def echo(self, files: list[Path]) -> list[str]:
        return [path.read_text() for path in files]
```

I used two distinct URL-backed fixtures whose contents were `ALPHA` and `BRAVO`. The clients were initialized in remote-schema mode (`service is None`; schema loaded from `/schema.json`; endpoint `input_spec is None`). I replaced only URL retrieval with an in-memory URL-to-`UploadFile` mapping so the test was deterministic and made no network request; client serialization, ASGI parsing, and service dispatch ran from the pinned source unchanged.

### Results

```text
DIRECT_URL_CONTROL        ["ALPHA", "BRAVO"]
LOCAL_FILE_CLIENT_CONTROL ["ALPHA", "BRAVO"]
HTTPX Sync remote client  ["BRAVO"]
HTTPX Async remote client ["BRAVO"]
proxy2 remote client      ["BRAVO"]
```

The direct control submitted:

```python
files = [
    ("files", (None, "https://files.example.invalid/a.txt")),
    ("files", (None, "https://files.example.invalid/b.txt")),
]
```

For the client cases, the equivalent call was:

```python
client.echo(files=[url_a, url_b])
```

### Expected

Each URL in a multipart list field should produce one same-name part, so the service should receive both values in order, matching the direct-request and local-file controls.

### Source-level cause

Both client implementations iterate the list but assign URL strings through `data[name] = file`. Because `data` is a mapping, each URL overwrites the previous value. Concrete local files instead use the repeated-entry `files` list. On the server side, `MultipartSerde.parse_request()` uses `form.getlist(k)`, so repeated values are preserved if the client actually sends them.

Affected implementations:

- `src/_bentoml_impl/client/http.py` (`_build_multipart` used by HTTPX sync/async clients)
- `src/_bentoml_impl/client/proxy2.py` (duplicated aiohttp/RemoteProxy multipart builder)

This report is limited to remote-schema URL-backed list inputs. Local file uploads and direct repeated multipart requests are unaffected. I am not making a security claim, and this does not appear to be an OpenAPI-schema defect.

I searched open and closed issues/PRs for the builder, URL-backed multipart lists, and last-value overwrite behavior and did not find a direct match. I have a focused service-level regression covering the direct control and all three client paths. Would a narrow fix and regression test be welcome?

feiiiiii5:
## What does this PR address?

Remote clients (HTTPX sync/async and the aiohttp proxy2 clients) serialise a `list[Path]` endpoint field with URL-backed values by writing each URL into the multipart `data` dict under the same field name. The dict can only hold one value per name, so every URL except the last is silently dropped before the request reaches the service.

Repro at current main (`73c4dbe`): a `list[Path]` field receiving `[url_a, url_b]` delivers only `[url_b]` to the service — reproduced through the HTTPX sync/async clients and the proxy2 client (issue #5675).

## Fix

- `http.py`: URL-backed values are appended to the multipart `files` list as text parts (`(name, (None, url))`) instead of overwriting the `data` dict entry, so a list field emits one same-name part per value.
- `proxy2.py`: URL-backed values are collected in `url_parts` and added to the `aiohttp.FormData` via repeated `add_field` calls, again producing one same-name field per value.

The server already aggregates repeated field names via `form.getlist(k)` in `MultipartSerde.parse_request`, so both clients now deliver the complete list.

## Test

New `tests/unit/_internal/client/test_multipart.py` builds multipart requests directly from both client implementations with an in-memory URL file manager and asserts one same-name part per URL:

```
PYTHONPATH=src python3 -m pytest tests/unit/_internal/client/test_multipart.py -q
3 passed
```

Neighbouring client and multipart io tests also pass (9 passed).

Fixes #5675

### DIFF
--- a/src/_bentoml_impl/client/http.py
+++ b/src/_bentoml_impl/client/http.py
@@ -336,7 +336,11 @@ class HTTPClient(AbstractClient, t.Generic[C]):
             for v in value:
                 file = self._file_manager.get_file(v)
                 if isinstance(file, str):
-                    data[name] = file
+                    # URL-backed values are sent as multipart text parts
+                    # (filename=None) so a list field emits one same-name
+                    # part per value instead of overwriting earlier values
+                    # in the ``data`` dict.
+                    files.append((name, (None, file)))
                 else:
                     files.append((name, file))
         headers.pop("content-type", None)
--- a/src/_bentoml_impl/client/proxy2.py
+++ b/src/_bentoml_impl/client/proxy2.py
@@ -482,6 +482,7 @@ class AsyncClient(AbstractClient):
             fields = {t.cast(str, k): getattr(model, k) for k in model.model_fields}
         data: dict[str, t.Any] = {}
         files: list[tuple[str, tuple[str, t.IO[bytes], str | None]]] = []
+        url_parts: list[tuple[str, str]] = []
 
         for name, value in fields.items():
             if not is_file_field(name):
@@ -493,13 +494,18 @@ class AsyncClient(AbstractClient):
             for v in value:
                 file = self._file_manager.get_file(v)
                 if isinstance(file, str):
-                    data[name] = file
+                    # URL-backed values are added as text parts directly so
+                    # a list field emits one same-name part per value instead
+                    # of overwriting earlier values in the ``data`` dict.
+                    url_parts.append((name, file))
                 else:
                     files.append((name, file))
         headers.pop("content-type", None)
         payload = aiohttp.FormData()
         for key, val in data.items():
             payload.add_field(key, val)
+        for key, url in url_parts:
+            payload.add_field(key, url)
         for key, (filename, fileobj, content_type) in files:
             payload.add_field(
                 key,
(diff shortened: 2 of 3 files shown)


feiiiiii5:
Pushed `7df1b13` to close an ordering edge in the aiohttp path: a mixed `list[File]` containing URL and binary values previously emitted all URL parts before all binary parts. Both now share one ordered `file_parts` list, preserving caller value order. New regression test pins a URL followed by a binary file in one field.

Local: multipart + container tests 26 passed; ruff 0.15.12 check + format clean.

feiiiiii5:

PR #5692, opened 2026-08-15
fix: sanitize newlines in docker env dict values and v2 env rendering

### LINKED ISSUE #5656: bug: Dockerfile injection via unsanitized docker.env dict values

feiiiiii5:
## What does this PR address?

Values supplied through the legacy `docker.env` dict (`bentofile.yaml` -> `docker.env: {KEY: VALUE}`) and through the new `envs` path are interpolated into generated Dockerfile `ARG` lines. A value containing a newline breaks out of the line and injects an arbitrary Dockerfile instruction (e.g. `RUN ...`), which executes during `docker build`.

Fixes #5656.

## Fix

Sanitize only at the Dockerfile render boundary, so the config model keeps user values byte-for-byte:

- `base.j2` (legacy dict path): `ARG {{ key|normalize_line }}={{ value|normalize_line|bash_quote }}` and the matching `ENV` reference.
- `base_v2.j2` (`envs` path): add `normalize_line` before `bash_quote` for the value.

## Test

- Newline injection regression renders both paths and asserts the value stays on the quoted `ARG` line.
- New model-preservation regression asserts `DockerOptions.env` keeps tabs, repeated spaces, and surrounding whitespace unchanged.

```bash
PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 PYTHONPATH=src python3 -m pytest tests/unit/_internal/container -q
24 passed
```

## Before submitting:

- [x] Does the Pull Request follow Conventional Commits specification naming?
- [x] Did you read through contribution guidelines and follow development guidelines?
- [x] Did your changes require updates to the documentation? Have you updated those accordingly?


shashvat-singham (commented):
Ran the branch — `tests/unit/_internal/client/test_multipart.py` is 3 passed.

**The template fix is the right one, and worth keeping.** Adding `normalize_line` to the `ARG` value is what actually closes #5656: `bash_quote` alone could never have, because Docker parses the Dockerfile line by line, so a raw newline in the value ends the `ARG` instruction no matter how the shell would later quote it. And it's sufficient on its own — the value appears only on the `ARG` line; the `ENV` line just references `${NAME}`, so there's no second path a value reaches.

**Given that, the `_convert_env` change looks redundant for security but lossy for data.** It collapses all whitespace in the *config model*, not at the render boundary, so it silently rewrites legitimate values that were never a problem:

```
'a\nRUN echo pwned'    -> 'a RUN echo pwned'    # the case you're targeting
'a\tb'                 -> 'a b'                 # tab destroyed
'-Xmx1g  -Xms512m'     -> '-Xmx1g -Xms512m'     # double space collapsed
' padded '             -> 'padded'              # surrounding whitespace stripped
```

Because it mutates `BentoEnvSchema` rather than the generated Dockerfile, the altered value is what the service sees at runtime too — so an env var whose value legitimately contains a tab or repeated spaces (`JAVA_OPTS`, a format string, an indented cert/PEM-ish blob) is quietly changed in a context where nothing unsafe was happening. That's a behaviour change for correct inputs in exchange for defence the template already provides.

My suggestion would be to drop the `_convert_env` hunk and keep the escaping at the point of rendering. If you'd rather have belt-and-braces at the model layer, *rejecting* a value containing a newline would be better than silently normalising it — a user who put a newline in an env value has made a mistake and would want to hear about it, and it leaves tabs and spacing alone.

**Separately: this PR appears to contain #5691.** The `http.py` / `proxy2.py` multipart changes here are the same fix as "send one multipart part per URL in list file fields", and the new test file covers that rather than the env sanitisation. If they're stacked deliberately that's fine, but as it stands the title and the linked issue (#5656) only describe the env half, so a reviewer landing here won't expect the client changes — and if #5691 merges first this will conflict. Worth either splitting them or retitling to say it carries both.

feiiiiii5:
Addressed in commit `a274616`:
- Dropped the `_convert_env` whitespace collapse. The config model now preserves tabs, repeated spaces, and surrounding whitespace; `test_docker_options_env_preserves_value_whitespace` pins that.
- Escaping moved entirely to the Dockerfile render boundary. `base.j2` (legacy dict path) now applies `normalize_line` to the name and `normalize_line | bash_quote` to the value; `base_v2.j2` keeps the same value treatment. A newline-bearing value stays on the quoted `ARG` line in both paths.
- Removed the multipart `http.py` / `proxy2.py` / `test_multipart.py` changes; those live only in #5691 now, so this PR contains exactly the env fix.

Local: `tests/unit/_internal/container` 24 passed; ruff 0.15.12 check + format clean.

feiiiiii5:
Follow-up commit `131ae0f` adds an explicit `base_v2.j2` regression test so the “both paths” claim is covered in-tree. Local container tests: 25 passed; ruff 0.15.12 check + format clean.

feiiiiii5:

feiiiiii5:
Reopening — the 08-28 close was a mistake on my side; this had a real review on it.

@shashvat-singham's 08-15 review confirmed the `normalize_line` addition on the `ARG` value is what actually closes #5656, since Docker parses the line before any shell quoting applies. The `_convert_env` whitespace collapse he flagged is already dropped in `a274616`, and `131ae0f` added the `base_v2.j2` regression test — so the diff is now the two templates plus tests.

Sorry for the noise.

feiiiiii5:
@shashvat-singham gentle nudge: both points from your 08-15 review are addressed on-branch (the `_convert_env` collapse is dropped with a whitespace-preservation test, and the branch is now template + tests only since #5691 closed). Anything else before merge? Thanks!

PR #5694, opened 2026-08-16
fix: mark task FAILED and persist a 500 response when the endpoint raises

feiiiiii5:
## What does this PR address?

In the new-SDK task pipeline (`ServiceAppFactory._run_task`, `src/_bentoml_impl/server/app.py`), when the endpoint wrapper raised an exception the handler only logged it and returned. The task row stayed `IN_PROGRESS` forever:

- `GET /status` never reached a terminal state.
- `GET /get` returned `400 "task is not completed yet"` indefinitely, so polling clients hang until the 24h retention silently deletes the row (then it becomes a confusing 404).
- `/retry` was unreachable (it reads the row via `get()`, which raises `RuntimeError` while `completed_at` is NULL).

`ResultStatus.FAILED` was only reachable when the endpoint *returned* an HTTP ≥400 response, never when it raised — so exception failures were invisible in the state machine.

**Fix**: in the exception path, persist a `FAILED` status and a 500 error response via `set_result` (which also sets `completed_at`), mirroring the message already used for unexpected errors elsewhere in the module. If persisting itself fails (e.g. DB down), log and let the background task end instead of raising from the handler.

## Test

Added `tests/unit/bentoml_io/test_task_failure_status.py`: builds a real `ServiceAppFactory` + `Sqlite3Store`, submits a task, makes the endpoint wrapper raise, runs `_run_task`, and asserts the stored row is `FAILED` with `completed_at` set and a 500 response. Fails on `main` (`RuntimeError: Task ... is not completed`), passes with the fix.

Local: `tests/unit/bentoml_io` 14 passed; the 2 pre-existing `test_mount_asgi_app{,_later}` failures are an opentelemetry `_HTTPStabilityMode` ImportError in the local environment (fail on `main` too, unrelated to this change). `ruff check` + `ruff format --check` clean (v0.15.12).

## Before submitting:

### DIFF
--- a/src/_bentoml_impl/server/app.py
+++ b/src/_bentoml_impl/server/app.py
@@ -592,6 +592,21 @@ class ServiceAppFactory(BaseAppFactory):
             )
         except Exception:
             logger.exception("Task(%s) %s failed", name, task_id)
+            try:
+                await self._result_store.set_result(
+                    task_id,
+                    JSONResponse(
+                        {
+                            "error": "An unexpected error has occurred, please check the server log."
+                        },
+                        status_code=500,
+                    ),
+                    ResultStatus.FAILED,
+                )
+            except Exception:
+                logger.exception(
+                    "Failed to persist failure status for task %s", task_id
+                )
         else:
             logger.info("Task(%s) %s is completed", name, task_id)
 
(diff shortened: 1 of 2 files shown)


feiiiiii5:

PR #5695, opened 2026-08-16
fix: stop cancel_task from wedging the task row in a half-canceled state

feiiiiii5:
## What does this PR address?

`cancel_task` (the `PUT /.../cancel` route in `ServiceAppFactory`) marked the task row `CANCELED` via `set_status`, then returned `400 "task cancellation is not supported in local development server"`. Two problems:

1. **The response contradicts the mutation** — the client is told cancellation failed, while the task row actually changed to `CANCELED`.
2. **The task ends up wedged** — `set_status` updates `executed_at` but never `completed_at`, so afterwards `/status` reports `canceled` while `/get` and `/retry` keep returning `400 "not completed yet"` forever. The row is stuck in a half-canceled state that no other endpoint can move.

Cancellation is genuinely not supported in the local development server (the background task keeps running), so the honest behavior is to return the not-supported error **without mutating the row**, letting the task run to completion.

## Fix

Removed the `set_status(CANCELED)` mutation from `cancel_task` (src/_bentoml_impl/server/app.py), with a comment explaining why the mutation must not come back.

## Test

Added `tests/unit/bentoml_io/test_task_cancel_noop.py`: builds a real `ServiceAppFactory` + `Sqlite3Store`, submits a task, calls `cancel_task`, and asserts the response is 400 **and** the stored status is still `IN_PROGRESS` (fails on main, where the status flips to `CANCELED`).

Local: the new test 1 passed (pre-fix FAIL verified); `test_task_failure_status.py` still passes; ruff 0.15.12 check + format clean on both changed files.

## Before submitting:

### DIFF
--- a/src/_bentoml_impl/server/app.py
+++ b/src/_bentoml_impl/server/app.py
@@ -571,7 +571,11 @@ class ServiceAppFactory(BaseAppFactory):
             resp = JSONResponse({"error": "task_id is required"}, status_code=400)
             self._add_response_headers(resp)
             return resp
-        await self._result_store.set_status(task_id, ResultStatus.CANCELED)
+        # Cancellation is not supported in the local development server:
+        # return the not-supported error WITHOUT mutating the task row. The
+        # previous set_status(CANCELED) left the task wedged — /status reported
+        # canceled while /get and /retry kept returning 400 "not completed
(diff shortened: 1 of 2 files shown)


feiiiiii5:

PR #5700, opened 2026-08-22
fix: reject control characters in docker.env values (Dockerfile injection)

### LINKED ISSUE #5656: bug: Dockerfile injection via unsanitized docker.env dict values

feiiiiii5:
Fixes #5656

## Root cause

Dict-form `docker.env` entries were interpolated into the generated Dockerfile completely raw (`ARG {{ key }}={{ value }}` in `base.j2`), while sibling paths were already hardened:

- list-form `env` → regex-validated in `_convert_env` (`build_config.py`)
- `system_packages` → `bash_quote`
- `base_image` → `normalize_line`

A newline in a value terminates the `ARG` instruction — Dockerfile instructions are newline-delimited and quotes do **not** span lines, so any following lines execute as instructions at build time (reproduced offline via `generate_containerfile`; the reporter's exact payload renders a live `RUN echo PWNED...` step).

## Fix (`src/bentoml/_internal/bento/build_config.py`, +45/-3)

1. **Dict branch of `_convert_env`**: keys/values are checked against a control-character class (`[\x00-\x1f\x7f]`) and raise `BentoMLException` with guidance to use the list form or an `.env` file for complex values. This is the single choke point every template renders from, so v1/v2/debian/alpine/ubi8 are all covered.
   - Plain values keep working byte-identically: spaces are legal unquoted on `ARG KEY=value` lines (`{"GREETING": "hello world"}` renders exactly as before).
2. **`BentoEnvSchema` (the v2 `envs:` path)**: same guard added as attr validators on `name`/`value`. The issue notes this path as "hardened", but `bash_quote` only wraps in single quotes — a raw newline inside still breaks the instruction, so the guard is needed at definition time.

## Tests (`tests/unit/_internal/container/test_generate.py`, +40 lines)

Following the existing same-family patterns (`test_generate_containerfile_quotes_system_packages`, `..._normalizes_custom_base_image`):

- newline payload in dict env raises `BentoMLException` at `DockerOptions` construction (before any template rendering)
- legit values (`"hello world"`, URL) render unchanged and no payload appears
- `BentoEnvSchema` rejects control chars in value and name, accepts legit entries

Red→green verified against the pre-fix tree: 5 passed → **8 passed** (the 3 new tests fail without the fix; the reporter's repro script also flips from "Injection present: True" to raising at construction). `ruff check`/`format --check`: zero new findings vs. baseline on both touched files.

### DIFF
--- a/src/bentoml/_internal/bento/build_config.py
+++ b/src/bentoml/_internal/bento/build_config.py
@@ -98,6 +98,41 @@ def _convert_cuda_version(
     )
 
 
+# Control characters (including newline/carriage return) must never appear in
+# env keys or values: Dockerfile instructions are newline-delimited and quotes
+# do not span lines, so a raw control character would terminate the ARG/ENV
+# instruction and any following lines are executed as instructions.
+_CONTROL_CHARS_RE = re.compile(r"[\x00-\x1f\x7f]")
+
+
+def _validate_env_entry(key: str, value: str) -> None:
+    if _CONTROL_CHARS_RE.search(key) or _CONTROL_CHARS_RE.search(value):
+        raise BentoMLException(
+            f"Invalid environment entry '{key}': environment names and values must not contain "
+            "newline or control characters, as they would break out of the generated Dockerfile's "
+            "ARG/ENV instruction. Use the `env` list form or an `.env` file for complex values."
+        )
+
+
+def _validate_bento_env_name(
+    instance: t.Any, attribute: attr.Attribute[str], value: str
+) -> None:
+    if not value or _CONTROL_CHARS_RE.search(value):
+        raise BentoMLException(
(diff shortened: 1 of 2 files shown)


feiiiiii5:

PR #5701, opened 2026-08-22
fix: send every URL-backed value of multipart list file fields

### LINKED ISSUE #5675: Remote clients drop preceding URL values for multipart list file inputs

feiiiiii5:
Fixes #5675

## Root cause

In both remote-schema multipart builders, URL-backed values were assigned into the scalar `data` mapping while iterating a list file field:

```python
for v in value:
    file = self._file_manager.get_file(v)
    if isinstance(file, str):
        data[name] = file        # last URL wins
    else:
        files.append((name, file))
```

So for e.g. `files: list[Path]` given two URLs, only the final one was sent — matching the report's httpx sync/async and aiohttp/RemoteProxy reproductions. Server-side `MultipartSerde.parse_request()` uses `form.getlist(k)` and preserves repeated values; the loss is purely client-side.

## Fix (`src/_bentoml_impl/client/`, +14/-4 across the two builders)

Collect URL values per field name and emit **one repeated plain multipart field each**:

- `_bentoml_impl/client/http.py`: `data.setdefault(name, []).append(file)`. httpx expands list-valued `data` entries into repeated `DataField`s (verified against httpx's `MultipartStream._iter_fields`), so the wire format is byte-identical to today's single-URL path — plain `Content-Disposition: form-data; name="..."`, no filename. (Appending raw strings to `files=` instead would flip them to `filename="upload"` octet-stream parts and change server-side typing — deliberately avoided.)
- `_bentoml_impl/client/proxy2.py`: same accumulation; the FormData build loop now iterates list values with repeated `add_field` calls.

Single-URL requests keep their exact previous encoding.

## Tests (`tests/unit/_bentoml_impl/test_multipart_url_values.py`, +91 lines)

Offline unit tests driving both real builders (`object.__new__` + injected `ClientFileManager`; no sockets):

- multi-URL + local-file mix: every URL survives as its own plain part (httpx body assertions; aiohttp `_fields` introspection)
- single-URL regression pinning the existing plain-field encoding
- red→green verified: 2 failed on pre-fix tree → **3 passed** after
- `ruff check`: finding count identical to baseline on both touched modules (9 pre-existing, none introduced)

### DIFF
--- a/src/_bentoml_impl/client/http.py
+++ b/src/_bentoml_impl/client/http.py
@@ -336,7 +336,10 @@ class HTTPClient(AbstractClient, t.Generic[C]):
             for v in value:
                 file = self._file_manager.get_file(v)
                 if isinstance(file, str):
-                    data[name] = file
+                    # Collect URL-backed values as a list so every entry is
+                    # emitted as a repeated multipart field (``data`` maps a
+                    # name to one scalar; assigning would keep only the last).
+                    data.setdefault(name, []).append(file)
                 else:
                     files.append((name, file))
         headers.pop("content-type", None)
--- a/src/_bentoml_impl/client/proxy2.py
+++ b/src/_bentoml_impl/client/proxy2.py
@@ -493,13 +493,20 @@ class AsyncClient(AbstractClient):
             for v in value:
                 file = self._file_manager.get_file(v)
                 if isinstance(file, str):
-                    data[name] = file
+                    # Collect URL-backed values as a list so every entry is
+                    # emitted as a repeated multipart field (assigning into
+                    # ``data`` keeps only the last one).
+                    data.setdefault(name, []).append(file)
                 else:
                     files.append((name, file))
         headers.pop("content-type", None)
         payload = aiohttp.FormData()
         for key, val in data.items():
-            payload.add_field(key, val)
+            if isinstance(val, list):
+                for item in val:
+                    payload.add_field(key, item)
+            else:
+                payload.add_field(key, val)
         for key, (filename, fileobj, content_type) in files:
             payload.add_field(
                 key,
--- a/tests/unit/_bentoml_impl/test_multipart_url_values.py
+++ b/tests/unit/_bentoml_impl/test_multipart_url_values.py
@@ -1,0 +1,99 @@
+"""URL-backed values of a multipart list file field must all survive encoding.
+
+Regression tests for the remote-schema client path: ``data[name] = url``
+assigned into a mapping and kept only the last URL of a ``list[Path]`` field.
+Both the httpx client builder and the aiohttp/RemoteProxy builder must emit
+one repeated plain multipart field per URL, matching the wire format that a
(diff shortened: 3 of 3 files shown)


feiiiiii5:

PR #5725, opened 2026-09-13
fix(sdk): handle bare iterator return annotations in IODescriptor.from_output (#5625)

### LINKED ISSUE #5625: BUG: IndexError in IODescriptor.from_output() with bare (unparameterized) iterator return annotations

Tyagiquamar:
Fixes #5625 by checking length of get_args(return_annotation) before indexing position 0 in IODescriptor.from_output, preventing IndexError on unparameterized iterator annotations.


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

